Issue #21, opened 2018-06-27
panic: couldn't find kind ClusterServiceBroker


kargakis:
```
2018/06/27 18:44:05 Create Secret/openshift-template-service-broker/templateservicebroker-client
panic: couldn't find kind ClusterServiceBroker

goroutine 1 [running]:
main.main()
	/home/kargakis/go/src/github.com/jim-minter/azure-helm/cmd/sync/sync.go:37 +0x1e8
exit status 2
```

During the addons installation

PR #23, opened 2018-06-28
Add CCP service dns name in SC certificate

kargakis:
Fixes https://github.com/jim-minter/azure-helm/issues/22

Also seems to fix https://github.com/jim-minter/azure-helm/issues/21

Fixes https://github.com/jim-minter/azure-helm/issues/21
(added by jim-minter)

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -166,7 +166,11 @@ func Generate(m *api.Manifest) (c *Config, err error) {
 			cert:         &c.OpenShiftMasterCert,
 		},
 		{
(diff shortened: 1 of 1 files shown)


jim-minter:
I could believe this fixes #21 as well, updating title.

jim-minter:
I don't see a problem with this - a) I can't see any alternative, b) the end point isn't directly user queriable so that certificate shouldn't leak out, c) if it does, it really doesn't say much anyway.

pweil-:
Agree, I don't see much alternative so no concerns here.

PR #26, opened 2018-06-28
Run SC controller manager in HCP

kargakis:
@jim-minter @pweil- this PR adds the SC controller manager which comes up successfully - we still need to fix the cluster service broker resource urls and of course deploy the service broker deployments.

Fixes https://github.com/jim-minter/azure-helm/issues/10

### DIFF
--- a/pkg/helm/chart/templates/Deployment.apps/servicecatalog-controller.yaml
+++ b/pkg/helm/chart/templates/Deployment.apps/servicecatalog-controller.yaml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: servicecatalog-controller
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: servicecatalog-controller
+  template:
+    metadata:
+      labels:
+        app: servicecatalog-controller
+    spec:
+      automountServiceAccountToken: false
+      containers:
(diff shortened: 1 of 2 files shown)


pkg/helm/chart/templates/Deployment.apps/servicecatalog-controller.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - kube-service-catalog
+        - --leader-elect-resource-lock
+        - configmaps
+        - --broker-relist-interval
+        - 5m
+        - --feature-gates
+        - OriginatingIdentity=true
+        - --k8s-kubeconfig
+        - /tmp/kubeconfig/admin.kubeconfig
+        - --service-catalog-kubeconfig
```
kargakis: If we don't specify this, SC tries to use the service account token. If we do, it needs to point to the core api server and aggregation will take care of the rest.

https://github.com/kubernetes-incubator/service-catalog/blob/300994316c0d323e52242836933f95b8f1660bca/cmd/controller-manager/app/controller_manager.go#L137

kargakis:
Opened https://github.com/jim-minter/azure-helm/issues/27 + https://github.com/jim-minter/azure-helm/issues/28 + https://github.com/jim-minter/azure-helm/issues/29

PR #31, opened 2018-06-28
reorg

jim-minter:
implement stub for upgrade
implement stub for health check
rearrange config files and append .yaml for better syntax highlighting


jim-minter:
@pweil- @kargakis ptal

jim-minter:
wanted to check you were ok with this before merging in case it causes much rebase pain - also you'll need to manually move around _in/_out -> _data or redeploy any running clusters

pweil-:
move/rename is no problem here.

PR #32, opened 2018-06-28
WIP Interfaces for OSA

pweil-:
Migrated from https://github.com/pweil-/azure-ccp-api/pull/1

### DIFF
--- a/pkg/osa/create.go
+++ b/pkg/osa/create.go
@@ -3,19 +3,21 @@ package osa
 import (
 	"io/ioutil"
 	"os"
-
-	"github.com/jim-minter/azure-helm/pkg/arm"
-	"github.com/jim-minter/azure-helm/pkg/config"
-	"github.com/jim-minter/azure-helm/pkg/helm"
 )
 
-func Create() error {
-	m, err := readManifest()
+type RP struct{}
+
+func (*RP) Run() error {
+	osa, err := NewOSAByPath("_data/manifest.yaml", "")
 	if err != nil {
 		return err
 	}
 
-	c, err := config.Generate(m)
+	if errs := osa.Validate(); errs != nil {
+		return err
+	}
+
+	c, err := osa.GenerateConfig()
 	if err != nil {
 		return err
 	}
@@ -30,7 +32,7 @@ func Create() error {
 		return err
 	}
 
-	values, err := helm.Generate(m, c)
+	values, err := osa.GenerateHelm()
 	if err != nil {
 		return err
 	}
@@ -39,7 +41,7 @@ func Create() error {
 		return err
 	}
 
-	azuredeploy, err := arm.Generate(m, c)
+	azuredeploy, err := osa.GenerateARM()
 	if err != nil {
 		return err
 	}
--- a/pkg/osa/create.go
+++ b/pkg/osa/create.go
@@ -5,9 +5,7 @@ import (
 	"os"
 )
 
-type RP struct{}
-
-func (*RP) Run() error {
+func Create() error {
 	osa, err := NewOSAByPath("_data/manifest.yaml", "")
 	if err != nil {
 		return err
--- a/pkg/osa/types.go
+++ b/pkg/osa/types.go
@@ -1,0 +1,111 @@
+package osa
+
+import (
+	"github.com/ghodss/yaml"
+	"github.com/jim-minter/azure-helm/pkg/api"
+	"github.com/jim-minter/azure-helm/pkg/arm"
+	"github.com/jim-minter/azure-helm/pkg/config"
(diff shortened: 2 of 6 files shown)


pkg/osa/types.go
```diff
@@ -0,0 +1,117 @@
...
+func unmarshallManifest(b []byte) (*api.Manifest, error) {
+	var m *api.Manifest
+	err := yaml.Unmarshal(b, &m)
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
```
pweil-: will clean up, just wanted to get it moved asap in case there are more comments :smile:

pkg/osa/create.go
```diff
@@ -4,18 +4,21 @@ import (
...
-	"github.com/jim-minter/azure-helm/pkg/arm"
-	"github.com/jim-minter/azure-helm/pkg/config"
-	"github.com/jim-minter/azure-helm/pkg/helm"
 )
 
-func Create() error {
-	m, err := readManifest()
+type RP struct {}
+
+func (*RP) Run() error {
```
pweil-: this is what I'd consider our "simulator".  Same should hold true for upgrade.go

pkg/osa/types.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	m.newManifest = n
+	m.oldManifest = o
+	return nil
+}
+
+func (m *osaManager) GenerateConfig() (*config.Config, error) {
+	c, err := config.Generate(m.newManifest)
+	if err != nil {
```
pweil-: == derp

pkg/osa/create.go
```diff
@@ -3,19 +3,21 @@ package osa
...
 	"os"
-
-	"github.com/jim-minter/azure-helm/pkg/arm"
-	"github.com/jim-minter/azure-helm/pkg/config"
-	"github.com/jim-minter/azure-helm/pkg/helm"
 )
 
-func Create() error {
-	m, err := readManifest()
+type RP struct{}
```
jim-minter: What do you foresee this type being used for?  What does it add?

pkg/osa/create.go
```diff
@@ -3,19 +3,21 @@ package osa
...
+
+func (*RP) Run() error {
```
jim-minter: I think `Create()` surely has to be a better name here?

pkg/osa/types.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/jim-minter/azure-helm/pkg/api"
+	"github.com/jim-minter/azure-helm/pkg/arm"
+	"github.com/jim-minter/azure-helm/pkg/config"
+	"github.com/jim-minter/azure-helm/pkg/helm"
+	"io/ioutil"
+)
+
+type OSA interface {
+	Validate() error
+	GenerateConfig() (*config.Config, error)
```
jim-minter: Should return `[]byte, error`?

pkg/osa/types.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/jim-minter/azure-helm/pkg/api"
+	"github.com/jim-minter/azure-helm/pkg/arm"
+	"github.com/jim-minter/azure-helm/pkg/config"
+	"github.com/jim-minter/azure-helm/pkg/helm"
+	"io/ioutil"
+)
+
+type OSA interface {
+	Validate() error
```
jim-minter: Should return `[]error`?

pkg/osa/types.go
jim-minter: The updating of config is going to be a pretty important piece as well, so `GenerateConfig` isn't a great name if it's going to do both.  I'm pretty tempted to have two completely different functions in the interface, one for Generate and one for Update.

pkg/osa/types.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		osa.oldManifestBytes = oldBytes
+	}
+	return osa, nil
+}
+
+var _ OSA = &osaManager{}
+
+type osaManager struct {
+	newManifestBytes []byte
```
jim-minter: I don't love storing new and old things side by side here - the "old" is only relevant for validation.  I think it'll be worse if we do the same for config.  I can get behind storing the internal representation manifest and config here though.

pweil-: the reason I went this route was so that I didn't have to validate in the New*** methods.  If that's part of versioned api loading I think these go away and we only have the internal representations to work with.

pweil-:
Latest improvements in: https://github.com/jim-minter/azure-helm/pull/32

PR #36, opened 2018-06-29
Enable dry-running addon sync

kargakis:
@jim-minter @pweil- adds dry run for addons. I don't feel strong on how outputing the resources works (stdout vs generating all files in a directory even though I guess we want to latter for convenience). I am going to merge this for now since I need it for debugging.


pweil-:
:+1:

PR #37, opened 2018-06-29
revert d73f564227310b324926ece8b314b671c8e40a59

jim-minter:


jim-minter:
@kargakis I'm reverting d73f564227310b324926ece8b314b671c8e40a59 because I don't want to force people to have admin on the underlying cluster at the point they run create.

kargakis:
The master api cannot run otherwise?

PR #39, opened 2018-06-29
add simple health check which waits for the console to be ready

jim-minter:
@pweil- ptal

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1,0 +1,55 @@
+package healthcheck
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/jim-minter/azure-helm/pkg/api"
+	"github.com/jim-minter/azure-helm/pkg/config"
+)
+
+func HealthCheck(ctx context.Context, m *api.Manifest, c *config.Config) error {
+	return waitForConsole(ctx, m, c)
+}
+
+func waitForConsole(ctx context.Context, m *api.Manifest, c *config.Config) error {
+	pool := x509.NewCertPool()
+	pool.AddCert(c.CaCert)
+
+	cli := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: pool,
+			},
+		},
+		Timeout: 10 * time.Second,
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,54 @@
...
+	for {
+		resp, err := cli.Do(req)
+		if err != nil {
+			return err
+		}
+
+		switch resp.StatusCode {
+		case http.StatusOK:
+			return nil
+		case http.StatusBadGateway:
```
pweil-: comment will be helpful here for future readers

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,54 @@
...
+			},
+		},
+		Timeout: 10 * time.Second,
+	}
+
+	req, err := http.NewRequest("HEAD", "https://"+m.PublicHostname+"/console/", nil)
+	if err != nil {
+		return err
+	}
+	req = req.WithContext(ctx)
```
pweil-: context must be non-nil here, do we want to provide a default just in case?

pkg/healthcheck/healthcheck.go
jim-minter: added

pkg/healthcheck/healthcheck.go
jim-minter: https://golang.org/pkg/context/: "Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use."

pweil-: so if/when an issue arises with this we'll just point them to go doc?  Seems dodgy

kargakis: Agreed on defaulting to TODO if we get a nil context.

pweil-: I think you're misinterpreting the documentation.  

> "Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use."

These are directions to the caller - MSFT's RP.  The "even if a function permits it" indicate it is perfectly valid to reject the call to `HealthCheck` if a nil context is given or default a timeout ourselves.  

You're defaulting to a runtime panic :smile: 

```
func (r *Request) WithContext(ctx context.Context) *Request {
	if ctx == nil {
		panic("nil context")
	}
```

PR #40, opened 2018-06-29
Use OpenShift Routes again instead of the LoadBalancer.

jim-minter:
This is faster to deploy and closer to running with an Ingress.
Also deploy helm/sync and arm in parallel as they are independent.


jim-minter:
@kargakis @pweil- with this PR I was able to deploy a cluster to the point where the console was working in 5m18 :-D

Hopefully that makes up a little for the fact that the PR itself is pretty ugly.  Need to think how to make it palatable.  Any thoughts?

pweil-:
close for #41? Sub 6 minute deploy is fantastic since we were creeping up in time.  That's good news

PR #41, opened 2018-07-02
Deploy OpenShift HCP on AKS

jim-minter:


jim-minter:
I'd suggest we might as well share the same underlying aks cluster - do `az aks get-credentials -g aks -n aks -f aks/admin.kubeconfig` to populate the kubeconfig

aks/create.sh
```diff
@@ -0,0 +1,25 @@
...
+    exit 1
+fi
+
+if [[ $# -eq 0 ]]; then
+    echo usage: $0 resourcegroup
+    exit 1
+fi
+
+RESOURCEGROUP=$1
+rm -f $(dirname $0)/admin.kubeconfig
```
kargakis: nit: seems like it should happen in `aks/delete.sh`

mjudeikis:
```
set +x
AppClientSecret=$(uuidgen)
AppClientID=$(tools/aad.sh app-create openshift.$RESOURCEGROUP.osadev.cloud $AppClientSecret)
set -x
```
In my test scenario I lacked right permissions and due `+x -x` errors is consusing. 
```
Directory permission is needed for the current user to register the application. For how to configure, please refer 'https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-service-principal-portal'. Original error: Insufficient privileges to complete the operation.
+ az ad sp create --id
az ad sp create: error: argument --id: expected one argument
```
should we add `AppClientID` check after this ran?

upgrade.sh
```diff
@@ -12,7 +12,7 @@ ImageResourceGroup=images ImageResourceName=centos7-3.10-201806231427 \
 
 # poor man's helm (without tiller running)
 helm template pkg/helm/chart -f _data/_out/values.yaml --output-dir _data/_out
-oc apply -Rf _data/_out/osa/templates
+KUBECONFIG=aks/admin.kubeconfig kubectl apply -n $RESOURCEGROUP -Rf _data/_out/osa/templates
```
kargakis: Not sure if `kubectl apply` works correctly when you have created the objects with `kubectl create`. I think `apply` needs to own creation too.

aks/create.sh
jim-minter: Yes - I put that in as a safety because `az aks get-credentials -f FILE` merges the kubeconfig it gets with an existing file, and that causes breakage in our case.
I think I can do `az aks get-credentials -f - >FILE` though and that'll fix :)

upgrade.sh
jim-minter: +1 will fix

jim-minter:
This works, except for waiting for an updated image to be published, so merging.
@pweil- @kargakis @mjudeikis please read the README.md and look at /env.example as I've messed around with create.sh again.

PR #46, opened 2018-07-03
Fully qualify SC apiserver DNS name

kargakis:
Fixes https://github.com/jim-minter/azure-helm/issues/17

@jim-minter


pkg/addons/translate.go
```diff
@@ -310,4 +310,10 @@ var Translations = map[string][]struct {
 			Template: "{{ .Config.RegistryServiceIP }}",
 		},
 	},
+	"Service/kube-service-catalog/apiserver": {
+		{
+			Path:     jsonpath.MustCompile("$.spec.externalName"),
+			Template: "servicecatalog-api.{{ .Manifest.ResourceGroup }}.svc.cluster.local",
```
jim-minter: This currently supposes that HCP namespace == Manifest.ResourceGroup, which may not be the case in generality.  We would need to include the HCP namespace in the manifest.  @amanohar is it OK for us to add another field to the manifest to include the HCP namespace?

pkg/config/generate.go
```diff
@@ -181,7 +181,10 @@ func Generate(m *api.Manifest) (c *Config, err error) {
 			cn: "servicecatalog-api",
 			dnsNames: []string{
 				"servicecatalog-api",
-				"apiserver.kube-service-catalog.svc", // TODO: unclear how safe this is
+				"servicecatalog-api." + m.ResourceGroup,
+				"servicecatalog-api." + m.ResourceGroup + ".svc",
+				"servicecatalog-api." + m.ResourceGroup + ".svc.cluster.local",
```
jim-minter: `"servicecatalog-api." + m.ResourceGroup + ".svc.cluster.local"` is the only one needed here, the others can be removed.

kargakis: The core api server is failing if `"apiserver.kube-service-catalog.svc"` is missing, that's why it's added here.

kargakis: The rest servicecatalog-api hostnames can probably be removed.

pkg/addons/translate.go
kargakis: I don't understand, why would it cause ordering problems?

jim-minter: As it stands, the plugin interface rather supposes that the manifest would be passed in once only at validation time, but it seems probable that an HCP namespace would not be known at that time.  @pweil- fyi

jim-minter:
NB, the same work is applicable for at least all references to master-etcd as well.

pkg/addons/translate.go
kargakis: right, I see

jim-minter:
@amanohar we discussed adding a side-channel struct that would have the hcp namespace available at the point where the helm generation took place, but it turns out that doesn't help us because it is after the point where we want to have finished generating certificates :(  Can we get that information earlier?

kargakis:
Interesting. Let's continue the discussion in https://github.com/jim-minter/azure-helm/issues/17, no reason to keep this open for now.

PR #47, opened 2018-07-04
Grab bag

jim-minter:


jim-minter:
@pweil- @kargakis @mjudeikis sorry this is a grab bag commit.  A few things of interest:
- returned openshift pod and service network ranges back to their defaults
- changed all the addon Secrets to use stringData rather than data (as long as the contents are valid utf-8)
- clean imagePullPolicy by default - at least for now would rather use kubernetes' default logic for this rather than getting it wrong ourselves when we override image names and versions
- don't record service IPs (except for the registry, whose IP gets baked into its cert -- can we stop doing that? --); have service IPs in clusters dynamically generated
- remove lots of default values from our addons files to make them easier to work with
- don't persist the kubernetes service
- don't persist autogenerated rolebindings
- log when sync.go updates an object - in the future we should be tracking the *updates* made both when a cluster is created and when sync.go runs twice to ensure that it is idempotent.

PR #49, opened 2018-07-04
Use separate kubeconfigs for SC api and bootstrap autoapprover

kargakis:
Fixes https://github.com/jim-minter/azure-helm/issues/11

@jim-minter ptal


jim-minter:
great stuff!

PR #52, opened 2018-07-04
Properly use helm for HCP pods

jim-minter:
* Install tiller into AKS cluster
* Use helm/tiller for HCP create/upgrade/delete
* Use checksums to auto-kick HCP pods when their configs change
  (and always lay down secrets before deployments on upgrade)

### DIFF
--- a/aks/tiller.yaml
+++ b/aks/tiller.yaml
@@ -1,0 +1,89 @@
+# from
+# https://github.com/kubernetes/helm/blob/master/docs/rbac.md
+# and
+# helm init --service-account tiller --dry-run --debug
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tiller
+  namespace: kube-system
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: tiller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: tiller
+  namespace: kube-system
+---
+
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app: helm
+    name: tiller
+  name: tiller-deploy
+  namespace: kube-system
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: helm
+        name: tiller
+    spec:
+      containers:
+      - env:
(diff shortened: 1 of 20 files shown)


mjudeikis:
lgtm
do we know if we will be able to use tiller in hcp? 
Helm security is such that "If you can talk to tiller, you can use that tiller" To be able to contact tiller, one needs to be able to create a pod in the namespace tiller resides in. Otherwise its errors.

jim-minter:
> do we know if we will be able to use tiller in hcp?

That's been my understanding, but you're right, we should absolutely check that.  I'll confirm today.

> To be able to contact tiller, one needs to be able to create a pod in the namespace tiller resides in

I don't think that's quite true, but I know what you're getting at.

aks/tiller.yaml
```diff
@@ -0,0 +1,89 @@
+# from
```
pweil-: are you maintaining this file to pin to a specific version (and can we confirm that version with MSFT) or would a `helm init --upgrade` be enough here?

pweil-: Just to call it out, the `helm init` approach is what @sozercan used in the bash script

create.sh
```diff
@@ -68,10 +68,7 @@ go run cmd/createorupdate/createorupdate.go
 
 az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _data/_out/azuredeploy.json --no-wait
 
-# poor man's helm (without tiller running)
-helm template pkg/helm/chart -f _data/_out/values.yaml --output-dir _data/_out
-KUBECONFIG=aks/admin.kubeconfig kubectl create namespace $RESOURCEGROUP
-KUBECONFIG=aks/admin.kubeconfig kubectl apply -n $RESOURCEGROUP -Rf _data/_out/osa/templates
+KUBECONFIG=aks/admin.kubeconfig helm install --namespace $RESOURCEGROUP pkg/helm/chart -f _data/_out/values.yaml -n $RESOURCEGROUP >/dev/null
```
pweil-: there is a `helm upgrade --install` flag that allows an upgrade or install if doesn't exist.  Maybe we can simplify to a single path.

aks/tiller.yaml
jim-minter: we should confirm the version with msft.
afaics helm init doesn't set up rbac so I just put it all in one place.

create.sh
jim-minter: it's true; having it error when someone unwittingly stomps something they already had may possibly be useful too.  Don't think I care too much either way.

PR #53, opened 2018-07-04
readding admin.kubeconfig secret to HCP

jim-minter:
@kargakis fyi - I think the sync pod fundamentally requires the admin.kubeconfig.


kargakis:
+1

PR #54, opened 2018-07-04
remove master-data mount

jim-minter:


jim-minter:
@pweil- you OK with this? afaics the master doesn't write anything there.

pweil-:
I’m fine with it if it isn’t used.

PR #55, opened 2018-07-05
Make config generation idempotent

### LINKED ISSUE #48: Figure out how to regenerate config for individual components
We should be able to regenerate config for individual components while retaining the rest of the config.

eg. I want to regenerate only the SC certificates, or the docker registry config, or ....

@jim-minter @pweil- @mjudeikis @kwoodson

jim-minter:
fixes #48


jim-minter:
@pweil- @kargakis @mjudeikis ptal

mjudeikis:
lgtm

pkg/config/generate.go
```diff
@@ -206,114 +202,171 @@ func Generate(m *api.Manifest) (c *Config, err error) {
...
+		},
 	}
-	if c.SessionSecretEnc, err = randomBytes(24); err != nil {
-		return
+	for _, kc := range kubeconfigs {
+		if kc.namespace == "" {
+			kc.namespace = "default"
+		}
+		if *kc.kubeconfig, err = makeKubeConfig(kc.clientKey, kc.clientCert, c.CaCert, kc.endpoint, kc.username, kc.namespace); err != nil {
+			return
```
pweil-: for other reviewers, the public interface method ensures nil is returned if an error is present, not a partially defined config.

PR #56, opened 2018-07-05
remove RegistryServiceIP config

jim-minter:
I don't think we should be preallocating the registry.default service IP and baking it into the registry TLS cert.
It encourages people to rely on the service IP being static which is an anti-pattern and potential support problem.
I believe that in OCP this is a hangover from the time before DNS was always used for registry access.  Given this is a new service I don't believe we need to support that.


jim-minter:
@pweil- @kargakis @mjudeikis fyi

@bparees do you have any view on this PR as described above in the context of Azure?

pweil-:
> I believe that in OCP this is a hangover from the time before DNS was always used for registry access.

Yes, IIRC registry dns as the default name was started in 3.6 and users won't have access to the OPENSHIFT_DEFAULT_REGISTRY setting

jim-minter:
rebased, thanks

bparees:
> @bparees do you have any view on this PR as described above in the context of Azure?

sounds fine to me.

PR #58, opened 2018-07-05
fix tunnel to work with scalesets

jim-minter:
@mjudeikis I'm not sure how you managed to run create.sh to completion with scalesets without this fix?


mjudeikis:
hmm. Thanks, now I start thinking we for sure need extended testing on feature like this.

PR #61, opened 2018-07-06
add BYO vnet option

mjudeikis:
At this point in time, I loop over scaleset so it has less code but mode complex logic. And this partially enables multiple compute pools in api model. Are we fine proceeding this way or better stick with bigger file and simpler json logic.
If we go this way of the loop we might want to create a separate security group for each pool? Or we just stick with 2 (nsg-infra and ngs-compute)? I would create a new one for each poll.

### DIFF
--- a/pkg/arm/data/azuredeploy.json
+++ b/pkg/arm/data/azuredeploy.json
@@ -4,82 +4,7 @@
     "parameters": {},
     "variables": {},
     "resources": [
-        {
-            "type": "Microsoft.Network/networkSecurityGroups",
-            "apiVersion": "2016-03-30",
-            "location": "{{ .ContainerService.Location | js }}",
-            "name": "nsg-compute",
-            "properties": {
-                "securityRules": [
-                    {
-                        "name": "allow_ssh",
-                        "properties": {
-                            "access": "Allow",
-                            "description": "Allow SSH traffic",
-                            "destinationAddressPrefix": "*",
-                            "destinationPortRange": "22-22",
-                            "direction": "Inbound",
-                            "priority": 101,
-                            "protocol": "Tcp",
-                            "sourceAddressPrefix": "*",
-                            "sourcePortRange": "*"
-                        }
-                    }
-                ]
-            }
-        },
-        {
-            "type": "Microsoft.Network/networkSecurityGroups",
-            "apiVersion": "2016-03-30",
-            "location": "{{ .ContainerService.Location | js }}",
-            "name": "nsg-infra",
-            "properties": {
-                "securityRules": [
-                    {
-                        "name": "allow_ssh",
-                        "properties": {
-                            "access": "Allow",
-                            "description": "Allow SSH traffic",
-                            "destinationAddressPrefix": "*",
-                            "destinationPortRange": "22-22",
-                            "direction": "Inbound",
-                            "priority": 101,
-                            "protocol": "Tcp",
-                            "sourceAddressPrefix": "*",
-                            "sourcePortRange": "*"
-                        }
-                    },
-                    {
-                        "name": "allow_http",
-                        "properties": {
-                            "access": "Allow",
-                            "description": "Allow http traffic",
-                            "destinationAddressPrefix": "*",
-                            "destinationPortRange": "80",
-                            "direction": "Inbound",
-                            "priority": 102,
-                            "protocol": "Tcp",
-                            "sourceAddressPrefix": "*",
-                            "sourcePortRange": "*"
-                        }
-                    },
-                    {
-                        "name": "allow_https",
-                        "properties": {
-                            "access": "Allow",
-                            "description": "Allow https traffic",
-                            "destinationAddressPrefix": "*",
-                            "destinationPortRange": "443",
-                            "direction": "Inbound",
-                            "priority": 103,
-                            "protocol": "Tcp",
-                            "sourceAddressPrefix": "*",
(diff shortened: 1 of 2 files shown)


create.sh
```diff
@@ -87,6 +87,10 @@ tools/dns.sh zone-create $RESOURCEGROUP
 tools/dns.sh a-create $RESOURCEGROUP openshift $HCPINGRESSIP
 tools/dns.sh a-create $RESOURCEGROUP openshift-tunnel $HCPINGRESSIP
 
+until $(curl --output /dev/null --silent --fail -k https://openshift.${RESOURCEGROUP}.${DNS_DOMAIN}/healthz); do
```
jim-minter: If this is needed it probably indicates that the code at https://github.com/jim-minter/azure-helm/blob/master/pkg/addons/addons.go#L290 is broken and needs to be fixed

pkg/config/generate.go
```diff
@@ -68,6 +68,21 @@ func selectContainerImages(cs *acsapi.ContainerService, c *Config) {
 	}
 }
 
+func selectVNET(cs *acsapi.ContainerService, c *Config) {
+	c.DefaultVnetSubnetID = "vnet"
+	c.DefaultSubnet = "default"
+	for key, pool := range cs.Properties.AgentPoolProfiles {
+		if pool.VnetSubnetID == "" {
+			cs.Properties.AgentPoolProfiles[key].VnetSubnetID = "vnet"
```
jim-minter: vnetSubnetID will look like `/subscriptions/SUB_ID/resourceGroups/RG_NAME/providers/Microsoft.Network/virtualNetworks/VNET_NAME/subnets/SUBNET_NAME`

create.sh
jim-minter: Create a follow-up issue if appropriate.

pkg/arm/data/azuredeploy.json
```diff
@@ -80,11 +81,12 @@
                 ]
             }
         },
+{{ if eq ( index .ContainerService.Properties.AgentPoolProfiles 0).VnetSubnetID $.Config.DefaultVnetSubnetID }}
```
jim-minter: for now, the PR you're based on will guarantee that there are 2 agent pool profiles and that the vnetsubnetID is identical across the two.
you can decide to create a vnet based on `{{ if not (index .ContainerService.Properties.AgentPoolProfiles 0).VnetSubnetID }}`

pkg/arm/data/azuredeploy.json
```diff
@@ -217,24 +220,28 @@
...
+{{- range $app := .ContainerService.Properties.AgentPoolProfiles }}
+        ,{
             "type": "Microsoft.Compute/virtualMachineScaleSets",
             "apiVersion": "2017-12-01",
-            "location": "{{ .ContainerService.Location }}",
+            "location": "{{ $.ContainerService.Location }}",
             "dependsOn": [
-                "[resourceId('Microsoft.Network/virtualNetworks', 'vnet')]",
-                "[resourceId('Microsoft.Network/networkSecurityGroups', 'nsg-compute')]"
+{{ if eq $app.Name `infra` }}
```
jim-minter: check {{ if eq $app.Role \`infra\` }}
(rebase)

pkg/arm/data/azuredeploy.json
```diff
@@ -217,139 +143,83 @@
...
-            "plan": {
-                "name": "{{ .Config.ImageSKU | js }}",
-                "publisher": "{{ .Config.ImagePublisher | js }}",
-                "product": "{{ .Config.ImageOffer | js }}"
-            },
-{{- end }}
+        }
+{{- range $app := .ContainerService.Properties.AgentPoolProfiles }},
+       {
+            "type": "Microsoft.Network/networkSecurityGroups",
```
pweil-: why are we moving from 2 NSGs to N NSGs based on the number of agent pools?  It seems like duplication when the NSG really is just tied to the role of the agent pool.

pweil-: > For now we can support only 2 compute pool

The agent pools are a slice in the external API PR.  We could validate that you can only have two but we don't restrict it now AFAICS.   In fact, I would suggest that in the (distant?) future more than 2 is probably a normal use case (infra, linux compute, windows compute)

```go
// AgentPoolProfiles represents all the AgentPoolProfiles.
type AgentPoolProfiles []*AgentPoolProfile```

mjudeikis: This was my intention behind the loop for (distant) distant future. As I understood we cant do more than 3 now because of this https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/azure/azure.go#L95 And I was told azure does not support multiple loadbalancers for scalesets. But original question stays, do you want to have like 2 NSG hardcoded? like `nsg-infa` with `ss-infra` pool and `nsg-compute` with `ss-compute` and future `ss-hpc`,`ss-windows` ?

pweil-: We can always change it later if necessary, leave it like you have it for now.

pweil-:
This is LGTM.  @kargakis I see you're poking around even though you're supposed to be on PTO, will give time if you want to review.  Otherwise I'll merge.

kargakis:
lgtm

kargakis:
Generated code added in https://github.com/jim-minter/azure-helm/commit/68621e2f77df711ec0f893cd66f745709f7e52f0

PR #65, opened 2018-07-09
add master-data pvc for etcd

mjudeikis:
Add pvc for master-data /etcd container

We could do without storageclass name but left it as we still dont know how final HCP will look like.


mjudeikis:
cc: @pweil- @kargakis @jim-minter

kargakis:
I would suggest switching to a statefulset.

mjudeikis:
@kargakis for etcd itself?

kargakis:
Close in favor of https://github.com/jim-minter/azure-helm/pull/67?

mjudeikis:
I would say merge and operator will be built ontop? This will enable people to do testing with what we have now. I think a working example for #67 is possible for tomorrow evening.

kargakis:
If we can't have the operator in time then I'd still vote for switching etcd to a statefulset.

pkg/config/generate.go
```diff
@@ -72,6 +72,7 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 	c.Version = versionLatest
 	c.ImageConfigFormat = "openshift/origin-${component}:${version}"
 	c.TunnelHostname = strings.Replace(cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname, "openshift", "openshift-tunnel", 1)
+	c.EtcdPVCStorageClassName = "managed-premium"
```
kargakis: Why not set this directly in the PVC? It's always going to be managed-premium or whatever gives etcd an SSD.

pweil-:
> I think a working example for #67 is possible for tomorrow evening.

Probably worth holding out then.

mjudeikis:
@pweil-  @kargakis  considering my email on etcd-operator and https://github.com/jim-minter/azure-helm/issues/75 I would merge this.

jim-minter:
@kargakis statefulset would be nice but when I tried to implement it I got stuck with https://github.com/kubernetes/kubernetes/issues/52653 - kubectl rollout status statefulset hangs forever if a rollout is not actually happening

kargakis:
https://github.com/kubernetes/kubernetes/pull/62943 needs to be cherrypicked in origin#master and origin#release-3.10

PR #67, opened 2018-07-10
[WIP] ETCD operator poc

mjudeikis:
Just a poc to show how we could use etcd operator instead of static deployment.

TODO:

- [ ] Add pvc config
- [ ] add restore, backup operators to the underlay
- [ ] clean certificates 
- [ ] test upgrades
- [ ] draft ha and non ha configuration (do we need it for sure)
- [ ] update config to corespond to [1]
- [ ] add quotas and limits for it. 
- [ ] investigate rh etcd image usage

1. 
```
ETCD_ADVERTISE_CLIENT_URLS=https://master-etcd:2379
ETCD_ELECTION_TIMEOUT=2500
ETCD_HEARTBEAT_INTERVAL=500
ETCD_QUOTA_BACKEND_BYTES=4294967296   
ETCD_TRUSTED_CA_FILE=/etc/etcd/ca.crt
```


pweil-:
:+1: Let's discuss this when everyone is back.  I think it buys a lot for us (backup procedures!).  My main concern is image versioning.  In the EtcdCluster type it notes:

```
// Repository is the name of the repository that hosts
	// etcd container images. It should be direct clone of the repository in official
	// release:
	//   https://github.com/coreos/etcd/releases
	// That means, it should have exact same tags and the same meaning for the tags.
	//
	// By default, it is `quay.io/coreos/etcd`.
```

I'm not sure if there are implications that if the repo layout doesn't match that there could be issues or if we're ok using this for custom images.

aks/etcd-operator.yaml
```diff
@@ -0,0 +1,109 @@
+# concatenation of
```
kargakis: Why is this part of the aks code? I expected we would deploy the operator as part of the helm templates.

pkg/config/generate.go
```diff
@@ -131,15 +132,28 @@ func Generate(m *api.Manifest, c *Config) (err error) {
 	}{
 		// Generate etcd certs
```
kargakis: Hrm, isn't the operator handling these?

upgrade.sh
```diff
@@ -12,7 +12,7 @@ KUBECONFIG=aks/admin.kubeconfig helm upgrade $RESOURCEGROUP pkg/helm/chart -f _d
 
 # TODO: when sync runs as an HCP pod (i.e. not in development), hopefully should
 # be able to use helm upgrade --wait here
-for d in master-etcd master-api master-controllers; do
+for d in master-api master-controllers; do
```
kargakis: OK, it seems we need to think about etcd upgrades separately. Follow-up issue for adding an etcd-upgrade.sh script? I expect in prod we are going to separate those anyway.

pkg/config/generate.go
mjudeikis: If you want TLS we need to generate it:
https://github.com/coreos/etcd-operator/blob/master/doc/user/cluster_tls.md

upgrade.sh
mjudeikis: We can helm resource to manage CRD. We just need different checking method to validate if it was updated ok.

aks/etcd-operator.yaml
kargakis: I think operator per client is what we want. Other operators are going to be installed via helm charts/addons and will be per client so I don't think there is any reason to do something different for etcd. The only thing I expect we will get for free from the aks cluster is the etcdcluster CRD is already created (or is it created by the operator?).

mjudeikis: CRD is created by operator itself. Need to confirm this, but sound reasonable. I just followed the ingress controller and tiller model.

pweil-: > I think operator per client is what we want.

agree

mjudeikis:
@pweil- @kargakis questions.
If we say we want to run one operator per customer (one operator managed one etcd cluster) we will need to create `ClusterRole` and `ClusterRoleBinding` for that operator serviceaccount (or default one). 

This implies we:
1. Will have cluster-admin kubeconfig to do so.
2. We will have to know namespace of HCP in the config when generating. 

If we would run one global operator we still need same, but this would be one-time setup for all HCP cluster. Ideas?

aks/etcd-operator.yaml
```diff
@@ -0,0 +1,17 @@
+# TODO: Move me to helm chart when we can sort ordering:
```
mjudeikis: There is ordering issues with CRD and HELM. 
In addition, if you delete/recreate CRD operators will kill etcd. You can delete operator, but you cant delete crd. 
Maybe we could maintain things like this with ROOT level chart for all HCP?

create.sh
```diff
@@ -50,7 +50,7 @@ fi
 # TODO: if the user interrupts the process here, the AAD application will leak.
 
 cat >_data/manifest.yaml <<EOF
-name: openshift
+name: $RESOURCEGROUP
```
mjudeikis: Operator etcd uses certificates with `master-etcd.namespace.svc` to communicate. Those names need to be in the certificates. We need to know name for those. Could we use manifest name field for a namespace?

kargakis: Hrm, they assume the operator runs on a different namespace than etcd?

mjudeikis: this should not be even a case. I assume because operator can be global or local it uses same code base. And we end with full DNS names.
```
 etcdserver: not healthy for reconfigure, rejecting member add {ID:e6a985970d419f37 RaftAttributes:{PeerURLs:[https://master-etcd-pkgtnptbmp.master-etcd.mjudeikis-hcp.svc:2380]} Attributes:{Name: ClientURLs:[]}}
2018-07-13 10:47:54.104837 W | etcdserver: timed out waiting for read index response
2018-07-13 10:48:01.390910 I | etcdserver/membership: added member 48e34665e30d3732 [https://master-etcd-m9glrkvqm2.master-etcd.mjudeikis-hcp.svc:2380] to cluster 6b80ae4a9ed28414
```
it advertises as full fqdn:
```
  - command:
    - /usr/local/bin/etcd
    - --data-dir=/var/etcd/data
    - --name=master-etcd-5lswnbnvvb
    - --initial-advertise-peer-urls=https://master-etcd-5lswnbnvvb.master-etcd.mjudeikis-hcp.svc:2380
    - --listen-peer-urls=https://0.0.0.0:2380
    - --listen-client-urls=https://0.0.0.0:2379
    - --advertise-client-urls=https://master-etcd-5lswnbnvvb.master-etcd.mjudeikis-hcp.svc:2379
    - --initial-cluster=master-etcd-5lswnbnvvb=https://master-etcd-5lswnbnvvb.master-etcd.mjudeikis-hcp.svc:2380
```
Where advertise peer url is:
```
func (m *Member) Addr() string {
	return fmt.Sprintf("%s.%s.%s.svc", m.Name, clusterNameFromMemberName(m.Name), m.Namespace)
}
```

kargakis: We may need to loop in someone from the etcd team. I would expect both etcd peers and the operator could assume they can run on the same namespace.

mjudeikis:
https://github.com/jim-minter/azure-helm/issues/75

PR #68, opened 2018-07-10
add ose image resolver

mjudeikis:
Initial draft to add OSE images function to resolve images

ToDo:
[ ] - confirm version after release.
[ ] - configure pre-release registry for testing before release
[ ] - pending on rhel image build to be working (other story)

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -59,18 +59,46 @@ func selectContainerImagesOrigin(cs *acsapi.ContainerService, c *Config) {
 	}
 }
 
+func selectContainerImagesOSA(cs *acsapi.ContainerService, c *Config) {
+	switch cs.Properties.OrchestratorProfile.OpenShiftConfig.OpenShiftVersion {
+	//TODO: confirm minor version after release
+	case "3.10", "3.10.15":
+		c.MasterEtcdImage = "rhel7/etcd:v3.10.15-1"
+		c.MasterAPIImage = "openshift3/ose-control-plane:v3.10.15-1"
+		c.MasterControllersImage = "openshift3/ose-control-plane:v3.10.15-1"
+		c.NodeImage = "openshift3/ose-node:v3.10.15-1"
+		c.ServiceCatalogImage = "openshift3/ose-service-catalog:v3.10.15-1"
+		c.TemplateServiceBrokerImage = "openshift3/ose-template-service-broker:v3.10.15-1"
+		c.PrometheusNodeExporterImage = "openshift3/prometheus-node-exporter:v3.10.15-1"
+		c.RegistryImage = "openshift3/ose-docker-registry:v3.10.15-1"
+		c.RouterImage = "openshift3/ose-haproxy-router:v3.10.15-1"
+		c.RegistryConsoleImage = "openshift3/registry-console:v3.10.15-1"
+		c.AnsibleServiceBrokerImage = "openshift3/ose-ansible-service-broker:v3.10.15-1"
+		c.WebConsoleImage = "openshift3/ose-web-console:v3.10.15-1"
+		c.OAuthProxyImage = "openshift3/oauth-proxy:v3.10.15-1"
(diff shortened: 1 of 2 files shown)


pkg/config/generate.go
```diff
@@ -59,18 +59,46 @@ func selectContainerImagesOrigin(cs *acsapi.ContainerService, c *Config) {
...
+
+}
+
 func selectContainerImages(cs *acsapi.ContainerService, c *Config) {
 	switch os.Getenv("DEPLOY_OS") {
 	case "":
-		// TODO: selectContainerImagesOSA(m, c)
+		c.ImageConfigFormat = "openshift3/ose-${component}:${version}"
+		selectContainerImagesOSA(cs, c)
 	case "centos7":
```
pweil-: let's have a default case with images or an error, a wrong setting will yield bad results here.

pkg/config/generate.go
```diff
@@ -59,18 +59,46 @@ func selectContainerImagesOrigin(cs *acsapi.ContainerService, c *Config) {
 	}
 }
 
+func selectContainerImagesOSA(cs *acsapi.ContainerService, c *Config) {
+	switch cs.Properties.OrchestratorProfile.OpenShiftConfig.OpenShiftVersion {
+	//TODO: confirm minor version after release
+	case "3.10", "3.10.15":
```
pweil-: this seems like it's going to a) get pretty repetitive and b) need a lot of updates.  What can we do here to make it easier to maintain?  Suggestions:

* semver comparison
* convention based on version for majority of images (not sure how possible this is)
* slice of images that take a coordinated version number (all the 3.10.15-1) and a helper that sets all of them that takes just a version

combination of the above, something completely different?

pweil-: Not a blocker since we have a data set of 1 so far :smile: 

Probably want a panic or error if we're not setting images though so we don't move along

mjudeikis: :+1:

mjudeikis: Question is how granular this will be. We would need to do some defaulting. Let's say a customer chose `3.10.` and the newest one is `3.10.15`, so we default to this. And we just got new image set (`3.10.20`). If we are running some arbitrary helm code (change secret permissions or deployment parameters) which does not involve image change we will need to make sure that we are running with `3.10.15` metadata for that run. Because I assume we want to control when we do minor patches upgrades so we would not have unplanned updated. 

Overall this will comes to decide if we want to be able to do deployment config/helm changes and upgrade changes as a separate action.

pkg/config/generate.go
mjudeikis: this is covered in `validate/validate.go` where it gives error if either `DEPLOY_OS` or `openShiftVersion` is set to non-supported or wrong value
```
+ go run cmd/createorupdate/createorupdate.go
panic: invalid openShiftVersion "3.9"
```

pkg/config/generate.go
pweil-: > openShiftVersion is a user-facing option and I think we shouldn't allow users to choose patch versions (ie. we should always update them to latest).

Seems like we need to support n versions even if the user cannot pick it so that we can perform upgrades correctly.  Limiting what the user can choose seems separate than the image selection.

mjudeikis: agreed. my biggest current `unknown` is how we make sure we dont do unintentional upgrades (even minor) when running sync container for lets say configuration change and not upgrade. This means sync will have to be aware of *current running version* and what we are asking it to do. Or im overthinking this one?

pweil-: sync is reading both the manifest and existing config.yaml.  The config.yaml has the image configuration already there.  Sync isn't running through the Generate portions.

kargakis: > Limiting what the user can choose seems separate than the image selection.

Agreed, my comment was specifically about handling openshiftVersion which should never contain a patch version.

mjudeikis: changed :)

kargakis:
If this deploys fine then lgtm

PR #70, opened 2018-07-12
Use centos as default until rhel is working

kargakis:
@mjudeikis @pweil- thoughts?


mjudeikis:
im in the middle of publishing rhel7 but from my own laptop (as there is no way yet how to inject secret for aws registry). Buts its a hack, not a solution. Give me 1-2 hours to test and lets see what we have...

mjudeikis:
on other hand, control plain will not be able to access aws registry. https://github.com/jim-minter/azure-helm/issues/69
lets merge this one :/

PR #72, opened 2018-07-12
[WIP] add OSA images and secret to test it

mjudeikis:
This adds most of the images for RHEL and Centos for 3.10 and 3.11.
Adds ability to inject pull secret to HCP layer so we could test it with AKS. 

Still work in progress but would like to get feedback on image versioning handling (@pweil- now we have more datasets than 1)

cc: @jim-minter @kargakis 

RHELimages still cant be built in CI but there is one image I published today in `images` which works.


mjudeikis:
cc: @kargakis @jim-minter @pweil- 

This brings 2 things:
1. Ability to inject dev pull secret to our HCP so we can test with a private registry (nodes will either have temporary secret baked or we will have a dev build with the permanent secret for non-publishing (?) 
2. Introduces a generic list of images in `generate.go` like `imagesOrigin`.
And for now uses resolver function `resolveImageOrigin("WebConsoleImage", defaultTag)`
If we want to have default case tag we just pass single tag into it. Else we can pass part of the image we overwriting (`name:tag`,`namespace/name:tag`,`registry/namespace/name:tag`). Ideally, when we go GA with version all images should point to `resolveImageOrigin("WebConsoleImage", defaultTag)` fr that particual release. 

RHEL ones does not have that because there is no point of calling resolver if you specify full image.

We spoke briefly about small helper, so the question is do we want just maintain multiple lists with tags or something like this here?

env.example
```diff
@@ -9,6 +9,12 @@ export DNS_RESOURCEGROUP=dns
 # Set the OpenShift version (3.10) (mandatory)
 export DEPLOY_VERSION=3.10
 
+# Set DEPLOYMENT_TYPE to dev to configure HCL to use image pull secret for private development images 
+export DEPLOYMENT_TYPE=dev
```
kargakis: Seems redundant - you can just use `DEV_PULL_SECRET` I think.

pkg/config/generate.go
```diff
@@ -18,6 +19,49 @@ import (
 	"github.com/jim-minter/azure-helm/pkg/tls"
 )
 
+// ORIGIN Images
```
jim-minter: this comment makes me think that a simplifying refactor is needed on the code, not an additional comment.

pkg/config/generate.go
```diff
@@ -35,68 +79,128 @@ func selectNodeImage(cs *acsapi.ContainerService, c *Config) {
 }
 
 func selectContainerImagesOrigin(cs *acsapi.ContainerService, c *Config) {
+	// TODO:
+	// Publish tunnel, sync, images
+	c.ImageConfigFormat = "openshift/origin-${component}:${version}"
 	switch cs.Properties.OrchestratorProfile.OpenShiftConfig.OpenShiftVersion {
+	case "3.11":
```
jim-minter: can we share the similar pieces between 3.10 and 3.11? e.g. `c.NodeImage = "docker.io/openshift/origin-node:v" + cs.Properties.OrchestratorProfile.OpenShiftConfig.OpenShiftVersion`

pkg/config/generate.go
```diff
@@ -35,68 +79,128 @@ func selectNodeImage(cs *acsapi.ContainerService, c *Config) {
...
 		c.ServiceCatalogImage = "docker.io/openshift/origin-service-catalog:v3.10"
+		c.TemplateServiceBrokerImage = "docker.io/openshift/origin-template-service-broker:v3.10"
+		c.PrometheusNodeExporterImage = "docker.io/openshift/prometheus-node-exporter:v0.15.2"
+		c.RegistryImage = "docker.io/openshift/origin-docker-registry:v3.10"
+		c.RouterImage = "docker.io/openshift/origin-haproxy-router:v3.10"
+		c.RegistryConsoleImage = "docker.io/cockpit/kubernetes:latest"
+		c.AnsibleServiceBrokerImage = "docker.io/ansibleplaybookbundle/origin-ansible-service-broker:latest"
+		c.WebConsoleImage = "docker.io/openshift/origin-web-console:v3.10"
+		c.OAuthProxyImage = "docker.io/openshift/oauth-proxy:v3.10"
+		c.PrometheusImage = "docker.io/openshift/prometheusv2.2.1"
```
jim-minter: looks wrong (missing `:`)

pkg/helm/chart/templates/Deployment.apps/bootstrap-autoapprover.yaml
```diff
@@ -71,6 +71,10 @@ spec:
         - name: kubeconfig
           mountPath: /tmp/kubeconfig
           readOnly: true
+{{if ne "" .Values.ContainerPullSecret }}
```
jim-minter: would it be better to do this via https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account ? I suspect it might be?

pkg/config/generate.go
```diff
@@ -35,68 +79,128 @@ func selectNodeImage(cs *acsapi.ContainerService, c *Config) {
...
-		c.PrometheusImage = "openshift3/prometheus:v3.10.15-1"
-		c.PrometheusAlertBufferImage = "openshift3/prometheus-alert-buffer:v3.10.15-1"
-		c.PrometheusAlertManagerImage = "openshift3/prometheus-alertmanager:v3.10.15-1"
-		c.TunnelImage = "docker.io/jimminter/tunnel:latest"      //TODO: We need to publish it somewhere.
-		c.SyncImage = "docker.io/jimminter/sync:latest"          //TODO: We need to publish it somewhere.
-		c.AzureCLIImage = "docker.io/microsoft/azure-cli:2.0.41" //TODO: create mapping for OSA release to any other image we use
+		c.ImageConfigFormat = "registry.reg-aws.openshift.com/openshift3/ose-${component}:${version}"
+		c.ContainerPullSecret = configureDevRegPullSecret("registry.reg-aws.openshift.com")
+
+		c.MasterEtcdImage = "registry.reg-aws.openshift.com/rhel3/etcd:3.2.22"
```
jim-minter: For sanity's sake/auditability I think it's going to make sense to somehow pull out strings like `registry.reg-aws.openshift.com/openshift3` here to avoid repetition

pkg/config/generate.go
mjudeikis: previously I had a helper which return pattern based what you pass into it:
```
registry/namespace/name:tag
namespace/name:tag
name:tag
tag
``` is it something you think about or what?

Comment (mjudeikis) [reply pkg/config/generate.go]: Or something like `fmt.Springf("%s/%s/%s", defaultRegistry,defaultNamespace, image:tag)` ?

File: pkg/config/generate.go (hunk 2 above)
Comment (mjudeikis) [reply pkg/config/generate.go]: I was thinking to keep this as a reference in the comments as list here is far from what we use. I would keep the comments and refactor code :)

File: pkg/helm/chart/templates/Deployment.apps/bootstrap-autoapprover.yaml (hunk 5 above)
Comment (mjudeikis) [reply pkg/helm/chart/templates/Deployment.apps/bootstrap-autoapprover.yaml]: We run all HCP via default service account. This means secret adding will be `kubectl patch` command in create.sh script. If this is sufficient (considering we will do this in dev only), yes its easier. Otherwise, we either need to move to other SA and create it as part of helm install.

File: pkg/config/generate.go (hunk 3 above)
Comment (mjudeikis) [reply pkg/config/generate.go]: Tag is not always will be version. There might be (and will be for sure) when we need to rollback one particular image from the list to the previous version.
I would better do:
```
dt:="v"+cs.Properties.OrchestratorProfile.OpenShiftConfig.OpenShiftVersion
dn:= "openshift3"
dr:="reg.redhat.com"
c.NodeImage = fmt.Sprintf("%s/%s/%s:%s", dr, dn, "node",dt)
// and we swap values where we need to from defaults? 
c.WebConsoleImage = fmt.Sprintf("%s/%s/%s:%s", dr, dn, "console","v3.11.0")
```
?

pkg/config/generate.go
kargakis: Yes, eventaually we will need this for PR testing too

mjudeikis:
as per chat with @kargakis, for now, I will be removing image naming changes and leaving only secret ones. Just need to agree on an approach for pull secret injection.

jim-minter:
discussed via IRC - kubectl patch from somewhere in create.sh (perhaps a short separate function) to add the secret into the default SA so that ContainerPullSecret doesn't go into config.go

PR #77, opened 2018-07-16
clean 

mjudeikis:
remove hardcode cookies value
remove deprecated settings


pkg/config/generate.go
```diff
@@ -499,10 +505,10 @@ func randomBytes(n int) ([]byte, error) {
 	return b, nil
 }
 
-func randomStorageAccountName() (string, error) {
+func randomString(lenght int) (string, error) {
```
kargakis: rollback this change and move your secret to be genereted as part of the secrets loop (L332). Also change your secret type to a byte slice.

mjudeikis:
TODO:  replace this with generated value in router service 
```
    prometheus.openshift.io/password: 75XmZa2KNV
    prometheus.openshift.io/username: admin
```

mjudeikis:
@kargakis @jim-minter @kwoodson ptal

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -16,66 +16,6 @@ data:
     # Kubernetes labels will be added as Prometheus labels on metrics via the
     # `labelmap` relabeling action.
 
-    # Scrape config for API servers.
```
kargakis: Why is this removed?

mjudeikis: Do we want to expose their api server and controller metrics to users? As they dont see them anywhere should they be concern on metrics of them? "metrics from the other side" ?

pkg/helm/data/values.yaml
```diff
@@ -80,6 +80,8 @@ Location: {{ .ContainerService.Location | quote }}
 RegistryStorageAccount: {{ .Config.RegistryStorageAccount | quote }}
 RegistryHTTPSecret: {{ Base64Encode .Config.RegistryHTTPSecret | quote }}
 AlertManagerProxySessionSecret: {{ Base64Encode .Config.AlertManagerProxySessionSecret | quote }}
+RegistryConsoleOAuthSecret: {{ Base64Encode .Config.RegistryConsoleOAuthSecret }}
```
kargakis: You need to be setting these in ` pkg/helm/chart/templates/Secret/sync.yaml`, too?

mjudeikis: I think so. I need to go one more round via all addon and sync process itself.

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
kargakis: We need to be gathering control plane metrics and the prometheus installed by us is for cluster monitoring. I think @pweil- and @jim-minter felt users need to deploy their own prometheus server for their own apps but I still haven't seen multitenant prom deployments to know how that will look like.

mjudeikis: I was under impression that this Prometheus will be "user one". For our control plain, we might need to have something like https://github.com/weaveworks/cortex . We need metrics, but this goes to the same conversation pool, "where to store etcd backups" kinda talks.

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -24,57 +24,6 @@ data:
     # default named port `https`. This works for single API server deployments as
     # well as HA API server deployments.
     scrape_configs:
-    - job_name: 'kubernetes-apiservers'
```
jim-minter: This should work.  I'm guessing it doesn't?  If so can you find out approximately why?

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -24,57 +24,6 @@ data:
...
-
-      # Keep only the default/kubernetes service endpoints for the https port. This
-      # will add targets for each API server which Kubernetes adds an endpoint to
-      # the default/kubernetes service.
-      relabel_configs:
-      - source_labels: [__meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
-        action: keep
-        regex: kubernetes;https
-
-    # Scrape config for controllers.
```
jim-minter: Comment this out and open an issue to re-enable

pkg/config/generate.go
```diff
@@ -319,6 +319,14 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 		{
 			secret: &c.RegistryHTTPSecret,
 		},
+		{
+			secret: &c.RegistryConsoleOAuthSecret,
+			n:      68,
+		},
+		{
+			secret: &c.RouterStatsPassword,
```
jim-minter: upstream sets [a-zA-Z0-9]{10}, then no base64 encode
https://github.com/openshift/origin/blob/master/pkg/oc/cli/admin/router/router.go#L1083-L1093

pkg/config/generate.go
```diff
@@ -319,6 +319,14 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 		{
 			secret: &c.RegistryHTTPSecret,
 		},
+		{
+			secret: &c.RegistryConsoleOAuthSecret,
```
jim-minter: upstream sets user[a-zA-Z0-9]{64}, then no base64 encode
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted_templates/files/v3.10/origin/registry-console.yaml#L109

pkg/config/generate.go
mjudeikis: ok, will revert as this was initial implementation.

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
mjudeikis: Do we want to give access to HCP metrics for clients? As this would be client side Prometheus scraping HCP component?

mjudeikis: @jim-minter this does not work due certificate. By default endpoints point to local address. In our case EP points to IP `https://10.244.x.xxx:443/metrics`

And certificate is valid only for:
`Get https://10.244.x.xxx:443/metrics: x509: certificate is valid for 172.30.0.1, not 10.244.x.xxx` 

`kubernetes         10.244.x.xxx:8053,10.244.x.xxx:8053,10.244.x.xxx:443   28m`

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -35,6 +35,8 @@ data:
       scheme: https
       tls_config:
         ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+        # TODO: Fix me
```
mjudeikis: Copy-paste from previous commit comment

@jim-minter this does not work due certificate. By default, endpoints point to local address. In our case EP points to IP https://10.244.x.xxx:443/metrics

And certificate is valid only for:
Get https://10.244.x.xxx:443/metrics: x509: certificate is valid for 172.30.0.1, not 10.244.x.xxx

kubernetes 10.244.x.xxx:8053,10.244.x.xxx:8053,10.244.x.xxx:443 28m

pkg/config/generate.go
```diff
@@ -341,6 +341,36 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
...
+			l:      68,
+		},
+		{
+			secret: &c.RouterStatsPassword,
+			l:      10,
+		},
+		{
+			secret: &c.RegistryStorageAccount,
```
mjudeikis: this is not a secret but uses the same pattern. tempted to rename all this to `strConfigurables` or smth like that.

pkg/config/generate.go
```diff
@@ -510,10 +534,10 @@ func randomBytes(n int) ([]byte, error) {
 	return b, nil
 }
 
-func randomStorageAccountName() (string, error) {
-	const letterBytes = "abcdefghijklmnopqrstuvwxyz0123456789"
+func randomString(lenght int) (string, error) {
```
mjudeikis: reluctant to keep code which does 90% the same. If a storage account name can live with capital letters in the name, this simplifies things a little bit.

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
jim-minter: I've been here before with the template service broker (see a long way down in this file).  Does setting tls_config: server_name: kubernetes help here instead of the insecure_skip_verify?

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -50,32 +52,32 @@ data:
...
+    #  relabel_configs:
+    #  - source_labels: [__meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
+    #    action: keep
+    #    regex: kubernetes;https
+    #  - source_labels: [__address__]
+    #    action: replace
+    #    target_label: __address__
+    #    regex: (.+)(?::\d+)
+    #    replacement: $1:8444
+    #
```
jim-minter: don't forget to open an issue to fix this

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -247,7 +249,8 @@ data:
 
     # Scrape config for the router
     - job_name: 'openshift-router'
-      scheme: https
+      #TODO: Fix me upstream (http->https)
```
jim-minter: open an issue here, refer to https://github.com/openshift/origin/issues/20375

pkg/config/generate.go
jim-minter: they're all subtly different - don't try to combine them.  storage accts must be [a-z0-9]{24} according to https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#storage , for now stick with upstream for the registry console secret (user[a-zA-Z0-9]{64}), etc.

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
mjudeikis: will do when squashing :)

jim-minter: we tend to open the issue then reference it here in the review, that way it doesn't get dropped

mjudeikis: https://github.com/jim-minter/azure-helm/issues/100

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
mjudeikis: https://github.com/jim-minter/azure-helm/issues/101

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -258,11 +260,11 @@ data:
         namespaces:
           names:
           - default
-
+
```
jim-minter: @mjudeikis your text editor should be configured not to do this.

jim-minter:
lgtm!

PR #78, opened 2018-07-17
[wip] move to LoadBalancer for Routers

mjudeikis:
This was jak shaving. 
Was doing something else but this was underway to make it happen fast, so I did.

cc: @jim-minter , @kargakis, @pweil-


mjudeikis:
test pending

mjudeikis:
`router             172.30.176.44   <pending>     80:31122/TCP,443:32428/TCP,1936:30245/TCP   2m`

Node ports needs to be fixed

jim-minter:
Shouldn't this be LoadBalancer and not ExternalIP?

mjudeikis:
Now, where is 2 doubts I have about this fo would be good to discuss implementation details:
1. IP address being assigned now is ephemeral. If Service gets deleted/recreated for some reason - we will need to update wildcard DNS. This goes to the second point
We cloud for maintenance reasons create services:
```
router - the default one
router-external - for current implementation 
router-internal - for point 3
```
So we just ignore `router-external` and `router-internal` in sync and manage them in some other process (recreate and update DNS flow)

2. Service with type: LoadBalancer becomes immutable and sync process (the current version of it as Im building on not yet refactored code) is not handling this well:
```

```
3. For future work, and BYO VNET we might want to introduce `azure-internal-loadbalancer`. This will enable users to get traffic straight to the router service  (separate Jira/trello ticket) 
https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/aks/internal-lb.md

cc: @jim-minter @kargakis  @pweil-  @kwoodson

pkg/addons/data/Deployment.apps/default/router.yaml
```diff
@@ -62,26 +52,23 @@ spec:
         imagePullPolicy: Always
         livenessProbe:
           httpGet:
-            host: localhost
             path: /healthz
             port: 1936
             scheme: HTTP
           initialDelaySeconds: 10
+        readinessProbe:
```
jim-minter: might as well move this back down because it'll just move itself when we next run tools/update-addons.go.

create.sh
```diff
@@ -93,7 +93,13 @@ KUBECONFIG=_data/_out/admin.kubeconfig go run cmd/sync/sync.go
 
 az group deployment wait -g $RESOURCEGROUP -n azuredeploy --created --interval 10
 
-ROUTERIP=$(az network public-ip list -g $RESOURCEGROUP --query "[?name == 'ip-router'].ipAddress | [0]" | tr -d '"')
+while true; do
```
jim-minter: wildcard cname approach should allow simplification

pkg/addons/data/Service/default/router.yaml
```diff
@@ -20,3 +20,4 @@ spec:
     port: 1936
```
jim-minter: 1936 should probably not be on the external service - separate service + update prometheus?

pkg/addons/data/Service/default/router.yaml
```diff
@@ -20,3 +20,4 @@ spec:
...
   selector:
     router: router
+  type: LoadBalancer
```
jim-minter: should service.beta.kubernetes.io/azure-load-balancer-mode be specified?

jim-minter:
On 1, I think you may be able to use the `service.beta.kubernetes.io/azure-dns-label-name: foo` annotation to specify a dns label associated with the external IP, e.g. specify foo and it gives you foo.eastus.cloudapp.azure.com.  Then you can use a wildcard CNAME against foo.eastus.cloudapp.azure.com.
On 2, I'm sure you can fix that.  I guess you'll need a small special case in `func write()` that detects this issue (Service with LoadBalancer type) and copies across the relevant key(s) (e.g. spec.clusterIP) from `existing` to `o`.
On 3, not a problem for now, but in the future see `service.beta.kubernetes.io/azure-load-balancer-internal`.

pkg/addons/data/Service/default/router.yaml
mjudeikis: So make it 2 services `router-external` and `router` ?

pkg/addons/data/Service/default/router.yaml
mjudeikis: tbc

pkg/addons/data/Service/default/router.yaml
jim-minter: router and router-stats?

mjudeikis:
> On 1, I think you may be able to use the service.beta.kubernetes.io/azure-dns-label-name: foo annotation to specify a dns label associated with the external IP, e.g. specify foo and it gives you foo.eastus.cloudapp.azure.com. Then you can use a wildcard CNAME against foo.eastus.cloudapp.azure.com.

Will check

> On 2, I'm sure you can fix that. I guess you'll need a small special case in func write() that detects this issue (Service with LoadBalancer type) and copies across the relevant key(s) (e.g. spec.clusterIP) from existing to o.

Was waiting for sync plugin to get merged to do this :)

> On 3, not a problem for now, but in the future see service.beta.kubernetes.io/azure-load-balancer-internal.

This was the intention too. Will raise and issue and later migrate to jira story

pkg/addons/data/ConfigMap/openshift-metrics/prometheus.yaml
```diff
@@ -247,10 +247,10 @@ data:
 
     # Scrape config for the router
     - job_name: 'openshift-router'
-      scheme: https
+      scheme: http
```
mjudeikis: This is part of clean PR, will rebase one of those later.

pkg/config/generate.go
```diff
@@ -101,6 +102,8 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 	c.Version = versionLatest
 	c.TunnelHostname = strings.Replace(cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname, "openshift", "openshift-tunnel", 1)
 
+	// TODO: Need unique name, potentially derivative from PublicHostname
+	c.RouterLBCName = fmt.Sprintf("router-%s", strings.Split(cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname, ".")[1])
```
mjudeikis: We need unique name here for `name.eastus.cloudapp.azure.com` . I imagine as past of PP or GA this should be MS responsibility extend their plugin or make zone configurable via annotations for cases like OSA. Or we can generate some short unique string like `router-[random]-osa.eastus.cloudapp.azure.com` based on the time of creation and some unique metadata (?)

mjudeikis: Issue for this to track?

pkg/addons/clean.go
```diff
@@ -147,3 +147,16 @@ func Clean(o unstructured.Unstructured) error {
 
 	return nil
 }
+
+// HandleSpecialObjects manages special object migration during upgrade state
+func HandleSpecialObjects(existing, o unstructured.Unstructured) {
```
jim-minter: lower case h on handleSpecialObjects (not exported, right?)

pkg/config/generate.go
jim-minter: I think this is fine for now.

PR #82, opened 2018-07-18
Set correctly image reference fields in ARM template

kargakis:


kargakis:
@mjudeikis @jim-minter fyi

PR #86, opened 2018-07-18
dev secret inject

mjudeikis:
This add one variables to the project:

`DEV_PULL_SECRET` to be set to private registry token for HCP layer to pull images.


env.example
```diff
@@ -9,6 +9,12 @@ export DNS_RESOURCEGROUP=dns
 # Set the OpenShift version (3.10) (mandatory)
 export DEPLOY_VERSION=3.10
 
+# Set DEV_PULL_SECRET when running pre-release/dev build of RHEL
+# You can get token string from docker login and cat ~/.docker/config.json
+#export DEV_PULL_SECRET=xxxxxxxxx
+# Set OREG_URL to overwrite registry and namespace. Used for testing.
+export OREG_URL='registry.reg-aws.openshift.com/openshift3/ose-${component}:${version}'
```
kargakis: I would avoid exporting this by default and would use it only on test jobs that need to export it (rhel image builds, pr tests).

mjudeikis: ack.

mjudeikis: ack.

tools/dev.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -ex
```
jim-minter: best not to use -x here

create.sh
```diff
@@ -76,6 +76,11 @@ az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _dat
 
 KUBECONFIG=aks/admin.kubeconfig helm install --namespace $RESOURCEGROUP pkg/helm/chart -f _data/_out/values.yaml -n $RESOURCEGROUP >/dev/null
 
+# Configure dev settings
```
jim-minter: - bring this above helm install - if dev.sh has to create the namespace to do so, that's fine
- delete the comment or make it meaningful
- rename dev.sh to something meaningful

tools/dev.sh
```diff
@@ -0,0 +1,16 @@
...
+    exit 1
+fi
+
+# Create docker dockerconfigjson auth string
+AUTH=$(echo -n "notused:$DEV_PULL_SECRET" | base64 -w 0)
+# Create k8s secret string
+SECRET=$(echo -n "{\"auths\":{\"registry.reg-aws.openshift.com\":{\"auth\":\"${AUTH}\"}}}" | base64 -w 0 )
+export KUBECONFIG=aks/admin.kubeconfig
+# Create secret in HCP
+sed "s/SECRET_PACEHOLDER/$SECRET/g" $(dirname $0)/secret.yaml | kubectl apply -f - -n $RESOURCEGROUP
```
jim-minter: - suggest making registry.reg-aws.openshift.com configurable
- choose an appropriate secret name
- use a here doc
```
kubectl apply -n $RESOURCEGROUP -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: devreg
stringData:
  .dockerconfigjson: '{"auths":{"registry.reg-aws.openshift.com":{"auth":"$AUTH"}}}'
type: kubernetes.io/dockerconfigjson
EOF
```

env.example
```diff
@@ -9,6 +9,10 @@ export DNS_RESOURCEGROUP=dns
 # Set the OpenShift version (3.10) (mandatory)
 export DEPLOY_VERSION=3.10
 
+# Set DEV_PULL_SECRET when running pre-release/dev build of RHEL
+# You can get token string from docker login and cat ~/.docker/config.json
```
jim-minter: is that true? isn't it just the password part?

mjudeikis:
@jim-minter post review changes done

PR #88, opened 2018-07-18
add ability to change and inject oreg_url and resolve multiple images

mjudeikis:
Image resolve part is needed for HCP layer to work with private and testing images. This is because AKS is ignoring oreg_url as such.

This works fine with RHEL builds, but I have doubts about CI testing:
when `OREG_URL=registry.svc.ci.openshift.org/ci-pr-images/prtest-89730d6-17-${component}:${version}`
Images get these names. I need to add logic to modify image names for this too, but I don't think we required all images and all of them will be available with the same name. For CI we might need to modify just subset of images.

```
MasterEtcdImage: "registry.svc.ci.openshift.org/ci-pr-images/etcd:v3.10"
MasterAPIImage: "registry.svc.ci.openshift.org/ci-pr-images/ose-control-plane:v3.10"
MasterControllersImage: "registry.svc.ci.openshift.org/ci-pr-images/ose-control-plane:v3.10"
NodeImage: "registry.svc.ci.openshift.org/ci-pr-images/ose-node:v3.10"
```

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"fmt"
 	"io"
 	"math/big"
 	"net"
@@ -17,6 +18,8 @@ import (
 
 	"github.com/jim-minter/azure-helm/pkg/tls"
 )
+
+const imageFormat = "%s/%s/%s:%s"
 
 func selectNodeImage(cs *acsapi.ContainerService, c *Config) {
 	c.ImagePublisher = "redhat"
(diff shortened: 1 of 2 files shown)


pkg/config/generate.go
```diff
@@ -37,48 +40,50 @@ func selectNodeImage(cs *acsapi.ContainerService, c *Config) {
...
-		c.RouterImage = "openshift3/ose-haproxy-router:v3.10.15-1"
-		c.RegistryConsoleImage = "openshift3/registry-console:v3.10.15-1"
-		c.AnsibleServiceBrokerImage = "openshift3/ose-ansible-service-broker:v3.10.15-1"
-		c.WebConsoleImage = "openshift3/ose-web-console:v3.10.15-1"
-		c.OAuthProxyImage = "openshift3/oauth-proxy:v3.10.15-1"
-		c.PrometheusImage = "openshift3/prometheus:v3.10.15-1"
-		c.PrometheusAlertBufferImage = "openshift3/prometheus-alert-buffer:v3.10.15-1"
-		c.PrometheusAlertManagerImage = "openshift3/prometheus-alertmanager:v3.10.15-1"
+		t := "v3.10"
+		c.MasterEtcdImage = resolveImage("etcd", t)
```
mjudeikis: /t/"3.2.22"/

jim-minter:
All of this is going in the right direction, but as you say it's too intrusive.  I've reworked it at #98.

PR #92, opened 2018-07-18
Enhance the healthcheck to provide healthz for api and check for pod status to be running. 

kwoodson:
Enhance the healthcheck to ensure the following:
1. The healthz check for the API returns a 200
2. The pods are in a Running state without restarts
3. The web console is available

@jim-minter, Please review

### DIFF
--- a/pkg/addons/client.go
+++ b/pkg/addons/client.go
@@ -1,23 +1,18 @@
 package addons
 
 import (
+	"context"
 	"errors"
-	"io"
 	"log"
-	"net"
-	"net/http"
-	"net/url"
-	"os"
 	"reflect"
-	"syscall"
-	"time"
+
+	"github.com/openshift/openshift-azure/pkg/checks"
 
 	"github.com/go-test/deep"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
@@ -69,7 +64,12 @@ func newClient(dryRun bool) (*client, error) {
 		cli:        cli,
 	}
 
-	if err := c.waitForHealthz(); err != nil {
+	transport, err := rest.TransportFor(c.restconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := checks.WaitForHTTPStatusOk(context.Background(), transport, c.restconfig.Host+"/healthz"); err != nil {
 		return nil, err
 	}
 
@@ -93,46 +93,6 @@ func (c *client) updateDynamicClient() error {
 	c.dyn = dynamic.NewClientPool(c.restconfig, rm, dynamic.LegacyAPIPathResolverFunc)
 
 	return nil
-}
-
-func (c *client) waitForHealthz() error {
-	if c.dryRun {
-		return nil
-	}
-
-	transport, err := rest.TransportFor(c.restconfig)
-	if err != nil {
-		return err
-	}
-
-	cli := &http.Client{
-		Transport: transport,
-		Timeout:   10 * time.Second,
-	}
-
-	req, err := http.NewRequest("GET", c.restconfig.Host+"/healthz", nil)
-	if err != nil {
-		return err
-	}
-
-	return wait.PollInfinite(time.Second, func() (bool, error) {
-		resp, err := cli.Do(req)
-
-		if err, ok := err.(*url.Error); ok {
-			if err, ok := err.Err.(*net.OpError); ok {
-				if err, ok := err.Err.(*os.SyscallError); ok {
-					if err.Err == syscall.ENETUNREACH {
-						return false, nil
-					}
-				}
-			}
-			if err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF {
-				return false, nil
-			}
-		}
-
-		return resp.StatusCode == http.StatusOK, err
-	})
 }
 
 // createResources creates all resources in db that match the provided filter.
--- a/pkg/checks/checks.go
+++ b/pkg/checks/checks.go
@@ -1,0 +1,167 @@
+package checks
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	appclient "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+// WaitForHTTPStatusOk poll until URL returns 200
+func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urltocheck string) error {
+	cli := &http.Client{
+		Transport: transport,
+		Timeout:   10 * time.Second,
+	}
+
+	req, err := http.NewRequest("GET", urltocheck, nil)
+	if err != nil {
+		return err
+	}
+	return wait.PollUntil(time.Second, func() (bool, error) {
+		resp, err := cli.Do(req)
+		if err, ok := err.(*url.Error); ok {
+			if err, ok := err.Err.(*net.OpError); ok {
+				if err, ok := err.Err.(*os.SyscallError); ok {
+					if err.Err == syscall.ENETUNREACH {
+						return false, nil
+					}
+				}
+			}
+			if err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF {
+				return false, nil
+			}
+		}
+		if err == io.EOF {
+			return false, nil
+		}
+		return resp.StatusCode == http.StatusOK, err
+	}, ctx.Done())
+}
+
+func checkNamespace(namespace string) bool {
+	if namespace == "default" {
+		return true
+	}
+
+	whitelistedNamespaces := []string{
+		"openshift-",
+		"kube-",
+	}
+	for _, ns := range whitelistedNamespaces {
+		if strings.HasPrefix(namespace, ns) {
+			return true
+		}
+	}
+	return false
+}
+
+// WaitForInfraServices verify daemonsets, statefulsets
+func WaitForInfraServices(ctx context.Context, appclient *appclient.AppsV1Client) error {
+
+	// Daemonsets
+	dsList, err := appclient.DaemonSets("").List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, ds := range dsList.Items {
+		if !checkNamespace(ds.Namespace) {
+			continue
+		}
+		for {
+			ds, err := appclient.DaemonSets(ds.Namespace).Get(ds.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if ds.Status.NumberMisscheduled > 0 {
+				return fmt.Errorf("Daemonset[%v] in Namespace[%v] has missscheduled[%v]", ds.Name, ds.Namespace, ds.Status.NumberMisscheduled)
+			}
+
+			if ds.Status.DesiredNumberScheduled == ds.Status.CurrentNumberScheduled &&
+				ds.Status.DesiredNumberScheduled == ds.Status.NumberReady &&
+				ds.Status.DesiredNumberScheduled == ds.Status.UpdatedNumberScheduled &&
+				ds.Generation == ds.Status.ObservedGeneration {
+				break
+			}
+			select {
+			case <-time.After(2 * time.Second):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+	}
+
+	// Statefulsets
+	ssList, err := appclient.StatefulSets("").List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, ss := range ssList.Items {
+		if !checkNamespace(ss.Namespace) {
+			continue
+		}
+		for {
+			ss, err := appclient.StatefulSets(ss.Namespace).Get(ss.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if ss.Status.Replicas == ss.Status.ReadyReplicas &&
+				ss.Status.ReadyReplicas == ss.Status.CurrentReplicas &&
+				ss.Spec.Replicas != nil &&
+				*ss.Spec.Replicas == ss.Status.Replicas &&
+				ss.Generation == ss.Status.ObservedGeneration {
+				break
+			}
+			select {
+			case <-time.After(2 * time.Second):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+	}
+
+	// Deployments
+	dList, err := appclient.Deployments("").List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, d := range dList.Items {
+		if !checkNamespace(d.Namespace) {
+			continue
+		}
+		for {
+			d, err := appclient.Deployments(d.Namespace).Get(d.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if d.Status.Replicas == d.Status.ReadyReplicas &&
+				d.Status.ReadyReplicas == d.Status.AvailableReplicas &&
+				d.Spec.Replicas != nil &&
+				*d.Spec.Replicas == d.Status.Replicas &&
+				d.Status.Replicas == d.Status.UpdatedReplicas &&
+				d.Generation == d.Status.ObservedGeneration {
+				break
+			}
+			select {
+			case <-time.After(2 * time.Second):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+	}
+
+	return nil
+}
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -10,11 +10,64 @@ import (
 	"time"
 
 	acsapi "github.com/Azure/acs-engine/pkg/api"
+	appsclient "k8s.io/client-go/kubernetes/typed/apps/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/tools/clientcmd/api/latest"
+	"k8s.io/client-go/tools/clientcmd/api/v1"
 
+	"github.com/openshift/openshift-azure/pkg/checks"
 	"github.com/openshift/openshift-azure/pkg/config"
 )
 
+// GetKubeconfigFromV1Config takes a v1 config and returns a kubeconfig
+func getKubeconfigFromV1Config(kc *v1.Config) (clientcmd.ClientConfig, error) {
+	var c api.Config
+	err := latest.Scheme.Convert(kc, &c, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	kubeconfig := clientcmd.NewDefaultClientConfig(c, &clientcmd.ConfigOverrides{})
+
+	return kubeconfig, nil
+}
+
+// HealthCheck function to verify cluster health
 func HealthCheck(ctx context.Context, cs *acsapi.ContainerService, c *config.Config) error {
+	kubeconfig, err := getKubeconfigFromV1Config(c.AdminKubeconfig)
+	if err != nil {
+		return err
+	}
+
+	restconfig, err := kubeconfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	t, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return err
+	}
+
+	// Wait for the healthz to be 200 status
+	err = checks.WaitForHTTPStatusOk(ctx, t, restconfig.Host+"/healthz")
+	if err != nil {
+		return err
+	}
+
+	appsclient, err := appsclient.NewForConfig(restconfig)
+	if err != nil {
+		return err
+	}
+	// Ensure that the pods in default are healthy
+	err = checks.WaitForInfraServices(ctx, appsclient)
+	if err != nil {
+		return err
+	}
+
+	// Wait for the console to be 200 status
 	return waitForConsole(ctx, cs, c)
 }
 


pkg/util/util.go
```diff
@@ -43,3 +49,16 @@ func Template(tmpl string, f template.FuncMap, cs *acsapi.ContainerService, c *c
 
 	return b.Bytes(), nil
 }
+
+// GetKubeconfigFromV1Config takes a v1 config and returns a kubeconfig
+func GetKubeconfigFromV1Config(kc *v1.Config) (clientcmd.ClientConfig, error) {
```
kwoodson: As per the comment above regarding the anti-pattern, should this function move into the `healthcheck.go`?  Seems like a useful function when you don't have a kubeconfig available.

pkg/checks/checks.go
```diff
@@ -0,0 +1,81 @@
...
+		if resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		time.Sleep(time.Second)
+	}
+}
+
+// CheckStatusOfPods verifies all pods are in running state
+func CheckStatusOfPods(coreclient *coreclient.CoreV1Client, namespace string) error {
```
jim-minter: use a context here and in WaitForHTTPStatusOk and return if the context has expired.

Caller:
```
ctx, cancel := context.WithTimeout(ctx.Background(), 5 * time.Minute)
defer cancel()
f(ctx, ...)
```

Callee:
```
func f(ctx context.Context, ...) {
	...
	for {
		...
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
		}
	}
}
```

jim-minter: (then no need for startTime, hard-coded timeout, etc)

jim-minter: How about `func WaitForPodRunning(coreclient *coreclient.CoreV1Client, namespace, name string)` and have a list of pods you want to check for now, a bit like in https://github.com/Azure/acs-engine/blob/master/test/e2e/openshift/openshift_test.go#L145-L173

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,54 @@ import (
...
+		return err
+	}
+
+	// Ensure that the pods in default are healthy
+	namespace, _, err := kubeconfig.Namespace()
+	if err != nil {
+		return err
+	}
+
+	err = checks.CheckStatusOfPods(coreclient, namespace)
```
jim-minter: this is where you need to iterate a list of pods to check

pkg/addons/addons.go
```diff
@@ -287,47 +285,17 @@ func getClients() (err error) {
...
 		if err != nil {
 			return err
 		}
-
-		err = waitForHealthz()
+		transport, err := rest.TransportFor(restconfig)
+		if err != nil {
+			return err
+		}
+		err = checks.WaitForHTTPStatusOk(transport, restconfig.Host+"/heathz")
```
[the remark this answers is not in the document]
kwoodson: fixed.

kwoodson:
The health check portion is working.  Seeing a panic when attempting to deploy inside of the sync.go:
```
+ go run cmd/sync/sync.go
panic: Get https://openshift.kwoodsonhcp1.osadev.cloud/api: EOF

goroutine 1 [running]:
main.main()
	/home/kwoodson/go/src/github.com/jim-minter/azure-helm/cmd/sync/sync.go:55 +0x49
exit status 2
```
@jim-minter, any ideas?

pkg/addons/addons.go
```diff
@@ -220,12 +223,20 @@ func writeDB(client *client, db map[string]unstructured.Unstructured) error {
...
 		return err
 	}
 
-	if err := client.waitForHealthz(); err != nil {
+	transport, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
jim-minter: just use context.Background() here for now, no timeout

pkg/addons/client.go
```diff
@@ -36,27 +32,27 @@ type client struct {
 	dryRun bool
 }
 
-func newClient(dryRun bool) (*client, error) {
+func newClient(dryRun bool) (*client, *rest.Config, error) {
```
jim-minter: restconfig is already a field in the returned client object, are these changes necessary?

pkg/checks/checks.go
```diff
@@ -0,0 +1,106 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(time.Second):
+		}
+	}
+}
+
+// WaitForPodsRunning verifies all pods are in running state
+func WaitForPodsRunning(ctx context.Context, coreclient *coreclient.CoreV1Client, namespace string) error {
```
jim-minter: namespace argument unnecessary

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,70 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	t, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return err
+	}
+
+	ctxHealthz, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
jim-minter: use the passed in context.  The point of context is the caller gets to decide how long the total timeout is.

pkg/checks/checks.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+
+	for {
+		resp, err := cli.Do(req)
+		if err, ok := err.(*url.Error); ok && (err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF) {
+			time.Sleep(time.Second)
+			continue
+		}
+		if err != nil {
```
jim-minter: w.r.t. your error, try adding here:
```
if err == io.EOF {
	time.Sleep(time.Second)
	continue
}
```

jim-minter: if that doesn't help add `log.Printf("the error was a %#v\n", err)` to work out what the error actually was.

pkg/checks/checks.go
```diff
@@ -0,0 +1,106 @@
...
+// WaitForPodsRunning verifies all pods are in running state
+func WaitForPodsRunning(ctx context.Context, coreclient *coreclient.CoreV1Client, namespace string) error {
+
+	whitelistedNamespaces := []string{
+		"openshift-",
+		"kube-",
+		"default",
+	}
+
+	pods, err := coreclient.Pods(namespace).List(metav1.ListOptions{})
```
jim-minter: we didn't get a chance to catch up today, sorry. I think you had some questions about this.  I'm expecting to see you looping through statefulsets, daemonsets, etc. using the apps client and not the core client.
you could use the clientset at https://github.com/jim-minter/azure-helm/blob/master/vendor/k8s.io/client-go/kubernetes/clientset.go#L399, that gives you nearly every k8s client there is.  Good to help developing.
For the final thing you're better off using more scoped clients (easier for testing).  In this case, https://github.com/jim-minter/azure-helm/blob/master/vendor/k8s.io/client-go/kubernetes/typed/apps/v1/apps_client.go#L63

pkg/addons/addons.go
kwoodson: done.

pkg/addons/client.go
kwoodson: I didn't realize that.  removing changes.

pkg/checks/checks.go
kwoodson: ok.  I will remove.

pkg/healthcheck/healthcheck.go
kwoodson: Ok.

pkg/checks/checks.go
kwoodson: I was able to find the statefulsets and daemonsets by using the apps_client.go you mentioned.  

My question would be, what is the advantage of getting those objects from the apps_client when the coreclient fetches the pods which are from those objects?  The owner reference on the pod has a `kind` field which tells you about the parent object.  

Is it specifically that daemonsets and statefulsets don't deploy when there are no matching selectors?  If this is the case, I'd lean to verifying there are nodes labeled appropriately so that the selectors are correct.

I am fine proceeding but would like your insight as to which is the preferred so that I understand.

Thanks @jim-minter

jim-minter: It's a mess and the answer is almost certainly that both clients are needed. I think it's probably best to iterate the daemonsets/deployments/etc and then resort to checking the pods only if necessary because the info isn't in the parent's status, but there are probably multiple ways to skin this particular cat.

pkg/checks/checks.go
kwoodson: The `io.EOF` did the trick.

pkg/checks/checks.go
```diff
@@ -0,0 +1,210 @@
...
+		case <-ctx.Done():
+			log.Printf("ctx.Done called.  Returning err")
+			return ctx.Err()
+		case <-time.After(time.Second):
+		}
+	}
+}
+
+// WaitForInfaServices verify daemonsets, statefulsets
+func WaitForInfaServices(ctx context.Context, appclient *appclient.AppsV1Client) error {
```
kwoodson: @jim-minter was this more in-line with what you had in mind?

Should I add the deeper dive (fetch pods and check status) when I find `ReadyReplicas != Replicas` with each type?

jim-minter: yes :)

jim-minter: Infra

pkg/checks/checks.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Daemonsets
+	dsList, err := appclient.DaemonSets("").List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, ds := range dsList.Items {
```
jim-minter: for each kind, discard it if it isn't in a namespace we care about

pkg/checks/checks.go
```diff
@@ -0,0 +1,210 @@
...
+				return err
+			}
+			if ds.Status.NumberMisscheduled > 0 {
+				return fmt.Errorf("Daemonset[%v] in Namespace[%v] has missscheduled[%v]", ds.Name, ds.Namespace, ds.Status.NumberMisscheduled)
+			}
+
+			if ds.Status.DesiredNumberScheduled == ds.Status.CurrentNumberScheduled && ds.Status.DesiredNumberScheduled == ds.Status.NumberReady {
+				break
+			}
+			time.Sleep(2 * time.Second)
```
jim-minter:
```
select {
	case <-time.After(2 * time.Second):
	case <-ctx.Done():
		return ctx.Err()
}
```

pkg/checks/checks.go
kwoodson: Fixed.

kwoodson:
@jim-minter, If you wouldn't mind taking a look.  If this level of detection is sufficient then I'll clean up and we can merge.  If you'd like a deeper dive into pods I can also call the pod function.

pkg/checks/checks.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	for _, ns := range whitelistedNamespaces {
+		if strings.HasPrefix(namespace, ns) {
+			return true
+		}
+	}
+	return false
+}
+
+// WaitForInfraServices verify daemonsets, statefulsets
```
kargakis: We probably need a hardcoded list of the expected workloads and cross-checking readiness vs mere listing. I am fine with a follow-up issue.

pkg/checks/checks.go
```diff
@@ -0,0 +1,249 @@
...
+			select {
+			case <-time.After(2 * time.Second):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+	}
+
+	// ReplicaSets
+	rsList, err := appclient.ReplicaSets("").List(metav1.ListOptions{})
```
kargakis: You shouldn't need to check for replicasets since you check for deployments, unless some replicasets run standalone.

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,75 @@ import (
...
+	t, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return err
+	}
+
+	ctxHealthz, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	// Wait for the healthz to be 200 status
+	err = checks.WaitForHTTPStatusOk(ctxHealthz, t, restconfig.Host+"/healthz")
```
kargakis: How about checking for `/healthz/ready` vs `/healthz`?

pkg/checks/checks.go
kwoodson: Currently they do not run standalone.  Is there any harm in doing so?

pkg/healthcheck/healthcheck.go
kwoodson: I'm not familiar with the difference.  I'll try it and see.

pkg/checks/checks.go
```diff
@@ -0,0 +1,249 @@
...
+			log.Printf("the err was %#v\n", err)
+			return err
+		}
+		if resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		select {
+		case <-ctx.Done():
+			log.Printf("ctx.Done called.  Returning err")
```
jim-minter: remove log.Printf?

pkg/checks/checks.go
```diff
@@ -0,0 +1,249 @@
...
+		case <-time.After(time.Second):
+		}
+	}
+}
+
+func checkNamespace(namespace string) bool {
+	whitelistedNamespaces := []string{
+		"openshift-",
+		"kube-",
+		"default",
```
jim-minter: edge case: `default` should be tested with equality, not hasprefix

pkg/checks/checks.go
```diff
@@ -0,0 +1,249 @@
...
+				return ctx.Err()
+			}
+		}
+	}
+
+	return nil
+}
+
+// WaitForPodsRunning verifies all pods are in running state
+func WaitForPodsRunning(ctx context.Context, coreclient *coreclient.CoreV1Client, appclient *appclient.AppsV1Client) error {
```
jim-minter: not used - remove (save in a scratch pad in case you need it one day ;)

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,75 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	// coreclient, err := coreclient.NewForConfig(restconfig)
```
jim-minter: remove

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,75 @@ import (
...
+
+	appsclient, err := appsclient.NewForConfig(restconfig)
+	if err != nil {
+		return err
+	}
+	// Ensure that the pods in default are healthy
+	ctxPods, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	err = checks.WaitForInfraServices(ctxPods, appsclient)
+	// err = checks.WaitForPodsRunning(ctxPods, coreclient, appsclient)
```
jim-minter: remove

pkg/checks/checks.go
kwoodson: done.

pkg/checks/checks.go
kwoodson: done

pkg/healthcheck/healthcheck.go
kwoodson: done

pkg/healthcheck/healthcheck.go
kwoodson: done

pkg/checks/checks.go
kargakis: My initial thought was that it's redundant since the deployment status api is an aggregation of all the replicasets under its control (I wrote that part :)). Now, I think it's also wrong because you may list old replicasets that should not have available replicas running.

pkg/checks/checks.go
```diff
@@ -0,0 +1,186 @@
...
+		if !checkNamespace(ds.Namespace) {
+			continue
+		}
+		for {
+			ds, err := appclient.DaemonSets(ds.Namespace).Get(ds.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if ds.Status.NumberMisscheduled > 0 {
+				return fmt.Errorf("Daemonset[%v] in Namespace[%v] has missscheduled[%v]", ds.Name, ds.Namespace, ds.Status.NumberMisscheduled)
```
kargakis: I wouldn't error out here at once but retry for some time until it converges to zero.

pkg/checks/checks.go
kargakis: Yeah, maybe that's enough

kargakis: Although I am worried about a missing workload for whatever reason, that's why I think a whitelist of the workloads is eventually required.

pkg/checks/checks.go
kargakis: We can reuse the existing context timeout as a retry mechanism so you can just log here.

pkg/checks/checks.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			if ds.Status.DesiredNumberScheduled == ds.Status.CurrentNumberScheduled && ds.Status.DesiredNumberScheduled == ds.Status.NumberReady {
```
kargakis:
```go
if ds.Status.DesiredNumberScheduled == ds.Status.CurrentNumberScheduled && 
  ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable &&
  ds.Status.DesiredNumberScheduled == ds.Status.UpdatedNumberScheduled &&
  ds.Metadata.Generation == ds.Status.ObservedGeneration {
```

pkg/checks/checks.go
kwoodson: Should I log or just remove this statement entirely and let the loop occur until we are satisfied with the valid logic?

kargakis: Removing is also probably fine. I would add a log that prints all status fields every time we retry but can be done in a follow-up across all loops if we want it.

pkg/checks/checks.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, ss := range ssList.Items {
+		if !checkNamespace(ss.Namespace) {
+			continue
+		}
+		for {
+			ss, err := appclient.StatefulSets(ss.Namespace).Get(ss.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if ss.Status.Replicas == ss.Status.ReadyReplicas && ss.Status.ReadyReplicas == ss.Status.CurrentReplicas {
```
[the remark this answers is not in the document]
kwoodson: Added.  Will test.

pkg/checks/checks.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, d := range dList.Items {
+		if !checkNamespace(d.Namespace) {
+			continue
+		}
+		for {
+			d, err := appclient.Deployments(d.Namespace).Get(d.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if d.Status.Replicas == d.Status.ReadyReplicas && d.Status.ReadyReplicas == d.Status.AvailableReplicas {
```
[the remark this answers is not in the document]
kwoodson: added.

pkg/checks/checks.go
kwoodson: Will remove replicaset testing.

kwoodson:
@jim-minter, @kargakis,

With a latest rebase everything completed successfully.

```
+ az group deployment wait -g kwoodsonhcp -n azuredeploy --created --interval 10
+ KUBECONFIG=_data/_out/admin.kubeconfig
+ go run cmd/healthcheck/healthcheck.go
```

pkg/addons/client.go
```diff
@@ -1,23 +1,17 @@
...
-	"net"
-	"net/http"
-	"net/url"
-	"os"
 	"reflect"
-	"syscall"
-	"time"
 
 	"github.com/go-test/deep"
+	"github.com/openshift/openshift-azure/pkg/checks"
```
jim-minter: we seem to have the style of keeping the github.com/openshift/openshift-azure imports in a separate block - separate for consistency

pkg/checks/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	for {
+		resp, err := cli.Do(req)
+		if err, ok := err.(*url.Error); ok && (err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF) {
+			time.Sleep(time.Second)
+			continue
+		}
+
+		if err == io.EOF {
```
jim-minter: the implementation changed upstream while your PR was in progress (see the `return wait.PollInfinite`...) and you'll need to merge my changes and yours.  The end result should include at least the `wait.PollInfinite`, the `syscall.ENETUNREACH` bit and your new `err == io.EOF` bit.

pkg/checks/checks.go
```diff
@@ -0,0 +1,167 @@
...
+	cli := &http.Client{
+		Transport: transport,
+		Timeout:   10 * time.Second,
+	}
+
+	req, err := http.NewRequest("GET", urltocheck, nil)
+	if err != nil {
+		return err
+	}
+	return wait.PollInfinite(time.Second, func() (bool, error) {
```
jim-minter: sorry, my bad, should be `wait.PollUntil(time.Second, func() (bool, error) { ... }, ctx.Done())`

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,69 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	t, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return err
+	}
+
+	ctxHealthz, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
jim-minter: remove this, use the passed in ctx instead of ctxHealthz

pkg/healthcheck/healthcheck.go
```diff
@@ -10,11 +10,69 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	appsclient, err := appsclient.NewForConfig(restconfig)
+	if err != nil {
+		return err
+	}
+	// Ensure that the pods in default are healthy
+	ctxPods, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
jim-minter: remove this, use the passed in ctx instead of ctxPods

pkg/healthcheck/healthcheck.go
kwoodson: done.

pkg/healthcheck/healthcheck.go
kwoodson: done.

PR #93, opened 2018-07-18
replace checksum/ with checksum/secret in helm

jim-minter:
@kargakis ptal


kargakis:
OK, now it's a good time for me to figure out how this works for helm. I assume helm has a mechanism to restart deployments, because merely updating object's metadata.annotations is not going to start new rollouts.

jim-minter:
fixed ;)

kargakis:
lgtm

PR #95, opened 2018-07-19
Use https in git pullspec

kargakis:
@jim-minter this allows us to run `glide install` inside a container w/o needing to mount any ssh keys


jim-minter:
thanks!

PR #106, opened 2018-07-24
Build plugins as part of a single target

kargakis:
Hold for on-going discussion in aos-cicd


kargakis:
I am going to add a job just for `sync` for now.

PR #107, opened 2018-07-25
Use VMs

jim-minter:


jim-minter:
@pweil- @kargakis @kwoodson @mjudeikis ptal

pkg/arm/data/azuredeploy.json
```diff
@@ -171,18 +254,21 @@
                                             "properties": {
                                                 "primary": true,
                                                 "subnet": {
-{{ if $app.VnetSubnetID }}
-                                                    "id": "{{ $app.VnetSubnetID }}"
-{{ else }}
-                                                    "id": "[concat(resourceId('Microsoft.Network/virtualNetworks', 'vnet'), '/subnets/default')]"
-{{ end }}
+                                                    "id": "{{if $app.VnetSubnetID}}{{$app.VnetSubnetID}}{{else}}[concat(resourceId('Microsoft.Network/virtualNetworks', 'vnet'), '/subnets/default')]{{end}}"
```
mjudeikis: +1

pkg/config/generate.go
```diff
@@ -121,9 +119,12 @@ func selectContainerImages(cs *acsapi.ContainerService, c *Config) {
 	}
 }
 
+func internalAPIServerHostname(cs *acsapi.ContainerService) string {
```
mjudeikis: I assume this will change when we get either dedicated zone or some identifier.

pkg/addons/client.go
```diff
@@ -109,11 +116,22 @@ func (c *client) waitForHealthz() error {
 
 	for {
 		resp, err := cli.Do(req)
-		if err, ok := err.(*url.Error); ok && (err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF) {
-			time.Sleep(time.Second)
-			continue
+		if err, ok := err.(*url.Error); ok {
```
kargakis: Create a helper to keep this loop cleaner
```go
func needsRetry(err error) bool {
			if err, ok := err.Err.(*os.SyscallError); ok {
				if err.Err == syscall.ENETUNREACH {
					return true
				}
			}
		}
		if err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF {
			return true
		}
		return false
	}
}
```

pkg/arm/data/azuredeploy.json
```diff
@@ -26,7 +26,93 @@
                 ]
             }
         },
-{{ end }}
+{{- end }}
+{{/* TODO: we may be able to avoid the IP and LB here if we stop using */}}
+{{/* DaemonSets for SDN.  Then Master nodes should come up as Ready */}}
```
kargakis: So how would you deploy SDN in this case?

pkg/config/generate.go
```diff
@@ -409,7 +410,7 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 			// sync kubeconfig has the same capabilities as admin kubeconfig, only difference
 			// is the use of HCP internal DNS to avoid waiting for the Azure loadbalancer to
```
kargakis: Follow-up issue to remove this in favor of using a service account.

glide.yaml
```diff
@@ -3,33 +3,3 @@ import:
...
-  - ssh
-- package: k8s.io/apimachinery
-  subpackages:
-  - pkg/api/errors
-  - pkg/api/meta
-  - pkg/apis/meta/v1
-  - pkg/apis/meta/v1/unstructured
-  - pkg/runtime/schema
-  - pkg/util/wait
-- package: k8s.io/client-go
```
kargakis: Why are these removed?

pkg/arm/data/azuredeploy.json
jim-minter: static pods

pkg/config/generate.go
jim-minter: not sure I understand:
a) sync effectively needs system:admin, so it might as well just use it
b) there's a bootstrapping issue whereby the first time sync runs, all there is is system:admin.

pkg/arm/data/master-startup.sh
```diff
@@ -0,0 +1,570 @@
...
+rm -rf /etc/etcd/* /etc/origin/master/*
+
+base64 -d <<< {{ CertAsBytes .Config.EtcdCaCert | Base64Encode }} >/etc/etcd/ca.crt
+base64 -d <<< {{ CertAsBytes .Config.EtcdServerCert | Base64Encode }} >/etc/etcd/server.crt
+base64 -d <<< {{ PrivateKeyAsBytes .Config.EtcdServerKey | Base64Encode }} >/etc/etcd/server.key
+base64 -d <<< {{ CertAsBytes .Config.EtcdPeerCert | Base64Encode }} >/etc/etcd/peer.crt
+base64 -d <<< {{ PrivateKeyAsBytes .Config.EtcdPeerKey | Base64Encode }} >/etc/etcd/peer.key
+
+base64 -d <<< {{ YamlMarshal .Config.AdminKubeconfig | Base64Encode }} >/etc/origin/node/node.kubeconfig
+base64 -d <<< {{ CertAsBytes .Config.CaCert | Base64Encode }} >/etc/origin/node/client-ca.crt
```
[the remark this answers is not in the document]
jim-minter: thanks!

pkg/config/generate.go
kargakis: Yeah, you are right about the boostrapping problem.

glide.yaml
jim-minter: They're all transitively implied by the acs-engine import, which is currently pinning a lot of versions.  That's a pretty horrible situation and this removal helps to make that evident.  We'll want to somehow cut that out so we can decide the versions of the libraries we need, not have it decided for us on the basis of importing a single API package :-/

kargakis: That's pretty bad, we use client-go directly in `sync` so it's not a transitive dependency and certain client-go versions map to certain k8s versions.

pkg/addons/client.go
jim-minter: yes, it was about time to do that ;-)

glide.yaml
mjudeikis: What about if we kick-off separate vendor tree/sub project for projects not dependent on acs-engine (like sync)?

jim-minter:
Thanks for the reviews.  Addressed.  Merging.

pkg/addons/data/StatefulSet.apps/openshift-metrics/prometheus.yaml
```diff
@@ -37,7 +37,6 @@ spec:
         - -openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
         - -skip-auth-regex=^/metrics
         image: '*** GENERATED ***'
-        imagePullPolicy: Always
```
kwoodson: for my understanding, why did these go away?  Looks like the default is to `IfNotPresent` which causes the kubelet to skip pulling if it exists vs forcing a pull each time.

Src:
https://kubernetes.io/docs/concepts/containers/images/

create.sh
```diff
@@ -57,13 +52,18 @@ properties:
   publicHostname: openshift.$RESOURCEGROUP.$DNS_DOMAIN
   routingConfigSubdomain: $RESOURCEGROUP.$DNS_DOMAIN
   agentPoolProfiles:
-  - name: compute
-    count: 1
+  - name: master
+    role: master
+    count: 3
```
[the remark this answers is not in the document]
jim-minter: in due course this should be more flexible https://github.com/openshift/openshift-azure/issues/119

pkg/addons/translate.go
```diff
@@ -343,19 +356,30 @@ var Translations = map[string][]struct {
 	},
 	"Secret/kube-service-catalog/apiserver-ssl": {
 		{
-			Path:     jsonpath.MustCompile("$.stringData.'tls.crt'"),
+			Path:     jsonpath.MustCompile("$.stringData.'apiserver.crt'"),
```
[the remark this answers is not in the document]
jim-minter: no - the single quotes are required because of the period.  we translate the Secret "data" to "stringData" at https://github.com/openshift/openshift-azure/blob/master/pkg/addons/clean.go#L45 to save extra hassle with base64 encodings.

pkg/arm/data/master-startup.sh
```diff
@@ -0,0 +1,570 @@
...
+resourceGroup: {{ .ContainerService.Properties.AzProfile.ResourceGroup | quote }}
+location: {{ .ContainerService.Location | quote }}
+securityGroupName: nsg-compute
+primaryScaleSetName: ss-compute
+vmType: vmss
+EOF
+
+# TODO: investigate the --manifest-url Kubelet parameter and see if it might
+# help us at all
+cat >/etc/origin/node/pods/etcd.yaml <<EOF
```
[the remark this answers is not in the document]
jim-minter: The direction I'd like to take is to get these under our control in openshift-azure rather than openshift-ansible - that's why

pkg/addons/data/StatefulSet.apps/openshift-metrics/prometheus.yaml
jim-minter: Yes, this isn't right. https://github.com/openshift/openshift-azure/issues/120.

PR #109, opened 2018-07-26
add unit tests (1)

mjudeikis:
We need to start at some point...


pkg/validate/validate_test.go
```diff
@@ -0,0 +1,116 @@
...
+						},
+					},
+					ServicePrincipalProfile: &api.ServicePrincipalProfile{},
+					AgentPoolProfiles:       nil,
+				},
+			},
+			expectedResult: fmt.Errorf("invalid number of agentPoolProfiles"),
+		},
+		{
+			name: "invalid number of agentPoolProfiles",
```
pweil-: copy paste miss

pkg/validate/validate.go
```diff
@@ -15,12 +15,6 @@ func Validate(cs, _ *acsapi.ContainerService) error {
 		return fmt.Errorf("malformed manifest")
 	}
 
-	for _, app := range cs.Properties.AgentPoolProfiles {
-		if app == nil {
-			return fmt.Errorf("malformed manifest")
-		}
-	}
-
```
mjudeikis: There is len(cs.Properties.AgentPoolProfiles) below which does the same thing. as even empty agentPoolProfile is not nil, so this does not get triggered.

jim-minter: not true - cs.Properties.AgentPoolProfiles is a slice of pointers.  This is checking that all the pointers in the slice are non-nil.  It is not checking the length of the slice.
The intention here is to do the non-nil sanity checking once in one place so we don't have to worry about it anywhere else in the codebase (perhaps that's worth a comment - perhaps you could add one?)

pkg/addons/clean_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestClean(t *testing.T) {
+	tests := []struct {
```
jim-minter: it must be possible to do something nicer? this will grow into misery.
testdata directory, subdirectory per test, %03d-input.yaml, %03d-expected.yaml, loop, unmarshal, clean, compare?

pkg/validate/validate.go
mjudeikis: got u.!

pkg/validate/validate_test.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, tt := range tests {
+		err := Validate(tt.cs, tt.cs)
+		if !reflect.DeepEqual(err, tt.expectedResult) {
+			t.Errorf("fail: %#v \n%#v\n%#v", tt.name, err, tt.expectedResult)
+		}
+	}
+
+}
+
+func TestValidateDevelopmentSwitches(t *testing.T) {
```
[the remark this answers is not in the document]
mjudeikis: is this even a thing? :)

pkg/addons/clean_test.go
mjudeikis: better?

pkg/validate/validate_test.go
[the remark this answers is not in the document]
jim-minter: emphatically yes.

pkg/addons/clean_test.go
jim-minter: not really - take a look at https://github.com/jim-minter/openshift-azure/commit/tests for what I had in mind

mjudeikis:
@jim-minter removed validation tets

jim-minter:
thanks @mjudeikis will squash merge

PR #111, opened 2018-07-27
remove v from image name

mjudeikis:
align to image versioning as we publish. 

alternative to republish all images


mjudeikis:
cc: @openshift/sig-azure 
/kind bug

jim-minter:
/lgtm

thanks!

PR #116, opened 2018-07-27
 Add validation script for generated code 

kargakis:
@openshift/sig-azure


kargakis:
Will merge to setup the verification test as part of the build job

hack/validate-generated.sh
```diff
@@ -0,0 +1,35 @@
...
+trap "rm -rf ${T}" EXIT
+
+cp -a "${DIR}/.." "${T}/"
+
+(cd "${T}/" && go generate ./...)
+
+
+GENERATED_FILES=(
+	"pkg/addons/bindata.go"
+	"pkg/arm/bindata.go"
```
mjudeikis: jsonpath too

hack/validate-generated.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
jim-minter: probably should combine hack and tools

jim-minter: don't care which stays

PR #118, opened 2018-07-27
#116 follow ups

jim-minter:
https://github.com/openshift/openshift-azure/pull/116 follow ups
@kargakis @mjudeikis


kargakis:
Should we move all shell scripts from the top level directory to `hack/`, too?

jim-minter:
Yeah, I think so; I was happy to do this one without a test run but I figured moving create.sh probably warranted doing some testing ;)

PR #121, opened 2018-07-27
bump external API

jim-minter:
This uses the (now committed!) external API.
It has the validation in one place.
I've tried to improve the code documentation in createorupdate.


jim-minter:
@pweil- @amanohar ptal, this makes some changes to the Plugin interface signature, for the better I think; is that OK?
@kargakis this removes the non-RP calls to Enrich. 
@mjudeikis @kwoodson fyi

cmd/createorupdate/createorupdate.go
```diff
@@ -7,82 +7,124 @@ import (
...
 	}
-	cs := ext.AsContainerService()
-	err = plugin.Enrich(cs)
+
+	// convert the external API manifest into the internal API representation
+	cs := acsapi.ConvertVLabsOpenShiftClusterToContainerService(oc)
+
+	// the RP will enrich the internal API representation with data not included
+	// in the original request
+	// TODO(mjudeikis): choose DNS names here
```
jim-minter: @mjudeikis fyi

create.sh
```diff
@@ -74,6 +79,8 @@ go run cmd/createorupdate/createorupdate.go
 
 az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _data/_out/azuredeploy.json --no-wait
 
+# TODO(mjudeikis): after your DNS work these commands should move to the end of
```
jim-minter: @mjudeikis fyi

pkg/api/api.go
```diff
@@ -5,13 +5,13 @@ import (
...
 )
 
-type NewPlugin func(cs, oldCs *acsapi.ContainerService, configBytes []byte) (Plugin, error)
-
 type Plugin interface {
-	Validate() error
-	GenerateConfig() ([]byte, error)
-	GenerateARM() ([]byte, error)
-	HealthCheck(context.Context) error
+	ValidateExternal(oc *vlabs.OpenShiftCluster) []error
```
jim-minter: @pweil- @amanohar changes to Plugin interface, are you OK with this?

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
+package validate
```
jim-minter: @pweil- @kargakis @mjudeikis combined validation as discussed

pkg/config/config.go
```diff
@@ -118,4 +118,9 @@ type Config struct {
 
 	// used for development purposes only
 	RunSyncLocal string
+
+	// enriched values which are not present in the external API representation
+	TenantID       string
```
jim-minter: @kargakis thinking we should stash these approximately here - the alternatives I saw were: in the public-facing API (not great), or carting around a different serialisation of the ContainerService (ugh).  wdyt?

cmd/createorupdate/createorupdate.go
```diff
@@ -1,88 +1,176 @@
...
+	// generate or update the OpenShift config blob
+	configBytes, err = p.GenerateConfig(cs, configBytes)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	err = p.Validate()
+	// persist the OpenShift config blob
+	err = ioutil.WriteFile("_data/config.yaml", configBytes, 0600)
```
mjudeikis: Should we keep some backups of n-1 configuration? would make debugging and rollbacks little bit easier.

pkg/validate/openshift_cluster.go
mjudeikis: for the future, what if we put all error messages to something like:
```
//go:generate stringer -type=ErrorType
type ErrorType int

const (
	ErrorMalformedManifest ErrorType = iota
	ErrorInvalidOpenshiftVersion
)
```
This would enable easier testing when `[]error` returned.

cmd/createorupdate/createorupdate.go
kargakis: We have briefly discussed storing separate blobs under the storage account that holds the config.

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+			errs = append(errs, fmt.Errorf("duplicate properties.routerProfiles %q", rp.Name))
+		}
+		rpmap[rp.Name] = rp
+
+		errs = append(errs, validateRouterProfile(rp)...)
+	}
+
+	for name := range validRouterProfileNames {
+		if _, found := rpmap[name]; !found {
+			errs = append(errs, fmt.Errorf("invalid properties.routerProfiles[%q]", name))
```
kargakis: The error message can probably be clearer that they are missing a valid router profile.

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+
+	for name := range rpmap {
+		if _, found := validRouterProfileNames[name]; !found {
+			errs = append(errs, fmt.Errorf("invalid properties.routerProfiles[%q]", name))
```
kargakis: ditto, they use an invalid name for the router profile

pkg/api/api.go
pweil-: yeah, now that we've settled on shape and location I'm fine with this.  

it looks like this has also changed the shape to a stateless plugin.  Looks good

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+	case "",
+		vlabs.Creating,
+		vlabs.Updating,
+		vlabs.Failed,
+		vlabs.Succeeded,
+		vlabs.Deleting,
+		vlabs.Migrating,
+		vlabs.Upgrading:
+	default:
+		errs = append(errs, fmt.Errorf("invalid properties.provisioningState %q", p.ProvisioningState))
```
pweil-: not sure I understand what this is validating.  State should be set by code.  I guess this is a fail safe but it seems a little odd to me.  If we want to protect against invalid state transitions that would make sense but probably not here.

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	if oc.Properties != nil {
+		errs = append(errs, validateProperties(oc.Properties)...)
+	}
+
+	return
+}
+
+func validateProperties(p *vlabs.Properties) (errs []error) {
```
pweil-: follow up: most of these errors don't give any suggestions on how to fix them (invalid formats don't show the format, invalid values don't show the acceptable set, etc)

pkg/validate/openshift_cluster.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	for name := range validAgentPoolProfileNames {
+		if _, found := appmap[name]; !found {
+			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q]", name))
+		}
+	}
+
+	for name := range appmap {
+		if _, found := validAgentPoolProfileNames[name]; !found {
```
pweil-: this check could be done in the first loop so that only valid app names are in the appmap (premature optimization ftw :smile:)

pkg/config/config.go
kargakis: `node-startup.sh` is still using `.ContainerService.Properties.AzProfile`. Do you want to change that to use these?

pkg/validate/openshift_cluster.go
jim-minter: MSFT have specific requirements around errors which they haven't shared with us yet (@amanohar please can you do this?) - we'll need to take these into account.

pkg/config/config.go
jim-minter: I did wonder about doing that; will do.

pkg/validate/openshift_cluster.go
jim-minter: going to open a follow-up from this.  I am expecting specific guidance on error formatting from MSFT hence I haven't been to worried about this up to now.

jim-minter: #127

pkg/validate/openshift_cluster.go
jim-minter: thanks ;)

jim-minter:
fixes added, merging
/lgtm

PR #128, opened 2018-07-30
[WIP] DNS integration 

mjudeikis:
first for dns integration


create.sh
```diff
@@ -79,8 +81,7 @@ go run cmd/createorupdate/createorupdate.go
 
 az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _data/_out/azuredeploy.json --no-wait
 
-# TODO(mjudeikis): after your DNS work these commands should move to the end of
-# this file.
+# This mimics RP dns record creation.
 hack/dns.sh zone-create $RESOURCEGROUP
```
mjudeikis: This is not at the end because healthcheck and sync in dev mode are blocking process and they need dns for checks (health check does)

jim-minter: they should not be using the vanity dns for that -- by moving the dns creation the end, it'll prove that they're not ;)

create.sh
```diff
@@ -49,10 +49,12 @@ name: openshift
 location: eastus
 properties:
   openShiftVersion: "$DEPLOY_VERSION"
+  FQDN: $RESOURCEGROUP.eastus.cloudapp.azure.com
```
jim-minter: this is all backwards - FQDN is the r/o field by which the RP indicates the chosen name.

pkg/arm/data/master-startup.sh
```diff
@@ -255,7 +255,7 @@ oauthConfig:
       kind: HTPasswdPasswordIdentityProvider
   masterCA: ca.crt
   masterPublicURL: {{ print "https://" .ContainerService.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname | quote }}
-  masterURL: {{ print "https://" .ContainerService.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname | quote }}
+  masterURL: {{ print "https://" .ContainerService.Properties.MasterProfile.FQDN | quote }}
```
jim-minter: I think we should have the RP set PublicHostname to FQDN if it wasn't set, and here we should continue using PublicHostname.
But in the master cert generation (for example), there should be SANs for both PublicHostname and FQDN, if they differ.

create.sh
mjudeikis: so whats wrong here?  
FQDN: mjudeikis.eastus.cloudapp.azure.com corresponds to random RP generated name, which should point to `ip-apiserver`. To fake it, we use loadbalancer default DNS name for now.

mjudeikis: This will be auto-allocated by RP:
https://github.com/jim-minter/acs-engine/blob/master/pkg/api/osa/vlabs/types.go#L39

pkg/arm/data/master-startup.sh
mjudeikis: PublicHostname, if set is in user control. Lets say FQDN is generated as `mjudeikis-9a4dc255.osa.westus.azmk8s.io` and user want to have `console.openshift.example.com` as the main URL. The user can set this on creation in PublicHostname and we will return FQDN and say "create a cname now please". In addition, if the user will want to have proper valid certificates at any point in future (question if this is even for GA, those values cant be set to the same value).

```
masterURL and masterPublicURL cannot be set to the same value when using custom named certificates. Both values need to be resolvable to the load balancer, but if masterURL is the same as masterPublicURL then hosts within the cluster will try to verify against the custom named certificates.

The masterURL needs to be resolvable by the nodes internal to the cluster; thus it does not need to be resolvable outside the cluster, whereas masterPublicURL must be externally resolvable.
```

create.sh
mjudeikis: It will work for cases where PublicHostname is not provided and we default to FQDN as PublicHostname. Otherwise, health-check gets this as it tries to reach OCP console via external url. And this DNS does not exist at the moment. And I think this is expected behavior to check external url as it

```
+ go run cmd/healthcheck/healthcheck.go
panic: Head https://openshift.mjudeikis.osadev.cloud/console/: dial tcp: lookup openshift.mjudeikis.osadev.cloud on 127.0.0.1:53: no such host
```
Because of this: https://github.com/openshift/openshift-azure/blob/master/pkg/healthcheck/healthcheck.go#L87
But I think it does what it expected to do.

We could change it to use internal url only: 
`req, err := http.NewRequest("HEAD", "https://"+cs.Properties.MasterProfile.FQDN+"/console/", nil)`
cc: @kwoodson

create.sh
kargakis: So we shouldn't set it here.

mjudeikis: got you! thanks :)

pkg/healthcheck/healthcheck.go
```diff
@@ -84,7 +85,7 @@ func waitForConsole(ctx context.Context, cs *acsapi.ContainerService, c *config.
 		Timeout: 10 * time.Second,
 	}
 
-	req, err := http.NewRequest("HEAD", "https://"+cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname+"/console/", nil)
```
mjudeikis: This is not quite right. But if PublicHostname is provided by user and not set by RP this test fails.

pkg/arm/data/master-startup.sh
mjudeikis: TODO: check customer_cert stuff https://docs.openshift.com/container-platform/3.5/install_config/certificate_customization.html

mjudeikis: TODO: Test if we can change    PublicHostname post fact

pkg/healthcheck/healthcheck.go
jim-minter: I think you can do this by using a custom HTTP Transport (see https://golang.org/pkg/net/http/#RoundTripper) and overriding Dial to dial the PublicHostname

mjudeikis:
TODO: Test if we can change PublicHostname post fact:
At the moment we cant due to https://github.com/openshift/openshift-azure/issues/135

pkg/checks/checks.go
```diff
@@ -32,6 +32,9 @@ func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urlto
 		resp, err := cli.Do(req)
 		if err, ok := err.(*url.Error); ok {
 			if err, ok := err.Err.(*net.OpError); ok {
+				if _, ok := err.Err.(*net.DNSError); ok {
```
mjudeikis: I dont like it as it might be valid error, but it fails because now with this it start fast and this record does not exist at that time... when moved to pod we might want to remove it and let pod to handle this?

pkg/config/generate.go
```diff
@@ -148,6 +143,11 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 			key:  &c.CaKey,
 			cert: &c.CaCert,
 		},
+		{
+			cn:   "openshift-console",
+			key:  &c.ConsoleCaKey,
+			cert: &c.ConsoleCaCert,
+		},
```
mjudeikis: Get a new CA so it would represent full BYO certificate model

pkg/config/generate.go
```diff
@@ -241,10 +241,9 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 			cert:        &c.MasterProxyClientCert,
 		},
 		{
-			cn: cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname,
+			cn: cs.Properties.MasterProfile.FQDN,
```
mjudeikis: We remove external name from all internal certificates

mjudeikis:
@jim-minter at the moment this removed an external name from internal certificates and adds external console certificate. Based on https://docs.openshift.com/container-platform/3.10/install_config/certificate_customization.html 

I still want to test if this certificate change didn't break down any functionality.
I would like to move RountTripper work to separate PR.

let me know what do you think. 

+ where will be one follow up story to this when we move console to be exposed as Route (not via API server). But I need some details from spasquie.

pkg/checks/checks.go
jim-minter: In the ARM template, I think you should pass in the first part of the FQDN into the apiserver public IP so that it is actually registered.  Then maybe this error catch isn't necessary.
You also need to check with @amanohar whether OSA will have its own separate dns domain which the RP will select a pre-created name from and how to handle that.

pkg/config/config.go
```diff
@@ -11,6 +11,11 @@ import (
 type Config struct {
 	Version int
 
+	ClusterId        string
+	ControlPlaneFQDN string
+	RouterFQDN       string
+	RouterFQDNShort  string
```
mjudeikis: Remove this later.

PR #129, opened 2018-07-30
Addings tests for pkg/validate.

kwoodson:
Adding unit tests for the `pkg/validate`.

### DIFF
--- a/pkg/validate/validate_test.go
+++ b/pkg/validate/validate_test.go
@@ -1,0 +1,119 @@
+package validate
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/Azure/acs-engine/pkg/api/osa/vlabs"
+	"github.com/ghodss/yaml"
+)
+
+/*
+name: openshift
+location: eastus
+properties:
+  openShiftVersion: "$DEPLOY_VERSION"
+  publicHostname: openshift.$RESOURCEGROUP.$DNS_DOMAIN
+  routerProfiles:
+  - name: default
+    publicSubdomain: $RESOURCEGROUP.$DNS_DOMAIN
+  agentPoolProfiles:
+  - name: master
+    role: master
+    count: 3
+    vmSize: Standard_D2s_v3
+    osType: Linux
+  - name: infra
+    role: infra
+    count: 1
+    vmSize: Standard_D2s_v3
+    osType: Linux
+  - name: compute
+    role: compute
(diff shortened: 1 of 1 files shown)


pkg/validate/validate_test.go
```diff
@@ -0,0 +1,185 @@
...
+    secret: $AZURE_CLIENT_SECRET
+*/
+func TestValidate(t *testing.T) {
+
+	tests := map[string]struct {
+		oc            *vlabs.OpenShiftCluster
+		expectedError string
+	}{
+		"invalid version": {
+			oc: &vlabs.OpenShiftCluster{
```
pweil-: turn this into a `getValidOpenShiftCluster` and use the closure to make the version invalid

pkg/validate/validate_test.go
```diff
@@ -0,0 +1,185 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	for name, test := range tests {
+		errs := OpenShiftCluster(test.oc)
+
+		for _, err := range errs {
```
pweil-: check error size, it should only be 1

pkg/validate/validate_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+			//if err.Error != test.expectedError {
+			t.Errorf("Test[%v] failed: expected %v", name, err)
```
pweil-: do a contains on the string here for your error message to ensure you got the right one (will account for wrapping if we end up wrapping in a MSFT helper or something)

jim-minter:
I e-mailed @kwoodson with a suggested alternative cleaner approach as a starting point for this

pkg/validate/validate_test.go
```diff
@@ -0,0 +1,119 @@
...
+			t.Fatal(err)
+		}
+
+		if test.f != nil {
+			test.f(oc)
+		}
+
+		errs := OpenShiftCluster(oc)
+		if !reflect.DeepEqual(errs, test.expectedErrs) {
+			t.Errorf("%v: OpenShiftCluster returned unexpected result", name)
```
pweil-: update this to show expected and received and this is LGTM.  Nice and clean, Kenny.  I like the non-returning function better than the returning one I proposed.

pweil-: this can be a follow up, would like working tests merged asap.  Approving!

PR #130, opened 2018-07-31
[WIP] OSA internal datamodel

amanohar:
This PR takes a first stab at the internal data model representing an OSA cluster meant to be used to generate ARM templates.


pkg/api/const.go
```diff
@@ -0,0 +1,57 @@
...
+	CoreOS Distro = "coreos"
+	// Supported distros by OpenShift
+	OpenShift39RHEL Distro = "openshift39_rhel"
+	OpenShiftCentOS Distro = "openshift39_centos"
+)
+
+// Availability profiles
+const (
+	// AvailabilitySet means that the vms are in an availability set
+	AvailabilitySet = "AvailabilitySet"
```
kargakis: I think we care only about vmss at this point.

pkg/api/const.go
```diff
@@ -0,0 +1,57 @@
...
+	// ScaleSetPriorityLow means the ScaleSet will use Low-priority VMs
+	ScaleSetPriorityLow = "Low"
+	// ScaleSetEvictionPolicyDelete is the default Eviction Policy for Low-priority VM ScaleSets
+	ScaleSetEvictionPolicyDelete = "Delete"
+	// ScaleSetEvictionPolicyDeallocate means a Low-priority VM ScaleSet will deallocate, rather than delete, VMs.
+	ScaleSetEvictionPolicyDeallocate = "Deallocate"
+)
+
+const (
+	// AgentPoolProfileRoleEmpty is the empty role.  Deprecated; only used in
```
kargakis: If acs-engine is not anywhere in the picture anymore, why do we need this?

pkg/api/types.go
```diff
@@ -0,0 +1,470 @@
...
+// CustomFile has source as the full absolute source path to a file and dest
+// is the full absolute desired destination path to put the file on a master node
+type CustomFile struct {
+	Source string `json:"source,omitempty"`
+	Dest   string `json:"dest,omitempty"`
+}
+
+// OpenShiftConfig holds configuration for OpenShift
+type OpenShiftConfig struct {
+	KubernetesConfig *KubernetesConfig `json:"kubernetesConfig,omitempty"`
```
kargakis: We don't use this and we actually meant to drop it from OpenShiftConfig. Now it's a good time to do it.

pweil-:
See https://github.com/openshift/openshift-azure/pull/143 for move of acs-engine types (pared down) into the repo.  :+1:

PR #131, opened 2018-07-31
Ignore all errors when healthchecking

kargakis:
Makes the healthchecking code more robust and relies solely on the context timeout for aborting.

/cc jim-minter kwoodson

### DIFF
--- a/pkg/checks/checks.go
+++ b/pkg/checks/checks.go
@@ -3,13 +3,8 @@ package checks
 import (
 	"context"
 	"fmt"
-	"io"
-	"net"
 	"net/http"
-	"net/url"
-	"os"
 	"strings"
-	"syscall"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -30,23 +25,9 @@ func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urlto
 	}
 	return wait.PollUntil(time.Second, func() (bool, error) {
 		resp, err := cli.Do(req)
-		if err, ok := err.(*url.Error); ok {
-			if err, ok := err.Err.(*net.OpError); ok {
-				if err, ok := err.Err.(*os.SyscallError); ok {
-					if err.Err == syscall.ENETUNREACH {
-						return false, nil
-					}
-				}
-			}
-			if err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF {
-				return false, nil
-			}
-		}
-		if err == io.EOF {
-			return false, nil
-		}
 		if err != nil {
-			return false, err
+			fmt.Println(err)
+			return false, nil
 		}
 		return resp != nil && resp.StatusCode == http.StatusOK, nil
 	}, ctx.Done())


pkg/checks/checks.go
```diff
@@ -30,23 +25,8 @@ func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urlto
...
-		}
-		if err == io.EOF {
-			return false, nil
-		}
 		if err != nil {
-			return false, err
+			return false, nil
 		}
 		return resp != nil && resp.StatusCode == http.StatusOK, nil
 	}, ctx.Done())
```
kargakis: The Background context is never cancelled so essentially this will loop forever until the apiserver is up. We may want to do something different in a follow-up.

jim-minter:
that is going to be very noisy and make it impossible to detect any permanent failure conditions.  I don't love it.

kargakis:
I don't think we can reliably define permanent failures. You know that you failed only when you have waited long enough. Apart from the fmt.Println addition this doesn't change anything before https://github.com/openshift/openshift-azure/pull/125 was merged.

kargakis:
> Apart from the fmt.Println addition this doesn't change anything before #125 was merged.

No, that's a lie, we still used to return errors. We either need this or we need to add retries in the client because the sync process is broken atm.

jim-minter:
I think that the errors we're encountering are volatile at the moment as we're developing.  But as we stabilise I think we might find classes of error which will be permanent - e.g. failure to resolve dns.  I feel like set of non-permanent errors which we will need to handle will stay the same or become smaller.

pkg/checks/checks.go
```diff
@@ -30,23 +25,9 @@ func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urlto
 	}
 	return wait.PollUntil(time.Second, func() (bool, error) {
```
mjudeikis: Can we increase PollUntil time?  5 second might be too much, 2? 
```
+ go run cmd/sync/sync.go -run-once=true
2018/08/06 08:58:49 Sync process started!
Get https://mjudeikis-test.eastus.cloudapp.azure.com/healthz: dial tcp 255.255.255.255:443: connect: network is unreachable
.... x ~50 times
Get https://mjudeikis-test.eastus.cloudapp.azure.com/healthz: dial tcp 255.255.255.255:443: connect: network is unreachable

Get https://mjudeikis-test.eastus.cloudapp.azure.com/healthz: EOF
... x ~20 times
Get https://mjudeikis-test.eastus.cloudapp.azure.com/healthz: EOF
Get https://mjudeikis-test.eastus.cloudapp.azure.com/healthz: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
2018/08/06 09:02:28 Update Namespace/default
2018/08/06 09:02:28 - Object.map[metadata].map[annotations]: <does not have key> != map[openshift.io/node-selector:]
2018/08/06 09:02:28 - Object.map[spec].map[finalizers].slice[1]: <no value> != openshift.io/origin
2018/08/06 09:02:28 Skip Namespace/kube-public

```

kargakis: I am seeing the same set of errors, only for longer than just 2 or 5 seconds. We need to ignore these errors inside the loop, increasing the interval is not a robust solution.

mjudeikis: They can't be ignored forever. There is a change to failure here. In example, if DNS was allocated for us, but we can't claim it from backend due ARM issues.  I think I need to set a deadline for this particular piece of code. Something like 5 min should be more than enough for infrastructure to converge?

kargakis: A timeout is acceptable I think.

mjudeikis:
I would say merge this, raise the following issue to add timeout here. lack of this making local development little bit hard?

kargakis:
> I would say merge this, raise the following issue to add timeout here. lack of this making local development little bit hard?

Agreed that local development is hard. I have to rerun the sync process everytime I am deploying from master.

pkg/checks/checks.go
mjudeikis:
``` 
ctxTimeout, cancel := context.WithTimeout(ctx, time.Minute*5)
defer cancel()
	return wait.PollUntil(time.Second, func() (bool, error) {
		resp, err := cli.Do(req)
		if err != nil {
			fmt.Println(err)
			return false, nil
		}
		return resp != nil && resp.StatusCode == http.StatusOK, nil
	}, ctxTimeout.Done())
``` 
?

mjudeikis:
@kargakis  

```
       ctxTimeout, cancel := context.WithTimeout(ctx, time.Minute*5)
	defer cancel()
	return wait.PollUntil(time.Second, func() (bool, error) {
		resp, err := cli.Do(req)
		if err != nil {
			fmt.Println(err)
			return false, nil
		}
		return resp != nil && resp.StatusCode == http.StatusOK, nil
	}, ctxTimeout.Done())
```
 ? 

lets do this maybe and merge? with @pweil- blessing/or not :)

pweil-:
I still haven't seen this error :smile: 

I'm not in love with ignoring everything and waiting 10 minutes.  There is also the issue of possibly masking errors here too since we're just logging them out and this is meant to return informational errors in the case of cluster issues.  

This really seems like an issue that should be fixed another way.  If we need a dev hack for now, ok, but we need to make sure that this is addressed in the correct manner before code freeze

kargakis:
I think this issue is moot as soon as we move the process inside a pod (soon) so no reason to keep this open.

/close

Issue #126, opened 2018-07-30
Node role label missing from master nodes


kargakis:
```
$ oc get no master-000000 -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2018-07-30T10:32:41Z
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: Standard_D2s_v3
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: eastus
    failure-domain.beta.kubernetes.io/zone: "0"
    kubernetes.io/hostname: master-000000
  name: master-000000
<-- snip -->
```
whereas it's set correctly on infra and compute nodes
```
$ oc get no infra-000000 -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2018-07-30T10:35:49Z
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: Standard_D2s_v3
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: eastus
    failure-domain.beta.kubernetes.io/zone: "0"
    kubernetes.io/hostname: infra-000000
    node-role.kubernetes.io/infra: "true"
    region: infra
  name: infra-000000
<-- snip -->
```
```
$ oc get no compute-000000 -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2018-07-30T10:35:49Z
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: Standard_D2s_v3
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: eastus
    failure-domain.beta.kubernetes.io/zone: "0"
    kubernetes.io/hostname: compute-000000
    node-role.kubernetes.io/compute: "true"
    region: primary
  name: compute-000000
<-- snip -->
```
/kind bug

kargakis:
@openshift/sig-azure

kwoodson:
I noticed that the labels are here:

[...]

PR #132, opened 2018-07-31
Bootstrap kubeconfig needs to exist for sync daemonset to work

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/126

@openshift/sig-azure ptal


pkg/arm/data/master-startup.sh
```diff
@@ -562,6 +562,9 @@ EOF
 
 sed -i -re "s#( *server: ).*#\1https://$(hostname)#" /etc/origin/master/openshift-master.kubeconfig
 sed -i -re "s#( *server: ).*#\1https://$(hostname)#" /etc/origin/node/node.kubeconfig
+# HACK: copy node.kubeconfig to bootstrap.kubeconfig so that openshift start node used in the sync
+# daemonset will not fail and set the master node labels correctly.
+cp /etc/origin/node/node.kubeconfig /etc/origin/node/bootstrap.kubeconfig
```
pweil-: should we not be doing something like we do in the node startup:

```
base64 -d <<< {{ Base64Encode (YamlMarshal .Config.NodeBootstrapKubeconfig) }} >/etc/origin/node/bootstrap.kubeconfig
base64 -d <<< {{ Base64Encode (CertAsBytes .Config.NodeBootstrapCert) }} >/etc/origin/node/node-bootstrapper.crt
base64 -d <<< {{ Base64Encode (PrivateKeyAsBytes .Config.NodeBootstrapKey) }} >/etc/origin/node/node-bootstrapper.key
chmod 0600 /etc/origin/node/node-bootstrapper.key /etc/origin/node/bootstrap.kubeconfig
```

kwoodson: Are these variables controlled by the `/etc/sysconfig/<origin|atomic-openshift>-node` file?  This might be another option.

kargakis: @pweil- we want to bootstrap master nodes using the admin kubeconfig (can't do otherwise because the api server runs as a static pod) and the bootstrap kubeconfig is not really used for anything when masters are bootstrapped (kubelet ignores bootstrap.kubeconfig when node.kubeconfig exists). Only `openshift start node --write-flags` breaks in the sync daemonset. Maybe deserves an upstream issue.

jim-minter:
/lgtm

good sleuthing!

PR #136, opened 2018-08-01
Pin kubernetes dependencies used directly by us

kargakis:
@openshift/sig-azure ptal


pweil-:
this is unnecessary when we remove the acs-engine dependency as proposed during the workflow call (by moving the api definitions into our repo)

kargakis:
@pweil- not sure it's unnecessary, we use these dependencies directly in `sync`. `acs-engine` goes away but not the k8s deps.

pweil-:
@kargakis yes, I see.  We want to pin those :+1:

PR #138, opened 2018-08-01
Add vmssrollout command for upgrading VMSS

0xmichalis:
@amanohar @openshift/sig-azure

### DIFF
--- a/cmd/vmssrollout/vmssrollout.go
+++ b/cmd/vmssrollout/vmssrollout.go
@@ -1,14 +1,16 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"log"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
 	"github.com/ghodss/yaml"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/equality"
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/plugin"
@@ -22,9 +24,11 @@ var (
 	name           = flag.String("name", "", "Name of the Virtual Machine Scale Set to upgrade")
 
 	// upgrade parameters
-	newCsPath = flag.String("new-config", "", "Path to the new container service config")
-	oldCsPath = flag.String("old-config", "", "Path to the old container service config")
-	inPlace   = flag.Bool("in-place", false, "Perform an in-place upgrade")
+	csPath          = flag.String("config", "", "Path to the latest container service config")
+	oldCsPath       = flag.String("old-config", "", "Path to the old container service config")
+	templateFile    = flag.String("template-file", "", "Path to the latest ARM template")
+	oldTemplateFile = flag.String("old-template-file", "", "Path to the old ARM template")
+	inPlace         = flag.Bool("in-place", false, "Perform an in-place upgrade")
 
 	// Kubernetes-specific
 	drain = flag.Bool("drain", false, "Perform a Kubernetes node drain")
@@ -66,6 +70,7 @@ func getUpgrader(newPath, oldPath string) (*upgrade.VMSSUpgrader, error) {
 			Sku:     &newCs.Config.ImageSKU,
 			Version: &newCs.Config.ImageVersion,
 		}
+		log.Print("Image updated")
 	}
 
 	if newCs.Config.ImageResourceGroup != oldCs.Config.ImageResourceGroup ||
@@ -75,9 +80,15 @@ func getUpgrader(newPath, oldPath string) (*upgrade.VMSSUpgrader, error) {
 		imageRef = &compute.ImageReference{
 			ID: &id,
 		}
+		log.Print("Image updated")
 	}
 
 	count, err := getCount(*role, newCs)
+	if err != nil {
+		return nil, err
+	}
+
+	script, err := getScript(*templateFile, *oldTemplateFile)
 	if err != nil {
 		return nil, err
 	}
@@ -87,6 +98,7 @@ func getUpgrader(newPath, oldPath string) (*upgrade.VMSSUpgrader, error) {
 		ResourceGroup:  *resourceGroup,
 		Name:           *name,
 
+		Script:   script,
 		ImageRef: imageRef,
 		Count:    int64(count),
 		Drain:    *drain,
@@ -101,7 +113,67 @@ func readFile(path string) ([]byte, error) {
 	return ioutil.ReadFile(path)
 }
 
-func getCount(role string, cs *api.ContainerService) (int, error) {
+func getScript(newPath, oldPath string) (map[string]interface{}, error) {
+	newScript, err := readScriptFromTemplate(newPath)
+	if err != nil {
+		return nil, err
+	}
+	oldScript, err := readScriptFromTemplate(oldPath)
+	if err != nil {
+		return nil, err
+	}
+	if equality.Semantic.DeepEqual(newScript, oldScript) {
+		return nil, nil
+	}
+	log.Print("Script updated")
+	return newScript, nil
+}
+
+func readScriptFromTemplate(path string) (map[string]interface{}, error) {
+	templateBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	templateJSON := make(map[string]interface{})
+	if err := json.Unmarshal(templateBytes, &templateJSON); err != nil {
+		return nil, err
+	}
+
+	resourceSlice := templateJSON["resources"].([]interface{})
+	var vmss *compute.VirtualMachineScaleSet
+	var result interface{}
+	for _, r := range resourceSlice {
+		res := r.(map[string]interface{})
+		for k, v := range res {
+			val, ok := v.(string)
+			if k == "name" && ok && val == *name {
+				result = r
+				break
+			}
+		}
+	}
+
+	data, err := json.Marshal(result)
+	if err != nil {
+		return nil, err
+	}
+	vmss = &compute.VirtualMachineScaleSet{}
+	if err := vmss.UnmarshalJSON(data); err != nil {
+		return nil, err
+	}
+
+	extensions := *vmss.VirtualMachineScaleSetProperties.VirtualMachineProfile.ExtensionProfile.Extensions
+	for _, ext := range extensions {
+		if ext.Name != nil && *ext.Name == "cse" {
+			return ext.ProtectedSettings.(map[string]interface{}), nil
+		}
+	}
+
+	return nil, nil
+}
+
+func getCount(role string, cs *api.OpenShiftManagedCluster) (int, error) {
 	for _, agentProfiles := range cs.Properties.AgentPoolProfiles {
 		if string(agentProfiles.Role) == role {
 			return agentProfiles.Count, nil
@@ -116,7 +188,7 @@ func main() {
 		log.Fatal(err)
 	}
 
-	ssu, err := getUpgrader(*newCsPath, *oldCsPath)
+	ssu, err := getUpgrader(*csPath, *oldCsPath)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/hack/upgrade.sh
+++ b/hack/upgrade.sh
@@ -47,8 +47,10 @@ export KUBECONFIG=_data/_out/admin.kubeconfig
 go run cmd/vmssrollout/vmssrollout.go -subscription $AZURE_SUBSCRIPTION_ID \
                                       -resource-group $RESOURCEGROUP \
                                       -name ss-master \
-                                      -new-config _data/containerservice.yaml \
+                                      -config _data/containerservice.yaml \
                                       -old-config _data/containerservice_old.yaml \
+                                      -template-file _data/_out/azuredeploy.json \
+                                      -old-template-file _data/_out/azuredeploy_old.json \
                                       -in-place \
                                       -role master
 
@@ -56,8 +58,10 @@ go run cmd/vmssrollout/vmssrollout.go -subscription $AZURE_SUBSCRIPTION_ID \
 go run cmd/vmssrollout/vmssrollout.go -subscription $AZURE_SUBSCRIPTION_ID \
                                       -resource-group $RESOURCEGROUP \
                                       -name ss-infra \
-                                      -new-config _data/containerservice.yaml \
+                                      -config _data/containerservice.yaml \
                                       -old-config _data/containerservice_old.yaml \
+                                      -template-file _data/_out/azuredeploy.json \
+                                      -old-template-file _data/_out/azuredeploy_old.json \
                                       -drain \
                                       -role infra
 
@@ -65,8 +69,10 @@ go run cmd/vmssrollout/vmssrollout.go -subscription $AZURE_SUBSCRIPTION_ID \
 go run cmd/vmssrollout/vmssrollout.go -subscription $AZURE_SUBSCRIPTION_ID \
                                       -resource-group $RESOURCEGROUP \
                                       -name ss-compute \
-                                      -new-config _data/containerservice.yaml \
+                                      -config _data/containerservice.yaml \
                                       -old-config _data/containerservice_old.yaml \
+                                      -template-file _data/_out/azuredeploy.json \
+                                      -old-template-file _data/_out/azuredeploy_old.json \
                                       -drain \
                                       -role compute
 
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -87,6 +87,19 @@ func createOrUpdate(oc *v1.OpenShiftManagedCluster, entry *logrus.Entry) (*v1.Op
 		return nil, err
 	}
 
+	log.Info("read old ARM template")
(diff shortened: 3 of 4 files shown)


kargakis:
Got master upgrades working. You can test out with:
```
make vmss
./vmssrollout -subscription-id $SUB_ID -resource-group $RESOURCEGROUP -name ss-master -max-unavailable 1 --in-place --replicas 3

**Comment by kargakis** [comment]
I think we need a separate interface than `Plugin` to wait for VM readiness during the upgrade. The problem is that WaitForReadiness needs to be unique based on the type of the scaleset we are upgrading (master VMs are considered ready when the API server running in the VM is healthy, compute/infra VMs are ready when they have joined the k8s cluster and are in Ready state). So I am proposing something along the lines of the following interface/implementations:

```go
// pkg/api/plugin.go

type Upgrade interface {
	WaitForReadiness(ctx context.Context, scaleSetName, instanceID string) error
}
```
```go
// pkg/plugins/plugins.go
type MasterUpgrade struct {
VirtualMachineScaleSetVMsClient // azure-sdk
dynamic.Interface // client-go
}

var _ api.Upgrade = &MasterUpgrade{}

func (u *MasterUpgrade) WaitForReadiness(ctx context.Context, scaleSetName, instanceID string) error {

}

type InfraUpgrade struct{
VirtualMachineScaleSetVMsClient // azure-sdk
dynamic.Interface // client-go
}

var _ api.Upgrade = &InfraUpgrade{}

func (u *InfraUpgrade) WaitForReadiness(ctx context.Context, scaleSetName, instanceID string) error {

}

type ComputeUpgrade struct{
VirtualMachineScaleSetVMsClient // azure-sdk
dynamic.Interface // client-go
}

var _ api.Upgrade = &ComputeUpgrade{}

func (u *ComputeUpgrade) WaitForReadiness(ctx context.Context, scaleSetName, instanceID string) error {

}
```

kargakis:
I've got readiness checks working; last thing is node draining.

pkg/vmssrollout/upgrade.go
```diff
@@ -0,0 +1,325 @@
...
+			// Wait for VM readiness; this may vary between VMSSs, eg. a master VM
+			// is ready when the API server pod running on it is ready whereas a compute
+			// VM is ready when it has joined the cluster as a Ready k8s node.
+			// TODO: Figure out the new node name
+			//if err := u.Plugin.WaitForReadiness(context.Background(), nodeName); err != nil {
+			//	return err
+			//}
+		}
+
+		canScaleDown, err := u.canScaleDown()
```
mjudeikis: why we are checking if we can scale down? Is there any reason why we should not be able to scale down?

pkg/vmssrollout/upgrade.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			return err
+		}
+		if !canScaleDown {
+			// TODO: Make this configurable.
+			time.Sleep(5 * time.Second)
+			continue
```
kargakis: This is wrong, I should't continue here since I am skipping old VMs.

pkg/vmssrollout/upgrade.go
```diff
@@ -0,0 +1,324 @@
...
+
+	for _, vmToUpgrade := range vmsToUpgrade {
+		// if possible, scale up first
+		canScaleUp, err := u.canScaleUp()
+		if err != nil {
+			return err
+		}
+		if canScaleUp {
+			log.Printf("Scale up VMSS %q", u.Name)
+			*vmss.Sku.Capacity = u.DesiredReplicas + u.MaxSurge
```
kargakis: I will change u.MaxSurge to 1 for now and I'll open an issue about extending it in the future.

kargakis:
I have simplified things a bit to ease reviews by removing the rolling parameters (maxSurge and maxUnavailable) in favor of in-place (the lack of an in-place upgrade now implicitly means scale up by 1 first). I hope to finish node drain today.

kargakis:
Draining is in place now, this should be ready for reviews @amanohar @openshift/sig-azure

kargakis:
azure sdk split into its own PR: https://github.com/openshift/openshift-azure/pull/169

pkg/plugin/plugin.go
```diff
@@ -50,3 +55,77 @@ func (p *Plugin) GenerateARM(cs *acsapi.ContainerService) ([]byte, error) {
 func (p *Plugin) HealthCheck(ctx context.Context, cs *acsapi.ContainerService) error {
 	return healthcheck.HealthCheck(ctx, cs)
 }
+
```
kargakis: @openshift/sig-azure ptal in the plugin code.

pkg/vmssrollout/drain.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+const cordonMaxRetries = 5
+
+func (u *VMSSUpgrader) drain(nodeName string) error {
+	var node *v1.Node
+	var err error
+
+	var cordoned bool
+	for i := 0; i < cordonMaxRetries; i++ {
```
mjudeikis: What the reason we try make node unschedulable few times? Does it fails sometimes?

kargakis: I guess it can fail on a conflict but I haven't seen anything so far. This part is a straight copy from acs-engine (although I continue in all error cases since we only retry 5 times).

pkg/vmssrollout/upgrade.go
```diff
@@ -0,0 +1,342 @@
...
+			// VM is ready when it has joined the cluster as a Ready k8s node.
+			nodeName := *vmToUpgrade.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName
+			if err := u.waitForReady(nodeName); err != nil {
+				return err
+			}
+			continue
+		}
+
+		log.Printf("Deleting VM %q in VMSS %q", *vmToUpgrade.InstanceID, u.Name)
+		res, err := u.vmc.Delete(context.Background(), u.ResourceGroup, u.Name, *vmToUpgrade.InstanceID)
```
mjudeikis: If we did InPace upgrade, why we should want to delete node?

kargakis: we don't delete on in-place

mjudeikis: And how azure behaves if you scale up to 4, and delete 1? will it initiates 1 creation again or its just leaves 3?

kargakis: VMSS is resized to 3.

pkg/plugin/plugin.go
mjudeikis: I like it. The only thing I miss is some high-level comment (maybe close to interfaces) on how each flow works. `inplace`, `reimage`, `scale-up scale-down`. Some code level docs on code capabilities and ideas behind.

kargakis: Scaling operations are orthogonal to plugin readiness but I can add some docs about the former.

pweil-: plugin structure looks good to me

cmd/vmssrollout/vmssrollout.go
```diff
@@ -18,11 +22,9 @@ var (
 	name           = flag.String("name", "", "Name of the Virtual Machine Scale Set to upgrade")
 
 	// upgrade parameters
-	count   = flag.Int64("count", -1, "Desired replica count of the scale set")
-	sku     = flag.String("sku", "", "New imageReference SKU to upgrade to")
-	version = flag.String("version", "", "New imageReference version to upgrade to")
-	id      = flag.String("id", "", "New imageReference id to upgrade to")
-	inPlace = flag.Bool("in-place", false, "Perform an in-place upgrade")
+	newCsPath = flag.String("new-config", "", "Path to the new container service config")
```
pweil-: aha, this is what I was looking for in the first commit!

hack/upgrade.sh
```diff
@@ -36,37 +36,30 @@ go generate ./...
 go run cmd/createorupdate/createorupdate.go
```
pweil-: ok, so now we have RP simulation of 

1.  new manifest comes in
2.  rp calls validateInternal (when merged)
3.  rp calls generate
4.  rp calls generateARM??  
5.  rp calls vmssrollout??
6.  new sync pod is deployed

I'm not clear on steps 4 and 5, I think we need to nail down the flow there.  It seems like we're making the assumption that ARM never changes and it's only vmss updates.

kargakis: I don't want to put words in @jim-minter's mouth but from what I inferred from our last conversation was that for PP it would be enough to have scaleset upgrades and we could fix applying the rest of the ARM template before GA. Even with something that applies the ARM template in place, we still need vmssrollout to manage VMSS rolling upgrades, since we cannot use the built-in rolling upgrade feature for a couple of reasons (lack of HTTPS readiness probes, and I am not sure whether in-place upgrades are supported).

kargakis:
Looking into a weird error when trying to use the kubernetes client
```
2018/08/09 15:10:07 Get https://mkargakisosa.eastus.cloudapp.azure.com/api/v1/namespaces/kube-system/pods/etcd-master-000000: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "openshift-signer")
```

pkg/vmssrollout/upgrade.go
```diff
@@ -0,0 +1,338 @@
...
+		if err != nil {
+			return fmt.Errorf("cannot delete VM %q: %v", *vmToUpgrade.InstanceID, err)
+		}
+		if err := res.WaitForCompletion(context.Background(), u.vmc.BaseClient.Client); err != nil {
```
kargakis: Just saw a VM deletion take more than 15m so we probably don't want to be waiting here.
```
2018/08/10 12:45:25 Deleting VM "0" in VMSS "ss-infra"
2018/08/10 13:00:25 cannot wait for VM "0" deletion: Future#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded
exit status 1
```

pkg/upgrade/drain.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	node.Spec.Unschedulable = cordon
+	_, err = u.kc.CoreV1().Nodes().Update(node)
+	return err
+}
+
+func (u *VMSSUpgrader) uncordon(nodeName string) error {
+	var updated bool
+	for i := 0; i < cordonMaxRetries; i++ {
+		if err := u.updateNode(nodeName, false); err == nil {
```
kargakis: !=

pkg/api/plugin.go
```diff
@@ -15,9 +15,17 @@ type Plugin interface {
 	// ValidateInternal is called and that it makes sense to do this.
 	ValidateInternal(new, old *ContainerService) []error
 
-	GenerateConfig(cs *ContainerService) error
+	// GenerateConfig ensures all the necessary in-cluster config is generated
+	// for an Openshift cluster. It performs merging of pre-existing old config
+	// to avoid incorrect generation of config that may not need to be regenerated
+	// during an upgrade (eg. certificates).
+	GenerateConfig(new, old *ContainerService) error
```
kargakis: remove

cmd/vmssrollout/vmssrollout.go
```diff
@@ -0,0 +1,151 @@
...
+func readFile(path string) ([]byte, error) {
+	if len(path) == 0 {
+		return nil, errors.New("a path must be specified")
+	}
+	return ioutil.ReadFile(path)
+}
+
+func getCount(role string, cs *api.ContainerService) (int, error) {
+	if role == "master" {
+		if cs.Properties.MasterProfile == nil {
```
kargakis: MasterProfile is no more internally

cmd/vmssrollout/vmssrollout.go
```diff
@@ -0,0 +1,151 @@
...
+	var newCs, oldCs *api.ContainerService
+	if err := yaml.Unmarshal(newCsBytes, &newCs); err != nil {
+		return nil, err
+	}
+	if err := yaml.Unmarshal(oldCsBytes, &oldCs); err != nil {
+		return nil, err
+	}
+
+	var imageRef *compute.ImageReference
+	// TODO: Validate updates
```
kargakis: Fixed elsewhere

pkg/upgrade/drain.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("failed to uncordon node %q", nodeName)
+	}
+	log.Printf("Node %q has been marked schedulable.", nodeName)
+	return nil
+}
+
+func (u *VMSSUpgrader) drain(nodeName string) error {
+	var cordoned bool
+	for i := 0; i < cordonMaxRetries; i++ {
+		if err := u.updateNode(nodeName, true); err == nil {
```
kargakis: !=

hack/upgrade.sh
kargakis: FWIW, I can't see how acs-engine applies azuredeploy.json changes on upgrades. They seem to be doing mere VM upgrades, too.

kargakis:
@openshift/sig-azure ptal

cmd/vmssrollout/vmssrollout.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/ghodss/yaml"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/plugin"
+	"github.com/openshift/openshift-azure/pkg/upgrade"
+)
+
+var (
+	// vmss parameters
+	subscriptionID = flag.String("subscription-id", "", "Azure subscription ID")
```
mjudeikis: nit: `s/--subscription-id/--subscription` ? follow az client pattern? 

`--resource-group` already match?

cmd/vmssrollout/vmssrollout.go
```diff
@@ -0,0 +1,143 @@
...
+		newCs.Config.ImageVersion != oldCs.Config.ImageVersion {
+		imageRef = &compute.ImageReference{
+			Sku:     &newCs.Config.ImageSKU,
+			Version: &newCs.Config.ImageVersion,
+		}
+	}
+
+	if newCs.Config.ImageResourceGroup != oldCs.Config.ImageResourceGroup ||
+		newCs.Config.ImageResourceName != oldCs.Config.ImageResourceName {
+		id := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/images/%s",
```
mjudeikis: `/images/` - if we want to test with our custom images not the in `images` RG? Its wider question is `images` is our dev RG or we will publish stuff from it...

hack/upgrade.sh
mjudeikis: I think if case like this appears we will just need to deploy new arm template. Should we move from `azuredeploy` as a name inside ARM to something like `OSA-Vxxxx`? This would allow us to have better tracking of arm deployed artifacts?

cmd/vmssrollout/vmssrollout.go
mjudeikis: Ignore, `/images` confused me.

mjudeikis:
lgtm, do we still need sign off for this?

kargakis:
@openshift/sig-azure script extensions are not applied correctly, I think we will need to diff between old and new azuredeploy.json as well as diffing between old and new containerservice.yaml (comparing latest azuredeploy.json with resources retrieved with the golang sdk is not going to fly)

kargakis:
@openshift/sig-azure @amanohar last commit is a strawman for upgrading in case of a diff in a scale set cse(.ProtectedSettings), thoughts?

cmd/vmssrollout/vmssrollout.go
```diff
@@ -101,6 +112,78 @@ func readFile(path string) ([]byte, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	if equality.Semantic.DeepEqual(newScript, oldScript) {
+		return nil, nil
+	}
+	return newScript, nil
+}
+
+func readScriptFromTemplate(path string) (*string, error) {
```
kargakis: Reading from ARM to a compute.VirtualMachineScaleSet is ugly, may be simpler to walk raw maps.

kargakis:
Getting a weird deserialization error when trying to update a CustomScriptExtension
```
INFO[0069] In-place upgrade VM "0" in VMSS "ss-master"  
2018/08/15 11:37:24 cannot wait for VM "0" update: Code="VMExtensionProvisioningError" Message="VM has reported a failure when processing extension 'cse'. Error message: \"Enable failed: failed to get configuration: error reading extension configuration: failed to parse protected settings: failed to unmarshal decrypted settings json: json: cannot unmarshal string into Go value of type map[string]interface {}\"."
exit status 1
```
@amanohar @sozercan @shrutir25

kargakis:
CSE is serialized as a string in a VMSS but VMs require it to be a map[string]interface{}?

kargakis:
Moving VM readiness checks into its own PR: https://github.com/openshift/openshift-azure/pull/198

kargakis:
Opened https://github.com/Azure/azure-sdk-for-go/issues/2466 for the empty error message.

kargakis:
Figured out the issue with the CSE update. I also realized that the master script will need some massaging for in-place upgrades.

jim-minter:
think it's probably worth refactoring this a little to separate the notion of pre-hook / workflow / post-hook

PR #139, opened 2018-08-01
Remove unnecessary go-test dependency

kargakis:
@openshift/sig-azure


kargakis:
Nevermind, we also use this for outputing diffs

PR #142, opened 2018-08-01
Move helper scripts in hack/

kargakis:
ref https://github.com/openshift/openshift-azure/pull/118#issuecomment-408511468

/hold


kargakis:
@openshift/sig-azure

PR #143, opened 2018-08-01
remove acs-engine dependency

pweil-:
@openshift/sig-azure 

TODO

- [x] remove unused api items that are just commented out right now
- [x] remove fields that aren't used from structs that are used

Follow ups: meld Config into the internal API struct (follow up PR)


pkg/api/types.go
```diff
@@ -0,0 +1,326 @@
...
+	VMSize         string `json:"vmSize" validate:"required"`
+	OSDiskSizeGB   int    `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1023"`
+	Username       string `json:"username,omitempty"`
+	PublicKey      string `json:"publicKey" validate:"required"`
+	StorageProfile string `json:"storageProfile,omitempty"`
+}
+
+// KubernetesConfig contains the Kubernetes config structure, containing
+// Kubernetes specific configuration
+type KubernetesConfig struct {
```
kargakis: I think we can get rid of this now; had the same comment in https://github.com/openshift/openshift-azure/pull/130

kargakis:
I would trim the API down to just what we use today. Other than that, I am all for merging.

pkg/api/types.go
```diff
@@ -30,36 +30,25 @@ type ContainerService struct {
...
+	OrchestratorProfile     *OrchestratorProfile     `json:"orchestratorProfile,omitempty"`
+	MasterProfile           *MasterProfile           `json:"masterProfile,omitempty"`
+	AgentPoolProfiles       []*AgentPoolProfile      `json:"agentPoolProfiles,omitempty"`
 	ServicePrincipalProfile *ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
-	AzProfile *AzProfile `json:"azProfile,omitempty"`
+	AzProfile               *AzProfile               `json:"azProfile,omitempty"`
 }
 
 // AzProfile holds the azure context for where the cluster resides
 type AzProfile struct {
```
kargakis: I thought we were getting rid of these options (they wont be provided by end users) and the RP would enrich the config with them?

pweil-: still used on the cmd type to simulate enrichment so we need them on the internal struct

pweil-: https://github.com/pweil-/openshift-azure/blob/d438b41ea1a2df590f208dfc388ee62352808a84/cmd/createorupdate/createorupdate.go#L106

pkg/api/types.go
```diff
@@ -85,117 +74,12 @@ const (
...
-
-// CustomFile has source as the full absolute source path to a file and dest
-// is the full absolute desired destination path to put the file on a master node
-type CustomFile struct {
-	Source string `json:"source,omitempty"`
-	Dest   string `json:"dest,omitempty"`
-}
-
 // OpenShiftConfig holds configuration for OpenShift
 type OpenShiftConfig struct {
```
kargakis: I don't think we use ConfigBundles anymore. Also the htpasswd options are not used, too, but probably we want a follow-up issue to use them? Same question for EnableAADAuthentication.

pweil-: nope, you're right, removing.  I went with the assumption that OpenShiftConfig was correct and didn't look at those.  Thanks!

pkg/api/types.go
kargakis: Hrm, so if these are our internal types, when do we plan to add the external? Can at least have a comment here to mention that this is only used internally and should be removed from external once we copy.

pweil-: external is in v1.  We need the RP to enrich the internal type for us and save it.  I'm not sure we need it on the external type.  WDYT?  I was thinking that if we state the RP will *always* enrich those fields then we don't need to worry about stability in conversion and if we (SRE) are always looking at the saved type then the values should always be there

pweil-: https://github.com/openshift/openshift-azure/pull/143/commits/1707b273cd9d0172bf0e3c49a6a9bb04df23784e for quick reference to v1 :smile:

kargakis: cool, not sure how I missed that. We don't need it in the external type.

kargakis:
lgtm, squash and add the label

pweil-:
need to rebase :smile:

pkg/api/types.go
```diff
@@ -0,0 +1,151 @@
...
+
+// TypeMeta describes an individual API model object
+type TypeMeta struct {
+	// APIVersion is on every object
+	APIVersion string `json:"apiVersion"`
+}
+
+// ResourcePurchasePlan defines resource plan as required by ARM
+// for billing purposes.
+type ResourcePurchasePlan struct {
```
amanohar: This is optional and can be removed.

pkg/api/types.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Properties represents the ACS cluster definition
+type Properties struct {
+	ProvisioningState       ProvisioningState        `json:"provisioningState,omitempty"`
+	OrchestratorProfile     *OrchestratorProfile     `json:"orchestratorProfile,omitempty"`
+	MasterProfile           *MasterProfile           `json:"masterProfile,omitempty"`
+	AgentPoolProfiles       []*AgentPoolProfile      `json:"agentPoolProfiles,omitempty"`
+	ServicePrincipalProfile *ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
+	AzProfile               *AzProfile               `json:"azProfile,omitempty"`
```
amanohar: @pweil- `AzProfile` doesn't need to be in `Properties` as the RP already has this metadata and can be passed separately to the OSA plugin.

pkg/api/types.go
```diff
@@ -0,0 +1,151 @@
...
+
+// OpenShiftRouterProfile represents an OpenShift router.
+type OpenShiftRouterProfile struct {
+	Name            string
+	PublicSubdomain string
+	FQDN            string
+}
+
+// MasterProfile represents the definition of the master cluster
+type MasterProfile struct {
```
amanohar: Master doesn't have a `VNET` defined. Is this because in the custom VNET scenario we are expecting the master to be a different VNET & subnet that is going to created on the fly instead of the one specified by customer?

pkg/api/types.go
```diff
@@ -0,0 +1,151 @@
...
+	// Master LB public endpoint/FQDN with port
+	// The format will be FQDN:2376
+	// Not used during PUT, returned as part of GET
+	FQDN string `json:"fqdn,omitempty"`
+}
+
+// Extension represents an extension definition in the master or agentPoolProfile
```
amanohar: This can be removed. I don't think it's using anywhere.

pkg/api/types.go
pweil-: See this conversation: https://github.com/openshift/openshift-azure/pull/143#discussion_r207224271  

I think if we can settle on how it is enriched (either by throwing it on the converted profile or as part of interface method args) we will know if it can be removed.  Which do you prefer?

cc @kargakis

pkg/api/types.go
pweil-: :+1:

PR #152, opened 2018-08-03
move to internal DNS name for certs and selfsign for console/external

mjudeikis:
Move to internal DNS and self sign certs for console 

My working branch of https://github.com/openshift/openshift-azure/pull/128/files is getting too big and want to split and start merging

cc: @openshift/sig-azure  PTAL


pkg/checks/checks.go
```diff
@@ -32,6 +32,9 @@ func WaitForHTTPStatusOk(ctx context.Context, transport http.RoundTripper, urlto
 		resp, err := cli.Do(req)
 		if err, ok := err.(*url.Error); ok {
 			if err, ok := err.Err.(*net.OpError); ok {
+				if _, ok := err.Err.(*net.DNSError); ok {
+					return false, nil
```
kargakis: Is this valid? I think @jim-minter wants DNS errors to be permanent failures (see https://github.com/openshift/openshift-azure/pull/131)

pkg/config/generate.go
```diff
@@ -316,6 +312,16 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
...
 		},
+		{
+			cn: cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname,
+			dnsNames: []string{
+				cs.Properties.OrchestratorProfile.OpenShiftConfig.PublicHostname,
+			},
+			extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+			key:         &c.ConsoleServerKey,
+			cert:        &c.ConsoleServerCert,
+			selfSign:    true,
```
kargakis: Can you add a comment as to why we want the console key to be self-signed?

pkg/checks/checks.go
mjudeikis: I think chose to merge https://github.com/openshift/openshift-azure/pull/131 we good. Need to test with POD approach, but at the moment it fails 100% of times when using a new dns approach because of DNS creation race condition. So it needs to be handled for sure. Question how.

mjudeikis:
@kargakis @pweil-  ptal. Removed checking as #131  will cover this part too.

pkg/config/generate.go
```diff
@@ -316,6 +312,17 @@ func Generate(cs *acsapi.ContainerService, c *Config) (err error) {
 			key:         &c.Certificates.Registry.Key,
 			cert:        &c.Certificates.Registry.Cert,
 		},
+		// Openshift Console is BYO type of certificate. In a long run we should enable users to configure their own certificates. For this reason we decouple it from all OCP certs and make it self-sign
```
kargakis: s/In a long run/In the long run/

Also, can you break this comment down to two or three separate lines?

mjudeikis: Done :) on my screen it takes only 1/3 of the screen :D

pkg/arm/data/master-startup.sh
```diff
@@ -288,6 +290,11 @@ servingInfo:
   keyFile: master.server.key
   maxRequestsInFlight: 500
   requestTimeoutSeconds: 3600
+  namedCertificates:
```
mjudeikis: When web-console will be officially accessed via route we will remove this and move to certs inside the secret mode

mjudeikis:
@openshift/sig-azure  please review :)

PR #155, opened 2018-08-03
move all external validation to internal and remove plugin interface …

pweil-:
…for external validation.

@openshift/sig-azure 
@amanohar


pweil-:
@kargakis @mjudeikis bump

@amanohar need sign off on interface change

cmd/createorupdate/createorupdate.go
```diff
@@ -56,7 +50,7 @@ func createOrUpdate(oc *v1.OpenShiftCluster) (*v1.OpenShiftCluster, error) {
 
 	// validate the internal API representation (with reference to the previous
 	// internal API representation)
-	errs = p.ValidateInternal(cs, oldCs)
+	errs := p.ValidateInternal(cs, oldCs)
```
amanohar: Here is an example of logger that's passed by the service to ACS Engine: https://github.com/Azure/acs-engine/blob/master/pkg/operations/kubernetesupgrade/upgrader.go#L45. It's from: `github.com/sirupsen/logrus`

pweil-: :+1: I've created a card for this since it applies to all of the methods.  Will work on that this sprint

pweil-: all good on the removal of `ValidateExternal` that this introduces? https://github.com/openshift/openshift-azure/pull/155/files#diff-866712c7c0e6cc75c39b753f2976c400L11

kargakis: +1 on the use of logrus. We should import it and change it across all the codebase, so it needs to be a separate PR.

mjudeikis: +1 For logrus :D my favorite one

pweil-:
@amanohar bump for sign off on the interface change in this one

pkg/validate/container_service.go
```diff
@@ -1,10 +1,258 @@
...
+	"default": struct{}{},
+}
+
+func isValidHostname(h string) bool {
+	return len(h) <= 255 && rxRfc1123.MatchString(h)
+}
+
 // ContainerService validates a ContainerService struct
-func ContainerService(new, old *acsapi.ContainerService) (errs []error) {
+func ContainerService(new, old *api.ContainerService) (errs []error) {
```
amanohar: This should be updated to `OpenShiftCluster` or something similar.

pkg/validate/container_service.go
```diff
@@ -1,10 +1,258 @@
...
+		if app.Count != 3 {
+			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].count %q", app.Name, app.Count))
+		}
+
+	default:
+		errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].role %q", app.Name, app.Role))
+	}
+
+	switch app.VMSize {
+	case "Standard_D2s_v3",
```
amanohar: Are these the only two VM size OSA will allow? And why?

pkg/plugin/plugin.go
```diff
@@ -19,10 +18,6 @@ type Plugin struct{}
 
 var _ api.Plugin = &Plugin{}
 
-func (p *Plugin) ValidateExternal(oc *v1.OpenShiftCluster) []error {
```
amanohar: The reason for removing this is to combine Create and Update validations into one? Also create validation will happen only on converted internal model and not the external request from the customer?

pkg/validate/container_service.go
pweil-: this is moving all validation to internal.  Internal data type is `ContainerService`.  We've talked about renaming it but I don't think it should be part of this PR

pkg/plugin/plugin.go
pweil-: No, this was based on our architecture discussion where we agreed to only validate the internal types.  Step 7 of the create process and step 6 of the upgrade process in the document.

pkg/validate/container_service.go
pweil-: These are the sizes we know are min required, we should allow anything bigger though.  Not really relevant to this code move (wasn't added in this PR, just moved).  We can expand if necessary.

pkg/plugin/plugin.go
pweil-: My response may be ambiguous :smile: 

> reason for removing this is to combine Create and Update validations into one

No, this was based on our architecture discussion where we agreed to only validate the internal types. 

> create validation will happen only on converted internal model and not the external request from the customer?

Correct, as discussed, MSFT will do a basic validation on external during create and update - Step 4 and 3 respectively in the doc.  MSFT and RH will do internal validations - step 7 and 6 respectively

amanohar: Step 4 was for MSFT to do validation on things like core quota count, SPN, or maybe DNS FQDN in the future. But validation for others parameters in PUT request body still needs to be done which should be in the plugin.

If this combined how will we know which what is create vs update?

pweil-: `ValidateInternal(new, old *ContainerService) []error`

if there is an old container service it is an update

Issue #156, opened 2018-08-03
Generated template has all hard coded values and is not configurable by passing paramerters


amanohar:
Generated template using `create.sh` has all hard coded values and is not configurable by passing parameters. This poses a challenge while doing development & testing via a managed app.

This is because each new scenario including a new dns name of a new cluster requires publishing a managed application which can take at least an hour in the best case scenario and could be longer. This slows down developer work flow. 

ARM template should be generated with parameters so that configurable values can be passed in during cluster/managed app creation (and is also the right & recommended way to use ARM templates). Note that this is needed while proxy resource work in under progress.

If the template can be parameterized to accept the following values instead of hard coding them then the same managed app can be used for multiple deployments:

- domainNameLabel/publicHostname
- publicSubdomain
- resource group name
- region
- clientID
- client secret

Making domainNameLabel, publicHostname, publicSubdomain, resource group name parameterized is most important for easy testing.

pweil-:
@amanohar I took a stab at getting some of this done in https://github.com/openshift/openshift-azure/pull/160.  PTAL

To address this bug directly, there are two considerations:

1.  parameterization of the ARM template is designed specifically so that you can deploy the same template over and over again with command line (or uploading) and simply changing the parameter file.  OSA has no intention of allowing publication of this template in that manner so current substitution seems fine

2.  the non-proxy app is a stop gap, we should not be designing around it.  My understanding is it is just enough to validate the deployment model

Adding the params in the PR covers all but client id and secret.  Those are embedded in the startup scripts so I don't have a good solution for abstracting them into something you can set without changing the signature of the extension to pass it in which seems like more than we should be doing in light of item 2.  

Are the parameters in #160 enough for you to validate that your deployment model?  I am inclined to not merge that PR since we don't get long term benefits from it (though it doesn't complicate anything since it's confined to 1 file at the moment)

pweil-:
Closing this based on scrum discussions.  We can reopen if necessary

PR #160, opened 2018-08-05
parameterize some of the arm template

pweil-:
@amanohar - takes a stab at adding parameters for some of the fields you mentioned in #156


mjudeikis:
Not the biggest fan of this as now we will have `{{- if eq $.Config.ImageResourceName "" }}` and parameters all around the place. Preference would be to stick to single method for injecting data?

pweil-:
Closing this based on scrum discussions.  We can reopen if necessary

PR #162, opened 2018-08-06
Add FQDN handling for master and router

mjudeikis:
Builds on top of https://github.com/openshift/openshift-azure/pull/152 as pre-req.
You are interested only in 1 commit:
https://github.com/openshift/openshift-azure/pull/162/commits/6ec47c43f1f556b8db083a42ae8957724db9244b 

@amanohar,  @openshift/sig-azure  ptal. This is DNS integration proposal.
We would you to provide us 2 inputs from MS side:
```
cs.Properties.MasterProfile.FQDN = random-cluster-master-prefix.eastus.cloudapp.azure.com
cs.Properties.OrchestratorProfile.OpenShiftConfig.RouterProfiles[0].FQDN = random-cluster-router-prefix.eastus.cloudapp.azure.com
	// We will:
	//    extract random-cluster-master-prefix and put it into ARM to use this for public-ip
	//    extract random-cluster-router-prefix and put it into service annotation so cloud-plugin could claim DNS name when provisioning IP
```
On MS side you would need to check if DNS is available and if we can use them from cloud-provider and ARM level.


cmd/createorupdate/createorupdate.go
```diff
@@ -121,16 +120,16 @@ func enrich(cs *acsapi.ContainerService) error {
 		return fmt.Errorf("must set RESOURCEGROUP")
 	}
 
+	// TODO: MS to provide FQDN values for OCP cluster to use.
```
kargakis: Shouldn't you simulate these steps in `hack/create.sh`?

pkg/config/generate.go
```diff
@@ -584,3 +584,18 @@ func randomString(length int) (string, error) {
 
 	return string(b), nil
 }
+
+func selectDNSNames(cs *acsapi.ContainerService, c *Config) {
```
kargakis: Config is no more. Also, it would be nice to start seeing unit tests about these functions. We can start with this one.

cmd/createorupdate/createorupdate.go
mjudeikis: yes. But all enrich function should be gone either way. Just choosing my tool until we agree on the right way.

mjudeikis:
@pweil-  @kargakis @kwoodson Im basing this on @pweil- change as Config is not more. If you have time review please only parts from DNS stuff. 

+ I didnt added dependencies here, as it would add few thousand files more. Will do this is separate PR.

mjudeikis:
@sozercan @amanohar  FIY for the review :)

pkg/checks/checks.go
```diff
@@ -168,3 +170,45 @@ func WaitForInfraServices(ctx context.Context, appclient *appclient.AppsV1Client
...
+	//check router FQDN by quering k8s LoadBalancer and ipAddress associated with it
+	rs, err = rc.Get(ctx, cs.Properties.AzProfile.ResourceGroup, "Microsoft.Network", "", "loadBalancers", "kubernetes")
+	if err != nil {
+		return err
+	}
+
+	// iterate over k8s LoadBalancers and check if we have a match. Match should always exist, so failed by default.
+	// TODO: With multiple router this need to check multiple conditions
+	exist := false
+	for _, v := range rs.Properties.(map[string]interface{})["frontendIPConfigurations"].([]interface{}) {
```
amanohar: Instead of using generic Azure Go client its more efficient to use SDK of a specific provider: https://github.com/Azure/azure-sdk-for-go/blob/master/services/network/mgmt/2018-05-01/network/loadbalancerfrontendipconfigurations.go#L49.

This way you won't need to cast into map etc. and you a get typed object.

mjudeikis: Gave it a shot. Not sure if this makes this look better. Logic still stays the same, but now I need to handle 2 separate Azure clients (`eipc network.PublicIPAddressesClient, lbc network.LoadBalancerFrontendIPConfigurationsClient`). 

I prefre this casting solution over having 2 clients. As soon as you have more than 2 clients, `struct` is good to abstract. Next think both packages `healthcheck` and `check` needs access to it, this means I need to externalise it to separate package so I could avoid circular imports. I think if at any point we decide to externalise all clients to separate package this would make sense, now - dont know.

@kargakis @pweil-  what do you think? 

Example:
https://gist.github.com/mjudeikis/1123f67e262c0d2559e4cbb2e30e6720
Clients:
https://gist.github.com/mjudeikis/1123f67e262c0d2559e4cbb2e30e6720

mjudeikis:
@sozercan @amanohar @kargakis  bumped with typed client as requested

pkg/checks/checks.go
```diff
@@ -12,6 +12,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-05-01/network"
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
```
kargakis: nit: separate into its own group

pkg/checks/checks.go
```diff
@@ -168,3 +170,43 @@ func WaitForInfraServices(ctx context.Context, appclient *appclient.AppsV1Client
...
+	lbIPList, err := lbc.List(ctx, cs.Properties.AzProfile.ResourceGroup, "kubernetes")
+	if err != nil {
+		return err
+	}
+	// iterate over k8s LoadBalancer and check if we have a match. Match should always exist, so failed by default.
+	// TODO: With multiple routers this need to check multiple conditions
+	exist := false
+	for _, v := range lbIPList.Values() {
+		//spew.Dump(v.PublicIPAddress.ID)
+		ipAddressName := strings.Split(*v.PublicIPAddress.ID, "/")[len(strings.Split(*v.PublicIPAddress.ID, "/"))-1]
```
kargakis: Needs to be dryed, should also be more readable. Do the strings.Split once:
```go
parts := strings.Split(*v.PublicIPAddress.ID, "/")
ipAddressName := parts[len(parts)-1]
```

pkg/healthcheck/client.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-05-01/network"
+
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
+)
+
+type AzureClients struct {
```
kargakis: Use a more specific name like `networkClient`

pkg/healthcheck/healthcheck.go
```diff
@@ -103,11 +93,43 @@ func waitForConsole(ctx context.Context, cs *acsapi.ContainerService) error {
 
 		switch resp.StatusCode {
 		case http.StatusOK:
+			fmt.Println("OK")
```
kargakis: remove

kargakis: or use logrus

mjudeikis: second.

pkg/healthcheck/client.go
mjudeikis: At the moments its only 2 clients, potentially more if we move to separate package. Its generic name?

kargakis: No reason for a centralized struct that holds all clients. I would not want to be importing networking libraries where I don't need them.

kargakis:
Can you also separate the dependency bump into its own commit? Should be cleaner to merge.

pkg/checks/checks.go
```diff
@@ -168,3 +170,43 @@ func WaitForInfraServices(ctx context.Context, appclient *appclient.AppsV1Client
 
 	return nil
 }
+
+func CheckDNS(ctx context.Context, eipc network.PublicIPAddressesClient, lbc network.LoadBalancerFrontendIPConfigurationsClient, cs *acsapi.ContainerService) error {
+
+	// get master FQDN from ip-apiserver loadbalancer
+	ip, err := eipc.Get(ctx, cs.Properties.AzProfile.ResourceGroup, "ip-apiserver", "")
```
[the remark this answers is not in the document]
pweil-: Yes

pkg/checks/checks.go
pweil-: yeah, typed is the way to go.  Compile time safety wins out here :-)

mjudeikis:
> Can you also separate the dependency bump into its own commit? Should be cleaner to merge.

Lets leave this for next time, please :)

pkg/checks/checks.go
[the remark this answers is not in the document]
mjudeikis: Will rebase when it goes in

kargakis:
@amanohar can you try /lgtm and see if it works? Thanks

PR #166, opened 2018-08-07
move config to internal type

pweil-:
Moves the config.Config type to the internal API type.  Persists a copy of the internal model and changes commands that expected an existing cluster (ones that read the manifest AND config) to use the persisted model.

@openshift/sig-azure 
@amanohar

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -4,8 +4,6 @@ package plugin
 import (
 	"context"
 
-	"github.com/ghodss/yaml"
-
 	"github.com/openshift/openshift-azure/pkg/api"
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/api/v1"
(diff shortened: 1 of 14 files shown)


pweil-:
> PR needs rebase.

:sob:

pkg/plugin/plugin.go
```diff
@@ -27,50 +25,28 @@ func (p *Plugin) ValidateInternal(new, old *acsapi.ContainerService) []error {
...
-		}
-	} else {
-		c = &config.Config{}
-	}
-
-	err := config.Upgrade(cs, c)
-	if err != nil {
-		return nil, err
+func (p *Plugin) GenerateConfig(cs *acsapi.ContainerService) error {
+	// TODO should we save off the original config here and if there are any errors we can restore it?
```
kargakis: I think we shouldn't be mutating stored config if we don't apply it in a cluster. Not sure what exactly does that mean for this case.

pweil-: +1, that's what I was thinking with this comment.  We should either return a new version of the config and expect the orchestrator to handle adding it to the struct or we should retain a copy and revert it if we return an error.

Since `GenerateConfig` calls both `Upgrade` and `Generate` I was leaning towards just preserving the config internally but would be open to other solutions

pweil-: then again, declarative is exactly that....mutating desired state prior to applying it :-)  That begs the question of versioned storage of the internal model.  Maybe @amanohar can comment on how a rollback might work in that case or if it is forever forward moving

kargakis:
@amanohar ptal, would be nice to get this merged asap, I need to build on top for the upgrade work.

mjudeikis:
lgtm too.

pweil-:
Signed off in IRC.  

/remove-label do-not-merge/needs-signoff
/approve
/lgtm

pweil-:
> @pweil-: you cannot LGTM your own PR.

yeah, but I can add the label manually, silly bot!

Issue #158, opened 2018-08-03
Remove BootstrapAutoapprover Key


pweil-:
> We can get rid of this now that we are back to a VM deployment and use the service account directly. Please open a follow-up issue.

https://github.com/openshift/openshift-azure/pull/153#discussion_r207631491

PR #167, opened 2018-08-07
Use boostrap autoapprover service account directly

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/158

/hold


kargakis:
@openshift/sig-azure any preferences over keeping the bootstrap-autoapprover in the `openshift-azure` vs `openshift-infra` namespace?

kwoodson:
Infra makes more sense to me.

/lgtm

pweil-:
infra is fine for me too

pkg/addons/data/SecurityContextConstraints.security.openshift.io/privileged.yaml
```diff
@@ -36,7 +36,6 @@ supplementalGroups:
 users:
 - system:admin
 - system:serviceaccount:openshift-infra:build-controller
-- system:serviceaccount:openshift-azure:default
```
kargakis: @jim-minter unclear why this was needed.

kargakis:
Needs one more review 

/cc @mjudeikis @kwoodson

pkg/addons/data/SecurityContextConstraints.security.openshift.io/privileged.yaml
mjudeikis: I think it is leftover from tunnel

kargakis: Ah, right, cool.

PR #169, opened 2018-08-08
Import azure SDK

kargakis:
@openshift/sig-azure split out of https://github.com/openshift/openshift-azure/pull/138 so we can make forward progress on other fronts where we need to use the azure sdk.


glide.yaml
```diff
@@ -10,3 +10,9 @@ import:
   version: ^1.1.0
   subpackages:
   - spew
+- package: github.com/Azure/azure-sdk-for-go
+  version: v19.1.0
+  subpackages:
+  - services/compute/mgmt/2018-06-01/compute
```
kargakis: @mjudeikis @kwoodson you should be adding the packages you are going to use here me thinks

PR #144, opened 2018-08-01
Enable MSI for deployments.

kwoodson:
Investigating enabling MSI for OSA deployments.   Very WIP.

PR #170, opened 2018-08-08
[WIP] MSI with azure go client.

kwoodson:


pkg/addons/addons.go
```diff
@@ -57,9 +62,23 @@ func readDB(cs *acsapi.ContainerService, c *config.Config) (map[string]unstructu
 			return nil, err
 		}
 
+		azs, err := GetStorageAccountsClient(c)
```
kargakis: run this outside of the loop?

kargakis: Or maybe move this code altogether in newClient. I think it makes more sense in there.

kwoodson: Where do you recommend this go?  As part of the client instantiation?  As a var inside of the addons.go?  Inside of addons.go main?

kargakis: it's client-related code so client instantiation would be my choice.

pkg/addons/addons.go
```diff
@@ -57,9 +62,23 @@ func readDB(cs *acsapi.ContainerService, c *config.Config) (map[string]unstructu
...
+		}
+
+		ex := &extra{
+			StorageAccountKey: *(((*response.Keys)[0]).Value),
+		}
+
 		ts := Translations[KeyFunc(o.GroupVersionKind().GroupKind(), o.GetNamespace(), o.GetName())]
 		for _, tr := range ts {
-			b, err := util.Template(tr.Template, nil, cs, c, nil)
+			b, err := util.Template(tr.Template, nil, cs, c, ex)
```
mjudeikis: Would it not be better to set it as c *Config object, so it will get persisted together with all config?

kwoodson:
Closing.  #144 was merged.

kargakis:
@kwoodson can you make sure we track this work in Jira? Thanks!

Issue #147, opened 2018-08-02
check for stale deps in CI


kargakis:
We should block merging PR that don't update deps correctly.

@openshift/sig-azure

PR #171, opened 2018-08-08
Check-in dep validation script

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/147

Note that we still need to pin some dependencies in glide.yaml - will open a separate issue for that.

@openshift/sig-azure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ sync-push: sync-image
 
 verify:
 	./hack/validate-generated.sh
+	./hack/validate-deps.sh
 
 unit:
 	go test ./...
--- a/hack/validate-deps.sh
+++ b/hack/validate-deps.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+####################################################
+SOURCE="${BASH_SOURCE[0]}"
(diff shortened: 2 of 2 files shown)


kargakis:
> Note that we still need to pin some dependencies in glide.yaml - will open a separate issue for that.

https://github.com/openshift/openshift-azure/issues/173

Makefile
```diff
@@ -21,6 +21,7 @@ sync-push: sync-image
 
 verify:
 	./hack/validate-generated.sh
+	./hack/validate-deps.sh
```
kargakis: We will probably need to use a different test base image than `openshift/release:golang-1.10` in our tests since it does not include glide I think.

mjudeikis: Im started working on assumption that we would maintain our own base image based on `centos7` where we will have `az, go, glide, etc` as our stack is quite unique.

kargakis:
@mjudeikis this should be ready to go now that we changed to the new test base, ptal

/hold cancel

mjudeikis:
/lgtm
Let the fun begin :)

PR #172, opened 2018-08-08
Separate master pool profile on external model

pweil-:
built on https://github.com/openshift/openshift-azure/pull/155

Moves master pool profile configuration into its own section in the external model.  Converts it to an agent pool profile on the internal model.

@openshift/sig-azure
@amanohar


pkg/api/v1/types.go
```diff
@@ -41,6 +41,9 @@ type Properties struct {
 	// RouterProfiles (in,optional/out): Configuration for OpenShift router(s).
 	RouterProfiles []RouterProfile `json:"routerProfiles,omitempty"`
 
+	// MasterPoolProfile (in): Configuration for OpenShift master VMs.
+	MasterPoolProfile MasterPoolProfile `json:"masterPoolProfile,omitempty"`
```
pweil-: Note for reviewers: this name was chosen because we have a `MasterProfile` on the internal spec already and I did not want to cause confusion.

pkg/validate/container_service.go
```diff
@@ -1,10 +1,258 @@
...
+	default:
+		errs = append(errs, fmt.Errorf("invalid properties.openShiftVersion %q", p.OrchestratorVersion))
+	}
+
+	errs = append(errs, validateOpenShiftConfig(p.OpenShiftConfig)...)
+
+	return
+}
+
+func validateMasterProfile(p *api.MasterProfile) (errs []error) {
```
amanohar: Should this have validation for `count` as well (3 or 5 as allowed values)?

pkg/api/types.go
```diff
@@ -133,6 +133,13 @@ const (
 // OSType represents OS types of agents
 type OSType string
 
+const (
```
amanohar: Should `MasterProfile` on line 95 have `count` field?

pkg/api/v1/types.go
```diff
@@ -111,8 +125,6 @@ const (
 	AgentPoolProfileRoleCompute AgentPoolProfileRole = "compute"
 	// AgentPoolProfileRoleInfra is the infra role.
 	AgentPoolProfileRoleInfra AgentPoolProfileRole = "infra"
-	// AgentPoolProfileRoleMaster is the master role.
-	AgentPoolProfileRoleMaster AgentPoolProfileRole = "master"
```
amanohar: Is this going to be read only field for masters or is it even needed in the public API model that `MasterPoolProfile` is a separate?

pkg/validate/container_service.go
```diff
@@ -1,10 +1,258 @@
...
+	}
+
+	switch app.Role {
+	case api.AgentPoolProfileRoleCompute,
+		api.AgentPoolProfileRoleInfra:
+		if app.Count < 1 || app.Count > 100 {
+			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].count %q", app.Name, app.Count))
+		}
+
+	case api.AgentPoolProfileRoleMaster:
```
amanohar: Here: https://github.com/pweil-/openshift-azure/blob/separate-masterprofile/pkg/api/types.go#L35 I see a separate `MasterProfile` being added and the validation for master should be on this object instead of `AgentPoolProfile`.

pkg/api/v1/types.go
amanohar: Same applies to other fields as well so this can be `MasterProfile`.

pkg/api/v1/types.go
pweil-: This constant is removed from the external api model with this line since it is not in the public `MasterPoolProfile`.  It is retained in the internal model (`pkg/types`) for use after conversion.

pkg/api/types.go
pweil-: no, at this point it is internal api and the `MasterPoolProfile` has been converted into an `AgentPoolProfile`

pweil-: (which has count)

pkg/validate/container_service.go
pweil-: That `MasterProfile` is different than the `MasterPoolProfile` as noted in https://github.com/openshift/openshift-azure/pull/172#discussion_r208613627

The `MasterPoolProfile` is converted to an `AgentPoolProfile`.  `MasterProfile` and `MasterPoolProfile` are separate.

pweil-:
@amanohar

For clarity here:

`api/v1/types.go ` == external types
`api/types.go` == internal types

`OpenShiftCluster.Properties.FQDN` converts to `ContainerService.Properties.MasterProfile.FQDN`
`OpenShiftCluster.Properties.MasterPoolProfile` converts to a `ContainerService.Properties.AgentPoolProfile`

I don't want to rewrite our templating with this change since it is specific to the external API (that was the motivation for separating out the master agent pool to begin with) so I don't want to try and repurpose `api.MasterProfile` at this point.

pkg/validate/container_service.go
pweil-: no, master is an agent pool in the internal representation so that validation occurs here: https://github.com/openshift/openshift-azure/pull/172/commits/17ac105e64003d35eaa5a0cd2acd182160832949#diff-33ccf5fc1c1988bc397bbbabb210cde4R134

pkg/validate/container_service.go
amanohar: v1 (public model): https://github.com/pweil-/openshift-azure/blob/separate-masterprofile/pkg/api/v1/types.go#L92 is as expected and FQDN is at the top level in `Properties`

Internal API model: https://github.com/pweil-/openshift-azure/blob/separate-masterprofile/pkg/api/types.go#L95-L100 has `FQDN` in `MasterProfile` but all the other properties of master are in `AgentPoolProfile` when role is `master`. This could be confusing. Any reason why internal representation needs `MasterProfile` at all?

pweil-: Yes, the FQDN field is going from external.Properties to internal.MasterProfile.FQDN.  I think this is a good candidate to move because it's easy to confuse internal.MasterProfile and external.MasterPoolProfile.  I didn't touch any of the existing fields since the objective was to get the external agent pool model in shape.  

If we want to change up the internal model I think we can but I'd prefer a follow up and limit the gating of this PR to making sure we are satisfied with the external representation. 

WDYT?

amanohar: @pweil- SGTM. This PR can be merged and followed up with another PR to remove `MasterProfile` from internal model and move `FQDN` to `Properties`.

kargakis:
We have an issue with webhooks in our CI cluster currently

pweil-:
removing label manually.  

@openshift/sig-azure  still needs a +1 from our team

pkg/api/converterfromosaapi.go
```diff
@@ -51,7 +51,8 @@ func ConvertVLabsOpenShiftClusterToContainerService(oc *v1.OpenShiftCluster) *Co
 			}
 		}
 
-		cs.Properties.AgentPoolProfiles = make([]*AgentPoolProfile, len(oc.Properties.AgentPoolProfiles))
+		// +1 because master pool profile becomes agent pool profile
```
kargakis: So internally we have a masterProfile and we store it also as an agentPoolProfile? Why?

pkg/validate/validate_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/ghodss/yaml"
 
+	"github.com/davecgh/go-spew/spew"
```
kargakis: nit: group with the import above

kargakis:
If there is no strong reason against having the same scheme internally and externally, I would prefer we align both to be the same.

pkg/api/converterfromosaapi.go
pweil-: > So internally we have a masterProfile and we store it also as an agentPoolProfile? Why?

I think I just need to remove that, it has confused everyone so far. :+1:  Will ping you to discuss options

pweil-: previous discussion on the topic https://github.com/openshift/openshift-azure/pull/172#discussion_r208959950

pweil-:
recap of discussion:  reason to convert to agent pool profile is reduce arm template duplication.  Will bite the bullet and remove MasterProfile by moving FQDN to Properties since it's confusing now.  If there is specific master config later we can use substructs on the agent pool profile as discussed previously

pweil-:
@kargakis @mjudeikis FQDN moved ready for re-review

PR #174, opened 2018-08-08
generate cluster id

pweil-:
Generates 8 digit id and adds them to the end of resource names in the template.

@openshift/sig-azure 
@amanohar FYI


pweil-:
id

@amanohar ready for your review 

cc @sozercan @shrutir25

kargakis:
Does this change the hostnames in the master VMs?

pweil-:
> Does this change the hostnames in the master VMs?

hmm, I don't think it did, I've been running ssh for every pr to verify node readiness and pod running.  Let me retest that just to be 100% sure

pweil-:
it does 

/hold

pweil-:
this has a lot more knock on effects than anticipated.  Since we've discussed that this is not a blocker I'm going to close for now.  We can revisit if necessary

PR #176, opened 2018-08-09
add cluster-reader sa and kubeconfig

mjudeikis:
Add dedicated SA and kubeconfig for Operations to use.


mjudeikis:
cc: @openshift/sig-azure

pkg/addons/data/ServiceAccount/openshift-infra/cluster-reader.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-reader
```
pweil-: how do you feel about being more specific here?  This is for Microsoft SRE, so maybe microsoft-sre?  I'm also wondering if we want to separate anything we do specifically for OSA into a different namespace.  Didn't we just get rid of the openshift-azure one? :rofl:  

What I'm wondering here is if we need to be wary of things in the future that make assumptions on what's in the infra namespace and remove or change things we've added.

mjudeikis: > Didn't we just get rid of the openshift-azure one? rofl 

Was thinking about this too. My vote would be `openshift-azure` namespace and dont distinguish our ops and their ops. Just `azure-cluster-reader`. We can create a dedicated role and role binding with `azure` prefix (or postfix). And so we know that anything with prefix `azure` is our made.

kargakis: +1 on using openshift-azure for OSA-specific things.

pkg/addons/data/Namespace/openshift-azure.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
```
kargakis: Is this needed?

pkg/addons/data/ServiceAccount/openshift-infra/cluster-reader.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
kargakis: Needs to be removed now?

pkg/addons/data/Namespace/openshift-azure.yaml
mjudeikis: This backfired us a few times already (lack of this annotation). I would be tempted to keep it as its default.

PR #177, opened 2018-08-09
Logrus

pweil-:
Add logrus and some basic logging

@kargakis @mjudeikis


cmd/createorupdate/createorupdate.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	"github.com/ghodss/yaml"
-	"k8s.io/apimachinery/pkg/util/errors"
+	log "github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/api/v1"
 	"github.com/openshift/openshift-azure/pkg/plugin"
 	"github.com/openshift/openshift-azure/pkg/tls"
+	"k8s.io/apimachinery/pkg/util/errors"
```
kargakis: nit: move back

pweil-: yeah, saw that when looking at the changes again, already taken care of.  Dunno why my IDE keeps doing that... :angry:

cmd/healthcheck/healthcheck.go
```diff
@@ -31,6 +32,7 @@ func healthCheck() error {
 }
 
 func main() {
+	log.SetLevel(log.DebugLevel)
```
amanohar: How would the logger be passed by the service/RP to the plugin e.g. in ACS E: https://github.com/Azure/acs-engine/blob/master/pkg/operations/kubernetesupgrade/upgradecluster.go#L64?

kargakis:
Yeah, it would be nice to make this configurable for internal use.

PR #179, opened 2018-08-09
Make roleassignments uniquely named with resource group.

kwoodson:
As a follow on to #144, we need to uniquely name our roleassignments that are associated with our vmscalesets.  This PR adds the resource group name to the roleassignment guid.


pweil-:
issues with CI right now, merging

PR #180, opened 2018-08-09
custom logger

pweil-:
update plugin to use custom logger, switch add on pieces that were static back to panics (like MustCompile), left other add on statements using logrus but expectation is that it will run in the pod so no custom logger support is there.

~~Not done: passing logger down to static impl methods that the plugin calls (example below).  Right now no logging is done there but we should add that.  When we do we will either pass the logger in the method signature (yuck) or move them to interface (`Generator`, etc) implementations so they can have state like the plugin.  If everyone is good with what is here now I think I'd like to go ahead and take care of moving them to have the logger available now so we don't forget later.~~  Edit: did it

@kargakis @mjudeikis @amanohar

```
func (p *plugin) GenerateARM(cs *acsapi.ContainerService) ([]byte, error) {
	return arm.Generate(cs)
}
```

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -4,6 +4,8 @@ package plugin
 import (
 	"context"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/openshift/openshift-azure/pkg/api"
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/arm"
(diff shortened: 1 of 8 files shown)


pweil-:
for fun I tested with a logger with custom fields.  You can see the cmd's logger showing a caller and the sync plugin (which wouldn't be part of the RP logging) showing no customization.  I think we should move the cmd logging into the plugin


```
INFO[0000] convert to internal                           caller=dev
INFO[0000] done                                          caller=dev
INFO[0000] enrich                                        caller=dev
INFO[0000] done                                          caller=dev
INFO[0000] read old config                               caller=dev
INFO[0000] done                                          caller=dev
INFO[0000] validate internal                             caller=dev
INFO[0000] done                                          caller=dev
INFO[0000] generate config                               caller=dev
INFO[0008] done                                          caller=dev
INFO[0008] persist config                                caller=dev
INFO[0008] done                                          caller=dev
INFO[0008] generate arm                                  caller=dev
INFO[0008] done                                          caller=dev
INFO[0008] write arm                                     caller=dev
INFO[0008] done                                          caller=dev
INFO[0008] write helpers                                 caller=dev
INFO[0008] done                                          caller=dev
+ az group deployment create -g pweilosa -n azuredeploy --template-file _data/_out/azuredeploy.json --no-wait
+ hack/dns.sh zone-create pweilosa
+ hack/dns.sh cname-create pweilosa openshift pweilosa.eastus.cloudapp.azure.com
+ hack/dns.sh cname-create pweilosa '*' pweilosa-router.eastus.cloudapp.azure.com
+ [[ true == \t\r\u\e ]]
+ KUBECONFIG=_data/_out/admin.kubeconfig
+ go run cmd/sync/sync.go -run-once=true
INFO[0000] Sync process started!                        
```

pweil-:
rebased, ready for review @mjudeikis @kargakis

pkg/plugin/plugin.go
```diff
@@ -12,21 +14,38 @@ import (
...
+	}
+
+	return &plugin{
+		log: log,
+	}
+}
 
-func (p *Plugin) ValidateInternal(new, old *acsapi.ContainerService) []error {
+func (p *plugin) ValidateInternal(new, old *acsapi.ContainerService) []error {
+	p.log.Info("validating internal data models")
```
mjudeikis: I dont like syntax like this, but not sure if anything could be done here :/

pweil-: yeah, that's why I was hoping we could just configure the global level logger but I didn't see a good way to do that when you want something like custom `WithFields` that was mentioned.  The other option I could think of would be to wrap logging our own package like

```go
package log

struct LogWrapper { log *logrus.Entry }

var logger *LogWrapper

//method to create logger once

//static wrapper methods for log.*
```

Then you could import our logging package and it would use the instantiated `LogWrapper`.

kargakis:
ci-operator bug

/retest

PR #181, opened 2018-08-10
Merge old and new config + validate updates

kargakis:
@openshift/sig-azure ptal

@pweil- comments on the merging<->validation order: 
1. we need to merge the old cs.Config into the new and hand that over to GenerateConfig (so there is no merging of the user request but of the generated config). 
2. validation only happens on the user request and not on generated config. 
3. In the close future, we probably want to add validation for updates and likely invalidate every request that causes a difference in fqdn, resourceGroup, and location. If we are going to merge user requests prior to validation, we will not be able to validate updates.  

Based on the above I don't see what merging gives us prior to validation (since the merged part is unaffected by validation) and I prefer to avoid the extra method in the plugin interface so I just opted on extending GenerateConfig. Thoughts?

cc @amanohar

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -6,6 +6,13 @@ import (
 )
 
 type Plugin interface {
+	// MergeConfig merges new and old config so that no unnecessary config
+	// is going to get regenerated during generation. It also handles merging
+	// partial user requests to allow reusing the same validation code during
+	// upgrades. This method should be the first one called by the RP, before
+	// validation and generation.
+	MergeConfig(new, old *OpenShiftManagedCluster)
+
 	// Validate exists (a) to be able to place validation logic in a
 	// single place in the event of multiple external API versions, and (b) to
 	// be able to compare a new API manifest against a pre-existing API manifest
@@ -14,6 +21,8 @@ type Plugin interface {
 	// should be excluded.
 	Validate(new, old *OpenShiftManagedCluster, externalOnly bool) []error
 
+	// GenerateConfig ensures all the necessary in-cluster config is generated
+	// for an Openshift cluster.
 	GenerateConfig(cs *OpenShiftManagedCluster) error
 
 	GenerateARM(cs *OpenShiftManagedCluster) ([]byte, error)
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -33,6 +33,34 @@ func NewPlugin(entry *logrus.Entry) api.Plugin {
 	}
 }
 
+func (p *plugin) MergeConfig(cs, oldCs *acsapi.OpenShiftManagedCluster) {
+	if oldCs == nil {
+		return
+	}
+	log.Info("merging internal data models")
+
+	// generated config should be copied as is
+	cs.Config = oldCs.Config
+
+	// user request data
+	// need to merge partial requests
+	if len(cs.Properties.AgentPoolProfiles) == 0 {
+		cs.Properties.AgentPoolProfiles = oldCs.Properties.AgentPoolProfiles
+	}
+	if cs.Properties.OrchestratorProfile == nil {
+		cs.Properties.OrchestratorProfile = oldCs.Properties.OrchestratorProfile
+	}
+	if len(cs.Properties.OrchestratorProfile.OrchestratorVersion) == 0 {
+		cs.Properties.OrchestratorProfile.OrchestratorVersion = oldCs.Properties.OrchestratorProfile.OrchestratorVersion
+	}
+	if cs.Properties.OrchestratorProfile.OpenShiftConfig == nil {
+		cs.Properties.OrchestratorProfile.OpenShiftConfig = oldCs.Properties.OrchestratorProfile.OpenShiftConfig
+	}
+	if len(cs.Properties.FQDN) == 0 {
+		cs.Properties.FQDN = oldCs.Properties.FQDN
+	}
+}
+
 func (p *plugin) Validate(new, old *acsapi.OpenShiftManagedCluster, externalOnly bool) []error {
 	log.Info("validating internal data models")
 	return validate.Validate(new, old, externalOnly)
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -74,6 +74,9 @@ func createOrUpdate(oc *v1.OpenShiftManagedCluster, entry *logrus.Entry) (*v1.Op
 		}
 	}
 
+	log.Info("merge old and new config")
+	p.MergeConfig(cs, oldCs)
+
 	// validate the internal API representation (with reference to the previous
 	// internal API representation)
 	// we set fqdn during enrichment which is slightly different than what the RP
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -38,7 +38,13 @@ func isValidHostname(h string) bool {
 func Validate(new, old *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
 	// TODO update validation
 	// TODO are these error messages confusing since they may not correspond with the external model?
-	return validateContainerService(new, externalOnly)
+	if errs := validateContainerService(new, externalOnly); len(errs) > 0 {
+		return errs
+	}
+	if old != nil {
+		return validateUpdateContainerService(new, old, externalOnly)
+	}
+	return nil
 }
 
 func validateContainerService(c *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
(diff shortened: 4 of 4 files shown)


pweil-:
> 1. we need to merge the old cs.Config into the new and hand that over to GenerateConfig (so there is no merging of the user request but of the generated config). 
> 2. validation only happens on the user request and not on generated config. 
> 3. In the close future, we probably want to add validation for updates and likely invalidate every request that causes a difference in fqdn, resourceGroup, and location. If we are going to merge user requests prior to validation, we will not be able to validate updates.  
>
> Based on the above I don't see what merging gives us prior to validation (since the merged part is unaffected by validation) and I prefer to avoid the extra method in the plugin interface so I just opted on extending GenerateConfig. Thoughts?

I was thinking the following (simplified to only list steps that apply to our code):

1.   user request comes in
2.   user request converted to internal format
3.   merge old config with new config.  Completely overwrite Config, merge in other fields if they are set in old but not new
4.  Validate (run all validations that you would on create since you now have a fully formed manifest)
5.  Validate updates (check fields that shouldn't have changed or changed improperly)
6.  Generate

4+5 are the same interface method since we take old and new.

The reason I like this is step 4.  It keeps the safety of things like master counts, duplicate profiles, etc (maybe most importantly - supported versions).  Everything we validate on create.  It's important to remember that what we receive for update might be missing more than just config based on our conversations with @amanohar so we need to be prepared to fully populate the manifest with the old data (granted that's not much right now :smile:)

What you propose is simple and works on the assumption that a full manifest (minus config) comes in each time but I don't think it's quite enough long term.  I would be happy to be wrong on this one.

Also I think extending one of the other methods to do a merge is a bit of a code smell since it's a side effect not obvious in the interface.  Merge is a first class step IMO.

kargakis:
> Also I think extending one of the other methods to do a merge is a bit of a code smell since it's a side effect not obvious in the interface. Merge is a first class step IMO.

Yeah, seems we need to make this first-class if we want to merge partial requests into olders. I will rework this based on our bluejeans discussion.

kargakis:
@pweil- added MergeConfig for merging and also started on update validation.

pkg/api/plugin.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 type Plugin interface {
+	MergeConfig(new, old *ContainerService) error
```
pweil-: @sozercan @amanohar @shrutir25 needs sign off

pkg/validate/container_service.go
```diff
@@ -38,7 +38,10 @@ func isValidHostname(h string) bool {
 func ContainerService(new, old *api.ContainerService) (errs []error) {
 	// TODO update validation
 	// TODO are these error messages confusing since they may not correspond with the external model?
-	return validateContainerService(new)
+	if errs := validateContainerService(new); len(errs) > 0 {
+		return errs
+	}
+	return validateUpdateContainerService(new, old)
```
pweil-: only if old is not nil

pkg/plugin/plugin.go
```diff
@@ -16,6 +16,30 @@ type Plugin struct{}
...
+	}
+	if cs.Properties.OrchestratorProfile == nil {
+		cs.Properties.OrchestratorProfile = oldCs.Properties.OrchestratorProfile
+	}
+	if len(cs.Properties.OrchestratorProfile.OrchestratorVersion) == 0 {
+		cs.Properties.OrchestratorProfile.OrchestratorVersion = oldCs.Properties.OrchestratorProfile.OrchestratorVersion
+	}
+	if cs.Properties.OrchestratorProfile.OpenShiftConfig == nil {
+		cs.Properties.OrchestratorProfile.OpenShiftConfig = oldCs.Properties.OrchestratorProfile.OpenShiftConfig
+	}
```
pweil-: cs.Properties.FQDN?

pkg/validate/container_service.go
kargakis: I am doing the check inside `validateUpdateContainerService`, no strong feelings on the one over the other.

pkg/plugin/plugin.go
kargakis: yeah, fixed

pkg/validate/container_service.go
kargakis: or so I thought

pweil-: maybe my github isn't refreshing, first line of that is an NPE if old is nil

```
func validateUpdateContainerService(cs, oldCs *api.ContainerService) (errs []error) {
	if cs.Location != oldCs.Location {
```

pkg/plugin/plugin.go
```diff
@@ -16,6 +16,33 @@ type Plugin struct{}
 
 var _ api.Plugin = &Plugin{}
 
+func (p *Plugin) MergeConfig(cs, oldCs *acsapi.ContainerService) {
```
pweil-: most of these are pointers, we're going to need to make a copy of the value rather than just assign the pointer.  Maybe it's worth having a look at https://github.com/imdario/mergo.  I believe Kubernetes uses it too.

pkg/validate/container_service.go
```diff
@@ -38,7 +38,13 @@ func isValidHostname(h string) bool {
 func ContainerService(new, old *api.ContainerService) (errs []error) {
 	// TODO update validation
 	// TODO are these error messages confusing since they may not correspond with the external model?
-	return validateContainerService(new)
+	if errs := validateContainerService(new); len(errs) > 0 {
+		return errs
+	}
+	if old != nil {
+		return validateUpdateContainerService(new, old)
```
amanohar: @pweil- regarding `MergeConfig` vs `validateUpdateContainerService` vs `ValidateInternal`: 
- I am assuming `validateUpdateContainerService`: is for updating the data model except config
- And  `MergeConfig` is for merging config only?

At what point would the RP need to call `MergeConfig`?

kargakis:
@amanohar cmd/createorupdate/createorupdate.go simulates how we expect the
RP to work

kargakis:
Also validateUpdate does not update anything, it just validates updates.

pkg/validate/container_service.go
pweil-: > At what point would the RP need to call MergeConfig

'MergeConfig' would be called during updates (so when you have an old and new config), after convert to internal and before `ValidateInternal`.  Step 5 in the doc for update flow.  Since we were unsure who was going to own this impl we figured we'd take a stab at it.  `MergeConfig` is for two items:

1.  ensuring we don't regen certs
2.  dealing with partial manifests received during update

> I am assuming validateUpdateContainerService: is for updating the data model except config
> And MergeConfig is for merging config only?

The RP will always just call `ValidateInternal`.  Since `ValidateInternal` has the old and new config it will call `validateUpdateContainerService` as appropriate (when old != nil).  `validateUpdateContainerService` is simply some update specific validations, no data side effects.

pkg/plugin/plugin.go
kargakis: Hrm, two comments: 1. if the API allows passing pointers then it's assumed that the implementation can muck with the pointers, so I think it's a responsibility of the caller to ensure that it provides copies rather than the original pointers, 2. why would we need to copy at all?

pkg/plugin/plugin.go
```diff
@@ -16,6 +16,33 @@ type Plugin struct{}
...
+	if oldCs == nil {
+		return
+	}
+
+	// generated config should be copied as is
```
mjudeikis: In the future this will have to have logic too. Question if it will be here, or in other places. In example image version changes, new kubeconfig, etc.

pkg/validate/container_service.go
amanohar: @pweil- . Yes,sgtm. I reread the whole `createOrUpdate` in `createorupdate.go` again.

kargakis:
I am going to take on deep-copying on a follow-up PR, @openshift/sig-azure @amanohar @sozercan ptal

kargakis:
> I am going to take on deep-copying on a follow-up PR

issue: https://github.com/openshift/openshift-azure/issues/193
PR: https://github.com/openshift/openshift-azure/pull/192

mjudeikis:
lgtm. I suspect we will see some logical gaps when we start rolling this out. I would encourage everybody @openshift/sig-azure to re-read PR again and familiarize with the code as I think its one of the most important flow, so when those gaps appear anybody could debug it fast...

pkg/api/plugin.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 type Plugin interface {
+	MergeConfig(new, old *ContainerService)
```
pweil-: bump for sign off on this addition @amanohar

kargakis:
@amanohar ptal

pweil-:
@sozercan @amanohar @shrutir25 ptal at the plugin interface.  I think it's all good based on the conversation in https://github.com/openshift/openshift-azure/pull/181#discussion_r209331108 but I don't want to assume.  Thanks!

kargakis:
@sozercan @amanohar @shrutir25 ping

jim-minter:
deepcopy will be done in #192

amanohar:
@kargakis signed off.

jim-minter:
thanks @amanohar 

/lgtm

kargakis:
failure is glide-related

PR #184, opened 2018-08-13
update glide dep

### LINKED ISSUE #230: uuid package revision
Looks like we use an older version of `satori/go.uuid` internally (before the breaking api change for `err`). Which is giving this error:
```
vendor/github.com/openshift/openshift-azure/pkg/config/generate.go:479:37: assignment mismatch: 2 variables but 1 values
```

Would it be possible to use `f58768cc1a7a7e77a3bd49e98cdd21419399b6a3` revision of  `satori/go.uuid`? 

I believe only place this is used is when generating service catalog cluster id: https://github.com/openshift/openshift-azure/blob/master/pkg/config/generate.go#L479 

cc @mjudeikis

mjudeikis:
Update glide deps 
fixes #230


mjudeikis:
Quesion, what do we want to do with 
```
golang.org/x/crypto/bcrypt
golang.org/x/crypto/ssh
golang.org/x/tools/cmd/goyacc
github.com/go-bindata/go-bindata/go-bindata
```

As first 3 are more or less backward compatible all the time. Should we still pin them? For the last one, its binary we need. Again, different case

@openshift/sig-azure

glide.yaml
```diff
@@ -15,4 +15,10 @@ import:
 - package: github.com/Azure/go-autorest
   version: 10.14.0
 - package: github.com/sirupsen/logrus
-  version: v1.0.6
\ No newline at end of file
+  version: v1.0.6
+- package: github.com/ghodss/yaml
```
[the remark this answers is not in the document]
pweil-: actually I don't see it in a quick vendor search....weird...ok

glide.yaml
```diff
@@ -15,4 +15,10 @@ import:
...
+  version: e9ed3c6dfb39bb1a32197cb10d527906fe4da4b6  # Master as per 2018-08-13
+- package: github.com/go-test/deep
```
kargakis: Just rambling but I think we can get rid of this package by using `k8s.io/apimachinery/pkg/equality.Semantic.DeepEqual` and `k8s.io/apimachinery/pkg/util/diff.ObjectGoPrintDiff`.

glide.yaml
```diff
@@ -15,4 +15,10 @@ import:
 - package: github.com/Azure/go-autorest
   version: 10.14.0
 - package: github.com/sirupsen/logrus
-  version: v1.0.6
\ No newline at end of file
+  version: v1.0.6
+- package: github.com/ghodss/yaml
+  version: e9ed3c6dfb39bb1a32197cb10d527906fe4da4b6  # Master as per 2018-08-13
```
kargakis: Why use latest vs pin to v1.0.0? Are there any known issues that have been fixed in between?

mjudeikis:
@kargakis  https://github.com/openshift/openshift-azure/issues/206 to remove `go-test/deep`. Dont want to do this here.

jim-minter:
@mjudeikis suggest pinning them at their latest current revision for now?
worth adding a validation to ensure that we have a pin for everything we import (maybe using `go list -f '{{join .Imports "\n"}}' ./pkg/*` mentioned by @kargakis?)

mjudeikis:
@kargakis 
> Why use latest vs pin to v1.0.0? Are there any known issues that have been fixed in between?

@jim-minter : 
> suggest pinning them at their latest current revision for now?

One of you says the latest revision, another releases? My vote is for releases :)

jim-minter:
I don't think I really care.  In the case you mention, `v1.0.0` refers to a branch not a tag, so it will pick up new commits anyway.

mjudeikis:
For which one? yaml is a Tag?

jim-minter:
@pweil- @kargakis @mjudeikis I'm having trouble understanding what we're aiming for with this.  Are we trying to pin to a specific commit hash or to a remote branch which is intended to preserve compatibility?  What is the purpose of hack/validate-deps.sh?

I think:
- we should not start failing our builds because some upstream has added a new commit
- we should make it easy and safe for ourselves to run `glide up` when we want to
- we should ensure we're pinning at versions which are compatible with the RP

Do those make sense and does this PR reflect that?

mjudeikis:
> we should not start failing our builds because some upstream has added a new commit

I dont think this will happen, as we lock to released versions/or commits in all deps.

> we should make it easy and safe for ourselves to run glide up when we want to. 

You can do it as long as you commit dependencies or updates too. It will fail if your code has some dependency which is not in glide. And this is what we want to avoid :)

> we should ensure we're pinning at versions which are compatible with the RP

We need to know what is compatible in this case :/

mjudeikis:
I think the main idea is "keep track of everything we bringing in" so if you use some new library, this test will fail if you dont add it to glide. This way we have always tracked dependency list. 

But this means we need to pin dependencies and do some lifecycle and update pinned versions from time to time. 

The only thing this PR is not reflecting is `glide up` fully. And this is due pinning of the version.

pweil-:
> What is the purpose of hack/validate-deps.sh

To make sure devs have updated glide.yaml has been updated properly and no dependencies have snuck in.  Same as a generated code check, making sure local dev processes are being followed.  Yes, it happened and yes, it was me :smile: 

> we should not start failing our builds because some upstream has added a new commit

agree

> we should make it easy and safe for ourselves to run glide up when we want to

and check that it has been done right (validate-deps.sh).  Not sure about the lock change here, why ignore it?

> we should ensure we're pinning at versions which are compatible with the RP

how are we going to verify this?  I think you've pointed out a gap we should resolve with MSFT

I don't see much controversy in pinning dependencies.

jim-minter:
suggest we talk about this on scrum to reach consensus.

hack/validate-deps.sh
```diff
@@ -21,7 +21,7 @@ cd "${T}/"
 glide up -v
 
 
-if [[ -n "$(git status --porcelain)" ]]; then
+if [[ -n "$(git status --porcelain | sed  "/M glide.lock/d")" ]]; then
```
jim-minter: still needed?

glide.yaml
```diff
@@ -7,7 +7,7 @@ import:
 - package: k8s.io/apimachinery
   version: kubernetes-1.10.6
 - package: github.com/davecgh/go-spew
-  version: ^1.1.0
+  version: v1.1.1
```
jim-minter: still needed?

hack/validate-deps.sh
mjudeikis: if we gonna use it in any CI - yes. If we will run locally only - no

glide.yaml
mjudeikis: Used in tests `validate_test.go`

kargakis:
We should figure out why the lock diff is produced in CI vs working around
it. Open a separate issue to track?

jim-minter:
@kargakis we may want to pin some imported packages to a specific commit SHA, but not necessarily all, because it unnecessarily makes it harder to consume upstream fixes.

We decided not to run validate-deps in CI because we don't want our build to break on new commits introduced upstream.  It is not a build-breaking error for our vendoring not to correspond exactly with the latest upstream.  We can catch this and fix asynchronously.

PR #188, opened 2018-08-13
Removing MSI

kwoodson:


kargakis:
@kwoodson you need to go generate ./... since the generated data has changed after we merged MSI

kwoodson:
@kargakis @mjudeikis @pweil- 

I believe these are the necessary bits to remove MSI.  I have tested a successful deployment on this branch.

pkg/addons/data/Deployment.apps/default/docker-registry.yaml
```diff
@@ -69,7 +69,7 @@ spec:
           set -o pipefail
 
           . <(sed -e 's/: */=/' /etc/origin/cloudprovider/azure.conf)
-          az login -i &>/dev/null
+          az login --service-principal -u "$aadClientId" -p "$aadClientSecret" --tenant "$aadTenantId" &>/dev/null
```
sozercan: Since we are moving back to service principal, we'll need to replace this by either Azure Go SDK or using resource functions in ARM templates (like `listKeys`: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-resource#listkeys)

pweil-: @sozercan I'm not sure I'm fully following this.  How is a login in Azure Go SDK different than a command line login with passed in SP if that SP is what is being used everywhere else?

kwoodson: @sozercan and I had a discussion.  This seems fine for now but longer term we'd like to see if the azure sdk would be a better solution or MSI will be enabled in managed applications.

kwoodson:
@pweil-, Please give this a lgtm.  Thanks

mjudeikis:
@kwoodson you need to do `go generate ./...` and update. Verify test is failing

pkg/addons/data/Deployment.apps/default/docker-registry.yaml
pweil-: thanks for following up on this.

kwoodson:
@mjudeikis @pweil- 

Ran `go generate ../...` which updated the bindata.go.

PR #189, opened 2018-08-13
 Run sync process in a pod

kargakis:
@openshift/sig-azure

/hold


mjudeikis:
Is dropping `RUN_SYNC_LOCAL` is final decision? It is very nice development feature.

kargakis:
> Is dropping RUN_SYNC_LOCAL final decision? It is a very nice development feature.

Moved it back just for `hack/create.sh` but I can't think of any nice way to disable creating it as a deployment when RUN_SYNC_LOCAL=true ..

kargakis:
Fails to deploy with
```
      "statusMessage": "{\"status\":\"Failed\",\"error\":{\"code\":\"ResourceDeploymentFailure\",\"message\":\"The resource operation completed with terminal provisioning state 'Failed'.\",\"details\":[{\"code\":\"VMExtensionProvisioningError\",\"message\":\"VM has reported a failure when processing extension 'cse'. Error message: \\\"Enable failed: The script's length (342648) exceeded the maximum allowed length of 262144!\\n[stdout]\\n\\n[stderr]\\n\\\".\"}]}}"
```
/shrug

sozercan:
@kargakis it's limited to 256kb, you can try to gzip it and check if that reduces much.
https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux#property-script

jim-minter:
@kargakis I had intended, but not communicated clearly (sorry), that the plugin should use an openshift client to inject the sync deployment into the cluster.  That way we don't add more to the master startup script and push it over the size limit.

It also means less potentially mutable state sitting in the custom script extension, which is surely a good thing.

kargakis:
I remember you were talking about running a container that would inject the
sync process but I am wondering whether the extra layer is worth it vs
using a config bundle which is going to help us tidy up the rest of the
script too.

jim-minter:
@kargakis #220 (currently testing) is what I'm thinking about

jim-minter:
a possible issue with having lots of stuff in a config bundle is handling what happens when some configurable in the bundle changes.  I wonder if long term we don't even want a config bundle.

jim-minter:
@kargakis I think this is all trickier than it looks.  Creating a deployment doesn't imply the pod getting created as the node is not ready.  The node doesn't become ready until after the sync pod runs (circular dependency).  It can be created via setting `openshift.io/node-selector: ""` + use DaemonSet (bypasses the scheduler) + possibly set nodeSelector: kubernetes.io/hostname: master-000000 + hostNetwork: true, but that's a miserable set of hacks.

jim-minter:
@pweil- see previous comment, we're going to need to think about this tomorrow.

kargakis:
Instead of using a DaemonSet, can't we use tolerations to schedule the pod even when the node is unready? DaemonSet is going to use the scheduler in the future, too. Also, instead of pinning to a single host, we could set `deploy.spec.template.spec.nodeSelector` to `node-role.kubernetes.io/master: "true"`.

PR #190, opened 2018-08-13
azure custom logger

mjudeikis:
Wrap logger in a custom package.

```
INFO[0000] convert to internal                           resourceGroup=mjudeikis
INFO[0000] enrich                                        resourceGroup=mjudeikis
INFO[0000] read old config                               resourceGroup=mjudeikis
INFO[0000] validating internal data models               resourceGroup=mjudeikis
INFO[0000] generating configs                            resourceGroup=mjudeikis
```

cc: @pweil-  this is what you had in mind?


mjudeikis:
@sozercan @amanohar @shrutir25 can you please take a look this one.

We want to keep our api as clean as possible. So instead of passing logger all around the place I have created log wrapper. By calling it once `log.New(entry)` it will configure all repo sub-packages with the same configuration.

This has one draw-back. If for any reason you would like to call 2 different part of code with different logging configuration, this would not work. Can you please confirm from RP side if this is ok?

pweil-:
> We want to keep our api as clean as possible. So instead of passing logger all around the place I have created log wrapper. By calling it once log.New(entry) it will configure all repo sub-packages with the same configuration.

couldn't we just keep the plugin instantiation with a logger rather than forcing the requirement on the RP?  We (internally) could call log.New from the plugin.  

That means that if we want to use logging in our wrapper code (which we do) we can still call log.New prior to plugin instantiation.  Having an additional plugin constructor that doesn't take a log would be fine too, it could try to initialize the logger with an entry from the package level logger and if sync.Once was used it would just be a no-op

pweil-:
> cc: @pweil- this is what you had in mind?

yes, this is the only way I can see us having a nicer api like 'log.Info' vs '<object>.log.Info'

mjudeikis:
> couldn't we just keep the plugin instantiation with a logger rather than forcing the requirement on the RP? We (internally) could call log.New from the plugin.

Check it now. I renamed `log` to `entry`, as log is our package name (less confusion).
And removed `log/entry` from any structs/implementation, where it does not need to be. Plugin code calls/initialize them, so config will be passed just there.

pkg/log/log.go
```diff
@@ -0,0 +1,54 @@
...
+// Wrapper wrapper logrus logger to enable global configuration
+type wrapper struct {
+	log *logrus.Entry
+}
+
+var logger *wrapper
+var once sync.Once
+
+// New takes in logrus.Entry and configures global logger
+func New(entry *logrus.Entry) {
```
kargakis: This seems like an inflexible interface for configuring a logger. Usually at the top level you start with a logger and a couple of known fields (eg. subscription ID) and as you progress down the rabbit hole you want to be able to inject more (eg. resource group) and more (eg. scale set name).

mjudeikis: We need requirements form MSFT on this. As per the current understanding, I was under impression that RP will provide its own `entry/logrus` config and it will apply to this particular call. Do we need this for our own use?

cmd/createorupdate/createorupdate.go
```diff
@@ -156,9 +157,12 @@ func writeHelpers(c *acsapi.Config) error {
 }
 
 func main() {
+	// mock logger configuration
 	logger := logrus.New()
 	logger.SetLevel(logrus.DebugLevel)
 	log := logrus.NewEntry(logger)
+	entry := logrus.NewEntry(logger)
```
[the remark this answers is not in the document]
mjudeikis: agree

PR #191, opened 2018-08-13
Updating since origin sku was removed.

kwoodson:
This modifies the base sku name since origin_ was removed by our automated tools.

I ran into this error until I updated:

```
msrest.http_logger : {"error":{"code":"MarketplacePurchaseEligibilityFailed","message":"Marketplace purchase eligibilty check returned errors. See inner errors for details. ","details":[{"code":"BadRequest","message":"Offer with PublisherId: redhat, OfferId: osa-preview cannot be purchased due to validation errors. See details for more information.
...
```

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -29,7 +29,7 @@ func selectNodeImage(cs *acsapi.ContainerService) {
 	case "":
 		c.ImageSKU = "osa_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
 	case "centos7":
-		c.ImageSKU = "origin_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
+		c.ImageSKU = "okd_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
 	}
 
 	c.ImageResourceGroup = os.Getenv("IMAGE_RESOURCEGROUP")


kwoodson:
@pweil-,

If this name isn't ideal we can choose something else but does require a republish.

pkg/config/generate.go
```diff
@@ -29,7 +29,7 @@ func selectNodeImage(cs *acsapi.ContainerService) {
 	case "":
 		c.ImageSKU = "osa_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
 	case "centos7":
-		c.ImageSKU = "origin_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
+		c.ImageSKU = "okd_" + strings.Replace(cs.Properties.OrchestratorProfile.OrchestratorVersion[1:], ".", "", -1)
```
pweil-: I don't think we're supposed to be using OKD in any packaging.  What about oso (openshift origin)?.  I'm not in love with anything but origin, I want it back.

kwoodson: If you'd prefer `oso' I can update it.  This process will cause another publish and is generally a 1-2 day process.  If we want to move forward with oso please let me know and I"ll kick that off.

Currently I'm awaiting to hear back on resurrecting `origin_`.

pweil-: I'd rather just get origin back.  :smile:

kwoodson: I agree 100%.  In the meantime this is broken for a few of us and I suspect will break for others when origin_ unavailability propagates.

pweil-: if it's temporary then...

/lgtm

kargakis: > if it's temporary then...

Famous last words

kwoodson:
@kargakis @pweil-, I have republished the origin sku.  When it goes live, we'll need to revert this PR. Guess @kargakis was right.

PR #192, opened 2018-08-14
WIP: Config deep-copying

kargakis:


jim-minter:
WIP, the difficult pieces here are around deep copying keys/certs etc

jim-minter:
worth looking at trying to get rid of `testOpenShiftClusterYAML` in our tests with this PR as well

mjudeikis:
@jim-minter gave a shot to update this, but I think its worth waiting until https://github.com/openshift/openshift-azure/pull/254 goes in to do this

kargakis:
@jim-minter @pweil- this is getting very very ugly (x509.Certificate contains `interface{}` values :/). I was suggested in IRC to switch our certificates to be stored as raw `[]byte` types in Config. That would ease deep-copying them. Thoughts?

jim-minter:
@kargakis I agree this is not looking like the right direction.  It's a little unsavoury, but for now, why not just implement a Config DeepCopy function which serialises the input to json and unserialises to a new object?  We're not going to be doing it that much, it gets us over the hurdle, and there really are much worse alternatives.

jim-minter:
@kargakis see https://github.com/openshift/openshift-azure/pull/315 for an alternative approach

kargakis:
I like it better

/close

PR #194, opened 2018-08-14
change from DNS to healthcheck

mjudeikis:
move to healthz checking for api server availability. 

@kargakis this still fails with random other errors sometimes. I think this is the best way for now for dev and test. What do you think?

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -81,10 +81,9 @@ az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _dat
 
 if [[ "$RUN_SYNC_LOCAL" == "true" ]]; then
     # will eventually run as an HCP pod, for development run it locally
+    # sleep until FQDN/healthz responds with OK. It takes up to 5 min for API server to respond. 
     FQDN=$(awk '/^  fqdn:/ { print $2 }' <_data/manifest.yaml)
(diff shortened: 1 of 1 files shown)


hack/create.sh
```diff
@@ -81,10 +81,9 @@ az group deployment create -g $RESOURCEGROUP -n azuredeploy --template-file _dat
 
 if [[ "$RUN_SYNC_LOCAL" == "true" ]]; then
     # will eventually run as an HCP pod, for development run it locally
+    # sleep until FQDN/healthz responds with OK. It takes up to 5 min for API server to respond. 
     FQDN=$(awk '/^  fqdn:/ { print $2 }' <_data/manifest.yaml)
-    # sleep until FQDN name is registered in Azure DNS
-    # 6 coresponds to "Could not resolve host"
-    while [[ "$(curl -s -k https://${FQDN} ; echo $?)" == "6" ]]; do sleep 5; done
+    while [[ "$(curl -s -k https://${FQDN}/healthz)" != "ok" ]]; do sleep 5; done
```
kargakis: use /healthz/ready over /healthz?

pweil-: +1 to healthz/ready.  I believe healthz could report more useful information if not ready (in the body) but I think it is possible to be healthy and not ready

Issue #149, opened 2018-08-02
Public and internal resource names


amanohar:
1) Internal data model: https://github.com/openshift/openshift-azure/blob/master/pkg/api/types.go is using type name `ContainerService`. Recommend changing the name to `OpenShiftService` or `OpenShiftCluster`.

2) Public name of the resource is going to be `openShiftManagedClusters`. This: https://github.com/openshift/openshift-azure/blob/master/pkg/api/v1/types.go#L5 should be updated to: `OpenShiftManagedCluster`

3) Azure public data model are usually of format `YYYY-MM-DD`. We should have a TODO to update `v1` to a version later: https://github.com/openshift/openshift-azure/blob/master/pkg/api/v1/types.go#L1

PR #195, opened 2018-08-14
Api cleanup

pweil-:
Fixes https://github.com/openshift/openshift-azure/issues/149

@openshift/sig-azure

### DIFF
--- a/pkg/api/converterfromosaapi.go
+++ b/pkg/api/converterfromosaapi.go
@@ -4,10 +4,10 @@ import (
 	"github.com/openshift/openshift-azure/pkg/api/v1"
 )
 
-// ConvertVLabsOpenShiftClusterToContainerService converts from a
-// v1.OpenShiftCluster to a ContainerService.
-func ConvertVLabsOpenShiftClusterToContainerService(oc *v1.OpenShiftCluster) *ContainerService {
-	cs := &ContainerService{
+// ConvertV1OpenShiftManagedClusterToOpenShiftManagedCluster converts from a
+// v1.OpenShiftManagedCluster to an OpenShiftManagedCluster.
+func ConvertV1OpenShiftManagedClusterToOpenShiftManagedCluster(oc *v1.OpenShiftManagedCluster) *OpenShiftManagedCluster {
+	cs := &OpenShiftManagedCluster{
 		ID:       oc.ID,
 		Location: oc.Location,
 		Name:     oc.Name,
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -17,13 +17,13 @@ import (
 )
 
 // createOrUpdate simulates the RP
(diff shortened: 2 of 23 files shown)


pkg/api/converterfromosaapi.go
```diff
@@ -6,8 +6,8 @@ import (
 
 // ConvertVLabsOpenShiftClusterToContainerService converts from a
 // v1.OpenShiftCluster to a ContainerService.
-func ConvertVLabsOpenShiftClusterToContainerService(oc *v1.OpenShiftCluster) *ContainerService {
-	cs := &ContainerService{
+func ConvertVLabsOpenShiftClusterToContainerService(oc *v1.OpenShiftManagedCluster) *OpenShiftCluster {
```
kargakis: Any reason why these don't use the same name?

pweil-: fixed and squashed

pkg/api/converterfromosaapi.go
```diff
@@ -4,10 +4,10 @@ import (
 	"github.com/openshift/openshift-azure/pkg/api/v1"
 )
 
-// ConvertVLabsOpenShiftClusterToContainerService converts from a
-// v1.OpenShiftCluster to a ContainerService.
-func ConvertVLabsOpenShiftClusterToContainerService(oc *v1.OpenShiftCluster) *ContainerService {
-	cs := &ContainerService{
+// ConvertV1OpenShiftManagedClusterToOpenShiftCluster converts from a
+// v1.OpenShiftManagedCluster to a OpenShiftCluster.
+func ConvertV1OpenShiftManagedClusterToOpenShiftCluster(oc *v1.OpenShiftManagedCluster) *OpenShiftCluster {
```
kargakis: Still different?

pweil-: Not following now.  Are you asking to remove the types from the name completely?  I was thinking you were asking to update the types in the name itself.  We can't use the same function name, the files are in the same package.

kargakis: Why is `v1.OpenShiftManagedCluster` different from `OpenShiftCluster`? Shouldn't the former be `v1.OpenShiftCluster` or the latter `OpenShiftManagedCluster`?

pweil-: ohhhh, I see what you're asking now.  External is named that way because of

> Public name of the resource is going to be openShiftManagedClusters. This: https://github.com/openshift/openshift-azure/blob/master/pkg/api/v1/types.go#L5 should be updated to: OpenShiftManagedCluster

https://github.com/openshift/openshift-azure/issues/149

We could use the same internal if we wanted to, I have no strong opinion there and was just keeping with the convention that they were different.

kargakis: I would prefer to use the same name internally. Can be a follow-up.

kargakis:
needs a rebase

pweil-:
> I would prefer to use the same name internally. Can be a follow-up.

will just do it now so it's settled :+1:  Ready for re-review

PR #198, opened 2018-08-15
 Add interface for VM readiness checks 

kargakis:
Split out of https://github.com/openshift/openshift-azure/pull/138

@openshift/sig-azure ptal


jim-minter:
need to get sign-off from MSFT; consider changing this to a single method in Plugin with an additional enum-based argument to indicate node type

Issue #119, opened 2018-07-27
don't hard-code 3 masters


jim-minter:
at the moment this is moderately hard coded

mjudeikis:
Question:
Is this expected:
https://github.com/openshift/openshift-azure/blob/master/pkg/arm/data/master-startup.sh#L161-L166
We configure each master to point only to its local copy of ETCD.

Upstream: each master contains a full list of etcd clients. 
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_control_plane/templates/master.yaml.v1.j2#L54-L56

I see point for upstream to have this, as in most cases ETCD is external to masters. Now what we want to do on our end? 
@openshift/sig-azure ?

kargakis:
I don't think we will be running etcd externally.

mjudeikis:
This would explain upstream reasoning for having all the members in the config. In our case, what benefits/losses we would get if we have "each master point to its own ETCD vs having full list"?

mjudeikis:
/close 
don't thin this is relevant anymore

PR #199, opened 2018-08-15
[WIP] remove hardcoded master count

mjudeikis:
Removes hardcoded master count


mjudeikis:
So this was initial shot to fix https://github.com/openshift/openshift-azure/issues/119
But as any good Yak Shaving exercise, I got stuck on our configuration issue. 

To solve this properly we need to remove ETCD hardcoded coding from a master startup script. 
Did small POC, but stopped, because I know how sensitive is our `Config` struct, and potentially this will put a path to other configs (scheduler, master, node) simplification.

Few questions:
1. I think we want to move out all the code generation and templating to separate sub-package? Or its good to keep it under `config` one?
2. I was a little bit reluctant to write lexer, parser for any type of config file we will need. I think templating is the way here? Any other ideas?
3. Master, Node configuration structure? At the moment I added new sub-struct `Etcd` -> `EtcdConfig`, and `EtcdConfig` has all the fields + one field for serialized output to be used in arm? I not sure if this is the best way? This enables to reuse the same templating and translation code, but it makes the config structure enormously big... 

Before I proceed any further would like to get input from others.

cc: @openshift/sig-azure

PR #200, opened 2018-08-15
Revert "Updating since origin sku was removed."

pweil-:
Reverts openshift/openshift-azure#191

@kwoodson


kwoodson:
origin_310 sku will be available soon.

PR #202, opened 2018-08-15
Generate capacity as number

kargakis:
/hold


kargakis:
@openshift/sig-azure ptal, this is needed if we want to be serializing scale sets from ARM templates to the azure sdk type (compute.VirtualMachineScaleSet). (I am already doing this in https://github.com/openshift/openshift-azure/pull/138)

kargakis:
had to rebase, reapplying lgtm

PR #207, opened 2018-08-15
Avoid regenerating self-signed cert

kargakis:


kargakis:
e2e is green but for some reason the test pod is not completing, merging

Issue #212, opened 2018-08-16
Bring config bundles back


kargakis:
The master script size is getting out of hand when processed (eg. see https://github.com/openshift/openshift-azure/pull/189#issuecomment-413471035). One solution is to bring the config bundles back, tar containerservice.yaml, templatize, untar inside the template, use.

@openshift/sig-azure

jim-minter:
we've now got the central storage account in place - having a new container which runs on the nodes to sync config correctly may be an alternative next step worth considering.  Having careful reliability here may be key for speeding up our upgrade times.

kargakis:
Not sure this is an issue anymore, please reopen if you think so.

PR #213, opened 2018-08-16
add etcd template 

mjudeikis:
Template etcd config. Want to know if this is feasible as we need something like this so we could start modifying configuration from the code. One example is a different number of masters/etcd needs this.

ptal: @openshift/sig-azure


kargakis:
Related: https://github.com/openshift/openshift-azure/issues/212

jim-minter:
I'm not sure about this but I can't help but think it's taking us in the wrong direction.

mjudeikis:
What do you have in mind for this? We need to make master script smaller and code manageable?

jim-minter:
I'm not saying never but I'm convinced this is overkill right now.

PR #214, opened 2018-08-16
added human readable logging level flag

thanasisk:
- added command line switch "loglevel" that  adjusts log level in a human readable way
- adjusted scripts to use it with loglevel=debug

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -75,7 +75,7 @@ properties:
 EOF
 
 go generate ./...
-go run cmd/createorupdate/createorupdate.go
+go run cmd/createorupdate/createorupdate.go -loglevel=debug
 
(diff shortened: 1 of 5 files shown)


hack/create.sh
```diff
@@ -75,7 +75,7 @@ properties:
 EOF
 
 go generate ./...
-go run cmd/createorupdate/createorupdate.go
+go run cmd/createorupdate/createorupdate.go -loglevel=debug
```
kargakis: No reason to touch these if you already default to debug?

thanasisk: I prefer to be explicit than implicit - if we change the default level later, we will need those. What do you think?

kargakis:
ack

/lgtm

PR #218, opened 2018-08-16
ensure that all prerequisite packages are vendored

jim-minter:
@kargakis @mjudeikis is doing this worthwhile?

### DIFF
--- a/hack/validate-vendor.sh
+++ b/hack/validate-vendor.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+####################################################
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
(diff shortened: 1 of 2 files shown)


hack/validate-vendor.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -x
+
+T="$(mktemp -d)"
+trap "rm -rf ${T}" EXIT
+
+mkdir -p $T/src/github.com/openshift/openshift-azure
+cp -a "${DIR}/.." "${T}/src/github.com/openshift/openshift-azure"
+
+GOPATH=$T go build github.com/openshift/openshift-azure/...
```
kargakis: How is this different from [`make build`](https://github.com/openshift/release/blob/f80e62f1b8ea73ef4b4c0951e1603777cf80c53d/ci-operator/config/openshift/openshift-azure/master.json#L24) that we already run in CI?

jim-minter: Agreed - I thought I must be missing something.
@mjudeikis I think you were a bit fast on the lgtm button there ;)
Will revert.

PR #221, opened 2018-08-17
Logging cleanup

kargakis:
@openshift/sig-azure tbh I am not in love with our log package at its current form but consolidating sanitizeLogLevel in there made sense.

### DIFF
--- a/cmd/sync/sync.go
+++ b/cmd/sync/sync.go
@@ -3,16 +3,16 @@ package main
 import (
 	"flag"
 	"io/ioutil"
-	"strings"
 	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/openshift/openshift-azure/pkg/addons"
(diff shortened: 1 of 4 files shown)


cmd/sync/sync.go
```diff
@@ -23,24 +23,8 @@ var (
...
-	case "error":
-		return log.ErrorLevel
-	default:
-		// silently default to info
-		return log.InfoLevel
-	}
-}
 func sync() error {
-	log.Print("Sync process started!")
+	logrus.Print("Sync process started!")
```
mjudeikis: Any particular reason you use logrus directly and not via wrapper `log.Stuff`? Same as other cmd packages?

kargakis: We used logrus directly even before my changes.

mjudeikis: in sync I think because it was under development. All other code should be using log from wapper now. Wrapper main idea was not to pass `hc.log.stuff`, `p.log.stuff` around in the structs and just use it like this https://github.com/openshift/openshift-azure/blob/master/pkg/plugin/plugin.go#L42

kargakis: Sure, although I dislike globals (and every logging library has already its own global), I am not planning on doing such a change here.

mjudeikis: Whats the alternative? We pass `entry` file down the pipe to each our subpackage in the structs?

PR #222, opened 2018-08-17
[wip] Add e2e skeleton

mjudeikis:
Small skeleton/proposal on e2e suite using ginko. 

Was looking into using k8s e2e framework, but we have some dependency nightmare to import it. 
What do people think about this kind of approach?


kargakis:
Nice, we definitely need this. Let me know when you want a proper review.

mjudeikis:
First I would like to agree on approach and pseudo-sign-off from @jim-minter and @pweil- so I could continue working. Don't want to invest weeks time and find out that it is not how we would like to do things :)

pweil-:
I'm wondering if we shouldn't invest more time in getting the OpenShift/Kube suites running on our cluster rather than redefining what an extended tests when we're talking specifically about CRUD for in-cluster objects.

In addition to the conformance suites it is probably also useful for us to be testing scale and upgrade in an automated fashion.  Those operations are specific to our code base so they wouldn't be covered in any other conformance suite.

pweil-:
> Should we say:
> E2E - target to run Origin/k8s test. Might require either create separate origin container where we pull it it and run on top of our cluster (I dont want to vendor it in, I would want just run it with our kubeconfig)
> OSA-Conformance - target our upgrade/scale set tests. This would be custom to us and our flow.

that's pretty much what I was thinking, yes.  And +1 to running the extended tests in a container.  All we should have to provide is the kube config and TEST_ONLY flag so it will run.

pweil-:
also of note: we're already going to run into some issues since we went our own way WRT deployment vs DCs.  Some of the tests for routers look specifically for a DC in the default namespace to know if a router is installed on the cluster.

mjudeikis:
> And +1 to running the extended tests in a container. All we should have to provide is the kube config and TEST_ONLY flag so it will run.

Sounds like a story for a backlog. We still need some small method to test scaleups and stuff, but it should be full fledge framework.

test/extended/deployment_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+var _ = g.Describe("Deployment Suite", func() {
+	defer g.GinkgoRecover()
+
+	// Create temp namespace based on ginko seed
+	g.BeforeEach(func() {
+		// Create temp namespace for tests
+		buf, err := Asset("Namespaces/test.yaml")
+		o, err := cl.Unmarshal(buf)
```
jim-minter: I'm not liking this pattern.  I'd be in favour of having `cl` include a proper k8s clientset; separate free-standing helper functions to create and tear down namespaces, no yaml files.

test/extended/deployment_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+		o.SetName(fmt.Sprintf("test-%d", g.GinkgoRandomSeed()))
+		cl.Delete(o)
+	})
+
+	g.It("should deploy deployment", func() {
+		for _, asset := range AssetNames() {
+			log.Println(asset)
+		}
+		buf, err := Asset("Deployment.apps/simple-deployment.yaml")
```
jim-minter: I think I would just write this out long form (`d := &v1.Deployment{ ... }`)

test/extended/client.go
```diff
@@ -0,0 +1,240 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/client-go/util/retry"
+	kaggregator "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
+)
+
+// Interface exposes the methods a client needs to implement
+// for the syncing process of the addons.
+type Interface interface {
```
jim-minter: I don't think this is necessary or appropriate.  This is test code and we're supposing `client` will be a singleton.  just use a struct with public members for all the clients you need.  Start with https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L127 because it contains nearly everything.

test/extended/client.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if err := c.UpdateDynamicClient(); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+// UpdateDynamicClient updates the client's server API group resource
```
jim-minter: I don't think we should have any of the copy/paste from here to the end.

pweil-:
Summary of above:

openshift-azure e2e should:

* spin up a cluster
* perform health check as normal (this is our sanity check)
* run origin e2e
* run scale and upgrade operations
* tear down the cluster

openshift-azure tests should not:

* test creation of in-cluster resources - this belongs in origin suites so we aren't duplicating effort

kargakis:
@pweil- @jim-minter fwiw we already can run conformance e2e and test upgrades w/o this framework. Maybe it's preferrable to move away from bash+openshift templates to a go binary that will be backed by this?

kargakis:
I feel we definitely need to think about adding coverage for things unique to our deployment, eg. ensure the sync pod is reentrant, test logbridge works as expected, try to make a storage account key rotation for the docker registry, etc.

kargakis:
/close

Will pick it up next week

PR #224, opened 2018-08-20
extend tls.go unit test coverage

thanasisk:
added partial test coverage for tls.go


pkg/tls/tls_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if newPK.Validate() != nil {
+		t.Errorf("Failed to validate private key")
+	}
+}
+
+//func newCert(key *rsa.PrivateKey, template *x509.Certificate,
+//signingkey *rsa.PrivateKey, signingcert *x509.Certificate)
+//(*x509.Certificate, error) {
+func testSelfSignedNewCert(t *testing.T) {
```
jim-minter: would need to be exported (capital T on `testSelfSignedNewCert`) to run

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
+	"math/big"
+	"testing"
+	"time"
+)
+
+// x509 expects quite a large number so 0xDEADBEEF will not cut it
```
jim-minter: I don't believe replicating these functions is the right approach.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+var cn = "dummy-test-certificate.local"
+
+//func NewPrivateKey() (*rsa.PrivateKey, error) {
+func TestNewPrivateKey(t *testing.T) {
+	newPK, err := NewPrivateKey()
+	if err != nil {
+		t.Errorf("failed to create private key")
+	}
+	if newPK.Validate() != nil {
```
jim-minter: I suppose you can also validate the private key length.

pkg/tls/tls_test.go
jim-minter: I would not test this function.  Test NewCA() and NewCert() instead.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+//func NewCA(cn string) (*rsa.PrivateKey, *x509.Certificate, error) {
+func testNewCA(t *testing.T) {
+	pk, cert, err := NewCA(cn)
+	if err != nil {
+		t.Errorf("failed to create new CA")
+	}
+	if pk.Validate() != nil {
```
jim-minter: this is somewhat duplicate.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,91 @@
...
+		t.Errorf("failed to validate RSA Private key")
+	}
+	opts := x509.VerifyOptions{DNSName: cn}
+	if _, err := cert.Verify(opts); err != nil {
+		t.Errorf("Failed to verify created self signed certificate")
+	}
+}
```
jim-minter: if you're going to do this, I think your approach here ends up being to zero out the parts of the returned cert which are dependent on time or the private key, then compare whatever is left to a known good value.  Not pretty.

jim-minter: Then the same for `TestNewCert()`.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,87 @@
...
+	dnsNames := []string{cn}
+	ipAddresses := []net.IP{net.ParseIP("192.168.0.1")}
+	keyUsage := []x509.ExtKeyUsage{x509.ExtKeyUsageAny}
+	selfSign := true
+
+	pk, cert, err := NewCert(cn, org, dnsNames, ipAddresses, keyUsage, nil, nil, selfSign)
+	if err != nil {
+		t.Errorf("error creating new cert %s", err)
+	}
+	if pk.Validate() != nil {
```
mjudeikis: I think this will fail always as its SelfSign. In self sign case it should be signed by itself. In this case we would need to check if `issued By: Common Name (CN)	*.badssl.com` matches `Issued to: Common Name (CN)	*.badssl.com` and is what we provided.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	knownGoodCerts, err := x509.ParseCertificates(block.Bytes)
+	if err != nil {
+		t.Errorf("unable to parse certificates %s", err)
+	}
+	goodCert := knownGoodCerts[0]
+	cert.SerialNumber = big.NewInt(0)
+	cert.NotBefore = time.Unix(0, 0)
+	cert.NotAfter = time.Unix(0, 0)
+	if reflect.DeepEqual(goodCert, cert) == false {
```
mjudeikis: `!reflect.DeepEqual(goodCert,cert)` or `reflect.DeepEqual(goodCert,cert)` depending on what you want.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"net"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/go-test/deep"
+)
+
+// x509 expects quite a large number so 0xDEADBEEF will not cut it
```
jim-minter: make comments useful/meaningful or don't add them please

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+		if err != nil {
+			t.Errorf("Cannot PEM encode certificate")
+		}
+		err = ioutil.WriteFile(knownGoodLocCA, pemEncodedCert, 0644)
+		if err != nil {
+			t.Errorf("unable to save PEM cert %s", err)
+		}
+	}
+	goodCert := loadGoodCert(knownGoodLocCA)
+	cert.SerialNumber = big.NewInt(0)
```
jim-minter: any pointer or slice fields that need to be nilled - set them to nil.  Others - set them directly to the public struct zero value

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+	cert.NotBefore = time.Unix(0, 0)
+	cert.NotAfter = time.Unix(0, 0)
```
jim-minter: time.Time{}

pkg/tls/tls_test.go
jim-minter: DRY: should be doing
```
for _, c := range []*x509.Certificate{cert, goodCert} {
	c.Thing = nil
	...
```

jim-minter: I calculate that you only need
```
		c.NotBefore = time.Time{}
		c.NotAfter = time.Time{}
		c.PublicKey.(*rsa.PublicKey).N = nil
		c.Raw = nil
		c.RawSubjectPublicKeyInfo = nil
		c.RawTBSCertificate = nil
		c.SerialNumber = nil
		c.Signature = nil
```

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+	goodCert.KeyUsage = 37
+	goodCert.DNSNames = nil
+	goodCert.IPAddresses = nil
+	goodCert.ExtKeyUsage = nil
+	goodCert.Issuer.CommonName = cn
+	goodCert.Subject.Names = nil
+	cert.Issuer.Names = nil
+	cert.Subject.CommonName = ""
+	cert.Subject.Names = nil
+	if diff := deep.Equal(cert, goodCert); diff != nil {
```
jim-minter: MUST use reflect.DeepEqual() for the test - unfortunately deep.Equal sometimes hides differences

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+func TestNewSelfSignedCert(t *testing.T) {
+	cn := cn
+	org := []string{"dummy-test-certificate.local"}
+	dnsNames := []string{cn}
+	ipAddresses := []net.IP{net.ParseIP("192.168.0.1")}
+	extKeyUsage := []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth | x509.ExtKeyUsageClientAuth}
+	selfSign := true
+
+	pk, cert, err := NewCert(cn, org, dnsNames, ipAddresses, extKeyUsage, nil, nil, selfSign)
```
jim-minter: Additional test required for CA signed certificate.  Look at cert.Verify().

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,241 @@
...
+const knownGoodLoc = "./testdata/known_good_cert.pem"
+const knownGoodLocCA = "./testdata/known_good_certCA.pem"
+const updateSSLVar = "UPDATE_KNOWN_SSL_CERT"
+
+var orgs = []string{cn}
+
+func TestNewPrivateKey(t *testing.T) {
+	newPK, err := NewPrivateKey()
+	if err != nil {
+		t.Errorf("failed to create private key")
```
jim-minter: no need to go to the effort of `t.Errorf("some new error message")` in tests.  The line number is already reported for you if the test fails.  Use `t.Error(err)` or `t.Fatal(err)` as a general rule.

pkg/tls/tls_test.go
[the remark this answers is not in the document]
thanasisk: my bad, fully understood

pkg/tls/tls_test.go
thanasisk: many thanks for this excellent suggestion

pkg/tls/tls_test.go
thanasisk: understood

PR #227, opened 2018-08-20
test generation required fields and merge

pweil-:
tests for generation setting the required fields

tests for merge of old/new config

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -464,8 +464,9 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
 
 	if len(c.RegistryConsoleOAuthSecret) == 0 {
 		if pass, err := randomString(64); err != nil {
+			return err
+		} else {
 			c.RegistryConsoleOAuthSecret = fmt.Sprintf("user%s", pass)
-			return nil
 		}
 	}
 
--- a/pkg/config/generate_test.go
+++ b/pkg/config/generate_test.go
@@ -5,8 +5,10 @@ import (
 	"testing"
 
 	"github.com/ghodss/yaml"
+	"github.com/satori/uuid"
 
-	acsapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/util/fixtures"
 )
 
 var testOpenShiftClusterYAML = []byte(`---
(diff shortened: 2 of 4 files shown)


pkg/config/generate_test.go
```diff
@@ -26,39 +27,162 @@ properties:
...
+			t.Errorf("%s received generation error %v", name, err)
+			continue
+		}
+		testRequiredFields(test.omc, t)
+		// check mutation
+	}
+}
+
+func testRequiredFields(omc *api.OpenShiftManagedCluster, t *testing.T) {
+	checkStringField := func(f string, name string) {
```
jim-minter: the top 3 of these condense down to `assert := func(b bool, msg string) { if(!b) { t.Error(msg) } }`

pkg/config/generate_test.go
```diff
@@ -26,39 +27,162 @@ properties:
...
+			t.Errorf("missing %s", name)
+		}
+	}
+	checkRandom := func(c bool, name string) {
+		if c {
+			t.Errorf("missing %s", name)
+		}
+	}
+	checkCert := func(c api.CertKeyPair, name string) {
+		if c.Key == nil {
```
jim-minter: then `assert(c.Key != nil, fmt.Sprintf("%s missing key", name)); assert(c.Cert != nil, fmt.Sprintf("%s missing cert", name))`

pkg/config/generate_test.go
```diff
@@ -97,3 +221,60 @@ func TestSelectDNSNames(t *testing.T) {
 
 	}
 }
+
+func testOpenShiftCluster() *api.OpenShiftManagedCluster {
```
jim-minter: testOpenShiftClusterYAML above took the tack of a small yaml representation and unmarshalling it.  This PR is adding two further testOpenShiftCluster() functions.  Can this be simplified?

pkg/config/generate.go
```diff
@@ -464,8 +464,9 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
 
 	if len(c.RegistryConsoleOAuthSecret) == 0 {
 		if pass, err := randomString(64); err != nil {
+			return err
```
jim-minter: whoops :(

pkg/config/generate_test.go
jim-minter: the reason is a lack of deepcopy implementation, will note to fix in #192
nonetheless can the two testOpenShiftCluster() functions be condensed?

pweil-: sure, it needs to be a third package though so we don't make weird dependencies.  I'll throw it as a fixture.  Having it not be in a test package is also a smell to me though, I'd rather accept duplication in the unit test and allow the creation be specific to the package itself since, strictly speaking, testing different plugin impl methods represent different stages of the config's lifecycle.

Done.

pkg/config/generate_test.go
```diff
@@ -26,39 +28,148 @@ properties:
...
+			continue
+		}
+		testRequiredFields(test.omc, t)
+		// check mutation
+	}
+}
+
+func testRequiredFields(omc *api.OpenShiftManagedCluster, t *testing.T) {
+	assert := func(c bool, name string) {
+		if c {
```
jim-minter: sorry, I'd specifically do !c here and reverse all the booleans in the callers, so that they read "assert(thing_that_should_be_true)" not "assert(thing_that_should_be_false)"

Issue #175, opened 2018-08-09
move to internal FQDN for sync kubeconfig


mjudeikis:
I think it would be more resilient to move to internal FQDN Loadbalancer for sync kubeconfig
https://github.com/openshift/openshift-azure/blob/master/pkg/config/generate.go#L406-L414

It gets created almost immediately, and in arm we already have this dependency form SS: 
` "[resourceId('Microsoft.Network/loadBalancers', 'lb-apiserver')]", `
In addition, if for some reason master0001 gets slower startup, it will impact the sync process. Never seen this for master1, but few times for master3 when it gets "late start to join cluster".

If we rotate master or take one out because of the "issues", sync process will get impacted too. So if we need to sync something meantime, we can't do it.

kargakis:
How exactly is the sync process impacted if it runs in a deployment? An upgrade is going to roll masters first, then infra and compute nodes, then kick the sync process, or maybe move compute node upgrade in the end.

mjudeikis:
Let's say MS has an issue with their infra, and master1 goes offline for few seconds/minutes. Or node gets not ready due to some storage issues. We cant run sync at this point.

mjudeikis:
closed via sync running in pod

PR #231, opened 2018-08-21
[WIP] Run sync as pod

jim-minter:
@pweil- @mjudeikis @kargakis replacement for #189 and #220.  This actually basically works although it needs some cleaning up ...

It creates the central config SA we've been discussing, then runs sync on master-000000 as a static pod.


cmd/sync/sync.go
```diff
@@ -23,17 +27,91 @@ var (
...
+	if err != nil {
+		return errors.Wrap(err, "cannot read _data/_out/azure.conf")
+	}
+
+	var m map[string]string
+	if err := yaml.Unmarshal(b, &m); err != nil {
+		return errors.Wrap(err, "cannot unmarshal _data/_out/azure.conf")
+	}
+
+	az, err := newAzureClients(context.Background(), m)
```
mjudeikis: So we are running sync with `cloudConfig` credentials?

pkg/arm/data/master-startup.sh
```diff
@@ -572,6 +562,34 @@ spec:
     name: master-cloud-provider
 EOF
 
+{{ if not (eq "true" .Config.RunSyncLocal) }}
+if [[ "$(hostname)" == "master-000000" ]]; then
```
mjudeikis: I really dont like this...  Partially related to this: https://github.com/openshift/openshift-azure/issues/175
Any issues with master0 takes out our syn process down.

pkg/config/generate.go
```diff
@@ -410,16 +411,6 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
 			kubeconfig: &c.NodeBootstrapKubeconfig,
 			namespace:  "openshift-infra",
 		},
-		{
-			clientKey:  c.Certificates.Admin.Key,
-			clientCert: c.Certificates.Admin.Cert,
-			// sync kubeconfig has the same capabilities as admin kubeconfig, only difference
-			// is the use of HCP internal DNS to avoid waiting for the Azure loadbalancer to
-			// come up in order to start creating cluster objects.
-			endpoint:   "master-000000",
```
mjudeikis: YAY! https://github.com/openshift/openshift-azure/issues/175 closed but re-introduced in previous comment in different form

pkg/config/generate.go
```diff
@@ -11,6 +11,7 @@ import (
 	"os"
```
mjudeikis: This file getting enormous. Maybe worth splitting it to separate files like `generate.go` - only generate function, `images.go` - image resolved `resolvers.go/hellpers.go` for all stuff like `selectNodeImage` `selectContainerImages` ? 

Not in this PR but I would like to see it split :/

pkg/initialize/client.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type azureStorageClient struct {
+	accounts storage.AccountsClient
+	storage  azstorage.Client
+}
+
+func newAzureClients(ctx context.Context, cs *api.OpenShiftManagedCluster) (*azureStorageClient, error) {
+	// TODO: needs to come from property bag
+	authorizer, err := auth.NewAuthorizerFromEnvironment()
```
mjudeikis: I imagine this will change to use credentials from `Context` ?

cmd/sync/sync.go
jim-minter: yes- the cloudConfig credential is a service principal set up in the management tenant which has full rights over the resource group in the customer's subscription.  This is able to read the config data stored in the SA.

pkg/arm/data/master-startup.sh
jim-minter: 100% agree.  It is abominably hard to get the sync pod to run at this stage in the cluster startup by any other means.  This is a stopgap measure.  If we ran the SDN workload as static pods, then nodes would immediately come up ready, which would speed setup time, and would then enable us to run the sync pod as a deployment.  #242.

pkg/config/generate.go
jim-minter: agree

glide.yaml
```diff
@@ -23,4 +23,6 @@ import:
 - package: github.com/go-test/deep
   version: v1.0.1
 - package: github.com/satori/go.uuid
-  version: f58768cc1a7a7e77a3bd49e98cdd21419399b6a3 # MS RP used version
+  version: f58768cc1a7a7e77a3bd49e98cdd21419399b6a3
+- package: github.com/marstr/guid
```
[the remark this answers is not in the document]
jim-minter: we started using the azure blob storage api.  glide detected this and now pulled in additional prereqs.

pkg/initialize/client.go
jim-minter: that's what the comment above is meant to mean ;)

jim-minter: #243

pkg/config/generate.go
mjudeikis: https://github.com/openshift/openshift-azure/issues/248

PR #233, opened 2018-08-21
remove dns check

mjudeikis:
rollback dns check. Will move to testing.


kargakis:
Did you also need to update glide.yaml to remove the network subpackage?

Issue #245, opened 2018-08-22
prometheus oauth proxies don't trust CA 


pweil-:
2018/08/22 15:52:37 oauthproxy.go:582: error redeeming code (client:10.131.0.8:32798): Post https://openshift.pweilosa.osadev.cloud/oauth/token: x509: certificate signed by unknown authority

kwoodson:
Looks to be the same issue:

https://github.com/openshift/openshift-azure/issues/240

Issue #240, opened 2018-08-22
prometheus login results in a 500 internal server error


kwoodson:
I attempted to login here:
https://prometheus-openshift-metrics.kwoodsontest.osadev.cloud/

I chose local password login and aad login.  I received:
```
500 Internal Error
Internal Error
```

```
2018/08/22 14:31:06 oauthproxy.go:635: error redeeming code (client:10.128.2.3:57846): Post https://openshift.kwoodsontest.osadev.cloud/oauth/token: x509: certificate signed by unknown authority
2018/08/22 14:31:06 oauthproxy.go:434: ErrorPage 500 Internal Error Internal Error
2018/08/22 14:31:06 provider.go:382: authorizer reason: User "system:anonymous" cannot get namespaces in project "openshift-metrics"
2018/08/22 14:31:06 provider.go:382: authorizer reason: User "system:anonymous" cannot get namespaces in project "openshift-metrics"
```

kwoodson:
Had a discussion with Jim, this is on purpose for the users.  We do not want users to login to prometheus but we do want to come up with a way for SRE to access prometheus.

PR #246, opened 2018-08-22
add web console cert to CAs for oauth proxy (prometheus and alert man…

pweil-:
…ager)

Fixes https://github.com/openshift/openshift-azure/issues/245
Fixes https://github.com/openshift/openshift-azure/issues/240

We need to consider how we'll do this when you can BYO certificates but this gets me to a 403 which is expected for the htpassword user due to not having get permissions on `openshift-metrics`

```
2018/08/22 17:08:21 provider.go:395: Permission denied for demo@cluster.local for check {
  "namespace": "openshift-metrics",
  "resource": "namespaces",
  "resourceName": "openshift-metrics",
  "scopes": [],
  "verb": "get"
}
```


pweil-:
> We need to consider how we'll do this when you can BYO certificates but this gets me to a 403 which is expected for the htpassword user due to not having get permissions on openshift-metrics

to follow on here, an ` oc adm policy add-role-to-user view demo` for the `openshift-metrics` namespace satisfies the SAR, we just need to decide how we want to get users into the system and assigned a role that grants that permission

PR #254, opened 2018-08-23
bring internal type into line with external type

### LINKED ISSUE #251: re-review API definitions
seems like we still have unexpected pointers
@mjudeikis fyi

jim-minter:
@pweil- @amanohar @mjudeikis @kargakis 

I'm worried about how different our internal and external types are and would really like to bring them into line.  It's painful but if we don't do this before PP, I don't know when we will.

1- add pointers to MasterPoolProfile and ServicePrincipalProfile in the external type.

2- remove ProfileSpec - don't believe it conforms with Azure API standards and it's not interchangeable between MasterPoolProfile and AgentPoolProfile.  If we really want a unit test to guarantee consistency of types I can add it.

3- refactor internal type to bring into line with the external type.  In particular, remove OrchestratorProfile (surely this is an ACSEngine-ism, defunct, and just increases code verbiage and nil pointer risk?)

4- along the way: add and correct `json:` tags on internal API - these are essential to ensure correct serialisation.  Add unit tests proving same.

5- colour me highly reluctant to have MasterProfile in external and not in internal.  I'm happy to stick with it in external if needs be, but I think it's best for bring internal into line.  Comments?

6- diff between pkg/api/types.go and pkg/api/v1/types.go is nearly tolerable to me at this point (notwithstanding 5).

fixes #251


mjudeikis:
:1st_place_medal:

pweil-:
> 2- remove ProfileSpec - don't believe it conforms with Azure API standards and it's not interchangeable between MasterPoolProfile and AgentPoolProfile. If we really want a unit test to guarantee consistency of types I can add it.

I do not like this change.  It is not DRY and just means that any field we want to add to both needs updated in two places.  Reuse is good practice, do you have a guide on the API standards?

> 5- colour me highly reluctant to have MasterProfile in external and not in internal. I'm happy to stick with it in external if needs be, but I think it's best for bring internal into line. Comments?

the reason I didn't do this to begin with is so that we can treat them the same in the template and everywhere else we iterate, validate, etc.  I'm not  against it but it's a copy/paste in parts of the code which doesn't sound great. 

:+1:  to the rest

kargakis:
I think I also prefer adding master profile in internal, it certainly would ease conversions and we can hope to move to generated one day.

PR #256, opened 2018-08-24
Property bag/Context

mjudeikis:
Suggestion to start using Context with all external methods we have. Some of them at the moment does not have use for context,  but I added ctx object too just to keep consistent api. 

The external consumer should not care about how context is being used. And by introducing it early we can use it later. 

At the moment context flow:
`createorUpdate` -> `plugin`-> `GenerateARM` -> `arm.Generator`
`createorUpdate` -> `plugin`-> `MergeConfig` 
`createorUpdate` -> `plugin`-> `Validate` 
`createorUpdate` -> `plugin`-> `GenerateConfig`
`createorUpdate` -> `plugin`-> `InitializeCluster` -> `intitialize.Initializer`
`createorUpdate` -> `plugin`-> `HealthCheck` -> `healthcheck.Healthchecker`
`createorUpdate` -> `plugin`-> `Drain` -> `upgrade.Upgrader`
`createorUpdate` -> `plugin`-> `WaitForReady` -> `upgrade.Upgrader`

I would keep the api consistent state for all plugin methods, and as per needs in secondary packages.

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -21,7 +21,7 @@ import (
 var logLevel = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
 
 // createOrUpdate simulates the RP
-func createOrUpdate(oc *v1.OpenShiftManagedCluster, entry *logrus.Entry) (*v1.OpenShiftManagedCluster, error) {
+func createOrUpdate(ctx context.Context, oc *v1.OpenShiftManagedCluster, entry *logrus.Entry) (*v1.OpenShiftManagedCluster, error) {
 	// instantiate the plugin
 	p := plugin.NewPlugin(entry)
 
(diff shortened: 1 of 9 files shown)


pkg/api/v1/types.go
```diff
@@ -2,6 +2,8 @@ package v1
 
 // TODO publish v1 to a dated api as in Azure conventions
 
+type PropertyBagKey string
```
pweil-: We will need defined constants so that the RP can use them.  I don't think it belongs in the external API though.  Perhaps in the plugin.

mjudeikis: Agree. But didnt knew where we want to see it so just added here as a placeholder. We can even use a string as a key, buts its suppose to be bad practice :D

cmd/createorupdate/createorupdate.go
```diff
@@ -202,9 +202,19 @@ func main() {
 		log.Fatal(err)
 	}
 
+	//simualte Context with property bag
+	pb := make(map[string]string)
+	pb["ClientID"] = os.Getenv("AZURE_CLIENT_ID")
```
jim-minter: any reason not to do:

in api.go:

```
type ContextKey string

const (
	ContextKeyClientID ContextKey = "ClientID"
	ContextKeyClientSecret ContextKey = "ClientSecret"
	...
)
```

here:

```
ctx := context.Background()
ctx = ctx.WithValue(ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID")
ctx = ctx.WithValue(ContextKeyClientSecret, os.Getenv("AZURE_CLIENT_SECRET")
...
```

at point of use:

```
config := auth.NewClientCredentialsConfig(ctx.Value(ContextKeyClientID).(string), ...)
```

Avoids v1.PropertyBagKey and the additional map.

mjudeikis: like it :)

mjudeikis:
@jim-minter added as suggested

PR #257, opened 2018-08-24
split generate

mjudeikis:
split generate to more readable chunks


pkg/config/util.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return "", err
+		}
+		b[i] = letterBytes[o.Int64()]
+	}
+
+	return string(b), nil
+}
+
+func selectDNSNames(cs *acsapi.OpenShiftManagedCluster) {
```
jim-minter: this function doesn't belong here

jim-minter: it should probably be refactored away

mjudeikis: @jim-minter , another place for this to be split would be a template and be using goTemplate. I would like to avoid constructing variables in that place and keep all config generation in here. What you had in mind for this?

jim-minter: I think RouterLBCNamePrefix, MasterLBCNamePrefix and other derived configs should probably be member functions on OpenShiftManagedCluster.
I don't know what the rationale is for the rest of the function - why is it there?

PR #258, opened 2018-08-24
add limits for node process

mjudeikis:
add a small reservation for node process to ring-fence system process if a user "fills the node" so we can avoid OOM


mjudeikis:
https://docs.openshift.com/enterprise/3.2/admin_guide/allocating_node_resources.html

kwoodson:
@mjudeikis, in dedicated we run with these:

`cpu=500m,memory=512M`

Either way I'm fine with what we choose and we can tune after-the-fact.

mjudeikis:
@kwoodson I did 500m but after this noticed that this is quite a lot for small boxes. I might add some logic into this to inject number based on size of the VM.

pkg/util/util.go
```diff
@@ -29,6 +29,17 @@ func Template(tmpl string, f template.FuncMap, cs *acsapi.OpenShiftManagedCluste
...
+			AgentPoolProfileRoleDummy := acsapi.AgentPoolProfileRole(r)
+			for i, ap := range cs.Properties.AgentPoolProfiles {
+				if ap.Role == AgentPoolProfileRoleDummy {
+					return i
+				}
+			}
+			// This should never happen
+			panic("template failed")
+			return 0
+		},
```
mjudeikis: I really really dont like how I did this part. I need to sleep on it to think something else, if @pweil- or @jim-minter have any suggestions, let me know. Either change API or logic to inject those values into the template 

Tests will follow when we settle down on implementation and API.

jim-minter:
the idea's right, the proactivity is great, the implementation went wobbly.

1. don't persist KubeReserved/SystemReserved, don't see why it's necessary

2. define a static map of valid vm sizes and associated settings
(e.g. KubeReserved/SystemReserved; maybe there'll be more of these
later) - check against that map in the validate and also to resolve
the settings given the vm size - encapsulate the data in 1 place ;-)

3. don't add the agentPoolIndex function - wrong direction

4. refactor readDB() and move the `for _, tr := range ts`
functionality into a new function in translate.go - that's where it
really belonged.  Take care not to break hack/update-addons.go along
the way.

5. now extend `var Translations` and add another field of type, say
`func(*acsapi.OpenShiftManagedCluster) string`.  In your new function
from 4, have it call that function field if the Template field is left
empty.  Now populate that function field with a suitable function to
look up in the static map to populate the kube configs.

jim-minter:
dunno if ^ will be perfect but it should be much closer ;-)

jim-minter:
hmm, ^ would possibly solve the problem of how to get rid of RouterLBCNamePrefix, but it doesn't solve how to get rid of MasterLBCNamePrefix.  Can you think whether ^ can be adapted to be able to do that?  Maybe don't add the new field type to Translations, but start adding functions like `func (c *Config) DeriveRouterLBCNamePrefix() string` which can return all these derived data *and* which can be called directly from templates.

mjudeikis:
@jim-minter will update tests when done and agreed on implementation. 
We should say rule, anything in `addons` goes via Derived config whats not in the main config.
And in the future same for `master-config`, `arm` template, when we can move them to similar translation maps... Maybe not arm...

pkg/addons/translate.go
```diff
@@ -25,6 +29,30 @@ const (
 	NestedFlagsBase64 NestedFlags = (1 << iota)
 )
 
+func translateAsset(o unstructured.Unstructured, cs *acsapi.OpenShiftManagedCluster) (unstructured.Unstructured, error) {
+
```
kargakis: nit: remove newline

pkg/addons/translate.go
```diff
@@ -25,6 +29,30 @@ const (
...
+			}
+		}
+
+		err := Translate(o.Object, tr.Path, tr.NestedPath, tr.NestedFlags, s)
+		if err != nil {
+			return unstructured.Unstructured{}, err
+		}
+	}
+	return o, nil
+
```
kargakis: nit: remove newline

jim-minter:
/hold cancel

have marked wip, because it is

pkg/addons/translate.go
```diff
@@ -72,6 +98,7 @@ var Translations = map[string][]struct {
 	NestedPath  jsonpath.Path
 	NestedFlags NestedFlags
 	Template    string
+	F           func(*acsapi.OpenShiftManagedCluster) string
```
kargakis: You can still use the derived lbc prefix without this, no?

pkg/api/config.go
```diff
@@ -71,7 +71,6 @@ type Config struct {
 	LoggingWorkspace           string // workspace for Azure Log Analytics resource
 
 	// DNS configurables
-	RouterLBCNamePrefix string
```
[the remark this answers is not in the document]
mjudeikis: It is being derived now. Any data which is the direct derivative result of other configuration items in the config is overpopulating our config file. This was one of those cases (same as MasterLBPrefix but we can remove it now). 

It just cleans our internal config little bit. Agree, this could be follow-up PR, but as we discussed with @jim-minter that this could be next candidate to move to `derivative data` m,odel I just threw it here.

pkg/addons/translate.go
mjudeikis: as a function to get value?

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -146,6 +146,16 @@ const (
 	AgentPoolProfileRoleInfra AgentPoolProfileRole = "infra"
 )
 
+// VMSize represents supported VMSizes
+type VMSize string
```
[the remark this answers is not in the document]
mjudeikis: Seen it, open for discussion. No hard feeling on this...

pkg/addons/translate.go
kargakis: I think this is fine, nvmd

kargakis:
/lgtm
/hold

for @jim-minter to take one last look

pkg/api/2018-09-30-preview/api/types.go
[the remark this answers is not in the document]
kargakis: On the upside, we avoid pulling in a bunch of code for a single type so we may just get on with our own.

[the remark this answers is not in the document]
mjudeikis: This was my intention too. We could just use `VirtualMachineSizeTypes` and `VirtualMachineSizeTypesStandardDS3` as naming standard in here and if this overgrows we will have straight migration path.

pkg/addons/translate.go
```diff
@@ -25,6 +29,28 @@ const (
 	NestedFlagsBase64 NestedFlags = (1 << iota)
 )
 
+func translateAsset(o unstructured.Unstructured, cs *acsapi.OpenShiftManagedCluster, ext *extra) (unstructured.Unstructured, error) {
+	ts := Translations[KeyFunc(o.GroupVersionKind().GroupKind(), o.GetNamespace(), o.GetName())]
+	for _, tr := range ts {
+		var s string
+		if tr.F != nil {
+			s = tr.F(cs)
```
jim-minter: in the future we can give tr.F the signature `func(*acsapi.OpenShiftManagedCluster) (string, error)` and if nil default it here to a closure wrapping util.Template, but this is fine for now.

pkg/addons/data/ConfigMap/openshift-node/node-config-compute.yaml
```diff
@@ -36,6 +36,10 @@ data:
       - /etc/origin/node/pods
       rotate-certificates:
       - "true"
+      kube-reserved:
```
jim-minter: should we do this on masters as well?

pkg/validate/validate.go
```diff
@@ -184,9 +184,10 @@ func validateAgentPoolProfile(app *api.AgentPoolProfile) (errs []error) {
 		errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].role %q", app.Name, app.Role))
 	}
 
+	// must match generate.go limit geneartion code
```
jim-minter: I was looking for something like `if _, found := DefaultVMSizeKubeArguments[app.VMSize]; !found { // error }` here

In which case you need to change the meaning of DefaultVMSizeKubeArguments a little - it's more like the map of valid VM names and various associated parameters for each.

I don't know if pkg/api is the right location for that map.

pkg/config/derive.go
```diff
@@ -0,0 +1,31 @@
+package config
+
+import (
+	"strings"
+
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
+)
+
+func DerivedSystemReserved(cs *acsapi.OpenShiftManagedCluster, role acsapi.AgentPoolProfileRole) string {
+	for _, pool := range cs.Properties.AgentPoolProfiles {
```
jim-minter: @kargakis more code that suggests AgentPoolProfiles should be a map internally?
@mjudeikis should have a single `GetAPP(cs *acsapi.OpenShiftManagedCluster, role acsapi.AgentPoolProfileRole) (*acsapi.AgentPoolProfile, error)` function in our codebase somewhere?

pkg/api/types.go
```diff
@@ -133,6 +133,16 @@ const (
 	AgentPoolProfileRoleMaster AgentPoolProfileRole = "master"
 )
 
+// VMSize represents supported VMSizes
```
jim-minter: this is probably the "right" thing to do, but one more thing to keep in sync ...

pkg/addons/data/ConfigMap/openshift-node/node-config-compute.yaml
mjudeikis: Master does not run any user workload, and kubelet static pods are the ones with heavy workload. Not sure if it worth limiting those (never did this on side for those specific reasons). Maybe @kwoodson can tell how they do in Dedicated?

pkg/validate/validate.go
mjudeikis: If not in `pkg/api`. I see only 2 places, either `pkg/config` or `pkg/api`. And we should agree where we keep all defaults in one place. So we would not end-up calling some defaults from `api` as const values, and some as variables from `config`.

pkg/api/types.go
mjudeikis: As discussed with @kargakis above, we could include azure library for this. But maybe lets start small and lets see.

pkg/config/derive.go
kargakis: > @kargakis more code that suggests AgentPoolProfiles should be a map internally?

Or create some boillerplate code that can abstract this from elsewhere? It should be a single helper that reads a []AgentPoolProfile and a role and returns a *AgentPoolProfile, right? Simpler than diverging between internal and external. Not simpler than adding a map in the API but saner I think.

PR #262, opened 2018-08-24
Add centralised logging :-D

jim-minter:


pkg/config/generate.go
```diff
@@ -74,6 +74,7 @@ func selectContainerImagesOrigin(cs *acsapi.OpenShiftManagedCluster) {
 		c.AzureCLIImage = "docker.io/microsoft/azure-cli:latest"
 
 		c.SyncImage = "quay.io/openshift-on-azure/sync:latest"
+		c.LogBridgeImage = "docker.io/jimminter/logbridge:latest"
```
kwoodson: If we decide to go with this `logbridge` should we move this to a more official location?

pweil-:
Word of caution here, we switched away from journal logging due to the fact that we could run into rate limiting in the journal with enough activity.  Need to confirm if that is still an issue or if it has been worked around.  The result is loss of logs.

pkg/config/generate.go
jim-minter: yes, I think I'll do that now.

jim-minter:
@pweil- indeed.  Our node image is set via openshift-ansible to rate limit the journal at 10k loglines/second, which I think should be OK for master and infra.

I didn't enable docker logging for the compute nodes on the grounds that I didn't want to hit this issue, or for customer logs to be contending for journald, or for us to be shipping customer logs.

The downside of not having docker logging for compute nodes is that we miss out on logging of DaemonSet pods.  We can revisit this if it becomes necessary.

jim-minter:
(Note we do get system journal and kubelet logging on compute nodes).

jim-minter:
(And we get the lot on infra nodes).

cmd/logbridge/main.go
```diff
@@ -0,0 +1,196 @@
...
+		hm.Write([]byte(sig))
+		req.Header.Add("Authorization", "SharedKey "+customerID+":"+base64.StdEncoding.EncodeToString(hm.Sum(nil)))
+
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return err
+		}
+
+		if resp.StatusCode/100 != 2 {
+			return fmt.Errorf("error: unexpected status code %d, expected 2xx", resp.StatusCode)
```
kargakis: Status contains more info, maybe better to print that instead of just the code/

cmd/logbridge/main.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		if time.Now().Sub(lastSync) > 10*time.Second {
```
kargakis: Worth making this configurable via a flag?

pkg/api/config.go
```diff
@@ -67,6 +68,7 @@ type Config struct {
 	RegistryStorageAccount     string
 	RegistryConsoleOAuthSecret string
 	RouterStatsPassword        string
+	LoggingWorkspace           string
```
kargakis: Unclear what this means just by seeing it here, mind adding godoc (which we should do for all fields but the rest can be done in another PR)?

pkg/arm/data/node-startup.sh
```diff
@@ -39,6 +46,48 @@ cat >/etc/origin/cloudprovider/azure.conf <<'EOF'
 {{ .Config.CloudProviderConf | String }}
 EOF
 
+mkdir -p /var/lib/logbridge
+cat >/etc/origin/node/pods/logbridge.yaml <<'EOF'
```
kargakis: Shouldn't you create this only on infra nodes?

cmd/logbridge/main.go
```diff
@@ -0,0 +1,196 @@
...
+			if err != nil {
+				return err
+			}
+
+			e := make(map[string]interface{}, len(entry.Fields))
+			for k, v := range entry.Fields {
+				e[k] = v
+			}
+
+			e["__CURSOR"] = entry.Cursor
```
kargakis: Worth cross-linking to documentation about these keys

jim-minter:
@kargakis changes addressed, thanks

jim-minter:
@openshift/sig-azure anyone who wants to lgtm this, feel free ...

pweil-:
Yes, I don't think this is our GA solution.  We know it has some pieces missing and may have operational issues with both performance and reliability.  Let's make sure we have a card to R&D the desired GA target

Issue #266, opened 2018-08-24
service catalog apiserver restarting


jim-minter:
SC apiserver, SC controllermanager and prometheus restarting regularly on master-built cluster

strace shows apiserver wilfully exits(0)
controller manager exits(1) - looks like because the apiserver connection went away
not sure about prometheus yet

jim-minter:
I think the sync pod keeps editing the deployments, triggering a redeploy

jim-minter:
the secret checksums are not stable across multiple invocations of sync, causing unwanted redeployments.

hack: added this to syncWorkloadsConfig:
```
```

kargakis:
Looking

kargakis:
Seems like an issue with the serving certs getting regenerated on every sync. For some reason, all services are updated on every sync and that seems to trigger regeneration of the serving certs which subsequently triggers redeployments of all services using serving certs (service catalog, prometheus).

kwoodson:
I'm a bit naive on the situation here but according to @kargakis comments it appears that we are not completely idempotent when running the sync.  The sync process can/should be run often but only change when items are updated/modified in the manifests.  This will ensure less churn and we can guarantee our changes only happen when our data changes.

kargakis:
We do not persist hashes for serving certs so there is a different issue that needs to be identified.

kargakis:
We will need to sort keys in configmaps/secrets before we hash them.

PR #268, opened 2018-08-27
Fix restarts on services that use multiple config keys

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/266

@openshift/sig-azure ptal

### DIFF
--- a/pkg/addons/client.go
+++ b/pkg/addons/client.go
@@ -191,22 +191,8 @@ func write(dyn dynamic.ClientPool, grs []*discovery.APIGroupResources, o *unstru
 		}
 		Default(*existing)
 
-		// this part of code should be executed only for updates
-		// TODO: Split flows to separate functions
-		handleSpecialObjects(*existing, *o)
-
-		if reflect.DeepEqual(*existing, *o) {
-			log.Infof("Skip " + KeyFunc(o.GroupVersionKind().GroupKind(), o.GetNamespace(), o.GetName()))
+		if !needsUpdate(existing, o) {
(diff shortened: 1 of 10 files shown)


kargakis:
/hold

Thinking about adding a unit test

kargakis:
Hrm, for a `write` unit test I would either have to bump client-go and remove dynamic.ClientPool in favor of dynamic.Interface so I can use the fake implementation (pretty intrusive refactor) or refactor `write` and split out a function that handles the cleanup of the special objects and the deep equality check. Ideally we want to have the former but for now I will resort to the latter and open a follow-up issue for using dynamic.Interface in the future.

kargakis:
Unit tests are done, @openshift/sig-azure ptal

kargakis:
@openshift/sig-azure ptal

pkg/addons/client.go
```diff
@@ -191,22 +191,9 @@ func write(dyn dynamic.ClientPool, grs []*discovery.APIGroupResources, o *unstru
...
+		if !needsUpdate(existing, o) {
 			return
 		}
-		// this part of code should be executed only for updates
-		// TODO: Split flows to separate functions
-		handleSpecialObjects(*existing, *o)
-
-		log.Infof("Update " + KeyFunc(o.GroupVersionKind().GroupKind(), o.GetNamespace(), o.GetName()))
-
-		// TODO: we should have tests that monitor these diffs:
```
jim-minter: don't think this comment should be deleted (yet)

Issue #275, opened 2018-08-28
glide install + make fails with "systemd/sd-journal.h: No such file or directory"


kargakis:
```
# github.com/openshift/openshift-azure/vendor/github.com/coreos/go-systemd/sdjournal
vendor/github.com/coreos/go-systemd/sdjournal/journal.go:27:33: fatal error: systemd/sd-journal.h: No such file or directory
 // #include <systemd/sd-journal.h>
```
@openshift/sig-azure

kargakis:
Fixed by installing `systemd-devel`.

PR #276, opened 2018-08-28
Document systemd-devel dep

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/275

@openshift/sig-azure


jim-minter:
thanks @kargakis !

PR #279, opened 2018-08-28
hide var checking

mjudeikis:
:)

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -1,5 +1,6 @@
 #!/bin/bash -ex
 
+set +x
 if ! az account show >/dev/null; then
     exit 1
 fi
@@ -33,6 +34,7 @@ if [[ -z "$DNS_RESOURCEGROUP" ]]; then
     echo error: must set DNS_RESOURCEGROUP
     exit 1
 fi
+set -x
 
 if [[ $# -ne 1 ]]; then
     echo usage: $0 resourcegroup


hack/create.sh
```diff
@@ -14,16 +14,6 @@ if [[ -z "$AZURE_TENANT_ID" ]]; then
     exit 1
 fi
 
-if [[ -z "$AZURE_CLIENT_ID" ]]; then
```
kargakis: Why this diff? Can you just add the +x -x block?

mjudeikis: for all vars? I was thinking moving all sensitive data under one block just after this one

kargakis: Not all vars, just these two.

kargakis: Also I think we need to gate subscription and tenant id

kargakis:
@stevekuznetsov @bbguimaraes these error messages don't make much sense - we need guidance on how to better structure our e2es in case this is an issue on our side or better insight into what's wrong.

bbguimaraes:
Artifact collection needs to be enabled to get the logs from the failed pods, see:

https://github.com/openshift/release/tree/aad6ec8e8ac14e2f6a468e9de896bd142d309e74/ci-operator#using-a-template

e.g.

https://github.com/openshift/release/blob/aad6ec8e8ac14e2f6a468e9de896bd142d309e74/ci-operator/jobs/openshift/origin/openshift-origin-presubmits.yaml#L219-L225

PR #282, opened 2018-08-28
Require and hide SP creds in upgrade.sh

kargakis:
Similar to https://github.com/openshift/openshift-azure/pull/279 for upgrade

@openshift/sig-azure


mjudeikis:
/lgtm 
I tempted to either hide all as I did or have 2 blocks as you did. This does not give any information so no point in showing output. But as long as we dont have sensitive data out im good :)

PR #238, opened 2018-08-22
[WIP] Upgrade

jim-minter:
@pweil- @mjudeikis @kargakis building on #231: upgrade.  I believe this PR has a cleaner/simpler approach compared to #138.

PR #283, opened 2018-08-28
Enable scale up/down

kargakis:
Ref https://jira.coreos.com/browse/AZURE-102
Fixes https://github.com/openshift/openshift-azure/issues/285

@openshift/sig-azure

### DIFF
--- a/cmd/createorupdate/update.go
+++ b/cmd/createorupdate/update.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
@@ -11,17 +12,72 @@ import (
 	"github.com/openshift/openshift-azure/pkg/log"
 )
 
-func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin) error {
+func update(ctx context.Context, cs, oldCs *api.OpenShiftManagedCluster, p api.Plugin, azuredeploy []byte) error {
 	config := auth.NewClientCredentialsConfig(ctx.Value(api.ContextKeyClientID).(string), ctx.Value(api.ContextKeyClientSecret).(string), ctx.Value(api.ContextKeyTenantID).(string))
 	authorizer, err := config.Authorizer()
 	if err != nil {
 		return err
 	}
-
 	ssc := compute.NewVirtualMachineScaleSetsClient(cs.Properties.AzProfile.SubscriptionID)
 	ssc.Authorizer = authorizer
 	vmc := compute.NewVirtualMachineScaleSetVMsClient(cs.Properties.AzProfile.SubscriptionID)
 	vmc.Authorizer = authorizer
+
+	// Need to determine whether the current update is a scale operation before
+	// applying the ARM template. For scale up, we need to figure out which are
+	// the new VMs that were created and wait for them to turn ready. For scale
+	// down, we need to drain the correct VMs before applying the ARM template
+	// scales them down.
+	isScaleOp := isScaleUpdate(cs, oldCs)
+	vmsBefore := map[string]struct{}{}
+	if isScaleOp {
+		for _, agent := range cs.Properties.AgentPoolProfiles {
+			vms, err := listVMs(ctx, cs, vmc, agent.Role)
+			if err != nil {
+				return err
+			}
+
+			if len(vms) > agent.Count {
+				for _, vm := range vms[agent.Count:] {
+					if err := drain(ctx, cs, p, vmc, agent.Role, *vm.InstanceID, *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName); err != nil {
+						return err
+					}
+				}
+			} else {
+				for _, vm := range vms {
+					vmsBefore[*vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName] = struct{}{}
+				}
+			}
+		}
+	}
+
+	// Apply the ARM template
+	if err := deploy(ctx, cs, p, azuredeploy); err != nil {
+		return err
+	}
+
+	if isScaleOp {
+		for _, agent := range cs.Properties.AgentPoolProfiles {
+			vms, err := listVMs(ctx, cs, vmc, agent.Role)
+			if err != nil {
+				return err
+			}
+
+			// wait for newly created VMs to reach readiness
+			for _, vm := range vms {
+				hostname := *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName
+				if _, found := vmsBefore[hostname]; !found {
+					log.Infof("waiting for %s to be ready", hostname)
+					err = p.WaitForReady(ctx, cs, agent.Role, hostname)
+					if err != nil {
+						return err
+					}
+				}
+			}
+		}
+
+		return nil
+	}
 
 	err = updateInPlace(ctx, cs, p, ssc, vmc, api.AgentPoolProfileRoleMaster)
 	if err != nil {
@@ -41,6 +97,33 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin) 
 	}
 
 	return nil
+}
+
+// isScaleUpdate returns whether the update is a scale operation.
+func isScaleUpdate(cs, oldCs *api.OpenShiftManagedCluster) bool {
+	newCounts := make(map[string]int)
+	for _, new := range cs.Properties.AgentPoolProfiles {
+		newCounts[new.Name] = new.Count
+	}
+
+	var isScaleUpdate bool
+	for i, old := range oldCs.Properties.AgentPoolProfiles {
+		if newCounts[old.Name] != old.Count {
+			isScaleUpdate = true
+		}
+		oldCs.Properties.AgentPoolProfiles[i].Count = newCounts[old.Name]
+	}
+
+	// No scale operations.
+	if !isScaleUpdate {
+		return false
+	}
+	// Includes other non-scaling related updates
+	if !reflect.DeepEqual(cs, oldCs) {
+		return false
+	}
+
+	return true
 }
 
 func getCount(cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) int {
@@ -126,20 +209,7 @@ func updatePlusOne(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.P
 			}
 		}
 
-		// remove surplus VM
-		log.Infof("draining %s", *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName)
-		err = p.Drain(ctx, cs, role, *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName)
-		if err != nil {
-			return err
-		}
-
-		log.Infof("deleting %s", *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName)
-		delFuture, err := vmc.Delete(ctx, cs.Properties.AzProfile.ResourceGroup, "ss-"+string(role), *vm.InstanceID)
-		if err != nil {
-			return err
-		}
-
-		if err := delFuture.WaitForCompletion(ctx, vmc.Client); err != nil {
+		if err := drain(ctx, cs, p, vmc, role, *vm.InstanceID, *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName); err != nil {
 			return err
 		}
 	}
@@ -223,4 +293,26 @@ func updateInPlace(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.P
 	}
 
 	return nil
+}
+
+func drain(
+	ctx context.Context,
+	cs *api.OpenShiftManagedCluster,
+	p api.Plugin,
+	vmc compute.VirtualMachineScaleSetVMsClient,
+	role api.AgentPoolProfileRole,
+	instanceID, nodeName string,
+) error {
+	log.Infof("draining %s", nodeName)
+	if err := p.Drain(ctx, cs, role, nodeName); err != nil {
+		return err
+	}
+
+	log.Infof("deleting %s", nodeName)
+	future, err := vmc.Delete(ctx, cs.Properties.AzProfile.ResourceGroup, "ss-"+string(role), instanceID)
+	if err != nil {
+		return err
+	}
+
+	return future.WaitForCompletion(ctx, vmc.Client)
 }
--- a/cmd/createorupdate/update_test.go
+++ b/cmd/createorupdate/update_test.go
@@ -1,0 +1,194 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+func TestIsScaleUpdate(t *testing.T) {
+	tests := []struct {
+		name      string
+		cs, oldCs *api.OpenShiftManagedCluster
+		exp       bool
+	}{
+		{
+			name: "no change",
+			cs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			oldCs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			exp: false,
+		},
+		{
+			name: "image change",
+			cs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			oldCs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.11",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			exp: false,
+		},
+		{
+			name: "scale up",
+			cs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "infra",
+							Count: 2,
+							Role:  "infra",
+						},
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			oldCs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "infra",
+							Count: 1,
+							Role:  "infra",
+						},
+						{
+							Name:  "compute",
+							Count: 3,
+							Role:  "compute",
+						},
+					},
+				},
+			},
+			exp: true,
+		},
+		{
+			name: "scale down",
+			cs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OpenShiftVersion: "v3.10",
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:  "infra",
+							Count: 1,
+							Role:  "infra",
+						},
+						{
+							Name:  "compute",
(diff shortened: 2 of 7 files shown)


cmd/createorupdate/update.go
```diff
@@ -43,6 +48,42 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
...
+	for _, agent := range oldCs.Properties.AgentPoolProfiles {
+		if counts[agent.Name] != agent.Count {
+			bm["scale"] = true
+		}
+	}
+
+	return bm
+}
+
+func waitForReady(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin, vmc compute.VirtualMachineScaleSetVMsClient) error {
```
kargakis: Should probably cross-check with scale set count

cmd/createorupdate/update.go
```diff
@@ -43,6 +52,42 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
 	return nil
 }
 
+// needUpdate returns a map with the exact operations that need to be
+// executed during an upgrade. Only scale is supported for now.
+func needUpdate(cs, oldCs *api.OpenShiftManagedCluster) map[string]bool {
```
kargakis: Eventually we will need to pull in here azuredeploy.json (both old and new) in order to figure out whether the CSE has changed.

pkg/validate/validate.go
```diff
@@ -68,8 +65,21 @@ func validateContainerService(c *api.OpenShiftManagedCluster, externalOnly bool)
 }
 
 func validateUpdateContainerService(cs, oldCs *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
-	if !reflect.DeepEqual(cs, oldCs) {
```
kargakis: @jim-minter not sure why these changes were reverted in https://github.com/openshift/openshift-azure/pull/238 - https://github.com/openshift/openshift-azure/pull/181 was a more reasonable approach to validate updates vs invalidating everything.

kargakis:
We need to identify requests for scale down before we apply the ARM template, in order to run drain.

cmd/manifest/manifest.go
```diff
@@ -0,0 +1,71 @@
+package main
```
jim-minter: This is grim.  Do we really need to solve this problem here?  End users can just edit their manifest yaml.  Scale testing can create its own manifest yaml and modify it using its own separate codebase

cmd/createorupdate/update.go
```diff
@@ -23,6 +23,15 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
 	vmc := compute.NewVirtualMachineScaleSetVMsClient(cs.Properties.AzProfile.SubscriptionID)
 	vmc.Authorizer = authorizer
 
+	bm := needUpdate(cs, oldCs)
+	if len(bm) == 1 && bm["scale"] {
```
jim-minter: I don't think this is right yet because needUpdate is only partially implemented.  It should fire only if nothing but counts (or a single count?) has changed.  Currently it would fire if a count + 0 or more additional things had changed.

cmd/createorupdate/update.go
```diff
@@ -23,6 +23,15 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
...
+		// At this point the scale set has already been updated so we
+		// only need to wait for things (VMs) to settle.
+		// TODO: Figure out whether scale down is going to be problematic
```
jim-minter: I think if we're doing a scale hack, it should include this too.

cmd/createorupdate/update.go
jim-minter: I can't say with any certainty that I think we'll actually do this.  For now I think this PR needs to run deploy() within update() instead of outside, and it needs a hack before that which detects if the only changes are scale-related and actions them.

cmd/createorupdate/update.go
jim-minter: if you have a custom hack which actions scales, it should be able to detect new vms created and just call WaitForReady() on them.  I don't think we shouldn't call it on all of them in case one is broken.
similarly it should have the code to drain vms that will be deleted.  So I think this will need to be refactored a little.

kargakis: >  similarly it should have the code to drain vms that will be deleted. So I think this will need to be refactored a little.

In case of scale down, drain needs to run before we apply the ARM template, as we discussed yesterday, right?

cmd/manifest/manifest.go
kargakis: I don't care where this lives but it's something that can evolve to a e2e library.

pkg/validate/validate.go
kargakis: There are a couple of extra fields that we need to gate apart from the current, `vmSize`, `publicHostname`, and `osType`. I can invalidate changes to those too, error messages are going to be friendlier.

cmd/createorupdate/update.go
kargakis: Right, this needs more work

cmd/createorupdate/update.go
kargakis: After moving `deploy` in `update` this made sense.

pkg/validate/validate.go
kargakis: Opted for your suggestion as it was much less code, better for review and the error message is ok.

cmd/createorupdate/update.go
kargakis: Right, removed in favor of a special cased function for scaling.

cmd/manifest/manifest.go
kargakis: removed for now, didnt add much indeed

kargakis:
@jim-minter updated as per your suggestions and added a unit test, ptal

cmd/createorupdate/update.go
```diff
@@ -11,7 +12,19 @@ import (
 	"github.com/openshift/openshift-azure/pkg/log"
 )
 
-func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin) error {
+func update(ctx context.Context, cs, oldCs *api.OpenShiftManagedCluster, p api.Plugin, azuredeploy []byte) error {
+	scaleOps := isScaleOp(cs, oldCs)
+	for _, scale := range scaleOps {
+		if scale.down {
+			// TODO: If we are scaling down, we need to drain first.
```
jim-minter: We probably want to do this always, regardless of whether it's a scale op or a more complex update.
The kernel of the code is at https://github.com/openshift/openshift-azure/blob/master/cmd/createorupdate/update.go#L131 - just factor it out, loop the roles here and call it.

cmd/createorupdate/update.go
```diff
@@ -11,7 +12,19 @@ import (
...
+		}
+	}
+
```
jim-minter: Here, loop roles and for each, record which VMs currently exist pre-deploy

cmd/createorupdate/update.go
```diff
@@ -23,6 +36,19 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
 	vmc := compute.NewVirtualMachineScaleSetVMsClient(cs.Properties.AzProfile.SubscriptionID)
 	vmc.Authorizer = authorizer
 
+	for role, scale := range scaleOps {
```
jim-minter: Post-deploy, detect new VMs and wait for them to be ready.  The kernel for all of this is in updatePlusOne, it just needs to be factored out.  I think it's much better for us just to await readiness on new VMs.

jim-minter: *Then*, if we were just doing a scale, return early.

pkg/validate/validate.go
```diff
@@ -68,7 +69,29 @@ func validateContainerService(c *api.OpenShiftManagedCluster, externalOnly bool)
...
+		if !ok {
+			continue
+		}
+		if old.Count != new.Count {
+			isScaleUpdate = true
+		}
+		oldCs.Properties.AgentPoolProfiles[i].Count = new.Count
+	}
+
+	if isScaleUpdate && oldCs.Properties.OrchestratorProfile.OrchestratorVersion != cs.Properties.OrchestratorProfile.OrchestratorVersion {
```
jim-minter: This is possibly premature.

pweil-: I'm sure someone will try to do it.  :-) Seems good if we want to limit it from the beginning

kargakis: @jim-minter I thought we wanted to start with as little update surface as possible :)

cmd/createorupdate/update_test.go
```diff
@@ -0,0 +1,182 @@
...
+		name      string
+		cs, oldCs *api.OpenShiftManagedCluster
+		exp       map[api.AgentPoolProfileRole]scaleOp
+	}{
+		{
+			name: "no change",
+			cs: &api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					OrchestratorProfile: &api.OrchestratorProfile{},
+				},
```
pweil-: this doesn't actually test the counts not changing

pkg/validate/validate.go
jim-minter: yes; at the moment we're not even allowing OrchestratorVersion to change.  I'd rather we start allowing it to change at the point where we have some confidence that it works.

cmd/createorupdate/update.go
```diff
@@ -43,6 +70,68 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
...
+		if newCounts[old.Name] > old.Count {
+			scaleOps[old.Role] = scaleOp{up: true}
+			hasScaleOps = true
+			log.Debug("scale up")
+		} else if newCounts[old.Name] < old.Count {
+			scaleOps[old.Role] = scaleOp{down: true}
+			hasScaleOps = true
+			log.Debug("scale down")
+		}
+		oldCs.Properties.AgentPoolProfiles[i].Count = newCounts[old.Name]
```
pweil-: why are we mutating `oldCs` here?

pweil-: oh, for the deepEqual.  This seems bad, anything after this that uses the pointer isn't going to have accurate info

kargakis: Agreed, we are in need of deep copies to do these kind of operations safely. For now, this should be fine since once we figure out a difference in the counts, we don't care about old anymore.

cmd/createorupdate/update.go
kargakis: Depends on whether we want to allow scale+upgrade in a single update.

cmd/createorupdate/update.go
kargakis: Does it matter if we just check readiness for all VMs now? Tidying up to check new VMs only can be a follow-up?

pkg/validate/validate.go
kargakis: It's much uglier though because we need to figure out a difference in counts in two different places: 1. validation (here) and 2. before apply the ARM template. So mutating counts in validation is a no-no. For now, I am copying the counts back after running deep equal. Eventually we should have a real deep copy for OpenShiftManagedCluster.

cmd/createorupdate/update.go
kargakis: Btw how do we know that we are draining the correct VM that will get deleted during the ARM update? Is it always the highest index that gets removed first?

cmd/createorupdate/update_test.go
kargakis: fixed

cmd/createorupdate/update.go
kargakis: done

cmd/createorupdate/update.go
kargakis: Fixed the readiness check. Still outstanding is whether we scale+upgrade in one go.

pkg/validate/validate.go
kargakis: > yes; at the moment we're not even allowing OrchestratorVersion to change. I'd rather we start allowing it to change at the point where we have some confidence that it works.

Seems like we need to discuss whether this is something we want to have for PP (don't we want image updates?). FWIW, when I started testing upgrades a couple of weeks ago, I was able to update from one image to another w/o any problem.

cmd/createorupdate/update.go
kargakis: This is not an issue after all, we just need to drain+delete the right number of VMs.

pkg/validate/validate.go
```diff
@@ -68,9 +69,40 @@ func validateContainerService(c *api.OpenShiftManagedCluster, externalOnly bool)
 }
 
 func validateUpdateContainerService(cs, oldCs *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
-	if !reflect.DeepEqual(cs, oldCs) {
+	newAgents := make(map[string]*api.AgentPoolProfile)
+	for _, newAgent := range cs.Properties.AgentPoolProfiles {
+		newAgents[newAgent.Name] = newAgent
```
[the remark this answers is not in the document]
jim-minter: WIP at #316

cmd/createorupdate/update.go
```diff
@@ -43,6 +106,49 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
 	return nil
 }
 
+type scaleOp struct {
+	up   bool
+	down bool
+	// diff is the count of VMs that are going
+	// to be scaled up or down.
+	diff int
```
jim-minter: pretty sure you can't do this - it's not idempotent

cmd/createorupdate/update.go
```diff
@@ -43,6 +106,49 @@ func update(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.Plugin)
 	return nil
 }
 
+type scaleOp struct {
```
jim-minter: I don't really like this structure at all

cmd/createorupdate/update.go
```diff
@@ -11,18 +12,80 @@ import (
...
 
+	// Need to determine whether the current update is a scale operation before
+	// applying the ARM template. For scale up, we need to figure out which are
+	// the new VMs that were created and wait for them to turn ready. For scale
+	// down, we need to drain the correct VMs before applying the ARM template
+	// scales them down.
+	scaleOps := isScaleOp(cs, oldCs)
+	vmsBefore := map[string]struct{}{}
+	for role, scale := range scaleOps {
+		vms, err := listVMs(ctx, cs, vmc, role)
```
jim-minter: should we not just always do this for all the roles?

pkg/api/types.go
```diff
@@ -119,6 +119,27 @@ type AgentPoolProfile struct {
 	Role         AgentPoolProfileRole `json:"role,omitempty"`
 }
 
+// TODO: Get rid of this in favor of generated deep copies
+func (a *AgentPoolProfile) DeepCopy() *AgentPoolProfile {
```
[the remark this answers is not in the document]
kargakis: agreed

cmd/createorupdate/update.go
kargakis: I can get rid of this but in order to make scale indempotent we always need to check for the health of all VMs, right?

kargakis: Can you walk me through a scenario where this is not idempotent?

cmd/createorupdate/update.go
jim-minter: the key action here is draining VMs that deploy() is about to cause to be deleted.  A potential case could be after a previously infra/compute upgrade failed where there are currently N+1 VMs all serving workload.
I think it'd be highly unexpected for this code to be doing anything with master VMs - you could just iterate infra and compute roles for now.

jim-minter: I think this is the case where storing the diff is invalid - should be storing the number of VMs that is expected post-deploy.

cmd/createorupdate/update.go
jim-minter: I don't think you need to store the delta here in order to be able solely to health check newly created VMs.  You do need to list the VMs pre- and post- deploy though but that code is elsewhere.

cmd/createorupdate/update.go
jim-minter: I don't think it adds anything - if you have it, you need additional unit tests, but if you don't have it, the unit tests aren't really lacking.
Clearly, you need a unit test around the function which detects if a diff between cs and oldCs is scale only.
Otherwise, if you robustly drain all excess VMs before deploy() and health check all new VMs post deploy(), then there is no diff-related code to unit test.  It's more robust and you're just in the desired-state->actual state reconciliation world.

kargakis: >  Clearly, you need a unit test around the function which detects if a diff between cs and oldCs is scale only.

If you have a function that detects you have a scale operation, you still need some extra logic to figure out the direction you are scaling. The point of `isScaleOp` is exactly that, to detect whether the current update is a scale operation and furthermore whether it's up or down. In order to make it unit-testable it has to be its own function. When `isScaleOp` returns, `update` needs to know what scale operation it is about to perform on which scale set, hence `isScaleOp` returns a `map[api.AgentPoolProfileRole]scaleOp` which should always be of size 1 but I don't do any sort of validation about it atm. So I am not sure how to better model this and make as much logic unit-testable as possible, other than use `scaleOp`. Probably holding `count` vs `diff` is more robust but that does not diminish our need for `scaleOp`, does it?

cmd/createorupdate/update.go
kargakis: FWIW, this is only used by scale down where we don't need to list post-deploy (deploy is actually a no-op). I went ahead and updated it to the desired count of the scale set.

pkg/validate/validate.go
```diff
@@ -68,9 +69,40 @@ func validateContainerService(c *api.OpenShiftManagedCluster, externalOnly bool)
...
+	}
+	oldVersion := oldCs.Properties.OrchestratorProfile.OrchestratorVersion
+	oldCs.Properties.OrchestratorProfile.OrchestratorVersion = cs.Properties.OrchestratorProfile.OrchestratorVersion
+
+	if !reflect.DeepEqual(cs.Properties, oldCs.Properties) {
 		errs = append(errs, fmt.Errorf("invalid change %s", deep.Equal(cs, oldCs)))
 	}
+	// Restore whatever we mutated because some of it (eg. scales) are needed
+	// down the road.
+	// TODO: Deep-copy OpenShiftManagedCluster
```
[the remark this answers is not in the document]
kargakis: done

cmd/createorupdate/update.go
kargakis: By "diff", do you mean the field in `scaleOp` which I just removed in favor of `scaleOp.count` or the diff between new and old VMs that we do now in case of a scale up to figure out which are the new VMs we should wait for? In case of the latter, I am not sure we have any use for the number of desired VMs post-deploy even in the case we are coming from a failed upgrade with a N+1 VMs currently running. `deploy` during the current scale up should either be a no-op (because we scaled up to the actual VMs) or we should scale up past the surge VM. In both cases, the diff between new and old VMs should still work as expected (we should wait for the VMs created during this scale up).

kargakis:
I've pushed a commit that removes `scaleOp.diff` in favor of `scaleOp.count` (we need to know how many VMs to remove during a scale down). This should be ready for another review. In a follow up I want to create interfaces for the VM, SSC clients, and our Plugin can use a fake implementation, to make the whole `update` fn unit-testable.

kargakis:
/hold

@jim-minter ptal in the last commit

cmd/createorupdate/update.go
```diff
@@ -11,18 +12,70 @@ import (
...
+	isScaleOp := isScaleUpdate(cs, oldCs)
+	vmsBefore := map[string]struct{}{}
+	if isScaleOp {
+		for _, agent := range cs.Properties.AgentPoolProfiles {
+			vms, err := listVMs(ctx, cs, vmc, agent.Role)
+			if err != nil {
+				return err
+			}
+
+			for _, vm := range vms[agent.Count:] {
```
kargakis: Getting a slice bount out of range with this
```
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
main.update(0x12bd3a0, 0xc42039e540, 0xc420383620, 0xc420452f00, 0x12c7600, 0xc42000e628, 0xc4205fe000, 0x2bad3, 0x54f6c, 0x0, ...)
	/home/kargakis/go/src/github.com/openshift/openshift-azure/cmd/createorupdate/update.go:40 +0x1065
main.createOrUpdate(0x12bd3a0, 0xc42039e540, 0xc4203830e0, 0xc4203225a0, 0xc42037bef0, 0xc420390810, 0x12bd3a0)
	/home/kargakis/go/src/github.com/openshift/openshift-azure/cmd/createorupdate/createorupdate.go:104 +0x481
main.main()
	/home/kargakis/go/src/github.com/openshift/openshift-azure/cmd/createorupdate/createorupdate.go:206 +0x7f5
exit status 2
```

kargakis: Happens when I want to scale up

cmd/createorupdate/update.go
```diff
@@ -2,27 +2,84 @@ package main
...
+	}
+
+	// Apply the ARM template
+	if err := deploy(ctx, cs, p, azuredeploy); err != nil {
+		return err
+	}
+
+	if isScaleOp {
+		for _, agent := range cs.Properties.AgentPoolProfiles {
+			vms, err := listVMs(ctx, cs, vmc, agent.Role)
```
kargakis: Another thing we avoided with `isScaleOp` was listing after deploy when we are scaling down.

kargakis:
/hold

Needs proper squashing

PR #288, opened 2018-08-28
Increasing infra node and infra pod replicas

### LINKED ISSUE #286: scale infra scale set to 2 nodes, scale core infra services to 2
/assign @kwoodson 

@openshift/sig-azure 

"core services" here being the web console, router, and potentially the registry. Thoughts?

kwoodson:
Increase number of infra nodes as well as replicas for infra node pods.

In the future we should query the API (kube or Azure) to determine the number of desired or configured nodes.

Fixes #286


kargakis:
I suggest we add soft pod anti-affinity (`preferredDuringSchedulingIgnoredDuringExecution`) in both deployments to make a best-effort try not to schedule new pods on nodes where another instance of the deployment is already running.

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature

kwoodson:
@kargakis, was this what you had in mind?

kwoodson:
@kargakis,

I added the affinity (anit-affinity) but when I oc delete node infra-00001 the pods came back on the same node:
```
NAME                                READY     STATUS    RESTARTS   AGE       IP            NODE
docker-registry-5fc4ccbf6f-9tbjp    1/1       Running   0          22m       10.128.2.2    infra-000000
docker-registry-5fc4ccbf6f-lcg5w    1/1       Running   0          2m        10.128.2.11   infra-000000
registry-console-69d7f666df-5chzr   1/1       Running   0          2m        10.128.2.8    infra-000000
registry-console-69d7f666df-qgjdm   1/1       Running   0          22m       10.128.2.3    infra-000000
router-ccb768fcb-pj459              1/1       Running   0          2m        10.128.2.9    infra-000000
router-ccb768fcb-td7jl              1/1       Running   0          22m       10.128.2.4    infra-000000
```

Will research the adding the affinity as well.

kargakis:
If you don't have a third node around, that is expected. If we scale infra
to two nodes, then during an upgrade we will have at least two infra nodes
running at all times. Not sure how hard anti-affinity behaves in your case,
it would be nice to see whether the second pod comes up as soon as a second
node turns ready. Maybe we should switch to hard.

pkg/addons/data/Deployment.apps/default/docker-registry.yaml
```diff
@@ -15,6 +15,18 @@ spec:
       labels:
         docker-registry: default
     spec:
+      affinity:
+        podAntiAffinity:
```
jim-minter: @kwoodson or @kargakis before this merges please could you explain to me in english what the below stanza does?

kwoodson: According to the docs, https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature: 
```
The pod anti-affinity rule says that the pod prefers not to be scheduled onto a node if that node is already running a pod with label having key
```

For our infrastructure pods, it will attempt to select another node with the same label.  If it does not find it, it will deploy onto the same node to satisfy the replica count.  If it does find the pod, the it attempts to find another node with the same label to satisfy the anti-affinity and the replica count.

pkg/addons/data/Deployment.apps/default/registry-console.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   name: registry-console
   namespace: default
 spec:
-  replicas: 1
+  replicas: 2
```
kargakis: I don't think we care about the registry console but the web console.

kargakis:
anti-affinity is missing from the web console. Also I am not sure we care about a HA registry console.

kwoodson:
Ok, I'll remove the registry console and add the web console to this pr.

pkg/addons/data/Deployment.apps/default/registry-console.yaml
pweil-: registry console is deprecated in 3.11 and won't be included in 4.0 (FYI)

pkg/addons/data/Deployment.apps/default/router.yaml
```diff
@@ -20,6 +20,20 @@ spec:
...
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: router
+                  operator: In
+                  values:
+                  - router
+              topologyKey: "kubernetes.io/hostname"
+
```
kargakis: Remove this whitespace

kargakis:
@kwoodson one comment, also please squash, then lgtm

kwoodson:
@kargakis, addressed and squashed.

pkg/addons/data/Deployment.apps/default/docker-registry.yaml
jim-minter: @kargakis I thought our scheduler was set up to do this already?  How does adding this change the behaviour?

jim-minter:
@sozercan 

@openshift/sig-azure be aware that committing this change forced MSFT to revendor because the validation that the sync pod does became incompatible with the version of the plugin MSFT were using.  We should discuss this on scrum tomorrow.

kargakis:
Is that the NodePreferAvoidPodsPriority? I think we want to switch to hard
anti-affinity eventually but we should discuss pros and cons first.

kargakis:
> Is that the NodePreferAvoidPodsPriority?

No, this is unrelated. `SelectorSpreadPriority` is probably what you are talking about.

https://docs.okd.io/latest/admin_guide/scheduling/scheduler.html#static-priority-functions

Looking into whether we don't really need soft anti-affinity..

PR #291, opened 2018-08-29
move to BYO AAD app flow

mjudeikis:
This moves AAD integration to use Bring-your-own app model

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,24 +32,19 @@
 
 1. **AAD Application / Service principal**.  The deployed OpenShift cluster
    needs a valid AAD application and service principal to call back into the
-   Azure API, and optionally in order to enable AAD authentication.  There are a
-   few options here:
+   Azure API, and optionally in order to enable AAD authentication. 
+   Application can be created using script `./hack/aad.sh app-create`.
 
-   1. (Ask your Azure subscription administrator to) precreate a generic AAD
-      application and service principal with secret and grant it *Contributor*
-      access to the subscription.  Record the service principal client ID and
-      secret.  Good enough to deploy OpenShift clusters, but AAD authentication
-      won't work.
+   AAD Flow:
+   1. Create an application.
+   2. Add `$AZURE_AAD_CLIENT_ID` variable with application ID to `env` file.
+   3. Create the cluster. `create.sh` script will update your application with 
+   required details. This can be done manually with `./hack/aad.sh app-update`
+   4. Get your application permissions approved by organization administrator.
+   Without approval cluster will start, just login will not work.
 
-   1. Automatically create an AAD application and service principal.  Your Azure
-      user will need *Contributor* and *User Access Administrator* roles, and
-      your AAD will need to have *Users can register applications* enabled.
-
-   1. (Ask your Azure subscription administrator to) precreate a specific AAD
-      application and service principal with secret.  You can use `hack/aad.sh`
-      to help with this process.  For AAD authentication to work, the public
-      hostname of the OpenShift cluster must match the AAD application created.
-      Record the service principal client ID and secret.
+  Once you have application with approved/granted permissions it can be re-used 
+  for all future clusters.  
 
 ### Deploy an OpenShift cluster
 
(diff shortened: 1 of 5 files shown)


hack/create.sh
```diff
@@ -36,16 +36,6 @@ if [[ -z "$DNS_RESOURCEGROUP" ]]; then
 fi
 set -x
 
-if [[ -z "$AZURE_CLIENT_ID" ]]; then
```
kargakis: needs a rebase?

env.example
```diff
@@ -17,8 +17,8 @@ export DEPLOY_VERSION=v3.10
 # If deploying CentOS/origin, uncomment this section, otherwise do nothing
 #export DEPLOY_OS=centos7
 
-# If not auto-creating AAD application and service principal, uncomment and
-# customise this section
+# For AAD to work Web API application has to be pre-created. Uncomment and
```
kargakis:
```sh
# For AAD to work, the Web API application ...
```

hack/aad.sh
```diff
@@ -18,55 +25,75 @@ app-create)
...
-        echo 'error: failed to assign contributor role to SP' >&2
-        exit 1
-    fi
-
     cat <<EOF
+AZURE_AAD_APP_NAME=$2
 AZURE_AAD_CLIENT_ID=$AZURE_AAD_CLIENT_ID
 AZURE_AAD_CLIENT_SECRET=$AZURE_AAD_CLIENT_SECRET
+
+Note: For the application to work Organization Administrator needs to grant permissions first.
```
kargakis: Note: For the application to work, an Organization Administrator ...

hack/aad.sh
```diff
@@ -18,55 +25,75 @@ app-create)
...
-        exit 1
-    fi
-
     cat <<EOF
+AZURE_AAD_APP_NAME=$2
 AZURE_AAD_CLIENT_ID=$AZURE_AAD_CLIENT_ID
 AZURE_AAD_CLIENT_SECRET=$AZURE_AAD_CLIENT_SECRET
+
+Note: For the application to work Organization Administrator needs to grant permissions first. 
+      Once it is approved application can be reused for other clusters using app-update functionality
```
kargakis: Once it is approved, it can be reused

hack/create.sh
```diff
@@ -48,12 +48,16 @@ mkdir -p _data/_out
...
+set +x
+if [[ "$AZURE_AAD_CLIENT_ID" ]]; then
+    . <(hack/aad.sh app-update $AZURE_AAD_CLIENT_ID https://openshift.${RESOURCEGROUP}.${DNS_DOMAIN}/oauth2callback/Azure%20AD)
+else
+    AZURE_AAD_CLIENT_ID=$AZURE_CLIENT_ID
+    AZURE_AAD_CLIENT_SECRET=$AZURE_CLIENT_SECRET
 fi
 
+set -x
+
```
jim-minter: delete the comment below

hack/aad.sh
```diff
@@ -18,55 +25,75 @@ app-create)
...
-        fi
-        sleep 5
-    done
-    if [[ -z "$success" ]]; then
-        echo 'error: failed to assign contributor role to SP' >&2
-        exit 1
-    fi
-
     cat <<EOF
+AZURE_AAD_APP_NAME=$2
```
jim-minter: suggest that here you just output AZURE_AAD_CLIENT_ID to stdout, and if you want to output an end user hint, do that to stderr

README.md
```diff
@@ -32,24 +32,19 @@
 
 1. **AAD Application / Service principal**.  The deployed OpenShift cluster
    needs a valid AAD application and service principal to call back into the
-   Azure API, and optionally in order to enable AAD authentication.  There are a
-   few options here:
+   Azure API, and optionally in order to enable AAD authentication. 
+   Application can be created using script `./hack/aad.sh app-create`.
 
-   1. (Ask your Azure subscription administrator to) precreate a generic AAD
```
jim-minter: I don't think the doc changes are right, and the updated doc is not very clear.
A SP is always needed; if using AAD, our workflow is to then have a second AAD enabled SP.
So this paragraph should stay, then there should be the (optional) AAD piece below?
Please read through the documentation from the perspective of a MSFT new joiner and check it would make sense to them.

PR #295, opened 2018-08-29
Pin vendor k8s.io/api to 73d903622b7391f3312dcbac6483fed484e185f8

sozercan:


jim-minter:
@sozercan I don't understand the rationale for this pin.  At the moment glide is automatically picking kubernetes-1.10.6 (0f11257a8a25954878633ebdc9841c67d8f83bdb) for us in line with our other k8s imports.  You want to us to pin at 73d903622b7391f3312dcbac6483fed484e185f8 (kubernetes-1.10.2).

The functional difference between the two versions is basically nil as far as we're concerned.  In the RP, I don't know why you wouldn't use the version which matches your other kubernetes libraries on principle.  I also don't know why you'd want to force the pinning in this repo?

PR #296, opened 2018-08-29
rename v1 to 2018-09-30-preview

jim-minter:


jim-minter:
@openshift/sig-azure @amanohar @sozercan FYI.  I have tried to do this in as painless a way possible bearingi n mind the retrospective Azure and Golang conventions.

Issue #270, opened 2018-08-27
sync process does not handle updates to config mounted on init containers


kargakis:
/kind bug
/assign

@openshift/sig-azure

kargakis:
Our docker registry mounts the registry config in its init container

kwoodson:
Should we go back to fetching the storage account key inside of the sync.go process? Something similar to what I attempted here: https://github.com/openshift/openshift-azure/pull/170

jim-minter:
Also saves having to worry about using an Azure az image - sounds well worth considering.

kargakis:
This is not an issue after all since the sync process is using volumes to hash content which are shared between containers and initContainers. Also we removed the registry init container in #300.

/close

PR #300, opened 2018-08-30
Remove registry init container

kargakis:
Move storage account key discovery in the sync process.

Ref https://github.com/openshift/openshift-azure/issues/270

We may want to fix applying config in init containers even though we won't have any after this PR merges anyway but separate PR.

@openshift/sig-azure ptal


kwoodson:
I like this approach since it is the approach I took in #170.

/lgtm

kargakis:
https://github.com/openshift/openshift-azure/issues/267

/retest

kargakis:
@stevekuznetsov @bbguimaraes @droslean seems that when I run retest, I get the same failed run in the report.

kargakis:
The e2e pod is mysteriously gone. Only related event in the namespace
```
10m         10m          1         e2e-azure.154fad4928814f27    Pod       spec.containers{teardown}                    Normal    Killing          kubelet, origin-ci-ig-n-cj00   Killing container with id docker://teardown:Need to kill Pod
```
Unclear why it's deleted.

kargakis:
bindata rebase, retagging

stevekuznetsov:
You should see all the inputs reused in the /retest -- but not the test itself. Is that not the case?

mjudeikis:
we have some serious flakes here... need to investigate on ci-operator jobs side :/

kargakis:
> You should see all the inputs reused in the /retest -- but not the test itself. Is that not the case?

Yeah, I saw a couple of times that the failed test is reused.

kargakis:
> Another strange thing, test that failed 2h ago and the namespace is already deleted from the cluster.

Correction: the test shows up as pending on Github, and in the logs it seems that it runs.

kargakis:
The sync process fails with the following log:
```
time="2018-08-31T09:35:18Z" level=info msg="Error while syncing: cannot sync cluster config: azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/XXXXXXXXXXXXXXXXXXXXXXXX/resourceGroups/ci-op-b2bhkln9-6cb1c/providers/Microsoft.Storage/storageAccounts/l8oleo6rdum5zslh0syia6hj/listKeys?api-version=2018-02-01: StatusCode=400 -- Original Error: adal: Refresh request failed. Status Code = '400'. Response body: {\"error\":\"invalid_request\",\"error_description\":\"Identity not found\"}"
```
Definitely caused by my PR. Trying to figure out why.

kargakis:
Ah, looks like the sync pod is lacking the azure credentials in its env.

PR #306, opened 2018-08-30
Removing hardcoded storageclass location.

### LINKED ISSUE #260: remove hard coded location in storage class
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  name: azure
parameters:
  kind: managed
  location: eastus
  skuName: Premium_LRS
provisioner: kubernetes.io/azure-disk
reclaimPolicy: Delete
volumeBindingMode: Immediate

```

kwoodson:
Add a json path update for storageclass location so that it can by dynamic.

Fixes #260


pkg/addons/data/StorageClass.storage.k8s.io/azure.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   name: azure
 parameters:
   kind: managed
-  location: eastus
+  location: "******* Generated *******"
```
pweil-: Nit: I think all of our generated fields use 3 stars and all caps.  Nothing depends on it but might as well keep them the same. 

Other than that, lgtm.  Tag when updated.

kwoodson:
@pweil-, updated.

kargakis:
:tada:

PR #308, opened 2018-08-31
Drop pod anti-affinity in favor of SelectorSpreadPriority

kargakis:
The default scheduler config is already doing this

https://docs.okd.io/latest/admin_guide/scheduling/scheduler.html#static-priority-functions

@openshift/sig-azure


jim-minter:
much preferred, thank-you.

PR #315, opened 2018-08-31
implement DeepCopy for OpenShiftManagedCluster

jim-minter:
@openshift/sig-azure ptal
@kargakis fyi - suggestion

### DIFF
--- a/pkg/api/deepcopy.go
+++ b/pkg/api/deepcopy.go
@@ -1,0 +1,19 @@
+package api
+
+import (
+	"encoding/json"
+)
+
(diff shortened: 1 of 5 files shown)


kargakis:
Mutate the copy in TestDeepCopy and expect reflect.DeepEqual to be false
after the first check.

kargakis:
https://github.com/openshift/openshift-azure/issues/317

/retest

kargakis:
ci-operator issue returns the same failure as the last one

/retest

kargakis:
> Mutate the copy in TestDeepCopy and expect reflect.DeepEqual to be false after the first check.

https://github.com/openshift/openshift-azure/pull/318

PR #316, opened 2018-08-31
WIP: make AgentPoolProfiles a map in the internal API

jim-minter:
not decided if we want to do this or not ...


kargakis:
Not a big fan of having maps in the API due to their lack of order.

FWIW, this is also a [k8s api convention](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) linking to issues with prior discussion on why not to use maps in the API.

jim-minter:
Looks like helper functions is the preferred way.  Closing.

PR #318, opened 2018-09-03
TestDeepCopy test after mutation

kargakis:
Follow-up to https://github.com/openshift/openshift-azure/pull/315

@openshift/sig-azure


jim-minter:
thanks!

Issue #223, opened 2018-08-17
use timestamps in logs instead of elapsed time


kargakis:
/kind cleanup

kargakis:
This is still an issue for `sync`

/reopen

jim-minter:
@thanasisk just needs a `logger.Formatter = &logrus.TextFormatter{FullTimestamp: true}` somewhere near the top of sync.go

PR #321, opened 2018-09-03
enabling timestamps in sync

thanasisk:
verified locally that:
- it gives the expected output

> ./sync 
INFO[2018-09-03T15:10:19+02:00] Sync process started                       

- make test - tests pass

Fixes https://github.com/openshift/openshift-azure/issues/223

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -24,6 +24,10 @@ func New(entry *logrus.Entry) {
 	})
 }
 
+func init() {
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -24,6 +24,10 @@ func New(entry *logrus.Entry) {
 	})
 }
 
+func init() {
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
```
kargakis: This should be set inside the main function of sync.

PR #322, opened 2018-09-03
simplify nil checks

jim-minter:


mjudeikis:
I like it. Less `if's`

Issue #319, opened 2018-09-03
Generate htpasswd when AAD is enabled


kargakis:
Or at the very least, generate the password instead of hardcoding to `demo`.

@openshift/sig-azure thoughts?

mjudeikis:
I think we agreed we need to generate one for now with intention we gonna remove it in the close future

jim-minter:
+1 on generate

mjudeikis:
@jim-minter if we generate it randomly. Its get stored in htpasswd format. Question is how we will know it? :D 
```
[mjudeiki@redhat openshift-azure]$ ./hack/config.sh get-config mjudeikis xxxxxxxxxxxx .config.HtPasswd
ZGVtbzokMmEkMTAkNEJxxxxxxxxxxxxxxxxxxxxxxxxxxSXV3NlBZSUpsQzlTUmp2NWNxbFc=
```

PR #323, opened 2018-09-03
Add generated htpasswd

mjudeikis:
Adds temporary password for PP UI access.

cc: @openshift/sig-azure please tell me better way to store password temporary for our use? :) 

Fixes https://github.com/openshift/openshift-azure/issues/319

### DIFF
--- a/hack/config.sh
+++ b/hack/config.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+
+usage() {
+    cat <<EOF >&2
+usage:
+$0 get-config resourcegroup
+EOF
+    exit 1
+}
+
+case "$1" in
+get-config)
+    if [[ "$#" -ne 2 ]]; then usage; fi
+
+    OSACONFIG=$(mktemp -d)
+    trap "rm -rf $OSACONFIG" EXIT
+
+    export AZURE_STORAGE_ACCOUNT=$(az storage account list -g $2 | jq '.[] | select(.tags.type == "config") | .name ' | tr -d '"')
+    export AZURE_STORAGE_KEY=$(az storage account keys list -n $AZURE_STORAGE_ACCOUNT -g $2 --query '[0].value' | tr -d '"')
+    az storage blob download --file $OSACONFIG/config -c config -n config --no-progress >/dev/null
+    cat $OSACONFIG/config
+    ;;
+
+*)
+    usage
+    ;;
+
+esac
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -53,6 +53,8 @@ type Config struct {
 	SessionSecretAuth []byte
 	SessionSecretEnc  []byte
 	HtPasswd          []byte
+	//TODO: Remove me before GA!
+	AdminPasswd       string
 	ImageConfigFormat string
 
 	// misc node configurables
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/satori/go.uuid"
+	"golang.org/x/crypto/bcrypt"
 	"k8s.io/client-go/tools/clientcmd/api/v1"
 
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
@@ -327,9 +328,16 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
 		}
 	}
 
-	if len(c.HtPasswd) == 0 {
-		if c.HtPasswd, err = makeHtPasswd("demo", "demo"); err != nil {
-			return
+	// TODO: Remove these password operations before GA
+	if len(c.AdminPasswd) == 0 {
+		if c.AdminPasswd, err = randomString(10); err != nil {
+			return err
+		}
+	}
+	if len(c.HtPasswd) == 0 || bcrypt.CompareHashAndPassword(getHashFromHtPasswd(c.HtPasswd), []byte(c.AdminPasswd)) != nil {
+		c.HtPasswd, err = makeHtPasswd("osadmin", c.AdminPasswd)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/pkg/config/generate_test.go
+++ b/pkg/config/generate_test.go
@@ -72,6 +72,7 @@ func testRequiredFields(omc *api.OpenShiftManagedCluster, t *testing.T) {
 
 	assert(c.ServiceAccountKey != nil, "service account key")
 	assert(len(c.HtPasswd) != 0, "htpassword")
+	assert(len(c.AdminPasswd) != 0, "admin password")
 	assert(c.SSHKey != nil, "ssh key")
 
 	assert(len(c.RegistryStorageAccount) != 0, "registry storage account")
--- a/pkg/config/util.go
+++ b/pkg/config/util.go
@@ -76,6 +76,10 @@ func makeHtPasswd(username, password string) ([]byte, error) {
 	return append([]byte(username+":"), b...), nil
 }
 
+func getHashFromHtPasswd(record []byte) []byte {
+	return []byte(strings.Split(string(record), ":")[1])
+}
+
 func randomBytes(n int) ([]byte, error) {
 	b := make([]byte, n)
 	if _, err := io.ReadAtLeast(rand.Reader, b, n); err != nil {


mjudeikis:
Just for a records. This is to start discussion. I cried when I wrote those 5 lines of code!

pkg/api/config.go
```diff
@@ -53,6 +53,8 @@ type Config struct {
 	SessionSecretAuth []byte
 	SessionSecretEnc  []byte
 	HtPasswd          []byte
+	//TODO: Remove me before GA!
+	HtPasswdPlain     string
```
jim-minter: OSAdminPassword?

pkg/config/generate.go
```diff
@@ -328,8 +328,13 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
...
 
 	if len(c.HtPasswd) == 0 {
-		if c.HtPasswd, err = makeHtPasswd("demo", "demo"); err != nil {
-			return
+		pass, err := randomString(10)
+		c.HtPasswdPlain = pass
+		if err != nil {
+			return err
+		}
+		if c.HtPasswd, err = makeHtPasswd("osadmin", pass); err != nil {
```
jim-minter: Open follow-up to make this a derived config

hack/config.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -e
+
+usage() {
+    cat <<EOF >&2
+usage:
+
+$0 get-config resourcegroup storageaccountname jsonpath
```
jim-minter: simplify! config.sh can work out the storageaccount name for itself (check the tags).  Have it output the config to stdout so you can do `config.sh get-config RG | jq .config.HtPasswd`.

pkg/config/generate.go
jim-minter: Cancel that - the hash will change each time - it can't be derived

mjudeikis: We just need to get it each time. Kinda password rotation? Prevents people from storing passwords for clusters?

hack/config.sh
mjudeikis: Was doing this as I saw your comments :)

hack/config.sh
```diff
@@ -0,0 +1,39 @@
...
+$0 get-config resourcegroup jsonpath
+
+example:
+    config.sh get-config resourcegroupname .config.HtPasswdPassword
+
+EOF
+    exit 1
+}
+
+if ! rpm -qa | grep -qw jq; then
```
jim-minter: remove the jq functionality, output to stdout

pkg/api/config.go
```diff
@@ -53,6 +53,8 @@ type Config struct {
 	SessionSecretAuth []byte
 	SessionSecretEnc  []byte
 	HtPasswd          []byte
+	//TODO: Remove me before GA!
+	HtPasswdPassword  string
```
jim-minter: Still don't like the name

pkg/config/generate.go
```diff
@@ -328,8 +328,13 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
 	}
 
 	if len(c.HtPasswd) == 0 {
-		if c.HtPasswd, err = makeHtPasswd("demo", "demo"); err != nil {
-			return
+		pass, err := randomString(10)
```
jim-minter: if len HtPasswdPassword == 0, generate the name
then if the given password doesn't correspond with the .htpasswd file (check bcrypt.CompareHashAndPassword), regenerate htpasswd file.

hack/config.sh
mjudeikis: We have multiple storage accounts, and selecting right one without jq get ambiguous. We can do as you suggest and output to stdout all storage accounts, but we need to chose one name so we can get key and download file. 
Without jq it becomes bash parsing operator to parse json/yaml based on jsonPath.

Alternative?

pkg/api/config.go
mjudeikis: :'(

pkg/config/generate.go
mjudeikis: @jim-minter please elaborate on this. Not very clear what you had in mind

hack/config.sh
jim-minter: see https://gist.github.com/jim-minter/a108288a581c053cdf2e62f6998b6339

pkg/api/config.go
jim-minter: please think of another one.

pkg/config/generate.go
jim-minter: something like
```
	if len(c.HtPasswdPassword) == 0 {
		if c.HtPasswdPassword, err = randomString(10); err != nil {
			return err
		}
	}

	if bcrypt.CompareHashAndPassword(c.HtPasswd, []byte(c.HtPasswdPassword)) != nil {
		if c.HtPasswd, err = makeHtPasswd("osadmin", c.HtPasswdPassword); err != nil {
			return err
		}
	}
```

hack/config.sh
mjudeikis: we still use jq here, but fine :)

hack/config.sh
```diff
@@ -0,0 +1,26 @@
+usage() {
```
pweil-: hashbang

mjudeikis: this is not the issue you are looking for....

pkg/config/generate.go
```diff
@@ -327,11 +328,19 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
...
+	if len(c.AdminPasswd) == 0 {
+		if c.AdminPasswd, err = randomString(10); err != nil {
+			return err
 		}
 	}
+	tHtPasswd, err := makeHtPasswd("osadmin", c.AdminPasswd)
+	if err != nil {
+		return err
+	}
+	if bcrypt.CompareHashAndPassword(c.HtPasswd, tHtPasswd) != nil {
```
jim-minter: can you confirm you checked this - I expect CompareHashAndPassword to need a hash and a password as arguments, I don't see how it would work with two hashes.

mjudeikis: sorry, just now got this AHA moment what you had in mind.

pkg/config/generate.go
```diff
@@ -327,9 +328,21 @@ func Generate(cs *acsapi.OpenShiftManagedCluster) (err error) {
...
+	if len(c.HtPasswd) > 0 && bcrypt.CompareHashAndPassword(getHashFromHtPasswd(c.HtPasswd), []byte(c.AdminPasswd)) != nil {
+		c.HtPasswd, err = makeHtPasswd("osadmin", c.AdminPasswd)
+		if err != nil {
+			return err
+		}
+	} else {
+		c.HtPasswd, err = makeHtPasswd("osadmin", c.AdminPasswd)
+		if err != nil {
+			return err
 		}
```
mjudeikis: Thinking if case `(c.AdminPasswd) >0` and `len(c.HtPasswd) == 0` is very likly?

jim-minter:
/lgtm

thanks

mjudeikis:
``` 
  Normal   Scheduled  8m               default-scheduler             Successfully assigned ci-op-7r3n99y1/release-latest to origin-ci-ig-n-fp3q
  Normal   Pulling    8m               kubelet, origin-ci-ig-n-fp3q  pulling image "busybox"
  Normal   Started    8m               kubelet, origin-ci-ig-n-fp3q  Started container
  Normal   Pulled     8m               kubelet, origin-ci-ig-n-fp3q  Successfully pulled image "busybox"
  Normal   Created    8m               kubelet, origin-ci-ig-n-fp3q  Created container
  Warning  Failed     7m (x3 over 8m)  kubelet, origin-ci-ig-n-fp3q  Failed to pull image "docker-registry.default.svc:5000/ci-op-7r3n99y1/stable:cli": rpc error: code = Unknown desc = Error: image ci-op-7r3n99y1/stable:cli not found
  Warning  Failed     6m (x5 over 8m)  kubelet, origin-ci-ig-n-fp3q  Error: ImagePullBackOff
  Normal   BackOff    6m (x5 over 8m)  kubelet, origin-ci-ig-n-fp3q  Back-off pulling image "docker-registry.default.svc:5000/ci-op-7r3n99y1/stable:cli"
  Normal   Pulling    6m (x4 over 8m)  kubelet, origin-ci-ig-n-fp3q  pulling image "docker-registry.default.svc:5000/ci-op-7r3n99y1/stable:cli"
  Warning  Failed     5m (x4 over 8m)  kubelet, origin-ci-ig-n-fp3q  Error: ErrImagePull
  Warning  Failed     5m               kubelet, origin-ci-ig-n-fp3q  Failed to pull image "docker-registry.default.svc:5000/ci-op-7r3n99y1/stable:cli": rpc error: code = Canceled desc = context canceled
```` 
ci-operator/infrastructure flake. never seen this container/error before. @stevekuznetsov whats this release cli stuff?

PR #324, opened 2018-09-03
only set public IPs on masters

jim-minter:


mjudeikis:
lgtm . Will for other to take a look

Issue #336, opened 2018-09-04
allow basic defaulting of some external api items


jim-minter:
At the moment, validation requires a routerprofile to be provided as input on cluster creation:
```
"routerProfiles": [
  {
    "name": "default"
  }
]
```

If the user is not providing a publicSubdomain it is a bit pointless to have to provide such a routerProfile - we could default that value appropriately.

jim-minter:
@openshift/sig-azure

jim-minter:
@sozercan thanks for raising!

jim-minter:
we may want to delay this post private preview so that we look at adding appropriate defaulting logic across the board

jim-minter:
another item that could be considered for defaulting is osType == Linux in agentpoolprofiles

jim-minter:
another one is name == "infra" for the infra agentpoolprofile

ehashman:
This PR should allow us to remove this defaulting code from the fakeRP:
https://github.com/openshift/openshift-azure/blob/5ee2f80b351e3f8c6b1e7549b7ed3de26c1dd70d/pkg/fakerp/fakerp.go#L275-L281

PR #338, opened 2018-09-05
remove router name

mjudeikis:
removed need for router profile default.

So this looks to me very "late in the process".
Alternatives:
1. We do this as a part of conversion. But conversion now will have to be smarter. As it gets empty `[]RouterProfiled` from external API. So it needs to create/init internal API object and populate it. 
2. MergeCofing part could add those too. But again, I dont like idea adding default to merge code.
3. `Validate` is out of scope as its `read-only & validate`.
4. `GenerateConfig` looks like the best place to do this. We already adding/defaulting things there. But this means we skip most of the validation, we rely on MSFT Enrich process to init `RouterProfiles` object in Internal representation (as it will be empty) and add anything else we need (FQDN). And we default it in generate.go same as we do with `publicHostname`.

But doing this so late in the plugin code we lose a lot of validation. Any ideas are welcome.. 

cc: @jim-minter @kargakis @openshift/sig-azure 

https://github.com/openshift/openshift-azure/issues/336

### DIFF
--- a/pkg/api/converterfromv20180930preview.go
+++ b/pkg/api/converterfromv20180930preview.go
@@ -32,6 +32,12 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster) *
 			FQDN:              oc.Properties.FQDN,
 		}
 
+		// if router profile is not set on the external api, default to minimum
+		if len(oc.Properties.RouterProfiles) == 0 {
+			oc.Properties.RouterProfiles = make([]v20180930preview.RouterProfile, 1)
+			oc.Properties.RouterProfiles[0].Name = "default"
+		}
+
 		cs.Properties.RouterProfiles = make([]RouterProfile, len(oc.Properties.RouterProfiles))
 		for i, rp := range oc.Properties.RouterProfiles {
 			cs.Properties.RouterProfiles[i] = RouterProfile{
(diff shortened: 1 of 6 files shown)


pkg/api/converterfromv20180930preview.go
```diff
@@ -32,6 +32,12 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster) *
 			FQDN:              oc.Properties.FQDN,
 		}
 
+		// if router profile is not set on the external api, default to minimum
+		if len(oc.Properties.RouterProfiles) == 0 {
+			oc.Properties.RouterProfiles = make([]v20180930preview.RouterProfile, 1)
+			oc.Properties.RouterProfiles[0].Name = "default"
```
mjudeikis: I think we should not enforce name "default" to router profiles. 
Now if I provide just bare minimum:
```
  routerProfiles:
  - publicSubdomain: $RESOURCEGROUP.$DNS_DOMAIN
```
 will get validation error:
```
[invalid properties.routerProfiles[""], invalid properties.routerProfiles[""].name "", invalid properties.routerProfiles["default"]]
```
Which implies now cusomter will need to know this "default" name to be default. Less customers need to worry, better. 

We could either:
1. Default first router to name `default` always
2. Remove validation on name == `default`.

mjudeikis: What was the idea behind `default` as a mandatory name?

jim-minter:
I think we should close this and re-visit post private preview.

mjudeikis:
archive to post PP

Issue #327, opened 2018-09-04
Need to tie RP vendoring to sync pod version


jim-minter:
This is critical to avoid mismatches in testing, and then in production
Longer term there will be a sync pod image per supported OSA release (e.g. at least sync-3.10, sync-3.11; maybe even sync-3.10.0, sync-3.10.1, ...). In the short term for PP we'll just start with one.

Proposal for now is:
1) on each commit, CI should build and push the sync pod tagged with the short commit ID.  Also tag :latest or :v310 or whatever.
2) plugin should accept a configurable indicating the short commit ID that should be used to pull. MSFT will set this up in the RP, based on the commit ID that they vendor at.
3) our fake RP/scaffolding can just default the configurable to :latest so that our CI works as expected

jim-minter:
@openshift/sig-azure @amanohar FYI
@kargakis if I don't get to this today please could you look at this urgently on Wednesday?

kargakis:
Pushing commits in their own tags in quay is going to be challenging. Today, we have two processes that comprise our sync image publishing process[1]: 

1. the [ci-operator postsubmit](https://github.com/openshift/release/blob/ba151f8f4a10aa0831d69a69eaa3e3c251f1e2f7/ci-operator/jobs/openshift/openshift-azure/openshift-openshift-azure-postsubmits.yaml#L16) which runs on every openshift-azure merged commit pushes the built image in an [imagestreamtag inside the CI cluster](https://github.com/openshift/release/blob/ba151f8f4a10aa0831d69a69eaa3e3c251f1e2f7/ci-operator/config/openshift/openshift-azure/master.json#L29).
2. a cronjob mirrors the imagestreamtag to [a set of static tags in quay](https://github.com/openshift/release/blob/ba151f8f4a10aa0831d69a69eaa3e3c251f1e2f7/projects/azure/image-mirror/image-mirror.yaml#L12) (atm `latest` and `v3.10`) on an hourly basis.

We could potentially mount quay credentials in the ci-operator job and try to shoehorn publishing to quay there but not sure how hacky it's going to be (cc @stevekuznetsov). It would be much easier to pin a certain commit that is vendored by the RP in a stable tag and publish that tag as an one-off but I believe we need an automated process for doing that whenever we want to. Thoughts?

[1] this mirrors how we publish images to quay and dockerhub elsewhere

kargakis:
[...]

PR #339, opened 2018-09-05
Configure sync image pullspec in plugin init 

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/327

@openshift/sig-azure this PR moves `SYNC_IMAGE` out of the plugin code and uses the plugin struct to hold the sync image.

### DIFF
--- a/pkg/config/images.go
+++ b/pkg/config/images.go
@@ -61,7 +61,6 @@ func selectContainerImagesOrigin(cs *acsapi.OpenShiftManagedCluster) {
 
 		c.AzureCLIImage = "docker.io/microsoft/azure-cli:latest"
 
-		c.SyncImage = os.Getenv("SYNC_IMAGE")
 		if c.SyncImage == "" {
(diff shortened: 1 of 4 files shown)


jim-minter:
As we talked about on the phone this would be best as a configurable admitted when the Plugin object is created.

kargakis:
@openshift/sig-azure updated, ptal

pkg/config/images.go
```diff
@@ -102,7 +102,6 @@ func selectContainerImagesOSA(cs *acsapi.OpenShiftManagedCluster) {
 
 		c.AzureCLIImage = "docker.io/microsoft/azure-cli:latest" //TODO: create mapping for OSA release to any other image we use
 
-		c.SyncImage = os.Getenv("SYNC_IMAGE")
```
pweil-: same for `selectContainerImagesOrigin`

kargakis: oops, missed that. fixed

PR #340, opened 2018-09-05
remove spp from external api

mjudeikis:
remove SSP from external api and add it to enrich


jim-minter:
you need to actually remove it from the external API definition as well ;-)

PR #351, opened 2018-09-05
 add tenantId field to AADIdentityProvider

### LINKED ISSUE #346: need to add TenantID to AADIdentityProvider
The tenant ID for the AAD integration is different to the cloud provider tenant ID.
While the tenant ID for the AAD integration is deducible in the RP, we might as well insist that users provided it up front from the beginning, that way if they want to use a different tenant for AAD integration, they can.

jim-minter:
fixes #346 

@sozercan


mjudeikis:
I assume this for callback url to be SP binding, not identity SP binding. lgtm for me

Issue #334, opened 2018-09-04
plugin API should have single function call for update()


jim-minter:
agreed approach with MSFT that the RP will call a single plugin update function which will handle everything from draining excess pods, calling deploy then all the update workflow.  Needs to be defined and refactored

jim-minter:
@openshift/sig-azure

PR #360, opened 2018-09-06
Add Update plugin method

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/334

### DIFF
--- a/pkg/upgrade/update.go
+++ b/pkg/upgrade/update.go
@@ -1,4 +1,4 @@
-package main
+package upgrade
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


kargakis:
@openshift/sig-azure

pkg/upgrade/update.go
```diff
@@ -295,16 +295,9 @@ func updateInPlace(ctx context.Context, cs *api.OpenShiftManagedCluster, p api.P
...
 
-func drain(
-	ctx context.Context,
-	cs *api.OpenShiftManagedCluster,
-	p api.Plugin,
-	vmc compute.VirtualMachineScaleSetVMsClient,
-	role api.AgentPoolProfileRole,
-	instanceID, nodeName string,
-) error {
+func (u *simpleUpgrader) delete(ctx context.Context, cs *api.OpenShiftManagedCluster, vmc compute.VirtualMachineScaleSetVMsClient, role api.AgentPoolProfileRole, instanceID, nodeName string) error {
```
jim-minter: not sure the s/drain/delete/ makes it clearer

kargakis: delete is closer than drain to what this is actually doing :)

PR #365, opened 2018-09-06
WiP - conversion of ovs/sdn daemonsets to static pods

thanasisk:


pkg/arm/data/master-startup.sh
```diff
@@ -571,6 +571,286 @@ spec:
...
+  - name: host-opt-cni-bin
+    hostPath:
+      path: /opt/cni/bin
+  - name: host-etc-cni-netd
+    hostPath:
+      path: /etc/cni/net.d
+  - name: host-var-lib-cni-networks-openshift-sdn
+    hostPath:
+      path: /var/lib/cni/networks/openshift-sdn
+EOF
```
mjudeikis: /n

pkg/arm/data/node-startup.sh
```diff
@@ -46,6 +46,286 @@ cat >/etc/origin/cloudprovider/azure.conf <<'EOF'
...
+      hostPath:
+        path: /run/openvswitch
+    - name: host-sys
+      hostPath:
+        path: /sys
+    - name: host-config-openvswitch
+      hostPath:
+        path: /etc/origin/openvswitch
+EOF
+cat >/etc/origin/node/pods/sdn.yaml <<'EOF'
```
mjudeikis: nit: /n

pkg/arm/data/master-startup.sh
```diff
@@ -571,6 +571,286 @@ spec:
...
+          # Restrict the number of pthreads ovs-vswitchd creates to reduce the
+          # amount of RSS it uses on hosts with many cores
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1571379
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1572797
+          if [[ `nproc` -gt 12 ]]; then
+              ovs-vsctl set Open_vSwitch . other_config:n-revalidator-threads=4
+              ovs-vsctl set Open_vSwitch . other_config:n-handler-threads=10
+          fi
+          while true; do sleep 5; done
+        image: '*** GENERATED ***'
```
jim-minter: you need to fix up all of the `*** GENERATED ***` parts and remove the corresponding templates from the sync pod.

mjudeikis:
Good start! You need to do translation of the `GENERATED` values now. Here we use simple golang Templates. 
It starts here: https://github.com/openshift/openshift-azure/blob/master/pkg/arm/arm.go#L49

And we inject values via `{{ .Config.ControlPlaneImage | quote }}`

You need to replace those with: `{{ .Config.NodeImage  | quote }}`

And remove https://github.com/openshift/openshift-azure/blob/master/pkg/addons/translate.go#L214-L225

jim-minter:
Unfortunately there are significant additional complexities here.

I don't know if static pods must be created in the kube-system namespace or not, but it needs to be checked.  Everything associated with the openshift-sdn namespace may need to be removed.

The SDN pods use the sdn service account which is bound to the sdn-reader cluster role.  We would need to generate a client cert and kubeconfig for the sdn pods to enable them to authenticate as a service account without a token, and somewhere pre-setup the service account, rolebinding and maybe SCC.

jim-minter:
Closing for https://github.com/openshift/openshift-azure/pull/450

PR #375, opened 2018-09-07
etcd backup operator

mjudeikis:
Etcd backup operator. Behavior:
backups each hour to same SA where config sits, just different blob `etcd` - file `backup`

minor nit needs fixing:
The first sync fails because, despite the fact we land CRD early, they are not getting registered on time.
So sync fails with error:
`Error while syncing: cannot sync cluster config: couldn't find group etcd.database.coreos.com"`

We need similar logic as we have for apiserver to wait for `CRD` resources to register to API before we create `CRDObjects`. 

But the second run completes so fast not even master nodes get ready. Not sure how big is the issue here.


pkg/addons/addons.go
```diff
@@ -200,6 +208,9 @@ func writeDB(client Interface, db map[string]unstructured.Unstructured) error {
 	if err := client.ApplyResources(cfgFilter, db, keys); err != nil {
 		return err
 	}
+
+	//TODO: validate if CRD and SC is available here.
```
mjudeikis: I think we should replace `client.ServiceCatalogExists` below with something more generic to check SC and CRD objects (or even second method). and check here, and only then they are available to move forward. We will need this as in example `Istio` will have around 28 CRDs... We will see more of this coming.

pkg/addons/data/EtcdBackup.etcd.database.coreos.com/openshift-etcd/etcd-backup.yaml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: openshift-etcd
+spec:
+  etcdEndpoints:
+  - "https://master-000000:2379"
+  - "https://master-000001:2379"
+  - "https://master-000002:2379"
+  storageType: ABS
+  abs:
+    absSecret: etcd-backup-abs-credentials
+    path: 'etcd/backup'
```
mjudeikis: `etcd` is a blob  `backup` is a file name here

mjudeikis:
Nees to validate node selectors. Now it run on masters

pkg/addons/addons.go
mjudeikis: Separate object better. APi server takes ages to aggregate as it needs infra nodes available. CRD on other hand is almost instant.

pkg/addons/data/Role.rbac.authorization.k8s.io/openshift-etcd/etcd-operator.yaml
```diff
@@ -0,0 +1,35 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - services
+  - endpoints
+  - events
+  - secrets
+  verbs:
+  - "*"
```
jim-minter: This is pretty wide.  Open follow-up issue to tune down later if possible.

jim-minter:
All lgtm - what else needs to be done before merging this?

jim-minter:
@mjudeikis I have force pushed to this branch logic which checks the etcd CRD is ready (haven't e2e tested as about to get on a flight).  Please don't overwrite my 2 commits ;-)

pkg/addons/data/Role.rbac.authorization.k8s.io/openshift-etcd/etcd-operator.yaml
jim-minter: #378

jim-minter:
@pweil- ptal

mjudeikis:
running final check with @jim-minter changes now.

pkg/addons/data/Role.rbac.authorization.k8s.io/openshift-etcd/etcd-operator.yaml
mjudeikis: It was original for od based etcd. I think we should be good only with `secrets` and maybe `events`. This should be straight forward task for anybody to pick

mjudeikis:
I still see error on first sync:
```
time="2018-09-07T20:04:21Z" level=info msg="Create Deployment.apps/openshift-web-console/webconsole"
time="2018-09-07T20:04:21Z" level=info msg="Error while syncing: cannot sync cluster config: couldn't find group etcd.database.coreos.com"
```

this passed before:
```
time="2018-09-07T20:04:11Z" level=debug msg="Waiting for the Etcd CRD to be ready"                                                                                                                                                                                                        
time="2018-09-07T20:04:12Z" level=debug msg="Etcd CRD is ready"                                                                                                                                                                                                                           
time="2018-09-07T20:04:12Z" level=info msg="Create APIService.apiregistration.k8s.io/v1beta1.servicecatalog.k8s.io"                                                                                                                                                                       
time="2018-09-07T20:04:12Z" level=info msg="Update ClusterRole.rbac.authorization.k8s.io/admin"             
```

pkg/addons/addons.go
```diff
@@ -200,6 +208,18 @@ func writeDB(client Interface, db map[string]unstructured.Unstructured) error {
...
 	}
+
+	log.Debug("Waiting for the Etcd CRD to be ready")
+	if err := wait.PollInfinite(time.Second, client.EtcdCRDReady); err != nil {
+		return err
+	}
+	log.Debug("Etcd CRD is ready")
+
+	// refresh dynamic client
+	if err := client.UpdateDynamicClient(); err != nil {
```
mjudeikis: If we wrap `EtcdCRDReady` and `ServiceCatalogExists` in subfunctions (or rewrite them) we can move update client logic into the clients itself. Methods should know if they need to update client or not. follow-up issue?

pweil-: follow up.  Taggiong

mjudeikis: https://github.com/openshift/openshift-azure/issues/380

jim-minter:
/retest

Excellent work @mjudeikis. Thanks and have a great weekend!

mjudeikis:
/test e2e
U shall Pass!

kargakis:
```
time="2018-09-08T20:16:54Z" level=fatal msg="Get https://ci-op-jgqgzg9w-5d5fb.eastus.cloudapp.azure.com/healthz: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"openshift-signer\")"
```

kargakis:
> time="2018-09-08T20:16:54Z" level=fatal msg="Get https://ci-op-jgqgzg9w-5d5fb.eastus.cloudapp.azure.com/healthz: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"openshift-signer\")"

This is probably an issue because of the ci-operator namespace reuse resulting in reusing an older cluster that was not cleaned up in time?

jim-minter:
@kargakis "ci-operator namespace reuse" ?

PR #389, opened 2018-09-10
don't use oldCs in update

jim-minter:


pkg/upgrade/update.go
```diff
@@ -56,74 +56,50 @@ func (u *simpleUpgrader) Update(ctx context.Context, cs, oldCs *api.OpenShiftMan
...
 		}
-
-		return nil
 	}
 
-	err = u.updateInPlace(ctx, cs, ssc, vmc, api.AgentPoolProfileRoleMaster)
-	if err != nil {
-		return err
-	}
+	// For PP day 1, scale is permitted but not any other sort of update.  When
```
kargakis: Why not make this configurable, at least to start soaking the upgrade plugin in tests?

jim-minter:
@pweil- @kargakis @mjudeikis ptal.

* at present the RP does not persist the "last cs that caused the cluster to successfully create or update" and MSFT is strongly against doing so based on experience.  The RP front end persists the new cs after validation as the goal state, then the back end reconciles this.
* so, if a PUT is tried, fails part way, then a second PUT is tried, inside the plugin oldCs is seen differently both times.  This means oldCs cannot really be used for anything but careful validation in the front end.  For example, validating that immutable fields are not changed, or that a user doesn't try to upgrade from a newer version to an older one.
* therefore, with good reason, oldCs is not available in the RP back end, only the front end.  Hence this PR to remove our use of it in the back end.

pkg/upgrade/update.go
jim-minter: Sure, happy to do that.

jim-minter:
tested, believed ready for merge

pkg/upgrade/update.go
```diff
@@ -13,7 +13,10 @@ import (
 	"github.com/openshift/openshift-azure/pkg/log"
 )
 
-func (u *simpleUpgrader) Update(ctx context.Context, cs, oldCs *api.OpenShiftManagedCluster, azuredeploy []byte) error {
+func (u *simpleUpgrader) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte) error {
+	// TODO: probably need to break this function into two pieces, pre-Deploy
```
jim-minter: #390

pkg/upgrade/update.go
```diff
@@ -57,74 +58,48 @@ func (u *simpleUpgrader) Update(ctx context.Context, cs, oldCs *api.OpenShiftMan
...
+	// the same time, current thinking is that we will add a hash-based
+	// mechanism to avoid unnecessary VM rotations as well.
 
-// isScaleUpdate returns whether the update is a scale operation.
-func isScaleUpdate(cs, oldCs *api.OpenShiftManagedCluster) bool {
-	newCounts := make(map[string]int)
-	for _, new := range cs.Properties.AgentPoolProfiles {
-		newCounts[new.Name] = new.Count
-	}
+	if os.Getenv("DEVELOPMENT_ALWAYS_ROTATE") != "" {
```
kargakis: Instead of adding one level of indentation, you can exit early.
```go
if os.Getenv("DEVELOPMENT_ALWAYS_ROTATE") == "" {
   return nil
}
// update code
```

kargakis:
Opened https://github.com/openshift/release/issues/1438 for the test failure.

pkg/upgrade/update.go
jim-minter: It's possible, but this is also a temporary hack and this highlights it.

PR #393, opened 2018-09-11
add check for password reset

mjudeikis:
add check to prevent pass reset for team account


mjudeikis:
ptal @kargakis

env.example
```diff
@@ -19,8 +19,8 @@ export DEPLOY_VERSION=v3.10
 
 # For AAD to work, the Web API application has to be pre-created. Uncomment and
 # customise this section with your Web API AAD application details
+# Do not this value same as AZURE_CLIENT_ID.
```
kargakis: Do not set this value to `AZURE_CLIENT_ID`. Also explain why not

hack/create.sh
```diff
@@ -48,14 +48,13 @@ mkdir -p _data/_out
 
 az group create -n $RESOURCEGROUP -l eastus --tags now=$(date +%s) >/dev/null
 
-set +x
```
kargakis: needs to be reverted back?

env.example
```diff
@@ -19,8 +19,9 @@ export DEPLOY_VERSION=v3.10
 
 # For AAD to work, the Web API application has to be pre-created. Uncomment and
 # customise this section with your Web API AAD application details
+# Do not set this value same as AZURE_CLIENT_ID. AZURE_AAD_CLIENT_ID application
+# instance secret is being reseted each time cluster is created
```
kargakis: s/reseted/reset/

hack/create.sh
```diff
@@ -48,8 +48,9 @@ mkdir -p _data/_out
 
 az group create -n $RESOURCEGROUP -l eastus --tags now=$(date +%s) >/dev/null
 
+# if AZURE_CLIENT_ID is used as AZURE_AAD_CLIENT_ID, script will reset global team account!
 set +x
-if [[ "$AZURE_AAD_CLIENT_ID" ]]; then
+if [[ "$AZURE_AAD_CLIENT_ID" ]] && [ "$AZURE_AAD_CLIENT_ID" != "$AZURE_CLIENT_ID" ]; then
```
jim-minter: `[[ "$AZURE_AAD_CLIENT_ID" && "$AZURE_AAD_CLIENT_ID" != "$AZURE_CLIENT_ID" ]]`

hack/ssh.sh
```diff
@@ -14,11 +14,11 @@ ID="$2"
 shift 2
 
 IP=$(az vmss list-instance-public-ips -g $RESOURCEGROUP -n ss-master --query "[$ID].ipAddress" | tr -d '"')
-
+echo $IP
```
jim-minter: why does this file change, out of interest?

mjudeikis: Because silent ssh bugs me so much as when it breaks and I don't know why. I have to go in, modify the file, add some debug messages, run again to check why it fails.  And later commit it by mistake with all other changes... :) Simple :D

jim-minter: it only breaks for you ;-)

Issue #396, opened 2018-09-11
e2e test failure due to missing deps


kargakis:
```
go test ./test/e2e -tags e2e
# github.com/openshift/openshift-azure/test/e2e
package github.com/openshift/openshift-azure/test/e2e (test)
	imports github.com/onsi/ginkgo
	imports github.com/onsi/ginkgo/internal/remote
	imports github.com/hpcloud/tail: cannot find package "github.com/hpcloud/tail" in any of:
	/go/src/github.com/openshift/openshift-azure/vendor/github.com/hpcloud/tail (vendor tree)
	/usr/local/go/src/github.com/hpcloud/tail (from $GOROOT)
	/go/src/github.com/hpcloud/tail (from $GOPATH)
FAIL	github.com/openshift/openshift-azure/test/e2e [setup failed]
make: *** [e2e] Error 1
```
/assign
/kind bug

PR #397, opened 2018-09-11
Add missing deps

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/396

@openshift/sig-azure


mjudeikis:
`hpcloud` O_o was it not being vendored with `ginko` deps?

mjudeikis:
ou, they dont vendor at all... :/ 
/lgtm

kargakis:
Even if they don't vendor, we are pruning nested vendored directories anyway, right?

mjudeikis:
am I missing something, where is glide.yaml update?

PR #401, opened 2018-09-11
added json tags to Config struct

### LINKED ISSUE #386: add `json` tags to Config struct members

charlesakalugwu:
part of #386

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -9,117 +9,117 @@ import (
 )
 
 type Config struct {
-	ImageOffer     string
-	ImagePublisher string
-	ImageSKU       string
-	ImageVersion   string
+	ImageOffer     string `json:"imageOffer,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/api/config.go
```diff
@@ -9,117 +9,117 @@ import (
...
 	// for development
-	ImageResourceGroup string
-	ImageResourceName  string
+	ImageResourceGroup string `json:"imageResourceGroup,omitempty"`
+	ImageResourceName  string `json:"imageResourceName,omitempty"`
 
 	Certificates CertificateConfig
 
 	// container images for pods
-	MasterEtcdImage             string
```
jim-minter: do you need a json tag on `Certificates CertificateConfig` above?

charlesakalugwu: done :)

jim-minter:
tests passed, squash merging.  Leaving #386 open to implement unit test.  Great 1st commit @charlesakalugwu!

Issue #407, opened 2018-09-11
disable log analytics on CI clusters


jim-minter:
@kargakis please can you look at this first thing in the morning?
suggest setting a generic env var for CI runs and simply disabling creation of the LogAnalytics resource when it's set

jim-minter:
Possibly for neatness the logbridge static pod as well.

Clearly, we'll need to have some test runs where it's enabled, but not all of them.

PR #409, opened 2018-09-12
Avoid running loganalytics on CI tests

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/407

@openshift/sig-azure


mjudeikis:
lgtm, the second pair of eyes for visibility

pkg/arm/arm.go
```diff
@@ -43,16 +44,23 @@ func (*simpleGenerator) Generate(ctx context.Context, cs *acsapi.OpenShiftManage
...
+				return util.Template(string(masterStartup), startupMap, cs, map[string]interface{}{"Role": role})
 			} else {
-				return util.Template(string(nodeStartup), nil, cs, map[string]interface{}{"Role": role})
+				return util.Template(string(nodeStartup), startupMap, cs, map[string]interface{}{"Role": role})
 			}
 		},
 		"IsUpgrade": func() bool {
 			return isUpdate
 		},
+		"DevelopmentMode": devModeFunc,
```
jim-minter: put this straight in util.Template, then you don't need the startupMap?

pkg/arm/arm.go
```diff
@@ -43,16 +44,23 @@ func (*simpleGenerator) Generate(ctx context.Context, cs *acsapi.OpenShiftManage
 	if err != nil {
 		return nil, err
 	}
+
+	devModeFunc := func() bool {
+		return os.Getenv("DEVELOPMENT_MODE") != ""
```
jim-minter: I'm on the fence about whether it makes sense to use the same env var for:
1- disabling LogAnalytics in at least most tests (permanent move; rationale: it's expensive)
2- enabling rotation for some dev/test? (temporary; rationale: getting more soak testing?)
I guess it's OK.

Clearly we want to minimise the number of these special cases over time.
Also would an env var like RUNNING_UNDER_TEST or something be a clearer name than DEVELOPMENT_MODE?

pkg/arm/data/master-startup.sh
```diff
@@ -595,6 +595,7 @@ spec:
 EOF
 fi
 
+{{ if not DevelopmentMode }}
```
jim-minter: `{{-` for all of these `if`s and `end`s

kargakis: I've also seen this in azuredeploy.json. What's the difference?

pweil-: eliminates whitespace

pkg/arm/arm.go
mjudeikis: `DEVELOPMENT_MODE` is more clear for me than `RUNNING_UNDER_TEST` :) so question if perception :)

jim-minter: Yes, but it's not an environment variable that developers should really be using.

pkg/util/util.go
```diff
@@ -29,6 +29,9 @@ func Template(tmpl string, f template.FuncMap, cs *acsapi.OpenShiftManagedCluste
 			replacer := strings.NewReplacer("$", "\\$")
 			return replacer.Replace(b)
 		},
+		"RunningUnderTest": func() bool {
+			return os.GetEnv("RUNNING_UNDER_TEST") != nil
```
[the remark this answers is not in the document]
kargakis: fixed, thanks

jim-minter:
/lgtm

@kargakis please could you do a follow up which sets the env var in our testing?

kargakis:
> @kargakis please could you do a follow up which sets the env var in our testing?

That's https://github.com/openshift/release/pull/1451

jim-minter:
thanks! had missed it!

Issue #402, opened 2018-09-11
cloudprovider identity changes causes masters to stop


mjudeikis:
to reproduce:
1. Create a cluster using if `AZURE_CLIENT_ID` `AZURE_CLIENT_SECRET` aka. `cs.Properties.ServicePrincipalProfile`.
2. Change password on application instance behind `AZURE_CLIENT_ID`.
3. Master kubelet should stop running all pods on the masters.

mjudeikis:
we noticed this only if ran service for a long time. I had a cluster running with the expired secret for 16h and its is still running. Might be either very long TTL (which would make sense) orr we have heisenbug here and it is not what we are looking for.
@jim-minter reopen if you think its relevant.

PR #410, opened 2018-09-12
make sp secret rotatable

mjudeikis:
@jim-minter @kargakis wdyt? This is fallout for https://github.com/openshift/openshift-azure/issues/402 me trying to update clusters without recreation


jim-minter:
This will absolutely happen, but not before PP day 1.  Close and we'll do this later.

mjudeikis:
:'(

PR #413, opened 2018-09-12
drop TenantID, SubscriptionID, ResourceGroup from Config struct - dup…

jim-minter:
…licates


kargakis:
https://github.com/openshift/release/issues/1469

/retest

Issue #157, opened 2018-08-03
Should kubeconfigs be part of the internal model


pweil-:
See https://github.com/openshift/openshift-azure/pull/153#discussion_r207622067 for discussion

cc @amanohar

amanohar:
This is mostly in the context of resource provider (RP)/service kubeconfig(s) to be used by operations team for access to OSA cluster and not the kubeconfigs used by components of OSA cluster.

The operations team kubeconfig can be part of `Config` and also be saved separately (implementation detail). It would be good to avoid duplication of data but not a blocker.

pweil-:
@sozercan @amanohar @shrutir25 - looking for guidance here and if there will be a plugin signature change.  Otherwise we can close.

PR #414, opened 2018-09-12
 Remove configurables for derived CIs:

jim-minter:
- ImageResourceGroup
- ImageResourceName
- ImageConfigFormat
- MasterLBCNamePrefix
- CloudProviderConf


jim-minter:
First commit is on its own in #413.

I'm leaving config fields in on the following basis:
- if they are uniquely derived during generate (e.g. certs, random secrets), OR
- if we want to latch their values to handle changing them on a timescale we choose (e.g. all the image pullspecs)

On that basis, we should remove the kubeconfigs.  This is a temptation because they are fully derivable and they bloat the config object a lot.  My concern is that it makes it a little harder to generate an admin kubeconfig from the config blob.  Q1: @kargakis @mjudeikis @pweil- wdyt what's your take on this?

Q2: should we do a `Images ImageConfig` to match `Certificates CertificateConfig` for PP?  Thoughts?

jim-minter:
hmm, if we got rid of the kubeconfigs we'd be able to avoid pulling in all the kubernetes api in our api package.  More temptation.

jim-minter:
cf #157

mjudeikis:
I like the idea. Im up for `Images ImageConfig`. 
/test e2e

jim-minter:
follow-up in #440

jim-minter:
@kargakis points out there's a third category of config field which is required not to be derived: anything dependent on ENV vars which runs in the sync pod.

PR #415, opened 2018-09-12
tie deployed images to specific version, and use published images

jim-minter:


jim-minter:
/hold

until we're confident MSFT have got access to the published images

jim-minter:
/hold

again - where we configure the base image is a long way away from where we configure the container images - need to tie these together better somehow

jim-minter:
/hold cancel

#452

jim-minter:
@kargakis ptal

Issue #408, opened 2018-09-12
Tune health check logic


mjudeikis:
@amanohar:
Total cluster deployment time (6 node cluster: 3 master, 1 agent, 2 infra) is: ~12-13 mins.

Template deployment took: ~4 mins

Approx. 6+ mins was spent on health check out of which:
`prometheus-node-exporter`: took ~1+ min
`prometheus`: took ~4+ mins

There might be scope for optimization via:
1) Doing health check of components in parallel where possible.
2) Possibly only check health of core components to declare success and do alerting on additional important components if they fail to come up like Prometheus but not gate create on it?
Also entirely possible that since `prometheus` was the long pole and `apiserver`, `controller-manager`, and other core components were checked after `prometheus` they were already in ready state so removing `prometheus` will not shave of 4 mins completely. 

cc: @openshift/sig-azure

pweil-:
The cli is expected to default to 4 agent nodes. Does that significantly
increase the time?

kargakis:
I am waiting more than 10m for a PV to get created for Prometheus on an install. We may want to revert using persistent storage for metrics in PP (some of the pain with not persisting metrics is going to be alleviated by the SRE Prometheus federation.

jim-minter:
I believe the principal issues affecting startup time are (in order):
1) Azure PV attach time
2) not running SDN as static pods
3) possibly I/O due to image sizes, possibly especially those associated with deploying Prometheus

@mjudeikis the fact that health checks run in serial does not affect cluster startup time.  The underlying deployments are in parallel, it's just that prometheus is the slowest because of the PV.

If customers are seeing slow PVs, we should see it and worry about it too.  I'm reluctant to stop using a PV for Prometheus.  We should be tracking and reporting cluster deploy time in tests.

jim-minter:
> The cli is expected to default to 4 agent nodes. Does that significantly increase the time?

@pweil- no.

jim-minter:
@mjudeikis I think this issue needs to be repurposed to concrete things that can be done.  Measure and report cluster startup / PV attach times in test would be a good example.

mjudeikis:
@jim-minter all these issues where raised by @amanohar in slack.

Wondering, is it worth trying to sping something like 100 clusters in random regions and check how many we fail and succ and problems with those? this should be scriptable?

kargakis:
[...]

PR #416, opened 2018-09-12
remove prometheus from health check, but add an e2e check

jim-minter:
I'm not in love with this, but it may be pragmatic.  @pweil- @kargakis @mjudeikis wdyt?


jim-minter:
cf #408

jim-minter:
there you go @amanohar :-)

Issue #399, opened 2018-09-11
add helper make targets for cluster create and delete


charlesakalugwu:
To create the cluster, we need to 

- create a copy of `env.example` as `env` (`cp env.example env`)
- fill in appropriate values into `env`
- source the `env` file (source ./env)
- call the create.sh script (`./hack/create.sh somename`)

Sometimes I might switch to a new terminal, forget to source the file `env` file and try to call the `delete.sh` script (`./hack/delete.sh somename`).

In addition I need to supply the same `somename` argument to the create.sh and delete.sh scripts. I might have forgotten the argument I used to call the create.sh script or might type it in wrong, possibly leading to either an error or deleting a wrong cluster (if user has sufficient permissions).

I propose:
- introducing a `RESOURCE_GROUP_NAME` variable into env.example
- developers would need to specify a value for this in their `env` file
- create a make `create-cluster` target which
  - sources the `env` file
  - calls `./hack/create.sh ${RESOURCE_GROUP_NAME}`
- create a make `delete-cluster` target which
  - sources the `env` file
  - calls `./hack/delete.sh ${RESOURCE_GROUP_NAME}`

This way, as long as i've created the ./env file I can switch between terminals, use `make create-cluster` and `make delete-cluster` with more deterministic results.

This change will not affect the current way of starting and deleting the cluster. So all existing CI processes, for example, depending on current functionality will continue to work.

jim-minter:
My personal take:

Agreed the situation isn't brilliant as it stands.

I'm reluctant about the `make` change because I don't like the idea of lots of layers to helper scripts - origin did this and it gets horrible quickly.  We're also trying to make the helper scripts etc. smaller as time goes on.

I recognise `delete.sh $RESOURCEGROUP` is irritating, but `ssh.sh $RESOURCEGROUP` will probably be useful to us more generally as a starting point if we need to ssh into random clusters.  Currently at least there's symmetry there.

It's problematic currently to run multiple clusters in parallel due to AAD applications needing an admin to approve them.  But again, continuing to specify $RESOURCEGROUP might make it easier for us to branch out that way in the future if we need to.

I do think that having the scripts automatically source env if it hasn't been done already and printing an error if env doesn't exist would be a nice touch - could you at least submit a PR for that?

kargakis:
Agreed with @jim-minter

charlesakalugwu:
@jim-minter @kargakis  I completely understand your stance. I'll submit a PR for sourcing `env` directly

PR #417, opened 2018-09-13
allow ssh into arbitrary clusters, stop requiring resourcegroup on delete

### LINKED ISSUE #411: have ssh.sh use config.sh to retrieve the cluster private key if it isn't available locally
this gives us an easy way ssh into arbitrary clusters with access to the cluster config in the storage account

jim-minter:
@openshift/sig-azure ptal

with this change
* `hack/ssh.sh` takes you straight in to your master-000000. `hack/ssh.sh othercluster` downloads the config for othercluster and takes you in to master-000000 on that cluster. `-n {0,1,2}` allows you to ssh onto a different master
* `hack/delete.sh` no longer requires an argument

shout out to @charlesakalugwu for pointing out the shortcomings of our hack/ scripts and to @mjudeikis for hack/config.sh.  These are your ideas.  @charlesakalugwu this doesn't replace you doing the `env` pieces of issue #399 although hopefully it helps.

fixes #411

### DIFF
--- a/hack/config.sh
+++ b/hack/config.sh
@@ -3,7 +3,9 @@
 usage() {
     cat <<EOF >&2
 usage:
+
 $0 get-config resourcegroup
+
 EOF
     exit 1
 }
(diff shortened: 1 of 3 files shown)


charlesakalugwu:
nice :+1:

PR #418, opened 2018-09-13
source env file in hack scripts

### LINKED ISSUE #399: add helper make targets for cluster create and delete

charlesakalugwu:
fixes #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,8 @@
 ### Deploy an OpenShift cluster
 
 1. Copy the `env.example` file to `env` and edit according to your requirements.
-   Source the `env` file: `. ./env`.
+   Source the `env` file: `. ./env`. Make sure the env file is placed in the root
+   directory of the project.
 
 1. Run `./hack/create.sh $RESOURCEGROUP` to deploy a cluster.
 
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -5,6 +5,17 @@ if ! az account show >/dev/null; then
     exit 1
 fi
 
+ENVIRONMENT_CONFIG=$(dirname $(dirname $0))/env
+
+# check if the environment config file exists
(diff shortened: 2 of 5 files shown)


charlesakalugwu:
```+++ dirname ./hack/delete.sh
++ dirname ./hack
+ ENVIRONMENT_CONFIG=./env
+ [[ ! -f ./env ]]
+ echo error: must setup an env config file in project root
+ exit 1
error: must setup an env config file in project root`
```
looking into the ci setup

jim-minter:
@charlesakalugwu please fix this or close it :-)

PR #419, opened 2018-09-13
osa test blacklist

mjudeikis:
depending on https://github.com/openshift/release/pull/1468 to go in first

### DIFF
--- a/test/overrides
+++ b/test/overrides
@@ -1,0 +1,3 @@
+# placeholder file to overwrite ci-operator variables for tests
+# example: 
+# export TEST_FOCUS="my-awesome-test-I-want-to-run"
--- a/test/overrides
+++ b/test/overrides
@@ -1,3 +1,7 @@
-# placeholder file to overwrite ci-operator variables for tests
+# Placeholder file to overwrite ci-operator variables for tests.
+# This file should be used ONLY for development. 
+# No permanent overrides should be added here. 
+# Permanent overrides should be in openshift/release repository
+#
 # example: 
 # export TEST_FOCUS="my-awesome-test-I-want-to-run"


mjudeikis:
Context, we need this so we can control test variables from our code repo. So if I working on certain test suite, and want to check why its failing ( due our code errors) I want to add either `TEST_FOCUS`, or 
`TEST_SKIP`. Origin controls this in release repo, and I think its very inflexible as we need this to be configurable for individual PR's

test/overrides
```diff
@@ -0,0 +1,7 @@
+# Placeholder file to overwrite ci-operator variables for tests.
+# This file should be used ONLY for development. 
+# No permanent overrides should be added here. 
+# Permanent overrides should be in openshift/release repository
+#
+# example: 
+# export TEST_FOCUS="my-awesome-test-I-want-to-run"
```
kargakis: This is not a blacklist but a whitelist. I would expect us to use both, default TEST_FOCUS to conformance, and mainly use TEST_SKIP for blacklist. If you wanted to test a single test you should still be able to do it, but I'd vote for extending our verification test to fail in case you touch `TEST_FOCUS`.

mjudeikis: I would say fail verification if this file changes in content. As its development file, as soon as you come up with changes you need/want to do - they should be reflected in release repository. This is just to make life easier when coming up with the list/changes.

mjudeikis:
ping @kargakis @charlesakalugwu

test/overrides
kargakis: I am fine with it, you need to add a validation script and plug it in `make verify`.

mjudeikis: https://github.com/openshift/openshift-azure/issues/435 will do this after, its just how I plan my work :) dont want to implement this before I have it running and working.

PR #427, opened 2018-09-13
add/enable kube-state metrics

mjudeikis:
This adds kube-state metrics deployment to be scrapped and accessed


mjudeikis:
ping @jim-minter  this will give as insights in what users are using. Number of projects, storages, pods, etc etc

mjudeikis:
@jim-minter agree on most of these. Kube state metrics is part of 3.11 stack, I didn't just it pulled it from the hat :) 
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_cluster_monitoring_operator/templates/cluster-monitoring-operator-config.j2#L60

mjudeikis:
@jim-minter conformance will fail. still WIP.

PR #428, opened 2018-09-13
Add PDB e2e test

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/309

### DIFF
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -5,6 +5,9 @@ package e2e
 import (
 	"time"
 
+	project "github.com/openshift/api/project/v1"
+	projectclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -14,6 +17,7 @@ var c *testClient
 
 type testClient struct {
 	kc        *kubernetes.Clientset
+	pc        *projectclient.ProjectV1Client
 	namespace string
 }
 
@@ -40,21 +44,38 @@ func newTestClient(kubeconfig string) *testClient {
 		panic(err)
 	}
 
+	pc, err := projectclient.NewForConfig(config)
+	if err != nil {
+		panic(err)
+	}
+
 	return &testClient{
 		kc: kc,
+		pc: pc,
 	}
 }
 
-func (t *testClient) createNamespace(namespace string) {
-	// TODO: Create a project request
+func (t *testClient) createNamespace(namespace string) error {
+	if _, err := t.pc.ProjectRequests().Create(&project.ProjectRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		}}); err != nil {
+		return err
+	}
 	t.namespace = namespace
 	// TODO: Wait for a successful SAR check
+	return nil
 }
 
-func (t *testClient) cleanupNamespace(timeout time.Duration) {
+func (t *testClient) cleanupNamespace(timeout time.Duration) error {
 	if t.namespace == "" {
-		return
+		return nil
 	}
 
-	// TODO: Do a project delete and wait for the namespace to cleanup
+	if err := t.pc.Projects().Delete(t.namespace, &metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+
+	// TODO: Wait for the namespace to cleanup
+	return nil
 }
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -3,11 +3,13 @@
 package e2e
 
 import (
+	"fmt"
 	"time"
 
 	project "github.com/openshift/api/project/v1"
 	projectclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -44,6 +46,7 @@ func newTestClient(kubeconfig string) *testClient {
 		panic(err)
 	}
 
+	// create a project client for creating and tearing down namespaces
 	pc, err := projectclient.NewForConfig(config)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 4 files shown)


test/e2e/osa_test.go
```diff
@@ -83,4 +129,23 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+			},
+			Spec: policy.PodDisruptionBudgetSpec{
+				MaxUnavailable: &maxUnavailable,
+				Selector:       selector,
+			},
+		}
+
+		_, err = kubeClient(c.cs.Config.AdminKubeconfig).Policy().PodDisruptionBudgets("e2e-tests").Create(pdb)
```
mjudeikis: Is this `system:admin` for now? Not yet delegated?

kargakis: Yes, there is a TODO in kubeClient to fix that

test/e2e/osa_test.go
```diff
@@ -30,10 +34,52 @@ var _ = BeforeSuite(func() {
...
+	// to test with different permissions.
+	c.kc, err = managedcluster.ClientsetFromConfig(c.cs.Config.AdminKubeconfig)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Create a namespace for running tests and assign osadmin
+	// as a project admin to it. Delegate permissions to osadmin
+	// to allow us to run tests in a non-privileged context.
+	_, err = c.kc.CoreV1().Namespaces().Create(&v1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "e2e-tests",
```
mjudeikis: are we using one namespace for all tests? if we introduce parallelism to the tests will it not clash?

mjudeikis: I have seen this now in origin conformance testing. If in example few tests use same namespace and one of them is testing "network policy" or rbac restriction it automatically can flake another test in the same namespace, despite the fact it uses different resources.

kargakis: > can flake another test in the same namespace, despite the fact it uses different resources.

Can you elaborate? Is it namespace cleanup that triggers the flake or something else?

mjudeikis: that could be too. I think each test case should create unique namespace for itself and clean after.

kargakis: I am against namespace cleanup. It's probably one of the flakiest things I have ever seen in origin e2e.

kargakis:
@mjudeikis @jim-minter @pweil- ptal, this is now ready to get merged

jim-minter:
I feel like a few things are problematic here:
* I think we should start down the line of being able to run selected tests straight with an end user kubeconfig - if you solve the associated problems, others can build tests in your footsteps.  Perhaps we should have two separate test suites - one expecting system:admin, one not?
* I want to avoid building stuff assuming the existence of `osadmin` or what its associated [cluster]rolebindings might be
* I agree with MJ about separate namespaces and at least trying to clean them up

mjudeikis:
> Perhaps we should have two separate test suites - one expecting system:admin, one not?

I think we need 3:
* Project admin
* System:Admin
* (future) azure-cluster-admin. And considering gingko uses tags to whitelist test, should we add relevant tagging to tests too? based on scope, type, etc

kargakis:
>  I think we should start down the line of being able to run selected tests straight with an end user kubeconfig - if you solve the associated problems, others can build tests in your footsteps. Perhaps we should have two separate test suites - one expecting system:admin, one not?

Do we need to precreate the end-user kubeconfig as part of cluster creation then? How do we bind it to users as they are getting access to the cluster?

> I want to avoid building stuff assuming the existence of osadmin or what its associated [cluster]rolebindings might be

I don't think the name of the user matters and should be easily swapped for whatever we want to come up with. The important part here is the binding to project admin.

> I agree with MJ about separate namespaces and at least trying to clean them up

Namespace cleanup, as I mentioned already, used to be the biggest source of flakes in Origin conformance tests. If we want to test namespace cleanup, that should be a scoped test. There is no other reason to do namespace cleanup if we are going to delete the cluster once the e2e tests are done. Separate namespace per test is fine.

kargakis:
> There is no other reason to do namespace cleanup if we are going to delete the cluster once the e2e tests are done. Separate namespace per test is fine.

I guess we want to do namespace cleanup if we plan to run these tests in production clusters :(

jim-minter:
> I don't think the name of the user matters and should be easily swapped for whatever we want to come up with. The important part here is the binding to project admin.

I think I got a bit confused because grafana in azure-misc was making osadmin a cluster reader.  It doesn't do that any more.  As long as we don't pull those kinds of tricks, using osadmin works for now.

kargakis:
Opened https://github.com/openshift/openshift-azure/pull/528 for splitting out the e2e framework refactoring.

kargakis:
@mjudeikis @jim-minter this should be ready now (I hope)

kargakis:
The tests wont pass until https://github.com/openshift/openshift-azure/pull/545 gets merged.

test/e2e/client_test.go
```diff
@@ -40,21 +44,38 @@ func newTestClient(kubeconfig string) *testClient {
...
+		}}); err != nil {
+		return err
+	}
 	t.namespace = namespace
 	// TODO: Wait for a successful SAR check
+	return nil
 }
 
-func (t *testClient) cleanupNamespace(timeout time.Duration) {
+func (t *testClient) cleanupNamespace(timeout time.Duration) error {
```
jim-minter: nit: timeout not used

test/e2e/client_test.go
```diff
@@ -40,21 +44,38 @@ func newTestClient(kubeconfig string) *testClient {
...
+func (t *testClient) createNamespace(namespace string) error {
+	if _, err := t.pc.ProjectRequests().Create(&project.ProjectRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		}}); err != nil {
+		return err
+	}
 	t.namespace = namespace
 	// TODO: Wait for a successful SAR check
+	return nil
```
jim-minter: may just be worth putting in some code that waits for the obvious service accounts to be created, before that bites a new starter

test/e2e/client_test.go
```diff
@@ -40,21 +44,38 @@ func newTestClient(kubeconfig string) *testClient {
...
 }
 
-func (t *testClient) cleanupNamespace(timeout time.Duration) {
+func (t *testClient) cleanupNamespace(timeout time.Duration) error {
 	if t.namespace == "" {
-		return
+		return nil
```
jim-minter: nit: simplifying this and have people to defer t.cleanupNamespace() is tempting, but it's not a blocker

kargakis: This is actually part of a AfterEach function so any new tests added should not care about namespace cleanup as it's handled automatically for them.

PR #437, opened 2018-09-14
[wip] disabled fqdn and publicSubdomain for router and apiserverprofile

charlesakalugwu:


pkg/util/fixtures/fixtures.go
```diff
@@ -10,14 +10,11 @@ func NewTestOpenShiftCluster() *api.OpenShiftManagedCluster {
...
 		Config:   &api.Config{},
 		Properties: &api.Properties{
-			FQDN:             "www.example.com",
 			OpenShiftVersion: "v3.10",
 			PublicHostname:   "openshift.test.example.com",
 			RouterProfiles: []api.RouterProfile{
 				{
 					Name:            "default",
-					FQDN:            "router-fqdn.example.com",
-					PublicSubdomain: "test.example.com",
```
mjudeikis: They are not removed, we just set PublicHostname to same value as FQDN. Which is logic in `config/utils.goL121` .

mjudeikis:
In addition to these: 
create.sh should be modified to have publicHostname removed. We will allow it post PP. FQDN should stay.  

Because we validate before we exeucte generate, we need to make sure that PublicHostname values for Router and API is not populated in the payload comming from the customer. So we fail if those are populated. There is one caviate here. When we generate and we set `PublicHostname/VanityDomain` to be same as FQDN. And we return those values back to user in `cmd/createororupdate.goL120`. So coversion code need to know that these values  should not be returned. 

In addition merge in `cmd/createorupdate.goL54` should check if when merging old and new we dont change FQDN as its not allowed (I think we have this already)

TL;DR: we need to make sure we fail if user specify PublicHostname value on validation, and we dont return it to the user back (as first not to consuse people and potential automation on customer side, where they get result and might reuse it for future clusters). And We keep `FQDN` and make sure that existing logic  default to `PublicHostname` under the hood.

pkg/validate/validate_test.go
```diff
@@ -18,7 +18,7 @@ name: openshift
 properties:
   openShiftVersion: v3.10
   publicHostname: openshift.test.example.com
-  fqdn: www.example.com
+  #fqdn: www.example.com
```
mjudeikis: either remove or keep. Dont comment things out :)

pkg/validate/validate_test.go
```diff
@@ -71,13 +71,13 @@ func TestValidate(t *testing.T) {
...
+				oc.Properties.FQDN = "example.com"
 			},
-			expectedErrs: []error{errors.New(`invalid properties.fqdn ""`)},
+			expectedErrs: []error{errors.New(`invalid properties.fqdn "example.com"`)},
 		},
 		"test external only false - invalid fqdn fails": {
-			f:            func(oc *api.OpenShiftManagedCluster) { oc.Properties.FQDN = "()" },
-			expectedErrs: []error{errors.New(`invalid properties.fqdn "()"`)},
+			f:            func(oc *api.OpenShiftManagedCluster) { oc.Properties.FQDN = "example.com" },
+			expectedErrs: []error{errors.New(`invalid properties.fqdn "example.com"`)},
```
mjudeikis: validation tests should check if PublicHostname is set - fail. FQDN logic should be the same

PR #441, opened 2018-09-14
use -X to store sync commit

jim-minter:
@kargakis @kwoodson I was hoping to do it this way so the commit goes straight into the logging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@
 /sync
 .vscode
 .idea
-COMMIT_ID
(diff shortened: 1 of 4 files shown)


mjudeikis:
need to rollback this too as it will break ci
https://github.com/openshift/release/pull/1475

kargakis:
https://github.com/openshift/release/pull/1506

/lgtm

mjudeikis:
@kargakis @jim-minter please merge this as CI now fails :)  small enough to deal with fallout post

kargakis:
Let's wait for the e2e test to finish? It should pass for this PR.

mjudeikis:
and it failed with the flake...

kargakis:
The image build was successful and `sync` seems to run in the cluster and all the infra resources are created so this is probably safe to green-button merge.

PR #444, opened 2018-09-14
always create StorageClasses before objects that use them

jim-minter:
@kargakis @kwoodson @mjudeikis @pweil- I'm fairly confident this will help our PV timeout problem.  I'm pretty sure that part/all of it is a race between the statefulset controller creating PVCs and the default StorageClass having been created (note, alphabetically speaking StorageClass is just after StatefulSet).

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -162,13 +162,14 @@ func setPodTemplateAnnotation(key, value string, o unstructured.Unstructured) {
 // resource filters
 var (
 	nsFilter = func(o unstructured.Unstructured) bool {
-		return o.GroupVersionKind().Kind == "Namespace"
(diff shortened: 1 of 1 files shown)


jim-minter:
might fix #361

kargakis:
I've thought about this too but it feels to me like a kube bug. Mind opening an issue upstream?

/lgtm

jim-minter:
change was fixing typo in commit, readding label

jim-minter:
for reference, https://github.com/kubernetes/kubernetes/issues/44370 is basically the issue upstream

mjudeikis:
/lgtm
the only concert is how fast each controller registers their components. So we would not have same condition as with crd

jim-minter:
there's no controller/status associated with the storageclass and it is expected that the create call does not return before a read call would be able to access it.

PR #448, opened 2018-09-15
don't wait 3 minutes if config blob read fails

jim-minter:
also convert Polls to PollImmediates - saves a few seconds

@kargakis ptal

### DIFF
--- a/cmd/sync/sync.go
+++ b/cmd/sync/sync.go
@@ -3,7 +3,9 @@ package main
 import (
 	"context"
 	"flag"
+	"io"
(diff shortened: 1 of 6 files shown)


stevekuznetsov:
@jim-minter there are people in the OWNERS for this that are not assignable to issues here -- are they collaborators on the repo? @sozercan @shrutir25

jim-minter:
good point @stevekuznetsov. #451.

mjudeikis:
lgtm  will leave unti somebody else to check it.

PR #450, opened 2018-09-15
[WIP] run ovs and sdn as static pods

jim-minter:


jim-minter:
This boots, but the ovs and sdn pods don't show as mirror pods because mirror pods aren't allowed to have service accounts.

This needs additional work to pass in a custom KUBECONFIG into the sdn pod in place of using the serviceAccountName.

jim-minter:
not convinced this helps speed startup that much, but it may be a prerequisite for running the sync pod as a deployment

mjudeikis:
I dont like this implementation, where we create another object than pods in master/node scripts. Makes us inconsistent. I vote for leaving them as pods until we found a better way to handle objects.

PR #458, opened 2018-09-17
Move imageConfig to its own struct

mjudeikis:
fixes: https://github.com/openshift/openshift-azure/issues/440


pkg/api/config.go
```diff
@@ -19,27 +19,7 @@ type Config struct {
...
-	RegistryConsoleImage        string `json:"registryConsoleImage,omitempty"`
-	AnsibleServiceBrokerImage   string `json:"ansibleServiceBrokerImage,omitempty"`
-	WebConsoleImage             string `json:"webConsoleImage,omitempty"`
-	OAuthProxyImage             string `json:"oAuthProxyImage,omitempty"`
-	PrometheusImage             string `json:"prometheusImage,omitempty"`
-	PrometheusAlertBufferImage  string `json:"prometheusAlertBufferImage,omitempty"`
-	PrometheusAlertManagerImage string `json:"prometheusAlertManagerImage,omitempty"`
-	LogBridgeImage              string `json:"logBridgeImage,omitempty"`
-	EtcdOperatorImage           string `json:"etcdOperatorImage,omitempty"`
+	Images       ImageConfig       `json:"images,omitempty"`
```
thekad: property name suggests this should be a list, shouldn't this be ``[]ImageConfig`` instead?

jim-minter:
@mjudeikis seeing as we love a good search and replace, is it not worth changing the field names `(.*)Image` to `\1` as well?

mjudeikis:
@jim-minter should be done

pkg/config/images.go
```diff
@@ -80,36 +78,34 @@ func selectContainerImagesOSA(cs *acsapi.OpenShiftManagedCluster) {
...
-		c.PrometheusAlertManagerImage = "registry.access.redhat.com/openshift3/prometheus-alertmanager:" + v
-		c.PrometheusNodeExporterImage = "registry.access.redhat.com/openshift3/prometheus-node-exporter:" + v
-
-		c.RegistryConsoleImage = "registry.access.redhat.com/openshift3/registry-console:" + v
-
-		c.AzureCLIImage = "docker.io/microsoft/azure-cli:latest" //TODO: create mapping for OSA release to any other image we use
-
-		if c.SyncImage == "" {
-			c.SyncImage = "quay.io/openshift-on-azure/sync:v3.10"
+		v := "v3.10.14" // TODO: perhaps we should calculate this from c.Images.ImageVersion
```
kwoodson: The latest image that we have produced is v3.10.34.  Is this something we should fix?

jim-minter: yes :( https://github.com/openshift/openshift-azure/issues/465 - please can you pick that up @kwoodson ?

PR #459, opened 2018-09-17
disable vanity domains for private preview

### LINKED ISSUE #370: disable vanity domains for apiserver and router for PP
check the validation

charlesakalugwu:
closes #370


hack/create.sh
```diff
@@ -75,7 +74,6 @@ properties:
         tenantId: $AZURE_TENANT_ID
   routerProfiles:
   - name: default
-    publicSubdomain: $RESOURCEGROUP.$DNS_DOMAIN
```
mjudeikis: @jim-minter this is "Defaulting code". As I understood we leaving this for Post PP?

mjudeikis:
So when this gets merged customer will get back PublicHostname back from API when submitting a request. Validation covers this if they take it and post back again. Dont know if we want to return PublicHostname or we want to obfuscate this field at all for now?

pkg/validate/validate.go
```diff
@@ -104,6 +104,10 @@ func validateProperties(p *api.Properties, externalOnly bool) (errs []error) {
 	if p.PublicHostname != "" && !isValidHostname(p.PublicHostname) {
 		errs = append(errs, fmt.Errorf("invalid properties.publicHostname %q", p.PublicHostname))
 	}
+	// for private preview
+	if p.PublicHostname != "" && p.PublicHostname != p.FQDN {
```
jim-minter: I think we should be failing validate if PublicHostname != "" here.
Then, everywhere where we use PublicHostname, we would need to reference a (new) helper function that returned FQDN if PublicHostname == "", otherwise PublicHostname.

pkg/validate/validate.go
```diff
@@ -257,6 +261,11 @@ func validateRouterProfile(rp api.RouterProfile) (errs []error) {
 		errs = append(errs, fmt.Errorf("invalid properties.routerProfiles[%q].fqdn %q", rp.Name, rp.FQDN))
 	}
 
+	// for private preview
+	if rp.PublicSubdomain != "" && rp.PublicSubdomain != rp.FQDN {
```
jim-minter: This logic is incorrect and should be removed.  For PP I think we're going to have to rely on the RP invalidating any user-provided rp.PublicSubdomain and populating the value here itself.  @sozercan is this realistic?

hack/create.sh
jim-minter: @charlesakalugwu if you remove this line, which you probably should, you'll need to add logic into enrich() to populate it.

pkg/validate/validate_test.go
```diff
@@ -29,7 +28,7 @@ properties:
         tenantId: aadTenantId
   routerProfiles:
   - name: default
-    publicSubdomain: test.example.com
+    fqdn: www.example.com
```
jim-minter: I think both fqdn and publicSubdomain should be here.

pkg/validate/validate.go
charlesakalugwu: I believe PublicHostname is already being defaulted to FQDN if it is "" during the plugin.Generate phase  (selectDNSNames in pkg/config/util.go).

jim-minter:
@charlesakalugwu to do this in time I've repushed this with updates.  ptal.
@mjudeikis or @kargakis please review and lgtm if appropriate

jim-minter:
sorry, or @kwoodson please review and lgtm if appropriate

PR #460, opened 2018-09-17
adds FQDN validation for cluster and for router

### LINKED ISSUE #348: need to validate fqdn (and router fqdn) end with <azure zone>.cloudapp.azure.com

thanasisk:
Fixes #348

### DIFF
--- a/pkg/util/fixtures/fixtures.go
+++ b/pkg/util/fixtures/fixtures.go
@@ -10,12 +10,12 @@ func NewTestOpenShiftCluster() *api.OpenShiftManagedCluster {
 		Name:     "openshift",
 		Config:   &api.Config{},
 		Properties: &api.Properties{
-			FQDN:             "www.example.com",
+			FQDN:             "example.eastus.cloudapp.azure.com",
 			OpenShiftVersion: "v3.10",
 			RouterProfiles: []api.RouterProfile{
 				{
 					Name:            "default",
-					FQDN:            "router-fqdn.example.com",
+					FQDN:            "router-fqdn.eastus.cloudapp.azure.com",
 					PublicSubdomain: "test.example.com",
 				},
 			},
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strings"
 
 	"github.com/go-test/deep"
 
@@ -39,6 +40,10 @@ var validRouterProfileNames = map[string]struct{}{
 
 func isValidHostname(h string) bool {
 	return len(h) <= 255 && rxRfc1123.MatchString(h)
+}
+
+func isAzureZone(fqdn string) bool {
+	return strings.HasSuffix(fqdn, ".cloudapp.azure.com") && len(strings.Split(fqdn, ".")) == 5
 }
 
 // Validate validates a OpenShiftManagedCluster struct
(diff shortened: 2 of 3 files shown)


pkg/validate/validate.go
```diff
@@ -29,6 +29,16 @@ var validAgentPoolProfileNames = map[string]struct{}{
 	string(api.AgentPoolProfileRoleMaster):  struct{}{},
 }
 
+// az account list-locations will fetch the latest additions, if needed
+var azureZones = regexp.MustCompile(`{` +
+	`eastasia|southeastasia|centralus|eastus|eastus2|westus|northcentralus|` +
+	`southcentralus|northeurope|westeurope|japanwest|japaneast|brazilsouth|` +
+	`australiaeast|australiasoutheast|southindia|centralindia|westindia|` +
```
jim-minter: I don't think we want to hard-code these ourselves and I'm not aware of a canonical location to pull these from.  Think it'll have to be something like `\w+\.cloudapp\.azure\.com`

charlesakalugwu: looks like there's an api to retrieve these from
`
az account list-locations | jq .[].name --raw-output
`

thanasisk: @charlesakalugwu true but how can we fetch it dynamically into the code without
1) requiring always on-line
2) requiring az credentials

I guess I will follow Jim's suggestion, which was my initial solution

jim-minter:
rebase onto #459 and this will be much simpler

kargakis:
/hold

can you add a unit test that uses an invalid hostname?

pkg/util/fixtures/fixtures.go
```diff
@@ -10,12 +10,12 @@ func NewTestOpenShiftCluster() *api.OpenShiftManagedCluster {
 		Name:     "openshift",
 		Config:   &api.Config{},
 		Properties: &api.Properties{
-			FQDN:             "www.example.com",
+			FQDN:             "www.example.eastus.cloudapp.azure.com",
```
jim-minter: will necessarily be foo.region.cloudapp.azure.com - can validate the number of domain parts

pkg/validate/validate.go
```diff
@@ -41,6 +43,10 @@ func isValidHostname(h string) bool {
 	return len(h) <= 255 && rxRfc1123.MatchString(h)
 }
 
+func isAzureZone(h string) bool {
+	return azureZones.MatchString(h)
```
jim-minter: use strings.HasSuffix() and also count the number of domain parts

pkg/validate/validate_test.go
```diff
@@ -314,3 +314,17 @@ func TestValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestInvalidAzureZone(t *testing.T) {
```
jim-minter: Condense to one function, TestIsAzureZone

kargakis: I imagined you'd be able to add a test case in `TestValidate`.

pkg/validate/validate.go
```diff
@@ -41,6 +42,10 @@ func isValidHostname(h string) bool {
 	return len(h) <= 255 && rxRfc1123.MatchString(h)
 }
 
+func isAzureZone(fqdn string) bool {
+	return strings.HasSuffix(fqdn, "cloudapp.azure.com") && len(strings.Split(fqdn, ".")) == 5
```
jim-minter: nits: need a preceding `.` on `".cloudapp.azure.com"`, can use strings.Count()

pkg/validate/validate_test.go
```diff
@@ -314,3 +314,14 @@ func TestValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestIsAzureZone(t *testing.T) {
```
jim-minter: add test for `"too.long.domain.cloudapp.azure.com"`

jim-minter:
lgtm, please squash the commits and will merge

PR #462, opened 2018-09-17
add basic prometheus rules

mjudeikis:
first shot to have some alerts for PP


jim-minter:
/hold

let's talk about this via bluejeans

jim-minter:
agreed that we're going to start off with the approach of trialling alerting rules at the centre and possibly push out to the ledge later.

mjudeikis:
close in favor of: https://github.com/openshift/azure-misc/pull/33

PR #468, opened 2018-09-18
Parsing image version.

### LINKED ISSUE #465: fix image version discrepancy
should be v3.10.34, not v3.10.14

kwoodson:
Parsing image version to determine correct version string.

fixes #465

### DIFF
--- a/pkg/config/images.go
+++ b/pkg/config/images.go
@@ -1,11 +1,23 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 )
+
+// openshiftVersion converts a VM image version (e.g. 310.14.20180101) to an
+// openshift container image version (e.g. v3.10.14)
+func openShiftVersion(imageVersion string) (string, error) {
+	parts := strings.Split(imageVersion, ".")
+	if len(parts) != 3 || len(parts[0]) < 2 {
+		return "", fmt.Errorf("invalid imageVersion %q", imageVersion)
+	}
+
+	return fmt.Sprintf("v%s.%s.%s", parts[0][:1], parts[0][1:], parts[1]), nil
+}
 
 func selectNodeImage(cs *acsapi.OpenShiftManagedCluster, deployOS string) {
 	c := cs.Config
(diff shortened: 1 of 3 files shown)


kwoodson:
@jim-minter, would you prefer a hard coded version string or the regexp?

pkg/config/images.go
```diff
@@ -1,12 +1,29 @@
...
+	"regexp"
 	"strings"
 
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 )
 
+func imageVersion(incVersion string) string {
+	// Version should be in xxx.xx.xxxxxxxx format
+	// e.g: 310.14.20180101
+	rxVersion := regexp.MustCompile(`^([0-9]){1}([0-9]{1,3})` +
```
jim-minter: use strings.Split

pkg/config/images.go
```diff
@@ -1,12 +1,29 @@
...
 import (
+	"fmt"
 	"os"
+	"regexp"
 	"strings"
 
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 )
 
+func imageVersion(incVersion string) string {
```
jim-minter: unit test

jim-minter: `func openShiftVersion(imageVersion string) string {`

pkg/config/images.go
kwoodson: I looked into string.split but wasn't aware of a clean way to divide the numbers exactly how we wanted them.
`310.14.20180101` to `3.10.14`

If you have an example that'd be great.

pkg/config/images_test.go
```diff
@@ -6,6 +6,40 @@ import (
...
+		testVersion string
+		testResult  string
+		testError   string
+	}{
+		"valid version string": {
+			testVersion: "310.14.20180101",
+			testResult:  "v3.10.14",
+		},
+		"invalid version string length": {
+			testVersion: "310.14.20180101",
```
kwoodson: Need to add more length here to cause the version string to be invalid.

pkg/config/images.go
```diff
@@ -1,12 +1,24 @@
...
 	"os"
 	"strings"
 
 	acsapi "github.com/openshift/openshift-azure/pkg/api"
 )
 
+// openshiftVersion converts a VM image version (e.g. 310.14.20180101) to an
+// openshift container image version (e.g. v3.10.14)
+func openShiftVersion(imageVersion string) (string, error) {
+	parts := strings.Split(imageVersion, ".")
```
kwoodson: I understand how you wanted to use split and use the returned string for `part[0]` and use array index for the version.

PR #470, opened 2018-09-18
Make AgentPoolProfile.Count a *int

jim-minter:
This is not pleasant, but the least worst mechanism to leave open the
possibility of being able to scale APPs down to 0 at the same time as
being able to specify a partial manifest on PUT.


jim-minter:
/hold

@kargakis @pweil- @mjudeikis thoughts?

jim-minter:
marked private-preview but need to make a call

kargakis:
Let me know when you need another review here if/when you rebase

/uncc

kargakis:
Will be handled as part of the merge card

PR #471, opened 2018-09-18
Implement merge

### LINKED ISSUE #434: MergeConfig overwrites existing model
In an` PUT` (update) operation, if user sends in a partial request such as the basic `routerProfile` below:

```
...
	  "routerProfiles": [
		  {
			  "name": "default"
		  }
	  ]
...
```

This will override the existing fields like `publicSubdomain` and `fqdn`, since it will skip copying because `len(cs.Properties.RouterProfiles)` will not be 0.
https://github.com/openshift/openshift-azure/blob/master/pkg/plugin/plugin.go#L55
This issue applies to other properties such as `AgentPoolProfiles` as well.

It might be better to take the existing saved model and update it rather then trying to constructing a new one based on the old model.

jim-minter:
Builds on #470 

fixes #434

### DIFF
--- a/pkg/api/merge.go
+++ b/pkg/api/merge.go
@@ -1,0 +1,368 @@
+package api
+
+// TODO: needs unit tests
+
+// Notes: shortcomings of the Azure-style API / its integration with Golang
+// include:
+// 1. There is no mechanism to delete an item from a slice or map without
+// implementing a separate CRUD endpoint.
+// 2. There is no way to set a plain value to its zero value (e.g. bool to
+// false, int to 0, string to "") without storing it as a pointer in the API.
+
+// At the moment Config isn't merged.  Perhaps what we actually need is a
+// merging converter f(*ext, *int) *int.  Perhaps this would mean we could get
+// rid of some pointers in our internal type.
+
+func Merge(complete, partial *OpenShiftManagedCluster) *OpenShiftManagedCluster {
+	if complete != nil {
+		complete = complete.DeepCopy() // required because Config isn't merged
+	}
+
+	return mergeOpenShiftManagedCluster(complete, partial)
+}
+
+func mergeOpenShiftManagedCluster(complete, partial *OpenShiftManagedCluster) *OpenShiftManagedCluster {
+	switch {
+	case complete == nil && partial == nil:
+		return nil
+	case complete == nil:
+		complete = &OpenShiftManagedCluster{}
+	case partial == nil:
+		partial = &OpenShiftManagedCluster{}
+	}
+
+	return &OpenShiftManagedCluster{
+		ID:         mergeString(complete.ID, partial.ID),
+		Location:   mergeString(complete.Location, partial.Location),
+		Name:       mergeString(complete.Name, partial.Name),
+		Plan:       mergeResourcePurchasePlan(complete.Plan, partial.Plan),
+		Tags:       mergeTags(complete.Tags, partial.Tags),
+		Type:       mergeString(complete.Type, partial.Type),
+		Properties: mergeProperties(complete.Properties, partial.Properties),
+		Config:     complete.Config, // NOTE: Config isn't merged
+	}
+}
+
+func mergeString(complete, partial string) string {
+	if partial == "" {
+		return complete
+	}
+	return partial
+}
+
+func mergeResourcePurchasePlan(complete, partial *ResourcePurchasePlan) *ResourcePurchasePlan {
+	switch {
+	case complete == nil && partial == nil:
+		return nil
+	case complete == nil:
+		complete = &ResourcePurchasePlan{}
+	case partial == nil:
+		partial = &ResourcePurchasePlan{}
+	}
+
+	return &ResourcePurchasePlan{
+		Name:          mergeString(complete.Name, partial.Name),
+		Product:       mergeString(complete.Product, partial.Product),
+		PromotionCode: mergeString(complete.PromotionCode, partial.PromotionCode),
+		Publisher:     mergeString(complete.Publisher, partial.Publisher),
+	}
+}
+
+func mergeTags(complete, partial map[string]string) map[string]string {
+	if complete == nil && partial == nil {
+		return nil
+	}
+
+	// TODO: is this the semantic that we want for mergeTags?
+
+	rv := make(map[string]string, len(complete))
+	for k, v := range complete {
+		rv[k] = v
+	}
+	for k, v := range partial {
+		rv[k] = v
+	}
+	return rv
+}
+
+func mergeProperties(complete, partial *Properties) *Properties {
+	switch {
+	case complete == nil && partial == nil:
+		return nil
+	case complete == nil:
+		complete = &Properties{}
+	case partial == nil:
+		partial = &Properties{}
+	}
+
+	return &Properties{
+		ProvisioningState:       mergeProvisioningState(complete.ProvisioningState, partial.ProvisioningState),
+		OpenShiftVersion:        mergeString(complete.OpenShiftVersion, partial.OpenShiftVersion),
+		PublicHostname:          mergeString(complete.PublicHostname, partial.PublicHostname),
+		FQDN:                    mergeString(complete.FQDN, partial.FQDN),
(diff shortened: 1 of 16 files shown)


jim-minter:
/hold

@kargakis @pweil- @mjudeikis thoughts?

jim-minter:
marked private-preview but need to make a call

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
...
+		Publisher:     mergeString(complete.Publisher, partial.Publisher),
+	}
+}
+
+func mergeTags(complete, partial map[string]string) map[string]string {
+	if complete == nil && partial == nil {
+		return nil
+	}
+
+	// TODO: is this the semantic that we want for mergeTags?
```
kargakis: Is there any use cases where someone would want to remove a tag? I can't think of any.

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
+package api
+
+// TODO: needs unit tests
+
+// Notes: shortcomings of the Azure-style API include:
+// 1. There is no mechanism to delete an item from a slice or map without
+// implementing a separate CRUD endpoint.
+// 2. There is no way to set a plain value to its zero value (e.g. bool to
```
kargakis: Isn't this a short-coming of Golang APIs rather than Azure?

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
+package api
+
+// TODO: needs unit tests
```
kargakis: Follow-up issue

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
...
+		IdentityProviders: mergeIdentityProviders(complete.IdentityProviders, partial.IdentityProviders),
+	}
+}
+
+func mergeIdentityProviders(complete, partial []IdentityProvider) []IdentityProvider {
+	if complete == nil && partial == nil {
+		return nil
+	}
+
+	// preconditions: each RouterProfile in partial is named uniquely; each
```
mjudeikis: copy-paste

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
...
+	complete = complete.DeepCopy() // required because Config isn't merged
+
+	return mergeOpenShiftManagedCluster(complete, partial)
+}
+
+func mergeOpenShiftManagedCluster(complete, partial *OpenShiftManagedCluster) *OpenShiftManagedCluster {
+	switch {
+	case complete == nil && partial == nil:
+		return nil
+	case complete == nil:
```
pweil-: does this need to move up to `Merge` since you're calling `DeepCopy` on a possibly nil `OpenShiftManagedCluster`?

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
...
+		FQDN:                    mergeString(complete.FQDN, partial.FQDN),
+		RouterProfiles:          mergeRouterProfiles(complete.RouterProfiles, partial.RouterProfiles),
+		AgentPoolProfiles:       mergeAgentPoolProfiles(complete.AgentPoolProfiles, partial.AgentPoolProfiles),
+		AuthProfile:             mergeAuthProfile(complete.AuthProfile, partial.AuthProfile),
+		ServicePrincipalProfile: mergeServicePrincipalProfile(complete.ServicePrincipalProfile, partial.ServicePrincipalProfile),
+		AzProfile:               mergeAzProfile(complete.AzProfile, partial.AzProfile),
+	}
+}
+
+func mergeProvisioningState(complete, partial ProvisioningState) ProvisioningState {
```
pweil-: nit: these are just strings we could reuse `mergeString` (thought it might look a bit messy since casting which is why you may not have)

pkg/api/merge.go
```diff
@@ -0,0 +1,368 @@
...
+		if p, found := partialMap[c.Name]; found {
+			rv = append(rv, *mergeRouterProfile(&c, &p))
+			delete(partialMap, c.Name)
+		} else {
+			rv = append(rv, c)
+		}
+	}
+
+	for _, p := range partial {
+		if _, found := partialMap[p.Name]; !found {
```
pweil-: Trying to work through this logic and it seems that partial still contains all the entries which may have been deleted from `partialMap` so I'd expect this is appending things it shouldn't.  For instance:

```
# after partialMap loop
partial := ["foo"]
complete := ["foo"]
partialMap := {"foo": {//partialFoo}}
rv := []

# after complete loop
partial := ["foo"]
complete := ["foo"]
partialMap := {} 
rv := [mergedFoo]

# after this loop
partial := ["foo"]
complete := ["foo"]
partialMap := {} 
rv := [foo (mergedFoo), foo (partialFoo)]

File: pkg/api/merge.go (hunk 3 above)
Comment (pweil-) [reply pkg/api/merge.go]: This is one of the most critical parts of code, I think we'd be wise to spend some time throwing in a unit test today to make sure we're performing how we expect.

File: pkg/api/merge.go (hunk 7 above)
Comment (jim-minter) [reply pkg/api/merge.go]: Still looking, but I don't think so.
In the complete loop, a p is deleted from partialMap whenever it corresponds to a c.
In the following loop, ps are added in order from partial only when they still exist in partialMap (i.e. did not correspond to a complete, were not previously added).

Comment (jim-minter) [reply pkg/api/merge.go]: The key piece here `if _, found := partialMap[p.Name]; !found` at the end

File: pkg/api/merge.go (hunk 5 above)
Comment (jim-minter) [reply pkg/api/merge.go]: I think the DeepCopy() just needs to be guarded with an `if complete != nil`.  Thanks, will fix.

File: pkg/api/merge.go (hunk 1 above)
Comment (jim-minter) [reply pkg/api/merge.go]: I'm sure there are.  Will ask the MSFT team about this.

File: pkg/api/merge.go (hunk 2 above)
Comment (jim-minter) [reply pkg/api/merge.go]: It's certainly a sort of nasty interaction between the two.  In principle the Golang json package provides access to the underlying parser, which would allow for fixing 2 without going to pointers everywhere, but it's quite a lot of effort.

File: pkg/api/merge.go (hunk 7 above)
Comment (pweil-) [reply pkg/api/merge.go]:
```go
func TestMergeRouterProfile(t *testing.T) {
	complete := []RouterProfile{RouterProfile{Name: "foo"}}
	partial := []RouterProfile{RouterProfile{Name: "foo"}}
	merged := mergeRouterProfiles(complete, partial)
	if len(merged) == 2 {
		t.Errorf("too many: %v", merged)
	}
}
```
```
GOROOT=/home/pweil/.gimme/versions/go1.9.linux.amd64 #gosetup
GOPATH=/home/pweil/codebase/openshift-azure #gosetup
/home/pweil/.gimme/versions/go1.9.linux.amd64/bin/go test -c -i -o /tmp/___TestMergeRouterProfile_in_github_com_openshift_openshift_azure_pkg_api github.com/openshift/openshift-azure/pkg/api #gosetup
/tmp/___TestMergeRouterProfile_in_github_com_openshift_openshift_azure_pkg_api -test.v -test.run ^TestMergeRouterProfile$ #gosetup
	merge_test.go:10: too many: [{foo  } {foo  }]

Process finished with exit code 1
```

jim-minter: Boo :(  Should be `found`, not `!found`.

pkg/api/merge.go
```diff
@@ -0,0 +1,371 @@
...
+		Name:         mergeString(complete.Name, partial.Name),
+		Count:        mergeIntPtr(complete.Count, partial.Count),
+		VMSize:       mergeVMSize(complete.VMSize, partial.VMSize),
+		VnetSubnetID: mergeString(complete.VnetSubnetID, partial.VnetSubnetID),
+		OSType:       mergeOSType(complete.OSType, partial.OSType),
+		Role:         mergeAgentPoolProfileRole(complete.Role, partial.Role),
+	}
+}
+
+func mergeIntPtr(complete, partial *int) *int {
```
pweil-: It's overwriting complete when partial is populated, seems like it could just be

```go
if partial != nil {
  return to.IntPtr(*partial)
} else if complete != nil {
  return to.IntPtr(*complete)
}
return nil

PR #472, opened 2018-09-18
Make kubeconfigs derived

jim-minter:
Avoids storing kubeconfigs in the cluster Config struct - they can be derived from the contents of the Config struct.


jim-minter:
also for consideration for private preview @mjudeikis @kargakis @pweil-

mjudeikis:
I see where you going with this, but this means now that each tool we want use kubeconfig needs to generate it. This is not a big blocker, but SRE's or MSFT now will need to generate one too when acceesing the cluster. I think this might be little bit inconveniant for PP.

mjudeikis:
+ MSFT was asking for cluster-reader kubeconfig for their side of monitoring/automation. If we merge this, they will need to generate one too.

jim-minter:
The reason I'm interested in doing this now is for config API cleanup.  We can straightforwardly make the kubeconfig generation part of a "libosa" which is reusable by anyone who has access to the Config struct.  It's not hard to generate a kubeconfig - for example create.sh still places the admin.kubeconfig in _data.

pweil-:
Is this a bug that needs merged on day of freeze?  Are we going to be in a spot where we can't change this next week?  Doesn't feel like something that needs last minute attention.

kargakis:
reopen if still relevant

PR #473, opened 2018-09-18
don't set a slice capacity with a negative number

### LINKED ISSUE #464: Convert panics on partial request
Looks like convert code panics on partial requests such as when `AgentPoolProfiles` and/or `AuthProfile` is empty.

jim-minter:
fixes #464


kargakis:
Needs to be unit-tested in a follow-up

/lgtm

PR #474, opened 2018-09-18
Add unit tests for validateUpdateContainerService

### LINKED ISSUE #217: unit test for validating diff between old and new config on different user requests
@openshift/sig-azure 

/assign pweil-

asalkeld:
closes #217

### DIFF
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -77,8 +77,6 @@ func validateContainerService(c *OpenShiftManagedCluster, externalOnly bool) (er
 }
 
 func validateUpdateContainerService(cs, oldCs *OpenShiftManagedCluster, externalOnly bool) (errs []error) {
-	// TODO: function needs unit testing.
-
 	newAgents := make(map[string]*AgentPoolProfile)
 	for i := range cs.Properties.AgentPoolProfiles {
 		newAgent := cs.Properties.AgentPoolProfiles[i]
--- a/pkg/api/validate_test.go
+++ b/pkg/api/validate_test.go
@@ -330,3 +330,74 @@ func TestIsAzureZone(t *testing.T) {
 		t.Errorf("Valid FQDN failed to pass test: %s", validFqdn)
 	}
 }
+
+func TestValidateUpdateContainerService(t *testing.T) {
+	var twoAgents = []byte(`---
+location: eastus
+name: openshift
+properties:
+  publicHostname: donotchange
+  routerProfiles:
+  - name: default
+    publicSubdomain: test.example.com
+  agentPoolProfiles:
+  - name: infra
+    role: infra
+    count: 1
+    vmSize: Standard_D2s_v3
+    osType: Linux
+  - name: myCompute
+    role: compute
+    count: 1
+    vmSize: Standard_D2s_v3
+    osType: Linux
+`)
+
+	tests := map[string]struct {
+		newAgentCount    int
+		oldAgentCount    int
+		newHostNameValue string
+		wantErrs         []error
+	}{
+		"good-2": {
+			newAgentCount: 1,
+			oldAgentCount: 1,
+		},
+		"good-2-count-not-important": {
+			newAgentCount: 5,
+			oldAgentCount: 2,
+		},
+		"bad-field-change": {
+			newAgentCount:    1,
+			oldAgentCount:    1,
+			newHostNameValue: "different",
+			wantErrs:         []error{errors.New("invalid change [Properties.PublicHostname: different != donotchange]")},
+		},
+	}
+
+	for name, tt := range tests {
+		var newCs *OpenShiftManagedCluster
+		var oldCs *OpenShiftManagedCluster
+
+		err := yaml.Unmarshal(twoAgents, &oldCs)
+		if err != nil {
+			t.Fatal(err)
+		}
+		newCs = oldCs.DeepCopy()
+
+		for i := range oldCs.Properties.AgentPoolProfiles {
+			oldCs.Properties.AgentPoolProfiles[i].Count = tt.oldAgentCount
+		}
+		for i := range newCs.Properties.AgentPoolProfiles {
+			newCs.Properties.AgentPoolProfiles[i].Count = tt.newAgentCount
+		}
+		if tt.newHostNameValue != "" {
+			newCs.Properties.PublicHostname = tt.newHostNameValue
+		}
+
+		gotErrs := validateUpdateContainerService(newCs, oldCs, false)
+		if !reflect.DeepEqual(gotErrs, tt.wantErrs) {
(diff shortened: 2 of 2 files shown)


pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateUpdateContainerService(t *testing.T) {
+
```
jim-minter: no additional newline after `{` or before `}` please

jim-minter: no additional newline after `{` or before `}` please

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+			oldTemplate:   threeAgents,
+			oldAgentCount: 1,
+			externalOnly:  false, // not used
+			wantErrs:      []error{errors.New("invalid change [Properties.AgentPoolProfiles.slice[2]: <no value> != {mySomethingElse 1 Standard_D2s_v3  Linux compute}]")},
+		},
+	}
+
+	for name, tt := range tests {
+		var newOc *v20180930preview.OpenShiftManagedCluster
+		err := yaml.Unmarshal(tt.newTemplate, &newOc)
```
jim-minter: lots of boilerplate here.  write yaml that unmarshals direct to a *api.OpenShiftManagedCluster to save the convert.

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		newCs := api.ConvertFromV20180930preview(newOc)
+		for _, o := range newCs.Properties.AgentPoolProfiles {
+			o.Count = tt.newAgentCount
```
jim-minter: unfortunately this code has no side effect (imo this is the biggest gotcha in Golang).
see https://bryce.is/writing/code/jekyll/update/2015/11/01/3-go-gotchas.html

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+		err = yaml.Unmarshal(tt.oldTemplate, &oldOc)
+		if err != nil {
+			t.Fatal(err)
+		}
+		oldCs := api.ConvertFromV20180930preview(oldOc)
+		for _, o := range oldCs.Properties.AgentPoolProfiles {
+			o.Count = tt.oldAgentCount
+		}
+
+		t.Run(name, func(t *testing.T) {
```
jim-minter: didn't know t.Run existed.  It's a bit overkill for this scenario.

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+		externalOnly  bool
+		wantErrs      []error
+	}{
+		"good-2": {
+			newTemplate:   twoAgents,
+			newAgentCount: 2,
+			oldTemplate:   twoAgents,
+			oldAgentCount: 1,
+			externalOnly:  false, // not used
+			wantErrs:      nil,
```
jim-minter: generally preferable not to initialise items to their zero value explicitly - ie leave out `foo: nil`, `bar: 0`, `baz: ""`

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+    vmSize: Standard_D2s_v3
+    osType: Linux
+`)
+
+	tests := map[string]struct {
+		newTemplate   []byte
+		newAgentCount int
+		oldTemplate   []byte
+		oldAgentCount int
+		externalOnly  bool
```
jim-minter: would remove and just hard tie the validateUpdateContainerService() argument to false for now

pkg/validate/validate_test.go
```diff
@@ -7,7 +7,6 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ghodss/yaml"
-
```
jim-minter: for better or worse as things stand we are using a convention of separating golang standard library imports, github.com/openshift/openshift-azure/* imports, and everything else into 3 separate blocks.

pkg/validate/validate_test.go
```diff
@@ -324,3 +323,114 @@ func TestValidate(t *testing.T) {
...
+		}
+		oldCs := api.ConvertFromV20180930preview(oldOc)
+		for _, o := range oldCs.Properties.AgentPoolProfiles {
+			o.Count = tt.oldAgentCount
+		}
+
+		t.Run(name, func(t *testing.T) {
+			gotErrs := validateUpdateContainerService(newCs, oldCs, tt.externalOnly)
+
+			if (gotErrs == nil && tt.wantErrs != nil) || !reflect.DeepEqual(gotErrs, tt.wantErrs) {
```
jim-minter: simplifies to `if !reflect.DeepEqual(gotErrs, tt.wantErrs)` ?

pkg/validate/validate_test.go
asalkeld: thanks for the link, interesting..

pkg/validate/validate_test.go
asalkeld: will do

pkg/validate/validate_test.go
asalkeld: ok

pkg/validate/validate_test.go
asalkeld: ok

pkg/validate/validate_test.go
asalkeld: it would be nice to have a linter that checked for this stuff

pkg/validate/validate_test.go
asalkeld: done

asalkeld:
thanks for the review @jim-minter I have gone through all your comments and fixed them up.

pkg/validate/validate_test.go
```diff
@@ -327,3 +327,109 @@ func TestIsAzureZone(t *testing.T) {
...
+    count: 1
+    vmSize: Standard_D2s_v3
+    osType: Linux
+  - name: myCompute
+    role: compute
+    count: 1
+    vmSize: Standard_D2s_v3
+    osType: Linux
+`)
+	var threeAgents = []byte(`---
```
jim-minter: This is a bit of a confusing test because this manifest would not pass validate (we currently do not permit multiple compute agentpools).  I feel like the test might be better off calling Validate() but checking the errors returned against those expected from validateUpdateContainerService.  Or at least ensuring that the input manifests are valid in the first place.

asalkeld:
@jim-minter ptal

jim-minter:
/lgtm

thanks!

Issue #463, opened 2018-09-17
review kube-state metrics rbac


mjudeikis:
```Sep 17 18:56:06 master-000000 dockerd-current[2030]: E0917 18:56:06.375043       1 nanny_lib.go:95] Error while querying apiserver for resources: pods "kube-state-metrics-587cf9796b-pn5bj" is forbidden: User "system:serviceaccount:openshift-metrics:kube-state-metrics" cannot get pods in the namespace "openshift-metrics": User "system:serviceaccount:openshift-metrics:kube-state-metrics" cannot get pods in project "openshift-metrics" ```

the functionality works but in the logs I noticed this. Might be something missing

PR #475, opened 2018-09-18
add missing rbac kube-state-metrics

mjudeikis:
@kargakis plat 
based on : https://github.com/kubernetes/kube-state-metrics/tree/master/kubernetes
solves: https://github.com/openshift/openshift-azure/issues/463


mjudeikis:
dam, generate

Issue #422, opened 2018-09-13
Increase control plane+node loglevel for CI runs and persist logs in test artifacts


kargakis:
https://github.com/openshift/openshift-azure/commit/4fd733dead7215e2528e13aab48cf9ee539838f8 reduced the control plane / kubelet loglevel from 4 to 2. https://github.com/openshift/openshift-azure/commit/cb21b24246dc9de706755594e5e034482fd71176 disabled loganalytics for CI runs so we should be able to re-increase the loglevel back to 4 for CI runs (use `RUNNING_UNDER_TEST`) and ensure we persist all those precious logs in the [test artifacts](https://github.com/openshift/release/blob/fbe4bfc55d411c0cd08e2138ab17061fce3906a4/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L194-L196). Ensure all [e2e templates](https://github.com/openshift/release/tree/master/ci-operator/templates/openshift/openshift-azure) get updated with the same change. 

@openshift/sig-azure

kargakis:
Persisting control plane logs is addressed in https://github.com/openshift/release/pull/1554

kargakis:
The last thing that remains is to get the kubelet log out of each and every machine.

PR #480, opened 2018-09-19
Increase loglevel for CI runs

kargakis:
@openshift/sig-azure ptal, part of https://github.com/openshift/openshift-azure/issues/422


pkg/arm/data/master-startup.sh
```diff
@@ -486,7 +488,11 @@ spec:
     - master
     - api
     - --config=/etc/origin/master/master-config.yaml
+{{- if not .Derived.RunningUnderTest }}
```
kwoodson: I think I'd prefer the logic swapped on this.  That way the default case is "normal".
```
{{- if .Derived.RunningUnderTest }}
    - --loglevel=4
{{ - else }}
    - --loglevel=2
{{ - end }}
```

Thoughts?

jim-minter:
@kargakis is this a good example of an issue where "@ openshift/sig-azure ptal" might not have been warranted? :)

pkg/arm/data/master-startup.sh
kargakis: sorry, had to repush since I cp'd the above stanza and it has some extra whitespace.

PR #483, opened 2018-09-19
Etcd cronjob

mjudeikis:
This is RAW implementation for cronjob based etcd backups. Just to prove point it works. Idea is to manage these resources via sync pod. For this sync pod needs:
1. To create resources only at certain time (each 1 hour)
2. Clean old resources, which are older than x (1 day old CRDs)

Currently, sync is not doing this. Do we want to start doing this or stay with raw implementation? 

cc: @openshift/sig-azure 
/hold


mjudeikis:
@jim-minter This was the reason I raised this. So it would be clear what is a current behavior. 

>what about deleting the old backups from the storage as well as old CRs?

This will require access to Azure API. So need to move to sync pod for sure

> correlated: what's the mechanism for knowing that if a backup is in the storage, that it is valid (like the valid=True on our image builds)? Is it the case that if the backup fails part-way through writing the storage, the item doesn't appear in the storage?

They use `io.Pipe()` to pass asyn writer. So I think it will depend on how ABS/S3 will act if the write is terminated in the middle. If it follows S3 model it should be `read-after-write-consistency.` this should be a case. 

> are we sure the sync pod won't mess with CRs that being created? 

it depends on how we write it :) In this model its job responsibility to create CR's. And Sync deploys job. 

> what is the test strategy?

Do e2e test where we set backup interval lets say 2 min, and purge 2 same. We need long-running test for this :/ 

> I'm about to say we need to step back from this and architect it out properly.

Happy to do this. Would like to do the implementation but at this point I agree with you on this.

PR #488, opened 2018-09-19
Allow the region to be passed in

### LINKED ISSUE #486: enable engineering in additional regions
suggest westeurope and australiasoutheast (recommendations based on availability of loganalytics service)

- [x] publish images in additional regions - try australia, germany, uk
- [x] parameterise eastus in create.sh, allow westeurope and australiasoutheast

at this point should be able to use published images outside of eastus

- [ ] copy published internal disk images to additional regions
- [ ] update purge and publish code accordingly

at this point should be able to use internal images outside eastus as well

asalkeld:
closes #486

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -39,7 +39,7 @@ for region in "${valid_regions[@]}"; do
 done
 if [[ $match = 0 ]]; then
     echo "Error invalid region: must be one of ${valid_regions[@]}"
-    exist 1
+    exit 1
 fi
 
 if [[ -z "$DNS_DOMAIN" ]]; then
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -25,6 +25,23 @@ if [[ -z "$AZURE_CLIENT_SECRET" ]]; then
     exit 1
 fi
 
+if [[ -z "$AZURE_REGION" ]]; then
+    echo error: must set AZURE_REGION
+    exit 1
+fi
+valid_regions=(eastus westeurope australiasoutheast)
+match=0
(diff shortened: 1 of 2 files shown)


hack/create.sh
```diff
@@ -25,6 +25,23 @@ if [[ -z "$AZURE_CLIENT_SECRET" ]]; then
     exit 1
 fi
 
+if [[ -z "$AZURE_REGION" ]]; then
```
jim-minter: I know it's painful and unfriendly but I'd prefer to replace all this with

```
if [[ -z "$AZURE_REGION" ]]; then
    echo error: must set AZURE_REGION
    exit 1
fi
```

Fundamentally I want this script to die, not become a user friendly crutch.

jim-minter: Doing the above may imply that a separate commit is needed to fix up our tests.  If so, please have a go - look in openshift/release.

asalkeld: ok, I had this initially then thought of the build job... https://github.com/openshift/release/pull/1588

asalkeld:
I must have missed something in release:

2018/09/20 23:26:46 Running pod e2e-azure
Using sync image registry.svc.ci.openshift.org/ci-op-7ks54fqk/stable:sync
+ set +x
error: must set AZURE_REGION
2018/09/20 23:27:01 Container setup in pod e2e-azure failed, exit code 1, reason Error

kargakis:
/lgtm

Please send an email in aos-azure and ask folks to update their env files to use the new variable.

PR #489, opened 2018-09-20
Basic unit tests for arm generate

asalkeld:


pkg/arm/arm_test.go
```diff
@@ -0,0 +1,75 @@
...
+		wantErr           bool
+		wantResourceTypes []string
+	}{
+		{
+			name:     "good-create",
+			cs:       fixtures.NewTestOpenShiftCluster(),
+			isUpdate: false,
+			wantErr:  false,
+			wantResourceTypes: []string{
+				"\"type\": \"Microsoft.OperationalInsights/workspaces\"",
```
jim-minter: use backticks here to avoid the backslashes

pkg/arm/arm_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+	"testing"
+
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/config"
+	"github.com/openshift/openshift-azure/pkg/util/fixtures"
+	"github.com/sirupsen/logrus"
+)
+
+func Test_simpleGenerator_Generate(t *testing.T) {
```
jim-minter: no underscores please

pkg/arm/arm_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/config"
+	"github.com/openshift/openshift-azure/pkg/util/fixtures"
+	"github.com/sirupsen/logrus"
```
kargakis: nit: group imports as discussed elsewhere

jim-minter: @kargakis I feel like if you really want this enforced, a lint tool is in order.  I don't think it's worth the effort.

Issue #487, opened 2018-09-19
Revert ImageConfigFormat move to derived


kargakis:
Partially revert https://github.com/openshift/openshift-azure/pull/414

ImageConfigFormat is reused by the sync pod and it's easier to share the variable via config rather than trying to keep in sync env variables

@jim-minter 
/kind bug

Issue #466, opened 2018-09-17
pull all os.Getenvs() out of plugin code and centralise in a plugin config struct


jim-minter:
It's important that none of the plugin code that is vendored into the RP should call os.Getenv.  We don't want to make any expectations about the RP environment, especially undocumented expectations.  They could bite us horribly later.

At the moment we check the following environment variables in the plugin:
RUNNING_UNDER_TEST
IMAGE_{RESOURCEGROUP,RESOURCENAME,OFFER,VERSION}
DEPLOY_OS
OREG_URL

All of these variables are expected to be empty when the code is running in the RP.

For these environment variables, only, that are actually used in the plugin, we should add a mechanism to set them in the PluginConfig.  Maybe for example:

```
type PluginConfig struct {
	SyncImage       string
	AcceptLanguages []string
	TestConfig      *TestConfig
}

type TestConfig struct {
	ImageResourceGroup string
	...
}
```

We would insist that the RP runs with TestConfig set to nil.  We could populate it otherwise if we wanted.

No environment variables used outside the plugin in the Fake RP (i.e. cmd/createorupdate) (e.g. AUTOACCEPT_MARKETPLACE_AGREEMENT, AZURE_CLIENT_SECRET) should be anywhere near the plugin config.

kargakis:
This is a prerequisite for sharing CreateOrUpdate properly between different processes/contexts, something needed in transforming our fake RP to an http server. @thekad let me know what I can do to help out. Happy to review once you have a PR out.

thekad:
@jim-minter there is a problem specifically with 3 vars: `RUNNING_UNDER_TEST` and `IMAGE_RESOURCE{NAME,GROUP}`: these are read directly from the templates via `GenerateARM`, that means we either a) change the GenerateARM API interface to include a `PluginConfig | TestConfig` or b) embed a `PluginConfig | TestConfig` in the Context or OpenShiftManagedCluster objects so they can still be read from the template.
Thoughs? cc @kargakis

thekad:
From our call this morning: It's fine to change the `api.plugin.GenerateARM` call.

PR #490, opened 2018-09-20
Add unit tests for derive

asalkeld:


pkg/config/derive_test.go
```diff
@@ -0,0 +1,298 @@
...
+			d := derived{}
+			os.Setenv("IMAGE_RESOURCEGROUP", tt.want)
+			if got := d.ImageResourceGroup(); got != tt.want {
+				t.Errorf("derived.ImageResourceGroup() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_derived_ImageConfigFormat(t *testing.T) {
```
kargakis: Don't test this, we need to revert its move in derive, see https://github.com/openshift/openshift-azure/issues/487

asalkeld: Okay

pkg/config/derive_test.go
```diff
@@ -0,0 +1,298 @@
...
+		want string
+	}{
+		{
+			want: "test-image1",
+		},
+		{
+			want: "test-image2",
+		},
+	}
+	defer os.Setenv("IMAGE_RESOURCEGROUP", os.Getenv("IMAGE_RESOURCEGROUP"))
```
jim-minter: this is a bad plan - it alters global state which lays us open to a flake some point down the road.  We need to centralise where we're doing os.Getenv()s to get round this.  That's #466 (@thekad fyi)

pkg/config/derive_test.go
```diff
@@ -0,0 +1,298 @@
...
+import (
+	"os"
+	"reflect"
+	"testing"
+
+	acsapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/util/fixtures"
+)
+
+func Test_derived_ImageResourceGroup(t *testing.T) {
```
jim-minter: `_` in go identifiers is considered bad style, please use camel case.

pkg/config/derive_test.go
```diff
@@ -0,0 +1,298 @@
...
+		},
+		{
+			want: "test-image2",
+		},
+	}
+	defer os.Setenv("IMAGE_RESOURCEGROUP", os.Getenv("IMAGE_RESOURCEGROUP"))
+
+	for _, tt := range tests {
+		t.Run(tt.want, func(t *testing.T) {
+			d := derived{}
```
jim-minter: should be able to use the singleton instance Derived here

pkg/config/derive_test.go
jim-minter: Short term could consider passing the env var in as a parameter to enable unit testing

pkg/config/derive_test.go
```diff
@@ -0,0 +1,298 @@
...
+			var cs = fixtures.NewTestOpenShiftCluster()
+			if got := d.RouterLBCNamePrefix(cs); got != tt.want {
+				t.Errorf("derived.RouterLBCNamePrefix() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_derived_PublicHostname(t *testing.T) {
+	tests := []struct {
```
jim-minter: In all of these, I'm a little worried about the load we're putting on NewTestOpenShiftCluster.  Also this isn't really achieving full coverage of the function under test.  Wouldn't something like the following be better:

```
{
	in: OpenShiftManagedCluster{Properties: &Properties{PublicHostname: "foo", FQDN: "bar"}},
	want: "foo",
},
{
	in: OpenShiftManagedCluster{Properties: &Properties{FQDN: "bar"}},
	want: "bar",
},
```

(need to reconsider multiple test functions in this file accordingly)

pkg/config/derive_test.go
kargakis: https://github.com/openshift/openshift-azure/pull/503

pkg/config/derive_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func TestDerivedMasterLBCNamePrefix(t *testing.T) {
+	tests := []struct {
+		cs   api.OpenShiftManagedCluster
+		want string
+	}{
+		{
+			cs: api.OpenShiftManagedCluster{
+				Properties: &api.Properties{FQDN: "bar"},
```
jim-minter: does not adequately test MasterLBCNamePrefix.  You should at least pass in bar.something and check it returns bar.

pkg/config/derive_test.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("derived.CloudProviderConf() = \"%v\", want \"%v\"", string(got), string(tt.want))
+			}
+		})
+	}
+}
+
+func TestDerivedMasterLBCNamePrefix(t *testing.T) {
+	tests := []struct {
```
jim-minter: 20 lines of boilerplate is not appropriate for a function which tests strings.Split, please simplify (comment probably applies to all tests in this file where the test matrix length is 1 and it is highly unlikely to increase)

pkg/config/derive_test.go
```diff
@@ -0,0 +1,219 @@
...
+`),
+		},
+	}
+
+	d := derived{}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := d.CloudProviderConf(&tt.cs)
+			if err != nil {
+				t.Errorf("derived.CloudProviderConf() error = %v", err)
```
jim-minter: t.Error(err) is fine here (check all tests)

pkg/config/derive_test.go
```diff
@@ -0,0 +1,219 @@
...
+			role: acsapi.AgentPoolProfileRoleCompute,
+		},
+		{
+			cs: api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{
+							Name:   "1",
+							Role:   acsapi.AgentPoolProfileRoleCompute,
+							VMSize: "Standard_D2s_v3",
```
jim-minter: test would probably be better if this were StandardD4sV3

pkg/config/derive_test.go
```diff
@@ -0,0 +1,219 @@
...
+			want: "",
+			role: acsapi.AgentPoolProfileRoleMaster,
+		},
+		{
+			cs: api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					AgentPoolProfiles: []api.AgentPoolProfile{},
+				},
+			},
+			want: "",
```
jim-minter: it would be better to change the signature of {System,Kube}Reserved to return (string, error) and return an appropriate error if we're going to be testing this.

asalkeld:
@jim-minter ptal

pkg/config/derive_test.go
```diff
@@ -0,0 +1,190 @@
...
+tenantId: tenant
+vmType: vmss
+`),
+		},
+	}
+
+	for _, tt := range tests {
+		got, err := Derived.CloudProviderConf(&tt.cs)
+		if err != nil {
+			t.Error(err)
```
jim-minter: nit could use t.Fatal(err) here

pkg/config/derive_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}{
+		{
+			cs: api.OpenShiftManagedCluster{
+				Properties: &api.Properties{
+					RouterProfiles: []api.RouterProfile{
+						{
+							FQDN: "one.two.three",
+						},
+						{
+							FQDN: "not.this",
```
jim-minter: (we currently validate against multiple RouterProfiles)

pkg/config/derive_test.go
jim-minter: @asalkeld bump - fixing this would improve the test

pkg/config/derive_test.go
asalkeld: ok, i have removed the second entry now

pkg/config/derive_test.go
asalkeld: done

PR #492, opened 2018-09-20
Add a plugin config and set the Accept-Languages header

### LINKED ISSUE #372: set locale for azure client (additional header)
example: https://github.com/Azure/acs-engine/blob/849e5fd1a26e1dda2e8532afb58123e6049e4638/pkg/armhelpers/azureclient.go#L429

asalkeld:
closes #372


asalkeld:
I haven't caught all the clients, just want to make sure I am on the right track here..

pkg/healthcheck/healthcheck.go
```diff
@@ -38,13 +38,13 @@ type HealthChecker interface {
 	HealthCheck(ctx context.Context, cs *acsapi.OpenShiftManagedCluster) error
 }
 
-type simpleHealthChecker struct{}
+type simpleHealthChecker struct{ pluginConfig *acsapi.PluginConfig }
```
kargakis: If you only need to use `AcceptLanguages`, pass just that instead of the whole PluginConfig.

cmd/createorupdate/createorupdate.go
```diff
@@ -21,15 +21,19 @@ import (
...
 	"github.com/openshift/openshift-azure/pkg/upgrade"
+	"github.com/openshift/openshift-azure/pkg/util"
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
 var logLevel = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
 
 // createOrUpdate simulates the RP
 func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCluster, entry *logrus.Entry) (*v20180930preview.OpenShiftManagedCluster, error) {
+
```
jim-minter: please don't do this

cmd/createorupdate/createorupdate.go
```diff
@@ -21,15 +21,19 @@ import (
...
 	// instantiate the plugin
-	p := plugin.NewPlugin(entry, os.Getenv("SYNC_IMAGE"))
+	var config = &acsapi.PluginConfig{SyncImage: os.Getenv("SYNC_IMAGE"),
+		AcceptLanguages: []string{util.LocaleLangToAcceptLanguage(os.Getenv("LANG"))}}
```
jim-minter: why not just hard code `en-us` here and drop all of LocaleLangToAcceptLanguage?  Does it really matter to us in our scaffolding implementation?

pkg/healthcheck/healthcheck.go
jim-minter: always multiple lines in type definitions

pkg/plugin/plugin_test.go
```diff
@@ -13,7 +13,10 @@ import (
 )
 
 func TestMerge(t *testing.T) {
-	p := NewPlugin(logrus.NewEntry(logrus.New()), "sync:latest")
+
+	var config = &api.PluginConfig{SyncImage: "sync:latest",
+		AcceptLanguages: []string{"en_US.UTF-8"}}
```
jim-minter: presumably this should be en-us here?

pkg/util/locale.go
```diff
@@ -0,0 +1,10 @@
+package util
```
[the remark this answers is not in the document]
jim-minter: agreed, but I don't think we need this helper anyway

pkg/initialize/client.go
```diff
@@ -2,29 +2,50 @@ package initialize
...
 type azureStorageClient struct {
-	accounts storage.AccountsClient
-	storage  azstorage.Client
+	accounts     storage.AccountsClient
+	storage      azstorage.Client
+	pluginConfig *api.PluginConfig
 }
 
-func newAzureClients(ctx context.Context, cs *api.OpenShiftManagedCluster) (*azureStorageClient, error) {
+func (az *azureStorageClient) addAcceptLanguages() autorest.PrepareDecorator {
```
jim-minter: I suspect something like this is going to have to sit somewhere central (invent a pkg/util/azureclient or something) for now; later we probably need to refactor to get our azure clients from one place generally, but that requires more thought

pkg/plugin/plugin.go
```diff
@@ -17,17 +17,18 @@ import (
 )
 
 type plugin struct {
-	entry     *logrus.Entry
-	syncImage string
+	entry        *logrus.Entry
+	pluginConfig *api.PluginConfig
```
jim-minter: if this were not a pointer we could avoid a pile of nil checks - thoughts?

pkg/plugin/plugin.go
```diff
@@ -105,18 +106,18 @@ func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluste
 
 func (p *plugin) InitializeCluster(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
 	log.Info("initializing cluster")
-	initializer := initialize.NewSimpleInitializer(p.entry)
+	initializer := initialize.NewSimpleInitializer(p.entry, p.pluginConfig)
```
jim-minter: am wondering whether we should just be passing `p` in each of these things and exporting the entry and config fields

pkg/plugin/plugin.go
jim-minter: plugin.pluginConfig stutters - plugin.config better

cmd/createorupdate/createorupdate.go
asalkeld: ok, it will make it simpler.

pkg/plugin/plugin_test.go
asalkeld: yes

pkg/plugin/plugin.go
asalkeld: yes, that would make it neater

asalkeld: I can't do this because we start getting import cycles

cmd/createorupdate/createorupdate.go
```diff
@@ -29,7 +28,9 @@ var logLevel = flag.String("loglevel", "Debug", "valid values are Debug, Info, W
 // createOrUpdate simulates the RP
 func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCluster, entry *logrus.Entry) (*v20180930preview.OpenShiftManagedCluster, error) {
 	// instantiate the plugin
-	p := plugin.NewPlugin(entry, os.Getenv("SYNC_IMAGE"))
+	var config = acsapi.PluginConfig{SyncImage: os.Getenv("SYNC_IMAGE"),
```
kargakis: Change this to be an input to `createOrUpdate`. It will be useful once I move `createOrUpdate` in its own library and reuse in our e2e framework.

pkg/upgrade/update.go
```diff
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
```
kargakis: should be grouped with the rest of the openshift-azure imports

pkg/util/azureclient/azureclient.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2018-02-01/storage"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+// AzureClients holds all the used clients used by openshift-azure
+type AzureClients struct {
```
kargakis: Doing this for our plugin code is probably fine since all plugins are used by a single codebase but I have resisted the urge for having a single struct that holds all of the clients we use. A single import of this in `sync` is going to pull with it a lot of unnecessary code.

pkg/util/azureclient/azureclient.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/marketplaceordering/mgmt/marketplaceordering"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	azmarketplaceordering "github.com/Azure/azure-sdk-for-go/services/marketplaceordering/mgmt/2015-06-01/marketplaceordering"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2018-02-01/storage"
+
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
```
kargakis: We probably need to start a document with the conventions we are following in our codebase (@jim-minter). Re imports we use the following convention:
```go
import (
  "stdlib imports"
  // empty line
  "third party imports"
  // empty line
  "same repo (openshift-azure) imports"
)
```
If we want to start viewing import graphs we will need to use `go list` but this conventions makes the code more readable and our packages are small enough to allow some utility out of this (you can quickly figure out third part deps in a package w/o using `go` magic).

pkg/api/plugin.go
```diff
@@ -41,4 +47,6 @@ type Plugin interface {
 	HealthCheck(ctx context.Context, cs *OpenShiftManagedCluster) error
 
 	Update(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte) error
+
+	GetPluginConfig() PluginConfig
```
kargakis: I don't think exposing this in the plugin interface is necessary. Both `Deploy` and `Update` can accept `PluginConfig` via an argument provided by our fake RP implementation.

pkg/util/azureclient/azureclient.go
asalkeld: @kargakis it was not my intention to use this in the sync code, just where it made sense.

pkg/api/plugin.go
asalkeld: @kargakis the problem is that we call Initialzer with a plugin instance in createorupdate. So if I remove this api I get the following error:

# github.com/openshift/openshift-azure/cmd/createorupdate
cmd/createorupdate/createorupdate.go:117:23: cannot use p (type "github.com/openshift/openshift-azure/pkg/api".Plugin) as type initialize.Initializer in argument to upgrade.Deploy:
	"github.com/openshift/openshift-azure/pkg/api".Plugin does not implement initialize.Initializer (missing GetPluginConfig method)
make: *** [Makefile:5: build] Error 2

asalkeld: don't worry, figured out what you meant now.. fixing..

pkg/plugin/plugin.go
```diff
@@ -105,18 +106,18 @@ func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluste
...
 
 func (p *plugin) HealthCheck(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
 	log.Info("starting health check")
-	healthChecker := healthcheck.NewSimpleHealthChecker(p.entry)
+	healthChecker := healthcheck.NewSimpleHealthChecker(p.entry, p.config)
 	return healthChecker.HealthCheck(ctx, cs)
 }
 
-func (p *plugin) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte) error {
+func (p *plugin) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte, config api.PluginConfig) error {
```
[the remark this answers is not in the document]
asalkeld: It gets messy because of the initializer. My initial plan was to add a Plugin api GetPluginConfig() to the Plugin and the initilizer, but @kargakis asked me to just add an arguement to this function.

pkg/plugin/plugin.go
```diff
@@ -91,8 +92,8 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
 	if err != nil {
 		return err
 	}
-	if p.syncImage != "" {
-		cs.Config.Images.Sync = p.syncImage
+	if p.config.SyncImage != "" {
```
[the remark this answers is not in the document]
asalkeld: ok

pkg/plugin/plugin.go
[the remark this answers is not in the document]
jim-minter: @asalkeld why can't you do https://gist.github.com/jim-minter/d1a266c870797a2b94a826b3218fdd7a ?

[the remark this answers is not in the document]
asalkeld: Have a look at this commit https://github.com/openshift/openshift-azure/pull/492/commits/514e71bf2a386538929d64f325273c2315db1ffc
Imagine the reverse of that.
The real issue is the Initializer. We call Deploy and Update with either a plugin or an initilizer instance. So your example is one of the easiest cases, when we just have the api instance then we don't have access to the private "config" so I initially wrote an accessor GetPluginConfig() that I placed on the Initializer, Updater and Plugin so you could safely call it anywhere.
see: https://github.com/openshift/openshift-azure/pull/492/commits/514e71bf2a386538929d64f325273c2315db1ffc#diff-c5388788ebacab203d51d71bbf0abf3bL22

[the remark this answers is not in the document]
jim-minter: I'm still not getting it :(  Got a sec to catch up on bluejeans in the next few hours?

PR #493, opened 2018-09-20
Tag resource groups with a TTL

kargakis:
@mjudeikis @jim-minter this should allow us to use a tight TTL for resource groups created in CI tests.

/cc mjudeikis


kargakis:
Opened https://github.com/openshift/azure-misc/pull/57 in azure-misc. One more PR in release needed to set a tighter TTL.

PR #494, opened 2018-09-20
add toLowerCamelCase function + unit tests

charlesakalugwu:
In preparation for json struct tags conformance tests... see #386

### DIFF
--- a/pkg/api/config_test.go
+++ b/pkg/api/config_test.go
@@ -21,11 +21,11 @@ func TestPadNumbers(t *testing.T) {
 		input    string
 		expected string
 	}{
-		{input: "single123number", expected: "single 123 number",},
-		{input: "single123numbers345spread", expected: "single 123 numbers 345 spread",},
-		{input: "number_at_end234", expected: "number_at_end 234 ",},
-		{input: "123number_at_start", expected: " 123 number_at_start",},
-		{input: "123number_at_start", expected: " 123 number_at_start",},
+		{input: "single123number", expected: "single 123 number"},
+		{input: "single123numbers345spread", expected: "single 123 numbers 345 spread"},
+		{input: "number_at_end234", expected: "number_at_end 234 "},
+		{input: "123number_at_start", expected: " 123 number_at_start"},
+		{input: "123number_at_start", expected: " 123 number_at_start"},
 	}
 
 	for _, test := range testcases {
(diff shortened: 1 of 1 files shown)


pkg/api/config_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	for _, test := range testcases {
+		got := padNumbers(test.input)
+		if got != test.expected {
+			t.Errorf("padNumbers(%v) != %v, expected %v", test.input, got, test.expected)
+		}
+	}
+}
+
+// Converts a string to lower camel case
```
asalkeld: wouldn't it be a good idea to move the functional code out into a separate file from the start?

pkg/api/config_test.go
```diff
@@ -0,0 +1,108 @@
...
+	prep := s
+	start, rest := s[0], s[1:]
+	if start >= 'A' && start <= 'Z' {
+		prep = strings.ToLower(string(start)) + rest
+	}
+	//  pad all numbers within the string with spaces
+	prep = padNumbers(prep)
+
+	// trim spaces at the edges of the string
+	prep = strings.Trim(prep, " ")
```
asalkeld: shouldn't this be a bit up on line 47? to catch the case " ABook"

pkg/api/config_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return res
+}
+
+func TestLowerCamelCase(t *testing.T) {
+
+	testcases := []struct {
+		input    string
+		expected string
+	}{
```
asalkeld: maybe add some test cases that start with  a space

kargakis:
> In preparation for json struct tags conformance tests... see #386

It's not clear from that issue what's the purpose of this PR. Maybe it would be clearer to try and use the functions you are unit-testing in prod code.

/hold
/cc jim-minter

pkg/api/config_test.go
charlesakalugwu: this function will only be used from another testcase which will be within this file. That is why this is currently just in config_test.go. already discussed this with @jim-minter

pkg/api/config_test.go
charlesakalugwu: done :)

pkg/api/config_test.go
charlesakalugwu: good catch! thanks. added more testcases for this condition.

charlesakalugwu:
@kargakis we would like to be able to walk through arbitrary structs in our project and verify that the json struct tags added to them conform to a standard that we set. That standard in our case is currently that all struct tags be set to the lower camel case format of the corresponding struct field. 

I am working on the next step, a function `jstCheck(o interface{}, f func(s string) string) []error` where f will be the `lowerCamelCase(string) string` function created in this pull request.

charlesakalugwu:
@asalkeld the work is much more than just having a lower camel case function. Please see #505. I will close this PR as the more complete solution is in #505

PR #496, opened 2018-09-20
[wip] externalize template 

mjudeikis:
externalize template so we can reuse it to template other config files, not only add-ons.
potentially `TemplateJson`, `TemplateIni`, etc

@pweil- @kargakis @jim-minter do you think its worth pursuing? 

Ignore random development code. Would like to know generic opinion about the approach.


jim-minter:
let's talk it through - I don't immediately see the rationale for this.

mjudeikis:
This will give us the ability to scale/rotate master, control scheduler, master config options, configure multiple auth providers easier, admission plugin, etc etc.. Might be little bit too early but I think we will need this.

PR #497, opened 2018-09-20
[WIP] Rewrite fake RP to be a library and use http

kargakis:
Still not ready but pushing this out for early feedback

TODO
- [ ] All the environment variables used in `pkg/fakerp/createorupdate.go` need to be read from the request context or passed down to `createOrUpdate` via other means.
- [ ] writting down files probably needs to be reworked? maybe not
- [x] Expose createOrUpdate to be used by the e2e framework

/cc jim-minter

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -1,276 +1,174 @@
 package main
 
 import (
+	"bytes"
 	"context"
-	"flag"
-	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 
-	"github.com/Azure/azure-sdk-for-go/services/marketplaceordering/mgmt/2015-06-01/marketplaceordering"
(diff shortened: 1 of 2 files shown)


kargakis:
>     * All the environment variables used in `pkg/fakerp/createorupdate.go` need to be read from the request context or passed down to `createOrUpdate` via other means.

Now I think https://github.com/openshift/openshift-azure/issues/466 is a prerequisite

kargakis:
@jim-minter @cjryan  I have started thinking that the e2e framework shouldn't worry about standing up a cluster and should only be responsible to run e2e tests given a kubeconfig. I don't think the cluster standup is going to be used outside of our CI (eg. MSFT, users) and even in our CI we may not use it and continue with our current scripts that we also use for development. Anyway, this PR is kind of blocked so I am going to close it for now.

PR #505, opened 2018-09-21
Create JSON struct tags conformance check function

### LINKED ISSUE #386: add `json` tags to Config struct members

charlesakalugwu:
follow up to work done in #494, closes #386


charlesakalugwu:
@jim-minter please lets sync up today to discuss the last steps required to get this merged.

pkg/api/config_test.go
```diff
@@ -0,0 +1,336 @@
+package api_test
```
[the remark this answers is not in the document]
charlesakalugwu: `package api_test` is also not wrong. this is best practice for testing to avoid cyclic dependencies and in fact many of the standard library tests use this pattern. e.g.https://golang.org/src/math/all_test.go

pkg/api/config_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"github.com/openshift/openshift-azure/pkg/api"
```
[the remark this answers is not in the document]
charlesakalugwu: point 1) i've rearranged the imports with goimport. 

point 2) this is not a problem if you use `package api_test` above

pkg/api/config_test.go
```diff
@@ -0,0 +1,336 @@
...
+			check: lowerCamelCase,
+			input: struct {
+				Name  string `json:"name,omitempty"`
+				Value string
+			}{
+				Name:  "name",
+				Value: "value",
+			},
+			expected: []error{
+				errors.New(fmt.Sprintf(`field "Value" does not have a json tag`)),
```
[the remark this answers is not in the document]
kargakis: fmt.Errorf instead of errors.New+fmt.Sprintf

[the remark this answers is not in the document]
kargakis: Actually you don't seem to need formatting so just errors.New?

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+		{
+			name:  "input is a struct but doesn't specify any json tags",
+			check: lowerCamelCase,
+			input: struct {
+				Name  string
+				Value string
+			}{},
+			expected: []error{
+				errors.New(fmt.Sprintf(`field "Name" does not have a json tag`)),
```
kargakis: no need for Sprintf. below likewise

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+		} else {
+			upper = false
+		}
+	}
+
+	return res
+}
+
+func TestLowerCamelCase(t *testing.T) {
+
```
kargakis: nit: remove

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+			name:     "input is the api.ImageConfig struct",
+			check:    lowerCamelCase,
+			input:    api.ImageConfig{},
+			expected: []error{},
+		},
+		{
+			name:     "input is the api.Config struct",
+			check:    lowerCamelCase,
+			input:    api.Config{},
+			expected: []error{},
```
kargakis: nit: this should be nil

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+	}
+
+	for _, test := range testcases {
+		got := jstCheck(test.input, test.check)
+		if !reflect.DeepEqual(got, test.expected) {
+			t.Errorf(`"%s" testcase expected errors %v but received %v`, test.name, test.expected, got)
```
kargakis: no need for backticks, use %q for quoting strings inside strings

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+}
+
+func jstCheck(o interface{}, check func(s string) string) []error {
+
```
kargakis: nit: remove

pkg/api/config_test.go
```diff
@@ -0,0 +1,334 @@
...
+	errs := []error{}
```
kargakis: var errs []error

is more idiomatic to define a slice

pkg/api/config_test.go
mjudeikis: ` -> "

PR #508, opened 2018-09-24
add basic configuration to plugin

### LINKED ISSUE #332: Add basic configuration to plugin
MSFT want to be able to change at least the VM image the plugin chooses to use without revendoring the plugin.  Need a mechanism to allow this to be specified, either via the plugin reading a config file, or by accepting configuration at plugin instantiation time.

charlesakalugwu:
closes #332

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -26,16 +26,8 @@ import (
 
 var logLevel = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
 
-var pc *plugin.Config
-
 // createOrUpdate simulates the RP
-func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCluster, entry *logrus.Entry) (*v20180930preview.OpenShiftManagedCluster, error) {
-	// initialize a new plugin config
-	pc = &plugin.Config{
-		NodeImage: os.Getenv("NODE_IMAGE"),
-		SyncImage: os.Getenv("SYNC_IMAGE"),
-	}
-
+func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCluster, entry *logrus.Entry, pc *plugin.Config) (*v20180930preview.OpenShiftManagedCluster, error) {
 	// instantiate the plugin
 	p := plugin.NewPlugin(entry, pc)
 
@@ -259,6 +251,12 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// initialize a new plugin config
+	pc := &plugin.Config{
+		VmImage:   os.Getenv("VM_IMAGE"),
+		SyncImage: os.Getenv("SYNC_IMAGE"),
+	}
+
 	//simulate Context with property bag
 	ctx := context.Background()
 	ctx = context.WithValue(ctx, api.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
@@ -267,7 +265,7 @@ func main() {
 
 	// simulate the API call to the RP
 	entry := logrus.NewEntry(logger).WithFields(logrus.Fields{"resourceGroup": os.Getenv("RESOURCEGROUP")})
-	oc, err = createOrUpdate(ctx, oc, entry)
+	oc, err = createOrUpdate(ctx, oc, entry, pc)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


charlesakalugwu:
@jim-minter @kargakis @mjudeikis is this going in the right direction? We could in addition wire in the viper package to add support for plugin config overrides via a combination of config file, environment variables (e.g. prefixed by OPENSHIFT_AZURE_PLUGIN_) and command line flags.

pkg/plugin/config.go
```diff
@@ -0,0 +1,10 @@
+package plugin
+
+// PluginConfig contains dynamic plugin configuration
+type Config struct {
+	// The sync image to use
+	SyncImage string `json:"syncImage,omitempty"`
+
+	// The node image to use
+	NodeImage string `json:"nodeImage,omitempty"`
```
kargakis: I would suggest `vmImage` or `machineImage`. `nodeImage` is confusing even though we don't run kubelet inside a container (and there is also `origin-node`).

cmd/createorupdate/createorupdate.go
```diff
@@ -26,10 +26,18 @@ import (
 
 var logLevel = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
 
+var pc *plugin.Config
+
 // createOrUpdate simulates the RP
 func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCluster, entry *logrus.Entry) (*v20180930preview.OpenShiftManagedCluster, error) {
```
kargakis: Can you move plugin.Config to be passed as an argument in `createOrUpdate`? This is going to be useful once we share `createOrUpdate` between this code and the e2e framework.

kargakis:
> We could in addition wire in the viper package to add support for plugin config overrides via a combination of config file, environment variables (e.g. prefixed by OPENSHIFT_AZURE_PLUGIN_) and command line flags.

Since the viper addition wouldn't actually be a part of the plugin code, I don't see any huge utility in bringing it in. Our e2e framework should reuse `createOrUpdate` so we are going to instantiate clusters in CI by passing pluginConfig. Where do you expect config overrides to be useful?

pkg/plugin/config.go
charlesakalugwu: changed to `VmImage`

cmd/createorupdate/createorupdate.go
charlesakalugwu: done

charlesakalugwu:
> > We could in addition wire in the viper package to add support for plugin config overrides via a combination of config file, environment variables (e.g. prefixed by OPENSHIFT_AZURE_PLUGIN_) and command line flags.
> 
> Since the viper addition wouldn't actually be a part of the plugin code, I don't see any huge utility in bringing it in. Our e2e framework should reuse `createOrUpdate` so we are going to instantiate clusters in CI by passing pluginConfig. Where do you expect config overrides to be useful?

we could, for example, have a pipeline where all stages are configured with a default config file representing pluginConfig but having arbitrary individual stages able to have plugin.config.vmImage passed via flag as --vm-image and having the passed in value take precedence over the value in the default config file. Or being able to set `OPENSHIFT_AZURE_VM_IMAGE` from another stage...It was just a thought to enable super flexible 12-factor-style config in our project. It is however truly not necessary at this point.

kargakis:
> but having arbitrary individual stages

Can you elaborate more with what you mean when you are saying "stages"? Are these stages during the install/scale/upgrade of a cluster or something else? The plugin config is supposed to hold configurables for a single request which means that during an install/scale/upgrade of a cluster, it's supposed to be the same all along the request (so the concept of "stages" does not make much sense in those scenarios)

pweil-:
is the idea here that we may need to give MSFT the flexibility to change anything we're defaulting in the plugin but not exposing in the API?  

I see these options:

1.  expose things in the api
2.  create a new configuration input that overwrites the fields (multiple impl options here)
3.  have them set the environment variables already in use for images and see what creeps up next - aka procrastinate 

I'd suggest option 2 requires a discussion in our arch call before implementation.  In any case I think we should prioritize testing and SRE tooling over this right now unless it's blocking something.

asalkeld:
also see https://github.com/openshift/openshift-azure/pull/492

kargakis:
/hold

until this is discussed in the next arch call as per @pweil-'s request

charlesakalugwu:
closing this and adding the work after #522 is merged since that is a basis for this work

Issue #400, opened 2018-09-11
consolidate clients into pkg


mjudeikis:
we starting to reuse the same client code base in a few places.  Without this our `e2e` tests will start duplicating client code in one more place

Suggestion: *Consolidate all clients under package `clients`* 
```
//reads local file config for testing and fake RP
clients.ReadConfig(path) (*api.OpenShiftManagedCluster, error)

//helps to coverts config file into context for testing fake RP
clients.ConfigToContext (cs *api.OpenShiftManagedCluster) (context.Context, error)

//returns kubeClient set for fake rp and testing and healthcheck, upgrade
clients.KubeClientsetFromConfig(cs *api.OpenShiftManagedCluster) (*kubernetes.Clientset, error) 

// azure storage clients for testing and initialization of the cluster. In the future etcd restore automation (?) 
clients.AzureStorageClientset(ctx context.Context, cs *api.OpenShiftManagedCluster) (*azureStorageClient, error)

// possible future azure clients extentions 
clients.AzureNetoworkClientset(ctx context.Context, cs *api.OpenShiftManagedCluster) (*azureNetworkClient, error) 

// for test and healthcheck. This might even be removed and replaces by one of the methods above
clients.KubeGetConfigFromV1Config(kc *v1.Config) (clientcmd.ClientConfig, error) 
```

WDYT @openshift/sig-azure  ?

kargakis:
I don't feel strong about creating a generic clients package that is basically a wrapper for the azure sdk and client-go. I am in favor of using the clients directly. client-go exposes interfaces and has fake implementations, that's the only issue with the azure sdk so for that we will need to create our own interfaces and fake implementations for unit-testability.

mjudeikis:
So maybe just for azure?

mjudeikis:
for future extensibility: https://github.com/Azure/azure-sdk-for-go/issues/409 
https://github.com/kubernetes/kubernetes/blob/34001d8c6a0ed4edf02d91c5b45920a2676418f9/pkg/cloudprovider/providers/azure/azure_client.go

mjudeikis:
close in favor of:  AZURE-148 - Consolidate clients pkg

PR #509, opened 2018-09-24
[wip] azure client consolidation

mjudeikis:
This would consolidate all our azure clients under our interfaces. 
Very similar as done here:
https://github.com/kubernetes/kubernetes/blob/34001d8c6a0ed4edf02d91c5b45920a2676418f9/pkg/cloudprovider/providers/azure/azure_client.go#L773-L814

There are still few places left in upgrade flow to remove.

This would allow testing of those interfaces and reusing same code in tests. For example, etcd backup or cluster init tests would need access to Azure blob client. Which means code duplication again. 

https://github.com/openshift/openshift-azure/issues/400

@jim-minter @kargakis @pweil- wdyt? before I jump to finish this one.


kargakis:
There is some work that starts on this in https://github.com/openshift/openshift-azure/pull/492

I would ensure no unnecessary client code gets pulled in sync or in the plugin code.

mjudeikis:
> I would ensure no unnecessary client code gets pulled in sync or in the plugin code.

For this I was thinking to keep interfaces and struct separate so if we modify sync pod we add only what is necessary.

mjudeikis:
close in favor: AZURE-148 - Consolidate clients pkg

PR #512, opened 2018-09-24
ssh add ability to pass command

mjudeikis:
so we could disable sync pod using this in ci as non-interactive mode


mjudeikis:
@jim-minter this will enable us to disable sync pod in CI:
` ./hack/ssh.sh -c "sudo rm /etc/origin/node/pods/sync.yaml" ${INSTANCE_PREFIX}" rg`

PR #515, opened 2018-09-25
Makefile updates

kargakis:
* avoid deleting binaries on every image build
* add targets for pushing e2e image in quay
* cleanup logbrdige binary


kargakis:
Necessary for building the e2e image in CI

PR #517, opened 2018-09-25
[WIP] move from os.Getenv to a plugin.Config struct

thekad:
This still doesn't work. Using env tags experimentally.


jim-minter:
something has gone wrong with this PR - failed rebase? it also looks like it's trying to be too smart - ping me in the morning and we can catch up on it

thekad:
Definitely botched rebase. I rebased my changes against master and they applied cleanly, didn't notice the mess until I saw github's "your diff is too large" sigh.

PR #519, opened 2018-09-25
regen certs on update

mjudeikis:
Regenerates certificate if dependency property changed.

### DIFF
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"reflect"
 
 	"github.com/satori/go.uuid"
 	"golang.org/x/crypto/bcrypt"
@@ -14,6 +15,18 @@ import (
 	api "github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/tls"
 )
+
+type certificate struct {
+	cn           string
+	organization []string
+	dnsNames     []string
+	ipAddresses  []net.IP
+	extKeyUsage  []x509.ExtKeyUsage
+	signingKey   *rsa.PrivateKey
+	signingCert  *x509.Certificate
+	key          **rsa.PrivateKey
+	cert         **x509.Certificate
+}
 
 func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (err error) {
 	c := cs.Config
@@ -65,20 +78,9 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 		}
 	}
 
-	certs := []struct {
-		cn           string
-		organization []string
-		dnsNames     []string
-		ipAddresses  []net.IP
-		extKeyUsage  []x509.ExtKeyUsage
-		signingKey   *rsa.PrivateKey
-		signingCert  *x509.Certificate
-		key          **rsa.PrivateKey
-		cert         **x509.Certificate
-		selfSign     bool
-	}{
-		// Generate etcd certs
+	certs := []certificate{
 		{
+			// Generate etcd certs
 			cn:          "etcd-server",
 			dnsNames:    []string{"master-000000", "master-000001", "master-000002"},
 			extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
@@ -227,12 +229,17 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 		if cert.signingKey == nil && cert.signingCert == nil {
 			cert.signingKey, cert.signingCert = c.Certificates.Ca.Key, c.Certificates.Ca.Cert
 		}
-		if *cert.key != nil && *cert.cert != nil &&
-			((*cert.cert).CheckSignatureFrom(cert.signingCert) == nil || cert.selfSign) {
-			continue
-		}
-		if *cert.key, *cert.cert, err = tls.NewCert(cert.cn, cert.organization, cert.dnsNames, cert.ipAddresses, cert.extKeyUsage, cert.signingKey, cert.signingCert, cert.selfSign); err != nil {
+		var k *rsa.PrivateKey
+		var c *x509.Certificate
+		if k, c, err = tls.NewCert(cert.cn, cert.organization, cert.dnsNames, cert.ipAddresses, cert.extKeyUsage, cert.signingKey, cert.signingCert, false); err != nil {
 			return
+		}
+		// compare certificate and replace if update is needed
+		// if at any point we start using self-sign certificates again
+		// (see console certificate comment), logic inside certEqual needs
+		// to be updated for this.
+		if !certEqual(c, *cert.cert) {
+			*cert.key, *cert.cert = k, c
 		}
 	}
 
@@ -387,4 +394,18 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 	}
 
 	return
+}
+
+func certEqual(certNew, certOld *x509.Certificate) bool {
+	if certOld == nil {
+		return false
+	}
+	if !reflect.DeepEqual(certNew.Subject, certOld.Subject) ||
+		!reflect.DeepEqual(certNew.DNSNames, certOld.DNSNames) ||
+		!reflect.DeepEqual(certNew.ExtKeyUsage, certOld.ExtKeyUsage) ||
+		!reflect.DeepEqual(certNew.IPAddresses, certOld.IPAddresses) ||
+		!reflect.DeepEqual(certNew.Issuer, certOld.Issuer) {
+		return false
+	}
+	return true
 }
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -26,6 +26,7 @@ type certificate struct {
 	signingCert  *x509.Certificate
 	key          **rsa.PrivateKey
 	cert         **x509.Certificate
+	selfSign     bool
 }
 
 func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (err error) {
@@ -229,7 +230,9 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 		if cert.signingKey == nil && cert.signingCert == nil {
 			cert.signingKey, cert.signingCert = c.Certificates.Ca.Key, c.Certificates.Ca.Cert
 		}
-		if !regenerate(cert) {
+		// if we start using self-singed certificates agaig (see console certificate comment)
+		// logic in this function needs to be updated
+		if needsGenerate(cert) {
 			if *cert.key, *cert.cert, err = tls.NewCert(cert.cn, cert.organization, cert.dnsNames, cert.ipAddresses, cert.extKeyUsage, cert.signingKey, cert.signingCert, false); err != nil {
 				return
 			}
@@ -389,17 +392,42 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 	return
 }
 
-func regenerate(cert certificate) bool {
+func needsGenerate(cert certificate) bool {
 	crt := *cert.cert
+	// it it is 1st run crt will be empty - generate
 	if crt == nil {
-		return false
+		return true
 	}
 
+	// if it is n+1 run, and if signingCert signature don't matche - generate
+	if crt.CheckSignatureFrom(cert.signingCert) != nil {
+		return true
+	}
+
+	// if main attributes do not match - regenerate
+	// net.IP does not comply with definition of deep equality.
+	// we use native Equal method within net package
 	if !reflect.DeepEqual(cert.dnsNames, crt.DNSNames) ||
 		!reflect.DeepEqual(cert.extKeyUsage, crt.ExtKeyUsage) ||
 		!reflect.DeepEqual(cert.cn, crt.Subject.CommonName) ||
-		!reflect.DeepEqual(cert.ipAddresses, crt.IPAddresses) {
+		!reflect.DeepEqual(cert.organization, crt.Subject.Organization) ||
+		!ipAddressEqual(cert.ipAddresses, crt.IPAddresses) {
+		return true
+	}
+	return false
+}
+
+func ipAddressEqual(ip1, ip2 []net.IP) bool {
+	eq := false
+	if len(ip1) != len(ip2) {
 		return false
 	}
-	return true
+	for _, ip1v := range ip1 {
+		for _, ip2v := range ip2 {
+			if ip1v.Equal(ip2v) {
+				eq = true
+			}
+		}
+	}
+	return eq
 }
--- a/pkg/config/generate.go
+++ b/pkg/config/generate.go
@@ -229,17 +229,10 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 		if cert.signingKey == nil && cert.signingCert == nil {
 			cert.signingKey, cert.signingCert = c.Certificates.Ca.Key, c.Certificates.Ca.Cert
 		}
-		var k *rsa.PrivateKey
-		var c *x509.Certificate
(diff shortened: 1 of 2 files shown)


pkg/api/plugin.go
```diff
@@ -32,7 +32,7 @@ type Plugin interface {
 
 	// GenerateConfig ensures all the necessary in-cluster config is generated
 	// for an Openshift cluster.
-	GenerateConfig(ctx context.Context, cs *OpenShiftManagedCluster) error
+	GenerateConfig(ctx context.Context, cs, oldCS *OpenShiftManagedCluster) error
```
jim-minter: doing this is unfortunately out of the question.  oldCS is not available on the RP backend.  You don't need it anyway.

pkg/config/generate.go
```diff
@@ -227,11 +230,14 @@ func Generate(cs *api.OpenShiftManagedCluster) (err error) {
...
-		if *cert.key != nil && *cert.cert != nil &&
-			((*cert.cert).CheckSignatureFrom(cert.signingCert) == nil || cert.selfSign) {
+		// if regeneration is not needed - skip
+		if !regenerate(cs, oldCs, cert) {
+			fmt.Printf("skip %v\n", cert.cn)
 			continue
 		}
+		fmt.Printf("gen %v\n", cert.cn)
 		if *cert.key, *cert.cert, err = tls.NewCert(cert.cn, cert.organization, cert.dnsNames, cert.ipAddresses, cert.extKeyUsage, cert.signingKey, cert.signingCert, cert.selfSign); err != nil {
+			fmt.Println("bumot error")
```
jim-minter: that's the worst kind of error.

pkg/config/generate.go
```diff
@@ -388,3 +394,40 @@ func Generate(cs *api.OpenShiftManagedCluster) (err error) {
 
 	return
 }
+
+// regenerate checks if certificate needs to be regenerated
+func regenerate(cs, oldCs *api.OpenShiftManagedCluster, cert certificate) bool {
```
jim-minter: this function needs to check the inputs against the pre-existing certificate in cs.Config to determine if a regeneration is needed.

pkg/config/generate.go
mjudeikis: sorry, debug message :D

pkg/config/generate.go
mjudeikis: This is where the challenge comes. My certificate might be for `old.example.com` but my `cs` is coming from `newawesomedomain.com`. So without `oldCs` im not able to check which certificate depends to where because values might be totally different. 
Alternatives are to keep some kind dependency between certificates by adding `type` to the certificate and mapping `type=masterFrontEnd` needs regeneration if `PublicHostname` or `FQDN` changes.

pkg/api/plugin.go
mjudeikis: See last comment

pkg/config/generate.go
mjudeikis: If you can think of any other way, let me know.

kargakis: Is VerifyHostname what you need here?

https://golang.org/pkg/crypto/x509/#Certificate.VerifyHostname

mjudeikis: The thing is I dont have a reference point to what to compare it. Without oldCS I need to know which cert uses which Properties field. In the example, routerPublicHostname is very similar to API FQDN, but they are not related. So verifyHostname will not give me this relationship.

kargakis: At this point all the hostnames in cs are updated to new, so if a certificate does not match **any** hostname then you probably want to regenerate it, right?

mjudeikis: This is just one case. In example:
```
 {
            cn: cs.Properties.RouterProfiles[0].PublicSubdomain,
            dnsNames: []string{
                cs.Properties.RouterProfiles[0].PublicSubdomain,
                "*." + cs.Properties.RouterProfiles[0].PublicSubdomain,
            },
            extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
            key:         &c.Certificates.Router.Key,
            cert:        &c.Certificates.Router.Cert,
        },
        {
            cn: "docker-registry-default." + cs.Properties.RouterProfiles[0].PublicSubdomain,
            dnsNames: []string{
                "docker-registry-default." + cs.Properties.RouterProfiles[0].PublicSubdomain,
                "docker-registry.default.svc",
                "docker-registry.default.svc.cluster.local",
            },
            extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
            key:         &c.Certificates.Registry.Key,
            cert:        &c.Certificates.Registry.Cert,
        },
        {
            cn: cs.Properties.FQDN,
            dnsNames: []string{
                cs.Properties.FQDN,
                "master-000000",
                "master-000001",
                "master-000002",
                "kubernetes",
                "kubernetes.default",
                "kubernetes.default.svc",
                "kubernetes.default.svc.cluster.local",
            },
            ipAddresses: []net.IP{net.ParseIP("172.30.0.1")},
            extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
            key:         &c.Certificates.MasterServer.Key,
            cert:        &c.Certificates.MasterServer.Cert,
            
        }, 
```
certificates does not have purpose/type. Just fields If I change Router FQDN to totally different value, how do I know that I need to regenerate router certificate, and not Master or Service Catalog? They are not related, but without mapping keeping somewhere just matching is not enough.

kargakis: Because master and service catalog are still going to match with the other hostname.

mjudeikis: So you suggest to take app properties values, iterate over small subset of certificates (the ones which depend on those properties) and if none of them matches properties field, regenerate it. If match found - skip

kargakis: I didn't suggest that but it's also an option to group certificates per "parent" hostname. I suggested that given a certificate, you can run VerifyHostname on it against all available hostnames (publicHostname, routerPublicHostname). If the cert cannot verify against any hostname, you want to regenerate it.

jim-minter: @mjudeikis you need to write a function which compares the current certificate's fields to the cn, organization, dnsNames, ipAddresses, extKeyUsage desired state.  If they don't match, regenerate.  We already do this to determine if the CA has changed and we need to regenerate, so what's left is the easier bit?

mjudeikis:
/hold
needs more testing as its sensitive change before full PP

mjudeikis:
I would like to add more tests to this, but not sure how. Any ideas?

mjudeikis:
@jim-minter @kargakis ptal

pkg/config/generate_test.go
```diff
@@ -112,3 +113,54 @@ func testRequiredFields(cs *api.OpenShiftManagedCluster, t *testing.T) {
 	assert(c.NodeBootstrapKubeconfig != nil, "NodeBootstrapKubeconfig")
 	assert(c.AzureClusterReaderKubeconfig != nil, "AzureClusterReaderKubeconfig")
 }
+
+func TestGenerateUpdateCertRegen(t *testing.T) {
+	var pluginConfig api.PluginConfig
+	var cs, oldCs *api.OpenShiftManagedCluster
+	cs = fixtures.NewTestOpenShiftCluster()
+	cs.Properties.FQDN = "example-new.eastus.cloudapp.azure.com"
```
kargakis: Can you move all of these where you start using cs? It's confusing to have them here.
In general I would try to factor down the loc
```go
	oldCs := fixtures.NewTestOpenShiftCluster()
	err := Generate(oldCs, api.PluginConfig{})
	if err != nil {
		t.Errorf("received generation error %v", err)
	}

	cs := fixtures.NewTestOpenShiftCluster()
	cs.Properties.FQDN = "example-new.eastus.cloudapp.azure.com"
	cs.Properties.RouterProfiles[0].FQDN = "router-fqdn-new.eastus.cloudapp.azure.com"
	cs.Properties.RouterProfiles[0].PublicSubdomain = "test-new.example.com"

	// copy config and regenerate everything else
	old := oldCs.DeepCopy()
	cs.Config = old.Config
	err = Generate(cs, pluginConfig)
	if err != nil {
		t.Errorf("received generation error %v", err)
	}

        // certificates should not match
	if reflect.DeepEqual(cs.Config.Certificates.MasterServer.Cert, oldCs.Config.Certificates.MasterServer.Cert) {
		t.Error("masterServer certificates matches, check test for details")
	}
	if reflect.DeepEqual(cs.Config.Certificates.Router.Cert, oldCs.Config.Certificates.Router.Cert) {
		t.Error("router certificates matches, check test for details")
	}
	if reflect.DeepEqual(cs.Config.Certificates.Registry.Cert, oldCs.Config.Certificates.Registry.Cert) {
		t.Error("registry certificates matches, check test for details")
	}
	if reflect.DeepEqual(cs.Config.Certificates.OpenshiftConsole.Cert, oldCs.Config.Certificates.OpenshiftConsole.Cert) {
		t.Error("openshiftConsole certificates matches, check test for details")
	}
	//certificates should match
	if reflect.DeepEqual(cs.Config.Certificates.Registry.Cert, oldCs.Config.Certificates.Registry.Cert) {
		t.Error("registry certificates matches, check test for details")
	}
	if !reflect.DeepEqual(cs.Config.Certificates.ServiceCatalogAPIClient.Cert, oldCs.Config.Certificates.ServiceCatalogAPIClient.Cert) {
		t.Error("serviceCatalogAPIClient certificates do not match, check test for details")
	}
	if !reflect.DeepEqual(cs.Config.Certificates.Admin.Cert, oldCs.Config.Certificates.Admin.Cert) {
		t.Error("admin certificates do not match, check test for details")
	}
}
```

pkg/config/generate.go
```diff
@@ -227,13 +229,18 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
...
+		var k *rsa.PrivateKey
+		var c *x509.Certificate
+		if k, c, err = tls.NewCert(cert.cn, cert.organization, cert.dnsNames, cert.ipAddresses, cert.extKeyUsage, cert.signingKey, cert.signingCert, false); err != nil {
 			return
 		}
+		// compare certificate and replace if update is needed
+		// if at any point we start using self-sign certificates again
+		// (see console certificate comment), logic inside certEqual needs
+		// to be updated for this.
+		if !certEqual(c, *cert.cert) {
```
jim-minter: This is the sole cpu-intensive part of the plugin.  It would be far better CPU wise to look at the old certificate and see if it is inconsistent with the new inputs, and only then generate a new certificate if needed.

pkg/config/generate.go
```diff
@@ -388,3 +395,17 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
...
+
+func certEqual(certNew, certOld *x509.Certificate) bool {
+	if certOld == nil {
+		return false
+	}
+	if !reflect.DeepEqual(certNew.Subject, certOld.Subject) ||
+		!reflect.DeepEqual(certNew.DNSNames, certOld.DNSNames) ||
+		!reflect.DeepEqual(certNew.ExtKeyUsage, certOld.ExtKeyUsage) ||
+		!reflect.DeepEqual(certNew.IPAddresses, certOld.IPAddresses) ||
+		!reflect.DeepEqual(certNew.Issuer, certOld.Issuer) {
```
jim-minter: Issuer is just a name.  This function will not fire correctly if the CA certificate is reissued and its name doesn't change.  That's why we use CheckSignatureFrom above.  You need to do the same.

pkg/config/generate.go
```diff
@@ -15,6 +16,18 @@ import (
 	"github.com/openshift/openshift-azure/pkg/tls"
 )
 
+type certificate struct {
```
jim-minter: unless you are actually really using this, revert this change

pkg/config/generate_test.go
```diff
@@ -112,3 +113,54 @@ func testRequiredFields(cs *api.OpenShiftManagedCluster, t *testing.T) {
 	assert(c.NodeBootstrapKubeconfig != nil, "NodeBootstrapKubeconfig")
 	assert(c.AzureClusterReaderKubeconfig != nil, "AzureClusterReaderKubeconfig")
 }
+
+func TestGenerateUpdateCertRegen(t *testing.T) {
```
jim-minter: This test will run pretty slowly.  Can't you just poke some pairs of certificates into certEqual and check that the return value is as expected?

pkg/config/generate.go
mjudeikis: the challenge is that if lets say I change API FQDN. I have `config` object with Old values. And I have 18 certificates, without any distinguished mapping which certificate is using FQDN in any field. 
So when logic comes to `Router` certificate - It will not match any of the values, so logic would say that it needs to be regenerated. When in reality - it does not.

If we want to compare certificates with values without generation - certificates should have list of attributes assigned to them to know which certs should contain one or another attribute. 

I might be getting this from the wrong angle, but each time I dont generate certificate I get "mapping problem" to check right values with right certificates.

pkg/config/generate.go
mjudeikis: It makes `regenerate` function cleaner. I can move it back but this means all logic from regenate goes to main function too...

pkg/config/generate_test.go
mjudeikis: go caches the test result, so not a big issue. And it shows how our other functions work in combined. I can do this, but I think this allows to test full flow too. WDYT?

pkg/config/generate.go
```diff
@@ -388,3 +388,18 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
...
+	crt := *cert.cert
+	if crt == nil {
+		return false
+	}
+
+	if !reflect.DeepEqual(cert.dnsNames, crt.DNSNames) ||
+		!reflect.DeepEqual(cert.extKeyUsage, crt.ExtKeyUsage) ||
+		!reflect.DeepEqual(cert.cn, crt.Subject.CommonName) ||
+		!reflect.DeepEqual(cert.ipAddresses, crt.IPAddresses) {
+		return false
```
mjudeikis: We might need to do some logic to check signing key too. but this is more complex.

pkg/config/generate.go
```diff
@@ -388,3 +388,18 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
 
 	return
 }
+
+func regenerate(cert certificate) bool {
```
jim-minter: call this needsGenerate and reverse the sense - ie where it's returning false currently return true, and vv

pkg/config/generate.go
jim-minter: also consider the organization

jim-minter: the signing key shouldn't be too bad - if !selfSign, then call CheckSignatureFrom (the code you deleted above)

pkg/config/generate_test.go
jim-minter: I disagree.

pkg/config/generate.go
mjudeikis: ok, signing key was not hard. IP address was. Can you advise anything on this, please?

pkg/config/generate.go
```diff
@@ -388,3 +391,43 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
...
+	}
+
+	// if main attributes do not match - regenerate
+	// net.IP does not comply with definition of deep equality.
+	// we use native Equal method within net package
+	if !reflect.DeepEqual(cert.dnsNames, crt.DNSNames) ||
+		!reflect.DeepEqual(cert.extKeyUsage, crt.ExtKeyUsage) ||
+		!reflect.DeepEqual(cert.cn, crt.Subject.CommonName) ||
+		!reflect.DeepEqual(cert.organization, crt.Subject.Organization) ||
+		!ipAddressEqual(cert.ipAddresses, crt.IPAddresses) {
```
mjudeikis: can this be done better? basically net.IP return not equal all the times when using DeepEqual (similar topic https://stackoverflow.com/questions/42921227/ip-part-of-net-ipnet-does-not-fulfil-reflect-deepequal-but-are-equal) Can this be done better?

jim-minter:
the functionality and testing here were incomplete.  I have completed and refactored into the tls package at  #561.

Issue #429, opened 2018-09-13
need check if published image is development or release


mjudeikis:
When publishing images we need to make sure we test if those images is build in "release" mode or "development" mode. Images should point to `registry.redhat.com` and not development aws registry

jim-minter:
e2e check should ensure that no reg-aws images are running on box

jim-minter:
also that centos boxes are running origin images, and rhel OCP images

mjudeikis:
Closed in favor of Azure-153 @kwoodson already assigned to you :)

PR #520, opened 2018-09-25
Adding tests for osa container images.

kwoodson:
Was recently an issue (#429) but moved to a feature https://jira.coreos.com/browse/AZURE-153

### DIFF
--- a/test/e2e/osa_test.go
+++ b/test/e2e/osa_test.go
@@ -83,4 +83,45 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
 		})
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	It("should run the correct image", func() {
+		var prefix string
+		if strings.HasPrefix(c.cs.Config.ImageSKU, "osa_") {
+			prefix = "registry.access.redhat.com/openshift3"
+		} else {
+			prefix = "docker.io/openshift"
+		}
+
+		// these pods should have the same prefix
+		podsToCheck := map[string]struct{}{
+			"oauth-proxy":                    {},
+			"origin-control-plane":           {},
+			"origin-docker-registry":         {},
+			"origin-haproxy-router":          {},
+			"origin-node":                    {},
+			"origin-service-catalog":         {},
+			"origin-template-service-broker": {},
+			"origin-web-console":             {},
+			"prometheus-alert-buffer":        {},
+			"prometheus-alertmanager":        {},
+			"prometheus-node-exporter":       {},
+			"prometheus":                     {},
+		}
+
+		pods, err := c.kc.CoreV1().Pods("").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		for _, pod := range pods.Items {
+			// e2e check should ensure that no reg-aws images are running on box
+			Expect(strings.Contains(pod.Status.ContainerStatuses[0].Image, "registry.reg-aws")).ToNot(BeTrue())
+			parts := strings.Split(pod.Status.ContainerStatuses[0].Image, "/")
+
+			if len(parts) < 3 {
+				continue // skip unexpected image formats
+			}
+			if _, ok := podsToCheck[parts[2]]; ok {
+				// also that centos boxes are running origin images, and rhel OCP images
+				Expect(strings.HasPrefix(pod.Status.ContainerStatuses[0].Image, prefix)).To(BeTrue())
+			}
+		}
+	})
 })
--- a/test/e2e/osa_test.go
+++ b/test/e2e/osa_test.go
@@ -3,6 +3,7 @@
 package e2e
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -85,12 +86,10 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
 	})
 
 	It("should run the correct image", func() {
-		var prefix string
-		if strings.HasPrefix(c.cs.Config.ImageSKU, "osa_") {
-			prefix = "registry.access.redhat.com/openshift3"
-		} else {
-			prefix = "docker.io/openshift"
-		}
+		// image format is either:
+		// docker.io/openshift || registry.access.redhat.com/openshift3
+		format := strings.Split(c.cs.Config.Images.Format, "/")
+		prefix := fmt.Sprintf("%v/%v", format[0], format[1])
 
 		// these pods should have the same prefix
 		podsToCheck := map[string]struct{}{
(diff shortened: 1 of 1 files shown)


jim-minter:
big apologies, I'm having difficulty remembering the rationale behind either of these requested tests now, or understanding their validity.  Specifically what things are we looking for here that can't be tested via unit testing @mjudeikis @kwoodson ?

mjudeikis:
> I'm having difficulty remembering the rationale behind either of these requested tests now, or understanding their validity. Specifically what things are we looking for here that can't be tested via unit testing 

Because we had a case were we published an image with dev credentials as a valid image or centos image with rhel config. To prevent mistakes like this in the future. especially when we changing our node build process. 

We cant test this in the unit test because of its VM image specific test case. So Azure e2e is the place

test/e2e/osa_test.go
```diff
@@ -83,4 +84,43 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
+		}
+
+		pods, err := c.kc.CoreV1().Pods("").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		for _, pod := range pods.Items {
+			// e2e check should ensure that no reg-aws images are running on box
+			Expect(strings.Contains(pod.Status.ContainerStatuses[0].Image, "registry.reg-aws")).ToNot(BeTrue())
+			parts := strings.Split(pod.Status.ContainerStatuses[0].Image, "/")
+
+			if len(parts) < 3 {
```
kwoodson: @mjudeikis, would it be better to loop through the pods in my `podsToCheck` map and ensure their registry is correct?

If by some reason their images don't align with this structure:
```
repo/namespace/container
```
then we skip the test as the image format is unknown.  

Let me experiment with this.

test/e2e/osa_test.go
```diff
@@ -83,4 +84,43 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
+	It("should run the correct image", func() {
+		// image format is either:
+		// docker.io/openshift || registry.access.redhat.com/openshift3
+		format := strings.Split(c.cs.Config.Images.Format, "/")
+		prefix := fmt.Sprintf("%v/%v", format[0], format[1])
+
+		// these pods should have the same prefix
+		podsToCheck := map[string]struct{}{
+			"oauth-proxy":                    {},
+			"origin-control-plane":           {},
```
kargakis: The suffix for RHEL is `ose-control-plane`. Similar I guess about all other origin instances.

kwoodson: Thanks for pointing this out.  I am currently reworking the test and will attempt to determine ose vs origin.

kwoodson:
After a discussion with @mjudeikis, we decided to simplify this check.  The issue that was raised has been addressed by the following:
1. Ensure that all pods are not running `reg-aws` registry.
2. Ensure that the node configs match the image format specified.

test/e2e/osa_test.go
```diff
@@ -83,4 +83,24 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	It("should run the correct image", func() {
+		pods, err := c.kc.CoreV1().Pods("").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		// e2e check should ensure that no reg-aws images are running on box
+		for _, pod := range pods.Items {
+			for _, cntr := range pod.Spec.Containers {
+				Expect(strings.Contains(cntr.Image, "registry.reg-aws")).ToNot(BeTrue())
```
jim-minter: can you use HasPrefix?

test/e2e/osa_test.go
```diff
@@ -83,4 +83,24 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
+			}
+		}
+
+		// Check all Configmaps for image format matches distribution
+		// format: registry.access.redhat.com/openshift3/ose-${component}:${version}
+		maps, err := c.kc.CoreV1().ConfigMaps("openshift-node").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		for _, cm := range maps.Items {
+			Expect(strings.Contains(cm.Data["node-config.yaml"], "format: "+c.cs.Config.Images.Format)).To(BeTrue())
```
jim-minter: As this stands I think this reduces to a unit test.
I think you should be determining the cluster OS (centos or rhel) by inspecting the Node object, then calculating what you expect the imageformat to be based on that, then testing it

test/e2e/osa_test.go
kwoodson: Done.

test/e2e/osa_test.go
kwoodson: Updated to check master-00000's os type and then test the image formats.

test/e2e/osa_test.go
```diff
@@ -83,4 +83,35 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	It("should run the correct image", func() {
+		pods, err := c.kc.CoreV1().Pods("").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		// e2e check should ensure that no reg-aws images are running on box
+		for _, pod := range pods.Items {
+			for _, cntr := range pod.Spec.Containers {
+				Expect(strings.HasPrefix(cntr.Image, "registry.reg-aws")).ToNot(BeTrue())
```
jim-minter: "registry.reg-aws.openshift.com/" ?

test/e2e/osa_test.go
```diff
@@ -83,4 +83,35 @@ var _ = Describe("Openshift on Azure e2e tests", func() {
...
+		} else {
+			registryPrefix = "docker.io/openshift"
+		}
+
+		// Check all Configmaps for image format matches master's OS type
+		// format: registry.access.redhat.com/openshift3/ose-${component}:${version}
+		maps, err := c.kc.CoreV1().ConfigMaps("openshift-node").List(metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		for _, cm := range maps.Items {
+			Expect(strings.Contains(cm.Data["node-config.yaml"], "format: "+registryPrefix)).To(BeTrue())
```
jim-minter: Hacky - better to parse the yaml and check it properly

kwoodson:
Tested on both origin and ose clusters.

kwoodson:
Rebased to resolve conflicts.

kwoodson:
Fixed rebase imports.

mjudeikis:
I think I good with it, if @jim-minter is fine too - squash and merge?

kargakis:
lgtm

Issue #390, opened 2018-09-11
(after PP) refactor plugin Update function


jim-minter:
1) It is a layering violation to call upgrade.Deploy in createorupdate.go.  plugin Update() should become CreateOrUpdate() and should handle both create and update reconciliation cases

2) CreateOrUpdate() needs to be split into two pieces (pre-deploy and post-deploy) so that MSFT can use custom ARM deployment logic in both cases.

3) The create case of CreateOrUpdate() should not return until all VMs are ready, to match the update case.  This saves the health check from worrying about whether its nodes are right or not.

asalkeld:
@jim-minter I have coded up part 1 of you description, part 3 looks OK, can you explain part 2 a bit more please?
I'll post what I have done so far.

jim-minter:
Sure.  For both create and update, we do some pre-work, then call `clients.Deployments.CreateOrUpdate`, then do some post-work.

In the RP, the deployment implementation is much more complicated than `clients.Deployments.CreateOrUpdate` because there's loads of custom error handling which is legitimate and which we don't want to know anything about in the plugin.

Effectively, we need to make `clients.Deployments.CreateOrUpdate` pluggable.

We could take in a function pointer for "the thing to call to deploy", which makes our public api quite nasty.  We could also split CreateOrUpdate into CreateOrUpdatePre and CreateOrUpdatePost.  That's not that nice either.  Can you think of any other alternatives?  We should possibly talk this through with MSFT.

jim-minter:
Maybe try implementing the function pointer route as a POC and see how ugly it looks?

PR #521, opened 2018-09-26
Move the call to Deploy out of createOrUpdate

asalkeld:
This is part 1 of #390


pkg/api/plugin.go
```diff
@@ -44,6 +44,8 @@ type Plugin interface {
 
 	InitializeCluster(ctx context.Context, cs *OpenShiftManagedCluster) error
 
+	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte, isUpdate bool) error
+
 	HealthCheck(ctx context.Context, cs *OpenShiftManagedCluster) error
 
 	Update(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte) error
```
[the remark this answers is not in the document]
asalkeld: @kargakis i believe so, but maybe keep it for a short while? I am not sure on our API guarantee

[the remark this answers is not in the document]
jim-minter: No API guarantee.  Revendoring will require fixing up right away.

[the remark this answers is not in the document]
jim-minter: tl;dr - so just delete it

jim-minter:
Looks spot on.  I'd fix up to remove Update then continue working on parts 2 and 3 as additional commits in this PR.

asalkeld:
@jim-minter I think it's ready for review..

mjudeikis:
Looks like it was image flake.

pkg/api/plugin.go
```diff
@@ -14,12 +14,17 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+// DeployFn makes it possible to plug in different logic to the deploy.
```
jim-minter: Provide suitable documentation here that is aimed at the implementor.  The implementor must initiate a deployment of the given template using mode resources.Incremental and wait for it to complete.

pkg/api/plugin.go
```diff
@@ -46,5 +51,7 @@ type Plugin interface {
 
 	HealthCheck(ctx context.Context, cs *OpenShiftManagedCluster) error
 
-	Update(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte) error
+	// CreateOrUpdate either deploys or runs the update depending on the isUpdate argument
+	// this will call the PluginConfig.Deployer
```
jim-minter: check comment (deployer no longer in plugin config)

pkg/api/plugin.go
```diff
@@ -46,5 +51,7 @@ type Plugin interface {
...
+	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte, isUpdate bool, Deployer DeployFn) error
```
jim-minter: lower case d on `Deployer`

cmd/createorupdate/createorupdate.go
```diff
@@ -130,6 +124,39 @@ func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 	return oc, nil
 }
 
+func defaultDeployer(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte) error {
```
jim-minter: A few points here:

1) I think we should assume the client will use a closure and use that fact to minimise the deployer function signature.

2) I think we should directly pass in the arguments that the client needs to pass through to CreateOrUpdate() (i.e. minimise any bit shuffling the client needs to do).

Therefore I think we should get the deploy function signature down to `func(ctx context.Context, template map[string]interface{}) error`.

I was tempted to include `mode resources.DeploymentMode` as well because it is an essential part of the contract that the client uses `resources.Incremental`.  However doing this is problematic because it ties the signature to a specific Azure API, and the alternative of passing in a string just confuses the issue (IMO).

3) Doing the above should halve the size of defaultDeployer.  At that point, either use a closure to fix things up and call defaultDeployer, or bring all of defaultDeployer inline into createOrUpdate, as you see fit.

```
deployer := func(ctx context.Context, template map[string]interface{}) error {
	return defaultDeployer(ctx, config, template)
}
err = p.CreateOrUpdate(ctx, cs, azuredeploy, oldCs != nil, deployer)
```

pkg/api/plugin.go
jim-minter: please open a follow up issue to replace `azuredeploy []byte` with `template map[string]interface{}` across the board?

pkg/util/managedcluster/managedcluster.go
```diff
@@ -42,3 +46,47 @@ func ClientsetFromConfig(cs *api.OpenShiftManagedCluster) (*kubernetes.Clientset
 
 	return kubernetes.NewForConfig(restconfig)
 }
+
+// getKubeconfigFromV1Config takes a v1 config and returns a kubeconfig
+func getKubeconfigFromV1Config(kc *v1.Config) (clientcmd.ClientConfig, error) {
```
jim-minter: this is now practically identical to ClientsetFromConfig above and can presumably go away?

pkg/util/managedcluster/managedcluster.go
```diff
@@ -42,3 +46,47 @@ func ClientsetFromConfig(cs *api.OpenShiftManagedCluster) (*kubernetes.Clientset
...
+		return nil, err
+	}
+
+	t, err := rest.TransportFor(restconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait for the healthz to be 200 status
+	err = wait.ForHTTPStatusOk(ctx, t, restconfig.Host+"/healthz")
```
jim-minter: The healthz check is an additional step that should only be required when the cluster is coming up, therefore it shouldn't really be moved into a general package.  Maybe a separate exported function here which returns the restconfig to the caller.  The caller can get the restconfig, poll for /healthz, then get the ClientSet.

pkg/healthcheck/healthcheck.go
```diff
@@ -10,70 +10,34 @@ import (
...
-		return err
-	}
-
-	// Wait for infrastructure services to be healthy
-	err = WaitForInfraServices(ctx, kc)
-	if err != nil {
-		return err
-	}
-
+func (hc *simpleHealthChecker) HealthCheck(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
```
jim-minter: think we can entirely get rid of this package and put the console check at the end of plugin.CreateOrUpdate()

pkg/plugin/plugin.go
```diff
@@ -99,24 +105,33 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
...
+			return err
+		}
+	} else {
+		log.Info("starting deploy")
+		err := p.clusterUpgrader.Deploy(ctx, cs, azuredeploy, deployFn)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
jim-minter: call WaitForInfraServices and console check here?

pkg/upgrade/deploy.go
```diff
@@ -2,44 +2,32 @@ package upgrade
...
-	err = future.WaitForCompletionRef(ctx, clients.Deployments.Client)
+	// ensure that all nodes are ready
+	err = WaitForNodes(ctx, cs, kc)
 	if err != nil {
 		return err
 	}
 
-	return i.InitializeCluster(ctx, cs)
+	// Wait for infrastructure services to be healthy
+	return WaitForInfraServices(ctx, kc)
```
jim-minter: you can remove this if the caller is calling it for you

pkg/plugin/plugin.go
```diff
@@ -10,14 +10,17 @@ import (
...
 )
 
 type plugin struct {
-	entry  *logrus.Entry
-	config api.PluginConfig
+	entry           *logrus.Entry
+	config          api.PluginConfig
+	configUpgrader  config.Upgrader
+	clusterUpgrader upgrade.Upgrader
+	healthChecker   healthcheck.HealthChecker
```
jim-minter: This is awesome :-D

pkg/plugin/plugin.go
```diff
@@ -99,24 +105,33 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
...
+	if isUpdate {
+		log.Info("starting update")
+		err = p.clusterUpgrader.Update(ctx, cs, azuredeploy, deployFn)
+		if err != nil {
+			return err
+		}
+	} else {
+		log.Info("starting deploy")
+		err := p.clusterUpgrader.Deploy(ctx, cs, azuredeploy, deployFn)
+		if err != nil {
```
jim-minter: move out of the if

pkg/upgrade/clients.go
```diff
@@ -0,0 +1,20 @@
+package upgrade
+
+import (
+	"context"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+)
+
+func (u *simpleUpgrader) getClients(ctx context.Context, cs *api.OpenShiftManagedCluster) (*azureclient.AzureClients, error) {
```
jim-minter: remove this for now

pkg/api/plugin.go
asalkeld: done https://github.com/openshift/openshift-azure/issues/547

cmd/createorupdate/createorupdate.go
asalkeld: done

pkg/util/managedcluster/managedcluster.go
asalkeld: i'll either fix this or raise an issue to track it.

pkg/healthcheck/healthcheck.go
asalkeld: done

pkg/plugin/plugin.go
asalkeld: done

pkg/upgrade/deploy.go
asalkeld: done

pkg/plugin/plugin.go
asalkeld: done

pkg/upgrade/clients.go
asalkeld: done

pkg/plugin/plugin.go
```diff
@@ -26,8 +27,11 @@ var _ api.Plugin = &plugin{}
 func NewPlugin(entry *logrus.Entry, pluginConfig api.PluginConfig) api.Plugin {
 	log.New(entry)
 	return &plugin{
-		entry:  entry,
-		config: pluginConfig,
+		entry:           entry,
+		config:          pluginConfig,
+		configUpgrader:  config.NewSimpleUpgrader(entry),
```
jim-minter: I don't know why we have config.Upgrader - it never did anything and I think we now know we won't need it. #556.

pkg/upgrade/types.go
```diff
@@ -6,24 +6,26 @@ import (
...
 
 	"github.com/openshift/openshift-azure/pkg/api"
-	"github.com/openshift/openshift-azure/pkg/initialize"
 	"github.com/openshift/openshift-azure/pkg/log"
 )
 
 type Upgrader interface {
-	initialize.Initializer
-	Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte, config api.PluginConfig) error
+	InitializeCluster(ctx context.Context, cs *api.OpenShiftManagedCluster) error
```
jim-minter: at first glance this looks pretty dodgy: healthcheck, initializecluster, waitforinfraservices can all be package-internal now. 
 Deploy and Update can condense to CreateOrUpdate.  The *plugin* external API gets simplified.  I don't think the public interface should expand like this.

jim-minter:
/lgtm

great work, thank-you!

pkg/plugin/plugin.go
```diff
@@ -99,24 +102,37 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
...
-	healthChecker := healthcheck.NewSimpleHealthChecker(p.entry, p.config)
-	return healthChecker.HealthCheck(ctx, cs)
+	return p.clusterUpgrader.HealthCheck(ctx, cs)
 }
 
-func (p *plugin) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte) error {
-	log.Info("starting update")
-	upgrader := upgrade.NewSimpleUpgrader(p.entry, p.config)
-	return upgrader.Update(ctx, cs, azuredeploy, p.config)
+func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedCluster, azuredeploy []byte, isUpdate bool, deployFn api.DeployFn) error {
```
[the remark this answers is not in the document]
asalkeld: Hi @amanohar 

What error format do you need?
If we can't get the error format to your liking, I could make those functions public again.
We were trying to make improve api (there was some inconsistencies between what deploy and update were doing that you may not have noticed since you didn't use them), so if it is not working for you I can sort it out. 

I assume you want to go back to:
(Plugin) GenerateARM
RP deploys template, handles errors
(Plugin) InitializeCluster
(Plugin) WaitForEverythingToComeUp (was healthcheck)

re: Side bar
HealthCheck doesn't call waitForNodes and waitForInfraServices any more so there shouldn't be any duplicated calls.
https://github.com/openshift/openshift-azure/blob/master/pkg/upgrade/healthcheck.go#L18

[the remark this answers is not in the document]
amanohar: Hey @asalkeld, it not as much about a particular kind of error format. RP does more than that. Based on success or failure of template deployment, RP gets all deployment results that were returned by ARM (by making another call) and parses all errors to determine if error should be QoS impacting or not.  Also based on the failure it determines what was the impacting downstream RP and logs it. 

Sometimes RP changes the error code that is reported to the customer because an error code returned by CRP for e.g. might not be relevant to OSA customer. Plus deploy function does a lot of other things like retries. and is constantly being improved. 

The template deployment code also has preflight error handling baked into it that we will enable for OSA at some point. For all of this to happen properly RP needs the flexibility to be able to get a template and deploy it.

Most importantly, running create or update cannot be a black box function to the RP because each step in the process in unique & functionally different and putting it all in one function and returning it as a generic error reduces flexibility of the process overall & takes away visibility into how each error needs to be handled, logged, and reported from RP's side which is essential.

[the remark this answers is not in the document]
amanohar: @asalkeld, I must have missed that there are no dup calls amidst all the scrolling and diffing. Sorry about that. I will reread the code.

PR #524, opened 2018-09-27
POC: Refactor plugin update part 2

asalkeld:
@jim-minter I think this looks quite good using the PluginConfig, what do you think? (depends on the other commit)

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -46,5 +46,6 @@ type Plugin interface {
 
 	HealthCheck(ctx context.Context, cs *OpenShiftManagedCluster) error
 
-	Update(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte) error
+	// CreateOrUpdate either deploys or runs the update depending on the isUpdate argument
+	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte, isUpdate bool) error
 }
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -14,12 +14,19 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+// DeployFn makes it possible to plug in different logic to the deploy.
+// this should call:
+// - clients.Deployments.CreateOrUpdate() to deploy the arm
+type DeployFn func(context.Context, *OpenShiftManagedCluster, []byte, PluginConfig) error
+
 // PluginConfig is passed into NewPlugin
 type PluginConfig struct {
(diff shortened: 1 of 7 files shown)


asalkeld:
Just look at this commit: https://github.com/openshift/openshift-azure/pull/524/commits/1b519da0375aee4ccfa8211e16c3cd512127a1c2

pkg/plugin/plugin.go
```diff
@@ -102,8 +101,8 @@ func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluste
 
 func (p *plugin) InitializeCluster(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
 	log.Info("initializing cluster")
-	initializer := initialize.NewSimpleInitializer(p.entry, p.config)
-	return initializer.InitializeCluster(ctx, cs)
+	upgrader := upgrade.NewSimpleUpgrader(p.entry, p.config)
+	return upgrader.InitializeCluster(ctx, cs)
```
mjudeikis: Maybe just a semantics but seeing `Upgrader` doing `InitializeCluster` does not look right. Dunno :)

pkg/api/plugin.go
```diff
@@ -14,12 +14,19 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+// DeployFn makes it possible to plug in different logic to the deploy.
+// this should call:
+// - clients.Deployments.CreateOrUpdate() to deploy the arm
+type DeployFn func(context.Context, *OpenShiftManagedCluster, []byte, PluginConfig) error
```
kargakis: Not sure DeployFn should need to know about PluginConfig.

pkg/plugin/plugin.go
asalkeld: Yeah, I think the upgrade package should be called "cluster"
cluster.Initialize()
cluster.Update()
cluster.Deploy() etc..
because we have a config package with a upgrader as well..

pkg/api/plugin.go
asalkeld: I need it at the moment to get at the acceptLanguages for the clients, but we *could* do without that.

pkg/plugin/plugin.go
mjudeikis: Im up for this rename :)

pkg/api/plugin.go
asalkeld: I am going to move the commits in this PR and add them to https://github.com/openshift/openshift-azure/pull/521 as Jim asked. I'll fold your requests into that.

PR #528, opened 2018-09-27
Split e2e framework to run different suites based on provided kubeconfig 

kargakis:
Split out of https://github.com/openshift/openshift-azure/pull/428

@jim-minter @mjudeikis the idea here is to run all e2e suites as part of our merge queue (hence `hack/e2e.sh`) while allowing to run tests both inside a cluster and given a kubeconfig. For now I am assuming that if the provided kubeconfig is not system:admin or the azure cluster reader, it's an end user. PTAL, I think we can get this merged as is and iterate on the e2e user tests with the addition of openshift client-go.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -1,0 +1,22 @@
+#!/bin/bash -e
+
+if [[ -z "$RESOURCEGROUP" ]]; then
+    RESOURCEGROUP=$(awk '/^    resourceGroup:/ { print $2 }' <_data/containerservice.yaml)
+fi
+
+echo "Running end user e2e tests"
+# Login as osadmin to simulate a regural user
+password=$(hack/config.sh get-config $RESOURCEGROUP | jq -r .config.adminPasswd)
+fqdn=$(hack/config.sh get-config $RESOURCEGROUP | jq -r .properties.fqdn)
+oc login $fqdn --username osadmin --password $password --insecure-skip-tls-verify=true
+oc new-project e2e-end-user-test-root
+# TODO: Run the e2e image inside a job. Figure out whether we run as part of ci-operator
+# or it's just a local run.
+
+# TODO: Wait for the job to finish, report results
+oc delete project e2e-end-user-test-root
+oc logout
+
+echo "Running azure cluster reader e2e tests"
+(hack/config.sh get-config $RESOURCEGROUP | jq -r .config.azureClusterReaderKubeconfig | base64 -d) > _data/_out/azure-cluster-reader.kubeconfig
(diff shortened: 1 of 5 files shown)


test/e2e/admin_test.go
```diff
@@ -8,34 +8,23 @@ import (
...
-	c.kc = clientset
+	c = newTestClient(*kubeconfig)
 })
 
-var _ = Describe("Openshift on Azure e2e tests", func() {
+var _ = Describe("Openshift on Azure admin e2e tests", func() {
+	defer GinkgoRecover()
+
 	It("should label nodes correctly", func() {
+		if !c.isAzureClusterReader() {
```
kargakis: I hope we can find a better way to filter these (maybe use tags in test names and run ginkgo by ourselves?). Not sure if it's a blocker for this PR.

hack/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [[ -z "$RESOURCEGROUP" ]]; then
+    RESOURCEGROUP=$(awk '/^    resourceGroup:/ { print $2 }' <_data/containerservice.yaml)
+fi
+
+echo "Running end user e2e tests"
+# Login as osadmin to simulate a regural user
+password=$(hack/config.sh get-config $RESOURCEGROUP | jq -r .config.adminPasswd)
+fqdn=$(hack/config.sh get-config $RESOURCEGROUP | jq -r .properties.fqdn)
+oc login $fqdn --username osadmin --password $password --insecure-skip-tls-verify=true
```
mjudeikis: We need to think what we gonna do when we ditch this one. @cjryan already raised the same request for their testing (need of test users). 
Im thinking to maybe add htpassws as identity provider which we can use in cases like this...

test/e2e/admin_test.go
mjudeikis: +1 for tags and use --focus and --skip if we need to.

hack/e2e.sh
kargakis: Unfortunately it's impossible to `oc login` using AAD, otherwise we could use a bot account in Azure to do the login. We should probably raise an issue in Origin about adding AAD support in `oc login`.

test/e2e/admin_test.go
kargakis: Updated to use tags, ptal

hack/e2e.sh
kargakis: For clarity, in the future if we want to get rid of htpasswd altogether and we still cannot use AAD via the cli, we could create a user here, grant it project admin, and pass system:admin in the test with delegation to user. Dirty, but could work. I hope we can figure out how to use AAD.

PR #532, opened 2018-09-28
Updated rfc1123 to rfc952.

### LINKED ISSUE #523: Cluster health check should validate that router FQDN creation succeeded
Router FQDN creation was failing due restriction that domain name label must start with a letter and not a number (randomly generated value). We made a fix to ensure that router FQDN always starts with a letter.

However, we also noticed that cluster health check was not failing due to this and create operation was succeeding. We discovered the issue by looking at controller manager logs when Jenkins app we created could not be reached. Health check should also validate that router FQDN is reachable.

```E0925 22:26:03.889803       1 azure_backoff.go:372] backoff: failure, will retry, HTTP response=400, err=network.PublicIPAddressesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidDomainNameLabel" Message="The domain name label 7e37591a6e974a0f93af is invalid. It must conform to the following regular expression: ^[a-z][a-z0-9-]{1,61}[a-z0-9]$." Details=[]```

### PR DESCRIPTION
**Pull request by kwoodson**
Update our rfc1123 to match MSFT's rfc952.

Fixes #523

### DIFF
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -24,6 +24,9 @@ var (
 		`$`)
 
 	rxAgentPoolProfileName = regexp.MustCompile(`(?i)^[a-z0-9]{1,12}$`)
+
+	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
+	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
 )
 
 var (
@@ -61,7 +64,7 @@ func isValidHostname(h string) bool {
 }
 
 func isValidCloudAppHostname(h, location string) bool {
-	if !isValidHostname(h) {
+	if !rxCloudDomainLabel.MatchString(h) {
 		return false
 	}
 	return strings.HasSuffix(h, "."+location+".cloudapp.azure.com") && strings.Count(h, ".") == 4
--- a/pkg/api/validate_test.go
+++ b/pkg/api/validate_test.go
@@ -550,3 +550,69 @@ properties:
 		}
 	}
 }
+func TestValidHostnames(t *testing.T) {
+	// adding tests to validate isValidCloudAppHostname
+	// ^[a-z][a-z0-9-]{1,61}[a-z0-9]$" + "."+location+".cloudapp.azure.com
+	tests := map[string]struct {
+		hostname string
+		valid    bool
+	}{
+		"good": {
+			hostname: "test",
+			valid:    true,
+		},
+		"good-2-longer-name": {
+			hostname: "test",
+			valid:    true,
+		},
+		"good-with-dashes": {
+			hostname: "this-is",
+			valid:    true,
+		},
+		"good-with-dashes-numbers": {
+			hostname: "this-is",
+			valid:    true,
+		},
+		"good-length-61": {
+			hostname: "abcdefghijklmnopqrstuvwxzyabcdefghijklmnopqrstuvwxzy",
+			valid:    true,
+		},
+		// bad start here
+		"bad-periods": {
+			hostname: "this.",
+			valid:    false,
+		},
+		"bad-start-with-numbers": {
+			hostname: "123",
+			valid:    false,
+		},
+		"bad-start-with-dashes": {
+			hostname: "-abc",
+			valid:    false,
+		},
+		"bad-length-64": {
+			hostname: "abcdefghijklmnopqrstuvwxzyabcdefghijklmnopqrstuvwxzyabcdefghijkl",
+			valid:    false,
+		},
+		"bad-char-slash": {
+			hostname: "a/b/c",
+			valid:    false,
+		},
+		"bad-0-length-first": {
+			hostname: "",
+			valid:    false,
+		},
+		"bad-capital": {
+			hostname: "Thisisatest",
+			valid:    false,
+		},
+	}
+
+	location := "eastus"
+	for name, tc := range tests {
+		hn := tc.hostname + "." + location + ".cloudapp.azure.com"
+		if isValidCloudAppHostname(hn, location) != tc.valid {
+			t.Errorf("%s failed for [%s]", name, tc.hostname)
+		}
+	}
+}
--- a/pkg/api/validate_test.go
+++ b/pkg/api/validate_test.go
@@ -393,15 +393,28 @@ func TestIsValidCloudAppHostname(t *testing.T) {
 		"too.long.domain.cloudapp.azure.com",
 		"invalid#characters#domain.westus2.cloudapp.azure.com",
 		"wronglocation.eastus.cloudapp.azure.com",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by kwoodson** [comment]
@jim-minter, some assistance checking the regex would be appreciated.

File: pkg/api/validate.go [hunk 1]
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 var (
-	rxRfc1123 = regexp.MustCompile(`(?i)^` +
-		`([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])` +
+	rxRfc952 = regexp.MustCompile(`(?i)^` +
+		`([a-z][-a-z0-9]{0,61}[a-z0-9])` +
 		`(\.([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]))*` +
 		`$`)
```
mjudeikis: Are we saying A-Z is not allowed in all blocks? `labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the minus sign ('-')`
Is the last part `{0.61}` is right? I understand this is `.com` part. There should be the limit for this too? 

This is really hard to understand I would like to see some simple unit test for this. And some comments on the "what's behind the regexp".

kwoodson: Added testing.

pweil-: I'm not sure this is actually RFC 952 compliant anyway.  952 specifies that the string can only be up to 24 characters.  Kube used to validate this though I think they may have changed.  Here is an old commit with the expression they were using: https://github.com/kubernetes/kubernetes/commit/54a3efb36428b6d77081d9a997b6b233f0e710e4#diff-4cb64063f15b780708c65e9c71922a5dR124 

https://tools.ietf.org/html/rfc952

kwoodson: Original Error:
```
 Service returned an error. Status=400 Code="InvalidDomainNameLabel" Message="The domain name label 7e37591a6e974a0f93af is invalid. It must conform to the following regular expression: ^[a-z][a-z0-9-]{1,61}[a-z0-9]$
```
It appears that the regex is just checking the first part of the string.  The storage account is the `7e37...` string.  So it might not be completely rfc952 compliant.  Should we update the regular expression variable name to be `rxHostname` instead of `rfc952`?

jim-minter: Urg, standards.

OK, so we are going to need a sane isValidHostname for RouterProfile.PublicSubdomain (and in the future, Properties.PublicHostname).  So let's just leave rxRfc1123 and isValidHostname right where they are.

Now, for properties.FQDN and routerProfile.FQDN.  These should validate against Azure's reading of the spec -- let's just use the regex they kindly gave us.

Thus: in isValidCloudAppHostname, stop calling isValidHostname but instead check that the first part matches Azure's regexp `(?i)^[a-z][-a-z0-9]{1,61}[a-z0-9]$`.  Continue to the suffix as we do now.

kwoodson:
@mjudeikis @jim-minter, 
Does this appear correct?

kwoodson:
e2e issues:
```
ERROR: The request is being throttled.
ERROR: az storage account keys list: error: argument --account-name/-n: expected one argument
usage: az storage account keys list [-h] [--verbose] [--debug]
                                    [--output {tsv,table,json,jsonc,yaml}]
                                    [--query JMESPATH]
                                    [--resource-group RESOURCE_GROUP_NAME]
                                    --account-name ACCOUNT_NAME
                                    [--subscription _SUBSCRIPTION]
ERROR: 
No credentials specified to access storage service. Please provide any of the following:
    (1) account name and key (--account-name and --account-key options or
        set AZURE_STORAGE_ACCOUNT and AZURE_STORAGE_KEY environment variables)
    (2) connection string (--connection-string option or
        set AZURE_STORAGE_CONNECTION_STRING environment variable)
    (3) account name and SAS token (--sas-token option used with either the --account-name
        option or AZURE_STORAGE_ACCOUNT environment variable)

ERROR: The request is being throttled.
ERROR: az storage account keys list: error: argument --account-name/-n: expected one argument
usage: az storage account keys list [-h] [--verbose] [--debug]
                                    [--output {tsv,table,json,jsonc,yaml}]
                                    [--query JMESPATH]
                                    [--resource-group RESOURCE_GROUP_NAME]
                                    --account-name ACCOUNT_NAME
                                    [--subscription _SUBSCRIPTION]
ERROR: 
No credentials specified to access storage service. Please provide any of the following:
    (1) account name and key (--account-name and --account-key options or
        set AZURE_STORAGE_ACCOUNT and AZURE_STORAGE_KEY environment variables)
    (2) connection string (--connection-string option or
        set AZURE_STORAGE_CONNECTION_STRING environment variable)
    (3) account name and SAS token (--sas-token option used with either the --account-name
        option or AZURE_STORAGE_ACCOUNT environment variable)

error: A server URL must be specified
make: *** [e2e] Error 1
```

pkg/api/validate.go
```diff
@@ -24,6 +24,10 @@ var (
...
 
 	rxAgentPoolProfileName = regexp.MustCompile(`(?i)^[a-z0-9]{1,12}$`)
+
+	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
+	rxCloudDomainLabel = regexp.MustCompile(`(?i)^` +
```
jim-minter: @kwoodson looks like we need to remove `(?i)` here - see https://github.com/openshift/openshift-azure/issues/553
@sozercan fyi

jim-minter:
@kwoodson please remove the (?i) and add a test case that verifies we don't allow capitals, then this can merge

pkg/api/validate_test.go
```diff
@@ -550,3 +550,69 @@ properties:
...
+	tests := map[string]struct {
+		hostname string
+		valid    bool
+	}{
+		"good": {
+			hostname: "test",
+			valid:    true,
+		},
+		"good-2-longer-name": {
+			hostname: "test",
```
jim-minter: duplicated?

pkg/api/validate_test.go
```diff
@@ -550,3 +550,69 @@ properties:
 		}
 	}
 }
+func TestValidHostnames(t *testing.T) {
```
jim-minter: nit: newline before

jim-minter: wouldn't it make sense to roll this into TestIsValidCloudAppHostname ?

kwoodson: Sure

pkg/api/validate_test.go
kwoodson: Cleaned up.

kwoodson:
@jim-minter, if lgtm I'll squash.

jim-minter:
tests passed, squash merging

Issue #204, opened 2018-08-15
Allow configuration of CIDRs


pweil-:
port from https://github.com/openshift/openshift-azure/issues/148 

Additionally, recommend adding `NetworkProfile` section to the public API model for supporting custom VNET. Documentation for AKS custom VNET: https://docs.microsoft.com/en-us/azure/aks/networking-overview

```
type NetworkProfile struct {
	NetworkPlugin    NetworkPlugin `json:"networkPlugin,omitempty"`
	NetworkPolicy    NetworkPolicy `json:"networkPolicy,omitempty"`
	PodCidr          string        `json:"podCidr,omitempty"`
	ServiceCidr      string        `json:"serviceCidr,omitempty"`
	DNSServiceIP     string        `json:"dnsServiceIP,omitempty"`
	DockerBridgeCidr string        `json:"dockerBridgeCidr,omitempty"`
}
```

`NetworkPlugin` can default to `kubenet`. And Azure CNI can be added later.

**Kubernetes Pod CIDR** (sometimes cluster-cidr): Address range managed by a Kubernetes cluster used to assign IP addresses to customer Pods. 
When an overlay network, for example, Kubenet, is used, this address range is given by customer. The range does not have to live within the VNET but it should not collide with the subnet used for cluster nodes to avoid traffic mis-routing.

**Kubernetes Service CIDR:** Address range used to allocate IP addresses for Kubernetes cluster Services. This range does not need to live within the containing VNET. Validation is also required to make sure there is no collision with address ranges defined in node and pod CIDR. It also is not usually directly routable by VMs that are not participating in the Kubernetes cluster. 

**Docker Bridge**: Address range used inside the docker bridge interfaces. This range defaults to 172.17.0.0/16 and need only be changed if the customer is using an overlapping/conflicting subnet within their larger VNET. 

cc: @qike-ms @pweil-  @jim-minter 

/kind feature

mjudeikis:
@pweil- @jim-minter whats priority on this one?

kargakis:
Closing in favor of https://jira.coreos.com/browse/AZURE-215

PR #533, opened 2018-09-28
vnet configuration rework

jim-minter:
We are going to use vnet peering instead of "custom vnet" for private preview.  This PR adds a NetworkProfile struct, also VnetCIDR, SubnetCIDR, PeerVnetID configurables.  It removes the vnetSubnetID configurable.

Every agent pool has its own SubnetCIDR.  Because of timing, for PP day 1 we will enforce that all agent pools' SubnetCIDRs are identical.  Later we will enforce that they are disjoint.

@amanohar @pweil- @mjudeikis @kwoodson ptal

I am going to prepare a backport of these two commits to the private-preview-1 branch, which I will push once this has a lgtm.


jim-minter:
Ah, even before we make serviceNetworkCIDR and the clusterNetworkCIDR configurable, I should validate that the given node subnet doesn't intersect with either of those.  Will fix.

pkg/api/validate.go
```diff
@@ -222,13 +282,28 @@ func validateAgentPoolProfile(app *AgentPoolProfile) (errs []error) {
...
 		errs = append(errs, fmt.Errorf("invalid properties.fqdn %q", p.FQDN))
 	}
 	return
 }
 
+func validateNetworkProfile(np *NetworkProfile) (errs []error) {
+	if np == nil {
+		errs = append(errs, fmt.Errorf("networkProfile cannot be nil"))
+		return
+	}
```
amanohar: Should this be required in a non custom VNET case?

pkg/api/validate.go
```diff
@@ -44,6 +44,34 @@ func isAzureZone(fqdn string) bool {
...
+}
+
+func vnetContainsSubnet(vnet, subnet *net.IPNet) bool {
+	if !vnet.IP.Equal(subnet.IP.Mask(vnet.Mask)) {
+		return false
+	}
+
+	vnetsize, _ := vnet.Mask.Size()
+	subnetsize, _ := subnet.Mask.Size()
+	if vnetsize > subnetsize {
```
amanohar: Should this be the opposite, i.e. `vnetsize < subnetsize` then `false` should be returned?

jim-minter:
^ Fixed.

pkg/api/validate.go
jim-minter: yes, because it contains the vnetCidr.

pkg/api/validate.go
jim-minter: I think this is correct, but it's not particularly clear.  Maybe I'll s/size/bits/ and add a comment.
Say the vnet is 10.0.0.0/16 and the subnet is 10.0.0.0/8.  The vnet does not contain the subnet, so we should return false.  Then vnetsize is 16, and subnetsize is 8.  16 > 8, therefore we return false.

jim-minter: Also there are unit test cases for this.

amanohar: Yes, agreed, I read this as `vnetSize` but it's actually VNET mask size so this check makes sense.

pkg/api/validate.go
```diff
@@ -15,17 +16,36 @@ var (
...
 
+var (
+	clusterNetworkCIDR *net.IPNet
+	serviceNetworkCIDR *net.IPNet
+)
+
+func init() {
+	var err error
+
+	// TODO: we probably need to bite the bullet and make these configurable.
```
jim-minter: https://github.com/openshift/openshift-azure/issues/536

pkg/api/validate.go
```diff
@@ -158,11 +229,11 @@ func validateAgentPoolProfiles(apps []AgentPoolProfile) (errs []error) {
 		}
 		appmap[app.Role] = struct{}{}
 
-		if i > 0 && app.VnetSubnetID != apps[i-1].VnetSubnetID {
-			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles.vnetSubnetID %q: all subnets must match when using vnetSubnetID", app.VnetSubnetID))
+		if i > 0 && app.SubnetCIDR != apps[i-1].SubnetCIDR { // TODO: in the future, test that these are disjoint
```
jim-minter: https://github.com/openshift/openshift-azure/issues/537

README.md
```diff
@@ -112,80 +112,37 @@ Basic OpenShift configuration:
 
 ```yaml
 name: openshift
-location: eastus
+location: $AZURE_REGION
 properties:
   openShiftVersion: v3.10
-  publicHostname: openshift.$RESOURCEGROUP.$DNS_DOMAIN
+  publicHostname: $RESOURCEGROUP.$AZURE_REGION.cloudapp.azure.com
```
kwoodson: Does it make sense to keep the `$DNS_DOMAIN` here?

hack/create.sh
```diff
@@ -93,21 +93,26 @@ properties:
         clientId: $AZURE_AAD_CLIENT_ID
         secret: $AZURE_AAD_CLIENT_SECRET
         tenantId: $AZURE_TENANT_ID
+  networkProfile:
+    vnetCidr: 10.0.0.0/8
```
mjudeikis: How does this work? All network profile is 10.0.0.0-254 and masters is same as compute nodes and infra? how does IP address allocation work? Should it be it not be something like vnetCidr `10.0.0.0/23` and we give `10.0.0.0/24` to masters. `10.0.0.1/24` to infras, etc etc ?

README.md
kargakis: Probably not as long as we invalidate everything except [`.cloudapp.azure.com`](https://github.com/openshift/openshift-azure/pull/535).

pkg/api/validate.go
```diff
@@ -44,6 +64,34 @@ func isAzureZone(fqdn string) bool {
...
+}
+
+func vnetContainsSubnet(vnet, subnet *net.IPNet) bool {
+	if !vnet.IP.Equal(subnet.IP.Mask(vnet.Mask)) {
+		return false
+	}
+
+	vnetsize, _ := vnet.Mask.Size()
+	subnetsize, _ := subnet.Mask.Size()
+	if vnetsize > subnetsize {
```
kargakis: nit:
```go
return subnetsize > vnetsize
```

kargakis: nit:
```go
return subnetsize >= vnetsize
```

pkg/arm/data/azuredeploy.json
```diff
@@ -29,20 +28,19 @@
...
-                        "10.0.0.0/8"
+                        "{{ .ContainerService.Properties.NetworkProfile.VnetCIDR | js }}"
                     ]
                 },
                 "subnets": [
                     {
                         "name": "default",
                         "properties": {
-                            "addressPrefix": "10.0.0.0/24"
+                            "addressPrefix": "{{ (index .ContainerService.Properties.AgentPoolProfiles 0).SubnetCIDR | js }}"
```
kargakis: It's fine to use the first agentPoolProfile subnet CIDR for now since we enforce that all of them need to be the same but in https://github.com/openshift/openshift-azure/issues/537 we should probably link this to save some time for the assignee.

kargakis:
@jim-minter I can't see how `PeerVnetID` is used.

pkg/api/validate.go
```diff
@@ -174,7 +245,7 @@ func validateAgentPoolProfiles(apps []AgentPoolProfile) (errs []error) {
 	return
 }
 
-func validateAgentPoolProfile(app *AgentPoolProfile) (errs []error) {
+func validateAgentPoolProfile(app AgentPoolProfile, vnet *net.IPNet) (errs []error) {
```
pweil-: why is `AgentPoolProfile` changing?  Seems appropriate to pass as a pointer rather than a copy.

pkg/api/validate.go
```diff
@@ -15,17 +16,36 @@ var (
...
 		`/providers/Microsoft\.Network` +
 		`/virtualNetworks/[^/]+` +
-		`/subnets/[^/]+` +
 		`$`)
 
 	rxAgentPoolProfileName = regexp.MustCompile(`(?i)^[a-z0-9]{1,12}$`)
 )
 
+var (
+	clusterNetworkCIDR *net.IPNet
```
pweil-: These don't seem appropriate here.  Seems like they should be subbed in other places in configuration (master-startup) as well if we're defining variables for them.  Probably means they should live outside of the validate files.  Likely need to be configuration items anyway as noted below.

pkg/api/validate.go
```diff
@@ -206,8 +277,23 @@ func validateAgentPoolProfile(app *AgentPoolProfile) (errs []error) {
...
-	if app.VnetSubnetID != "" && !rxAgentPoolProfileVNetSubnetID.MatchString(app.VnetSubnetID) {
-		errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].vnetSubnetID %q", app.Name, app.VnetSubnetID))
+	if !isValidIPV4CIDR(app.SubnetCIDR) {
+		errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].subnetCidr %q", app.Name, app.SubnetCIDR))
+	}
+
+	_, subnet, _ := net.ParseCIDR(app.SubnetCIDR)
+	if vnet != nil && subnet != nil {
+		// we are already going to fail validation if one of these is nil.
+
```
pweil-: nit: extra space

README.md
jim-minter: Thanks @kwoodson this was a typo, it should have read `fqdn: $RESOURCEGROUP.$AZURE_REGION.cloudapp.azure.com`.  Fixed

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -91,32 +106,18 @@ type RouterProfile struct {
...
-	// VnetSubnetID is expected to be empty or match
-	// `^/subscriptions/[^/]+
-	//   /resourceGroups/[^/]+
-	//   /providers/Microsoft.Network
-	//   /virtualNetworks/[^/]+
-	//   /subnets/[^/]+$`
-	VnetSubnetID string `json:"vnetSubnetID,omitempty"`
+	Count      int    `json:"count,omitempty"`
+	VMSize     VMSize `json:"vmSize,omitempty"`
+	SubnetCIDR string `json:"subnetCidr,omitempty"`
```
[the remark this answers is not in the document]
jim-minter: Yes.

hack/create.sh
jim-minter: This matches what we were doing before, but it becomes user configurable so that the user can use a network range that fits with their peered vnet.  The precise details of what we use as defaults in create.sh are unimportant - and also do not change in this PR.  In the future we should probably hand out different ranges to master/infra/compute - that's #537.

pkg/api/validate.go
jim-minter: I'm trying to bound the change because we need to backport.  #204 covers doing this work fully.

hack/create.sh
mjudeikis: But does this works now if you hand the same subnet to each Pool? Just wondering

pkg/api/validate.go
jim-minter: It seemed to make sense on Friday but I think I agree with you on Monday.  I'll fix it up.

pweil-:
> Previously with custom vnet we could get away without adding these IP configurables. But now we're moving to using "vnet peering", and we need them.

Yep, got that part.  My question was 
> are we sure this gets us far enough and down the road we want to be for network profile config?

For instance, if we must change this now should we be putting in more structure and eliminating this:

```
// PeerVnetID is expected to be empty or match
	// `^/subscriptions/[^/]+
	//   /resourceGroups/[^/]+
	//   /providers/Microsoft.Network
	//   /virtualNetworks/[^/]+$`
```

pkg/api/validate.go
jim-minter: Ah, the Friday reasoning was, that doing it your way would introduce a nil check for something that cannot possibly be nil in the actual structure as passed in by the user - so you end up with a synthetic theoretically-user-facing error message for something that can only be caused by a logic fault by the programmer.  That's why I removed the pointer.  Thoughts on what to do @pweil- ?

pweil-:
https://github.com/openshift/openshift-azure/pull/533#discussion_r221608404 probably answers the question above.

pkg/api/validate.go
pweil-: if it is validated as non-nil elsewhere then we shouldn't have to worry about the check.  There are other examples of it in this PR (ie. oldCS can never be nil when it gets to the internal update validation method, it's checked in the public one) but they're just nits IMO.  It doesn't hurt either way, to me.  

I lean on the side of being more defensive, even if it shouldn't happen, so long as it's not making the code hard to read.

jim-minter:
@pweil-, @amanohar and I discussed it fairly extensively and we believe that we're well set for the future.  The PeerVnetID string is only used by the RP and well matches the input to the appropriate APIs that it calls, so I don't immediately believe it would need to change.  We think the NetworkProfile could become home to the pod/service CIDR configs as well (#204), so there's some extensibility capability there.

jim-minter:
@pweil- if you're happy to lgtm this, please do.

kargakis:
lgtm

PR #539, opened 2018-10-01
client consolidation

mjudeikis:
client consolidation to enable code reuse and test.

Still todo: 
- [ ] https://github.com/golang/mock integration
- [ ] tide up names, interfaces, methods. 
- [ ] add methods testing
- [ ] potentially move from utils to separate pkg

/hold

This code was laying in one of my branche for a week now. The initial intention was it to become https://jira.coreos.com/browse/AZURE-148 implementation. So just putting it out. Its on pause now due to SRE activities.


pkg/util/azureclient/virtualmachines.go
```diff
@@ -0,0 +1,107 @@
...
+		return result, err
+	}
+	err = future.WaitForCompletionRef(ctx, az.client.Client)
+	if err != nil {
+		return future, err
+	}
+	return future, err
+}
+
+func (az azVirtualMachineScaleSetsClient) Update(ctx context.Context, resourceGroupName string, VMScaleSetName string, parameters compute.VirtualMachineScaleSetUpdate) (result compute.VirtualMachineScaleSetsUpdateFuture, err error) {
```
asalkeld: these should just mirror the API, so like this
return az.client.Update(ctx, resourceGroupName, VMScaleSetName, parameters)

pkg/upgrade/update.go
```diff
@@ -210,53 +208,33 @@ func (u *simpleUpgrader) updateInPlace(ctx context.Context, cs *api.OpenShiftMan
...
 		}
 
 		{
 			log.Infof("starting %s", *vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName)
-			future, err := vmc.Start(ctx, cs.Properties.AzProfile.ResourceGroup, "ss-"+string(role), *vm.InstanceID)
-			if err != nil {
-				return err
-			}
-
-			err = future.WaitForCompletionRef(ctx, vmc.Client)
```
asalkeld: I was chatting to @jim-minter and he seems to think the mirrored functions should have the same behaviour (be non-blocking). It will probably reduce confusion in the long run even if the code looks a bit better like this now.

asalkeld:
@mjudeikis also don't try do it all in one pr, I can help with some items if we get what you have here in.

mjudeikis:
@pweil- @jim-minter @kargakis can you please give in a review? @asalkeld has like tone of tests aligned to go after this :)

jim-minter:
@mjudeikis I have rebased and force-pushed another commit onto your branch, please don't overwrite it.

This is all good stuff; I wanted to straighten it out a bit more.  Please see the detailed comment on 84b2bd947c22fe4ada699a4ec0d44fd39f09a51e to see what I did.

@mjudeikis / @asalkeld I'd be happy to see this merge, please feel free to tag with lgtm.

asalkeld:
failing continuously because of https://github.com/openshift/openshift-azure/issues/550

pkg/util/azureclient/storage/storage.go
```diff
@@ -0,0 +1,87 @@
...
+	// public cloud when a default client is created.
+	DefaultBaseURL = storage.DefaultBaseURL
+
+	// DefaultAPIVersion is the Azure Storage API version string used when a
+	// basic client is created.
+	DefaultAPIVersion = storage.DefaultAPIVersion
+)
+
+// Client is a minimal interface for azure Client
+type Client interface {
```
kargakis: nit: typically we name this interface as `Interface` instead of `Client` in the k8s/openshift client libraries.

pkg/util/azureclient/resources.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// DeploymentsClient is a minimal interface for azure DeploymentsClient
+type DeploymentsClient interface {
+	CreateOrUpdate(ctx context.Context, resourceGroupName string, deploymentName string, parameters resources.Deployment) (result resources.DeploymentsCreateOrUpdateFuture, err error)
+	Client
```
kargakis: The interface for this client (and probably some of the rest too) is not that intuitive. Here we either need to CreateOrUpdate a deployment and we need to WaitForCompletion. I feel it would be better to expose WaitForCompletion in the interface as opposed to Client.

kargakis: Same comment applies for the `VirtualMachineScaleSetVMsClient` and `VirtualMachineScaleSetsClient` interfaces.

kargakis: No blockers but please fix in a follow-up

/lgtm

PR #542, opened 2018-10-01
fix missing NetworkProfile merge

### LINKED ISSUE #541: networkProfile is not included in MergeConfig
`networkProfile` is not included in mergeConfig

https://github.com/openshift/openshift-azure/blob/master/pkg/plugin/plugin.go#L34

jim-minter:
fixes #541 

This should have been caught by a test.  AZURE-135 will rework all of this area, and I have added the requirement for such a test there.


pweil-:
> This should have been caught by a test.

Did you find why it wasn't?  We validate that `NetworkProfile` is not nil and that CIDRs are set.

PR #545, opened 2018-10-02
Vendor openshift api

kargakis:
Of course we will also need openshift/api and not just client-go

/cc @mjudeikis @jim-minter


kargakis:
Seems we don't vendor the kube api types. Need to fix that in a follow-up.

kargakis:
> Seems we don't vendor the kube api types. Need to fix that in a follow-up.

Correction: we do vendor k8s.io/api but we werent pinning it in glide.yaml. Fixed here.

jim-minter:
Bot went to sleep? Merging.

PR #548, opened 2018-10-03
Define APIVersion in the api packages

### LINKED ISSUE #433: Add API version string to each package the contains datamodel for that API version
Currently API version is available in folder name only e.g. `2018-09-30-preview` here: `https://github.com/openshift/openshift-azure/tree/master/pkg/api/2018-09-30-preview/api`

However, for RP to be able to conclusively know the correct API version name it is desirable to have each package also contain a const string with the value of version e.g.: https://github.com/Azure/acs-engine/blob/master/pkg/api/agentPoolOnlyApi/v20180331/const.go#L3-L6 that RP can refer to since API version values needs to be used in a no. of places in the service.

asalkeld:
closes #433 

@amanohar can you let me know if this is what you are after?
I was unsure what we should put into the unversioned api, maybe "unversioned" instead of "master"..

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1,5 +1,10 @@
 package api
 
+const (
+	// APIVersion is the version of this API
+	APIVersion = "internal"
+)
+
 // OpenShiftManagedCluster complies with the ARM model of resource definition in
 // a JSON template.
 type OpenShiftManagedCluster struct {
--- a/pkg/api/2018-09-30-preview/api/types.go
+++ b/pkg/api/2018-09-30-preview/api/types.go
@@ -1,5 +1,10 @@
 package api
 
+const (
+	// APIVersion is the version of this API
+	APIVersion = "2018-09-30-preview"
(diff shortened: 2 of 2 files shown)


pkg/api/types.go
```diff
@@ -1,5 +1,10 @@
 package api
 
+const (
+	// APIVersion is the version of this API
+	APIVersion = "master"
```
pweil-: we've use "unstable" in the past.  I don't have a preference.

thanasisk:
this lgtm but the reasons @pweil- stated still apply

pkg/api/types.go
jim-minter: @asalkeld what about "internal" ?

asalkeld: ok going with internal.

jim-minter:
/lgtm

going to merge this, will mention to MSFT verbally, it can be changed if needed

PR #554, opened 2018-10-04
Replace Getenvs in the code that ends up vendored in RP

### LINKED ISSUE #466: pull all os.Getenvs() out of plugin code and centralise in a plugin config struct

thekad:
Fixes #466 

Third time's the charm. Important points:

* This changes `api.plugin/GenerateARM` to add a `PluginConfig` struct
* There is a `LoadTestConfigFromEnv` that attempts to keep the same defaults the code currently uses, please help me check those defaults are OK
* I removed `Derived.*` calls that did `Getenv` and updated the templates accordingly (put the `TestConfig` as an extra variable)
* There is no validation on the values loaded from environment because all of them are optional/can have defaults (please point out if I'm mistaken)
* Updated tests

cc @jim-minter


pkg/plugin/plugin.go
```diff
@@ -97,10 +129,10 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
 	return nil
 }
 
-func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool) ([]byte, error) {
+func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool, pc *api.PluginConfig) ([]byte, error) {
 	log.Info("generating arm templates")
 	generator := arm.NewSimpleGenerator(p.entry)
-	return generator.Generate(ctx, cs, isUpdate)
+	return generator.Generate(ctx, cs, isUpdate, pc)
```
jim-minter: s/pc/p.config/ then you don't have to change the signature

pkg/config/generate_test.go
```diff
@@ -18,7 +18,10 @@ func TestGenerate(t *testing.T) {
 		},
 		//TODO "test generate doesn't overwrite": {},
 	}
-	var pluginConfig api.PluginConfig
+	tc := api.TestConfig{}
+	pluginConfig := api.PluginConfig{
+		TestConfig: &tc,
```
jim-minter: revert this, the plugin has to work with TestConfig: nil, hence so should the test

pkg/api/plugin.go
```diff
@@ -14,10 +14,22 @@ const (
...
+	ImageVersion       string
+	DeployOS           string
+	ORegURL            string
+}
+
 // PluginConfig is passed into NewPlugin
 type PluginConfig struct {
 	SyncImage       string
 	AcceptLanguages []string
+	TestConfig      *TestConfig
```
jim-minter: Best not make this a pointer otherwise we're bound to get a NPE in production

pkg/arm/arm.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 type Generator interface {
-	Generate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool) ([]byte, error)
+	Generate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool, pluginConfig *api.PluginConfig) ([]byte, error)
```
jim-minter: fwiw this addition of *PluginConfig is fine, the one on GenerateARM isn't.

pkg/config/images.go
```diff
@@ -19,16 +18,16 @@ func openShiftVersion(imageVersion string) (string, error) {
 	return fmt.Sprintf("v%s.%s.%s", parts[0][:1], parts[0][1:], parts[1]), nil
 }
 
-func selectNodeImage(cs *api.OpenShiftManagedCluster, deployOS string) {
+func selectNodeImage(cs *api.OpenShiftManagedCluster, testConfig *api.TestConfig) {
```
jim-minter: passing TestConfig all over the place is pretty horrible.  Ideally I think we should make these functions methods on a new struct type which includes TestConfig, so we don't have to pass it anywhere any more.  Either do that as a new commit in this PR, or open a follow-up to do it (you could assign it to yourself).

pkg/config/images_test.go
```diff
@@ -41,14 +42,18 @@ func TestOpenShiftVersion(t *testing.T) {
...
 	for _, deployOS := range []string{"", "rhel7", "centos7"} {
 		cs := api.OpenShiftManagedCluster{
 			Properties: &api.Properties{
 				OpenShiftVersion: "v3.10",
 			},
 			Config: &api.Config{},
 		}
-		selectNodeImage(&cs, deployOS)
+		tc.DeployOS = deployOS
+		log.Printf("Deploy OS is '%s'", tc.DeployOS)
```
jim-minter: t.Logf()

pkg/plugin/plugin.go
```diff
@@ -22,6 +24,36 @@ type plugin struct {
 
 var _ api.Plugin = &plugin{}
 
+// GetEnvOrDefault returns an env var's value or the given default
+func GetEnvOrDefault(variable string, defaultValue ...string) string {
```
jim-minter: This is over-engineered.  Move to cmd/createorupdate.  Just do
```
tc := api.TestConfig{
	RunningUnderTest: os.GetEnv("RUNNING_UNDER_TEST"),
	...
}
```

pkg/plugin/plugin_test.go
```diff
@@ -13,8 +13,12 @@ import (
 )
 
 func TestMerge(t *testing.T) {
-	var config = api.PluginConfig{SyncImage: "sync:latest",
-		AcceptLanguages: []string{"en-us"}}
+	tc := api.TestConfig{}
+	config := api.PluginConfig{
+		SyncImage:       "sync:latest",
+		AcceptLanguages: []string{"en-us"},
+		TestConfig:      &tc,
```
jim-minter: you'll be able to get rid of this

pkg/api/plugin.go
```diff
@@ -14,10 +14,22 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+// TestConfig is embedded into PluginConfig
+type TestConfig struct {
```
jim-minter: nits - can you place this below `type PluginConfig struct` for readability and add a comment on PluginConfig that TestConfig should not be populated in production.

pkg/config/images.go
jim-minter: @thekad please can you do this as a new commit in this PR.  Make the store the *plugin* config in the new struct type, that way @charlesakalugwu will be able to use it for implementing #332 as well.

pkg/config/generate.go
```diff
@@ -19,9 +18,9 @@ import (
 func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (err error) {
 	c := cs.Config
 
-	selectNodeImage(cs, os.Getenv("DEPLOY_OS"))
+	selectNodeImage(cs, &pluginConfig)
```
jim-minter: pass pluginConfig via a pointer into func Generate()

cmd/createorupdate/createorupdate.go
```diff
@@ -265,11 +265,21 @@ func main() {
...
+	tc := api.TestConfig{
+		RunningUnderTest:   os.Getenv("RUNNING_UNDER_TEST") != "",
+		ImageResourceGroup: os.Getenv("IMAGE_RESOURCEGROUP"),
+		ImageResourceName:  os.Getenv("IMAGE_RESOURCENAME"),
+		DeployOS:           os.Getenv("DEPLOY_OS"),
+		ImageOffer:         os.Getenv("IMAGE_OFFER"),
+		ImageVersion:       os.Getenv("IMAGE_VERSION"),
+		ORegURL:            os.Getenv("OREG_URL"),
+	}
+	config := api.PluginConfig{
```
jim-minter: style nit: use &api.PluginConfig here rather than & at the point of passing to createOrUpdate

pkg/api/plugin.go
```diff
@@ -23,9 +23,20 @@ type DeployFn func(context.Context, map[string]interface{}) error
 type PluginConfig struct {
 	SyncImage       string
 	AcceptLanguages []string
+	TestConfig      TestConfig // this should be empty in production
+}
+
+// TestConfig is embedded into PluginConfig
```
jim-minter: can you improve this docstring?

pkg/api/plugin.go
```diff
@@ -23,9 +23,20 @@ type DeployFn func(context.Context, map[string]interface{}) error
...
+	RunningUnderTest   bool
+	ImageResourceGroup string
+	ImageResourceName  string
+	ImageOffer         string
+	ImageVersion       string
+	DeployOS           string
+	ORegURL            string
 }
 
-// Plugin is the main interface to openshift-azure
```
jim-minter: we want all exported identifiers to have a docstring.  Improve it, but don't remove it.

pkg/arm/arm.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 type Generator interface {
-	Generate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool) ([]byte, error)
+	Generate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool, pluginConfig api.PluginConfig) ([]byte, error)
```
jim-minter: use pointer to api.PluginConfig

pkg/config/generate_test.go
```diff
@@ -18,7 +18,7 @@ func TestGenerate(t *testing.T) {
 		},
 		//TODO "test generate doesn't overwrite": {},
 	}
-	var pluginConfig api.PluginConfig
+	pluginConfig := api.PluginConfig{}
```
jim-minter: this change has no effect and the previous syntax is fine

pkg/config/images.go
```diff
@@ -19,16 +18,16 @@ func openShiftVersion(imageVersion string) (string, error) {
 	return fmt.Sprintf("v%s.%s.%s", parts[0][:1], parts[0][1:], parts[1]), nil
 }
 
-func selectNodeImage(cs *api.OpenShiftManagedCluster, deployOS string) {
+func selectNodeImage(cs *api.OpenShiftManagedCluster, pluginConfig *api.PluginConfig) {
```
jim-minter: this works but is ugly.  As mentioned before I'd like to see all these functions that take `cs *api.OpenShiftManagedCluster, pluginConfig *api.PluginConfig` being methods on a base struct which contains cs and pluginConfig as fields.  That way they don't need to be passed around everywhere.

pkg/config/images_test.go
```diff
@@ -41,14 +41,20 @@ func TestOpenShiftVersion(t *testing.T) {
 }
 
 func TestNodeImageVersion(t *testing.T) {
+	pc := api.PluginConfig{
+		TestConfig: api.TestConfig{},
+	}
```
jim-minter: this boils down to `var pc api.PluginConfig`.  Prefer that syntax.

pkg/plugin/plugin.go
```diff
@@ -102,7 +102,8 @@ func (p *plugin) GenerateConfig(ctx context.Context, cs *api.OpenShiftManagedClu
 
 func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool) ([]byte, error) {
 	log.Info("generating arm templates")
-	return p.armGenerator.Generate(ctx, cs, isUpdate)
+	generator := arm.NewSimpleGenerator(p.entry)
+	return generator.Generate(ctx, cs, isUpdate, p.config)
```
jim-minter: this change is due to a mishandled rebase - it shouldn't be here - see what `InitializeCluster` looks like below.

pkg/plugin/plugin_test.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func TestMerge(t *testing.T) {
-	var config = api.PluginConfig{
+	config := api.PluginConfig{
```
jim-minter: prefer &api.PluginConfig here

thekad:
@jim-minter I created a struct to hold cluster and image configuration, but naming sucks, what do you think so far?

thekad:
@jim-minter @charlesakalugwu please do take a look, I have created a new struct to hold both the OpeshiftManagedCluster and PluginConfig(+TestConfig) structs. At this point I sort of feel I should take the TestConfig _out_ of PluginConfig and into the same new struct I created, but eager to know what your thoughts are.

pkg/api/plugin.go
```diff
@@ -23,6 +23,24 @@ type DeployFn func(context.Context, map[string]interface{}) error
...
+	ImageResourceGroup string
+	ImageResourceName  string
+	ImageOffer         string
+	ImageVersion       string
+	DeployOS           string
+	ORegURL            string
+}
+
+// ClusterImageConfig holds all config fields needed for images to be processed
+type ClusterImageConfig struct {
```
kargakis: Why do you need this new type? It should be possible to do what you want w/o it.

jim-minter: +1. Definitely don't do this

pkg/api/plugin.go
jim-minter: bump

pkg/config/images.go
jim-minter: @thekad I think I didn't explain myself well enough here.  In pkg/config (not pkg/api) create a new unexported struct type just like in the other packages.  That should contain cs *api.OpenShiftManagedCluster and pluginConfig *api.PluginConfig.  Then all of these functions here should be member functions of that struct type.

jim-minter: @kargakis fyi ^

kargakis:
There is a failure in pkg/fakerp
```
pkg/fakerp/fakerp.go:142:19: config.Derived.ImageResourceName undefined (type config.derived has no field or method ImageResourceName)
```

thekad:
@kargakis yep I guess I missed your fakerp stuff, fixed already let's see if CI is happy now

pkg/config/images.go
```diff
@@ -19,50 +42,50 @@ func openShiftVersion(imageVersion string) (string, error) {
 	return fmt.Sprintf("v%s.%s.%s", parts[0][:1], parts[0][1:], parts[1]), nil
 }
 
-func selectNodeImage(cs *api.OpenShiftManagedCluster, deployOS string) {
-	c := cs.Config
+func (cg *configGenerator) selectNodeImage() {
+	c := cg.cs.Config
 	c.ImagePublisher = "redhat"
 	c.ImageOffer = os.Getenv("IMAGE_OFFER")
```
kargakis: Shouldn't this flow from TestConfig?

pkg/config/images.go
```diff
@@ -19,50 +42,50 @@ func openShiftVersion(imageVersion string) (string, error) {
...
 	if c.ImageOffer == "" {
 		c.ImageOffer = "osa"
 	}
 
 	c.ImageVersion = os.Getenv("IMAGE_VERSION")
```
kargakis: Shouldn't this flow from TestConfig?

pkg/api/plugin.go
kargakis: The last part of this comment is still outstanding.

pkg/util/mocks/mock_arm/arm.go
```diff
@@ -9,7 +9,6 @@ import (
 	reflect "reflect"
 
 	gomock "github.com/golang/mock/gomock"
-
```
kargakis: This generated diff is invalid - you can either run `goimports -local=github.com/openshift/openshift-azure -e -w pkg` to fix it or wait for https://github.com/openshift/openshift-azure/pull/640 to merge, rebase on top, and rerun `go generate ./...`

jim-minter:
@thekad I've pushed a commit with fixups to this PR.  Please review, ask any questions, then squash.  Then this can be lgtmed.

kargakis:
/hold

needs a squash

thekad:
Squashed, this looks cleaner, thanks @jim-minter

PR #558, opened 2018-10-05
Clean up the upgrader interface

### LINKED ISSUE #557: Cleanup the Upgrader interface
from Jim's comment here: https://github.com/openshift/openshift-azure/pull/521#discussion_r222862362

At first glance this looks pretty dodgy: healthcheck, initializecluster, waitforinfraservices can all be package-internal now.
Deploy and Update can condense to CreateOrUpdate. The plugin external API gets simplified. I don't think the public interface should expand like this.

and drop the equivalent calls from the main plugin interface since no-one'll use them

asalkeld:
closes #557


mjudeikis:
/approve 
I think we spoke about this that current name is not right. I like this one. @pweil- are we good to remove those 2 methods without impact? Just checking before lgtm
```
InitializeCluster(ctx context.Context, cs *OpenShiftManagedCluster) error
```

jim-minter:
all looks good, @asalkeld can you think of a better name than Upgrader ?

jim-minter:
"deployer"?

jim-minter:
@asalkeld are this PR and issue #557 superceded?  Is there anything to salvage here or should it be closed?

asalkeld:
> @asalkeld are this PR and issue #557 superceded? Is there anything to salvage here or should it be closed?

@jim-minter I rebased and just included 2 commits that might still be nice, ptal.

kargakis:
Doesn't look like this PR needs a signoff anymore

/remove-label do-not-merge/needs-signoff
/retest
/lgtm

PR #561, opened 2018-10-05
regenerate certificates if any of their input configurations change

jim-minter:


pkg/config/generate.go
```diff
@@ -66,173 +67,224 @@ func Generate(cs *api.OpenShiftManagedCluster, pluginConfig api.PluginConfig) (e
...
+				ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+				SigningKey:  c.Certificates.ServiceCatalogCa.Key,
+				SigningCert: c.Certificates.ServiceCatalogCa.Cert,
 			},
-			extKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
-			signingKey:  c.Certificates.ServiceCatalogCa.Key,
-			signingCert: c.Certificates.ServiceCatalogCa.Cert,
-			key:         &c.Certificates.ServiceCatalogServer.Key,
-			cert:        &c.Certificates.ServiceCatalogServer.Cert,
+			key:  &c.Certificates.ServiceCatalogServer.Key,
```
kargakis: Unrelated to this PR but do we need these anymore?

pkg/tls/tls.go
```diff
@@ -81,27 +88,71 @@ func NewCert(
...
+// CertMatchesParams returns true if the given key and cert match the CertParams
+// struct.
+func CertMatchesParams(key *rsa.PrivateKey, cert *x509.Certificate, params *CertParams) bool {
+	if key == nil || cert == nil {
+		return false
+	}
+
+	// check the key and cert match eachother
+	certPublicKey, ok := cert.PublicKey.(*rsa.PublicKey)
+	if !ok || !reflect.DeepEqual(key.PublicKey, *certPublicKey) {
```
kargakis: Is it possible to get a different public key when the only variable input to our certgen process is the dns names? IOW, isn't it enough to compare the dns names only?

kargakis: I guess we also need to account for different signing keys.

pkg/tls/tls.go
```diff
@@ -81,27 +88,71 @@ func NewCert(
...
+
+	return true
+}
+
+func netIPsEqual(ip1, ip2 []net.IP) bool {
+	// reflect.DeepEqual doesn't work on net.IP - see net.IP.Equal() for details
+	if len(ip1) != len(ip2) {
+		return false
+	}
+	for i := range ip1 {
```
kargakis: Do we need to sort here?

pkg/tls/tls.go
jim-minter: it's belt and braces

PR #563, opened 2018-10-08
Use gomock to unit test

asalkeld:
* vendor gomock
* show a simple example of using it.
Once the client consolidation PR lands we can add mocks for the Azure clients and test more.
This PR is mostly getting gomock in.


Makefile
```diff
@@ -11,7 +11,13 @@ clean:
 
 test: unit e2e
 
-generate:
+MOCK_FILES=pkg/util/mocks/mock_upgrade/types.go pkg/util/mocks/mock_arm/arm.go
+
+pkg/util/mocks/mock_%.go: pkg/%.go
```
jim-minter: any chance this can be done via `//go:generate` ?

jim-minter: and can you update hack/validate-generated.sh (it would be nice if it could be changed such that it didn't need to be updated in a case like this)

asalkeld: will do, I have a look at go generate

asalkeld:
hi @pweil- 

I just had a look at the test files and found these links:
https://codeengineered.com/blog/2016/go-why-not-strip-unused-pkgs/
(basically for legal reasons they don't want to "modify" projects)
if we want to go ahead with this, then there is a separate tool: https://github.com/sgotti/glide-vc (vendor cleaner).

mjudeikis:
my personal take is while we are playing in `x0Mb` sizes, we should not strip down things. If we see that our binary becomes heavy - when. Its not Java with 200Mb runtime and 100Mb code dragging half the world around :D

pweil-:
> (basically for legal reasons they don't want to "modify" projects)
> if we want to go ahead with this, then there is a separate tool: https://github.com/sgotti/glide-vc (vendor cleaner).

nah, I can buy the reasoning behind not modifying.  That makes sense to me and since glide is the tool of choice right now we'll run with it.  Thanks for the link and explanation!

pweil-:
> Its not Java with 200Mb runtime and 100Mb code dragging half the world around :D

```java
System.out.println("are you proposing rewrite? :-D")
```

pkg/arm/arm.go
```diff
@@ -3,6 +3,9 @@ package arm
 //go:generate go get github.com/go-bindata/go-bindata/go-bindata
 //go:generate go-bindata -nometadata -pkg $GOPACKAGE -prefix data data/...
 //go:generate gofmt -s -l -w bindata.go
+//go:generate mockgen -destination=../util/mocks/mock_arm/arm.go -package=mock_arm -source arm.go
```
jim-minter: add a `//go:generate go get` or `go install` or whatever line to get mockgen above each line in each file where you call mockgen

hack/validate-generated.sh
```diff
@@ -24,6 +24,10 @@ GENERATED_FILES=(
 	"pkg/addons/bindata.go"
 	"pkg/jsonpath/jsonpath.go"
 	"pkg/arm/bindata.go"
+	"pkg/util/mocks/mock_upgrade/types.go"
+	"pkg/util/mocks/mock_arm/arm.go"
+	"pkg/util/mocks/mock_azureclient/azureclient.go"
+	"pkg/util/mocks/mock_azureclient/mock_storage/storageclient.go"
```
jim-minter: can we check the output of `git clean -ndx` or equivalent as a strategy here rather than maintaining this file list?

asalkeld: i am now searching for the "generated by.."

pkg/arm/arm.go
asalkeld: done

pkg/arm/arm.go
```diff
@@ -3,6 +3,11 @@ package arm
 //go:generate go get github.com/go-bindata/go-bindata/go-bindata
 //go:generate go-bindata -nometadata -pkg $GOPACKAGE -prefix data data/...
 //go:generate gofmt -s -l -w bindata.go
+//go:generate go get github.com/golang/mock/gomock
+//go:generate go install github.com/golang/mock/mockgen
+//go:generate mockgen -destination=../util/mocks/mock_arm/arm.go -package=mock_arm -source arm.go
```
jim-minter: nit: could probably s/arm/$GOPACKAGE/ here fwiw

pkg/plugin/plugin_test.go
```diff
@@ -96,3 +103,79 @@ func testPluginRun(p api.Plugin, newCluster *api.OpenShiftManagedCluster, oldClu
 	}
 	return azuretemplate
 }
+
+func TestGenerateARM(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	var testData map[string]interface{}
+	err := json.Unmarshal([]byte("{\"test\": \"data\"}"), &testData)
```
jim-minter: nit `testData := map[string]interface{}{"test": "data"}`

pkg/upgrade/initialize.go
```diff
@@ -10,28 +10,31 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
 )
 
-func (si *simpleUpgrader) InitializeCluster(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
-	authorizer, err := azureclient.NewAuthorizerFromContext(ctx)
-	if err != nil {
-		return err
+func (u *simpleUpgrader) InitializeCluster(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+	if u.accountsClient == nil {
```
jim-minter: this can move to NewSimpleUpgrader, can't it?

pkg/upgrade/initialize.go
```diff
@@ -10,28 +10,31 @@ import (
...
+		}
+
+		u.accountsClient = azureclient.NewAccountsClient(cs.Properties.AzProfile.SubscriptionID, authorizer, u.pluginConfig.AcceptLanguages)
 	}
-
-	accounts := azureclient.NewAccountsClient(cs.Properties.AzProfile.SubscriptionID, authorizer, si.pluginConfig.AcceptLanguages)
-	keys, err := accounts.ListKeys(ctx, cs.Properties.AzProfile.ResourceGroup, cs.Config.ConfigStorageAccount)
-	if err != nil {
-		return err
+	if u.storageClient == nil {
```
jim-minter: I don't like this, but I guess it's because the SA doesn't exist when NewSimpleUpgrader is called.  I'm not immediately coming up with any better alternatives though - can you?
Certainly warrants a comment.

pkg/util/azureclient/storage/storage.go
```diff
@@ -52,6 +52,11 @@ type blobStorageClient struct {
 	storage.BlobStorageClient
 }
 
+// NewBlobStorageClient just for mocking purposes
+func NewBlobStorageClient() BlobStorageClient {
```
jim-minter: AFAICS this isn't used in the PR?  I'm a little reluctant about this because the upstream API doesn't have this functionality.

pkg/upgrade/initialize.go
asalkeld: it uses the context and we don't have that in the NewSimpleUpgrader

pkg/upgrade/initialize.go
asalkeld: unless you pass the context and cs into NewSimpleUpgrader

pkg/util/azureclient/storage/storage.go
asalkeld: we need the mock function NewMockBlobStorageClient(), the mock generator only makes functions based on public functions, so if i remove this i can't create my own mock blob client..

pkg/upgrade/initialize.go
jim-minter: @amanohar or @sozercan is the Context object containing the Azure credentials (api.ContextKeyClientID etc) available when the RP calls plugin.NewPlugin?

jim-minter: opened follow up issue #602

pkg/upgrade/initialize.go
jim-minter: opened follow up issue #602

PR #564, opened 2018-10-08
Smoke test

thekad:
This reorganizes the e2e suite a bit plus new test suite. About half of this got taken out of the okd docs. This adds a new focus in the e2e test suite that:

* Instantiate a template already in the box
* Wait for it to finish
* Hit it a few times to make sure the visit counter increments


Makefile
```diff
@@ -51,7 +51,7 @@ cover:
 	go tool cover -html=coverage.out
 
 e2e: generate
-	./hack/e2e.sh
+	./hack/e2e.sh "${SUITE}"
```
kargakis: This script needs to run all e2e tests that we have so we should not make this change. Otherwise, we need to update the way we run e2e tests in openshift/release.

test/e2e/end_user_test.go
```diff
@@ -16,20 +14,6 @@ import (
 var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 	defer GinkgoRecover()
 
-	BeforeEach(func() {
-		namespace := nameGen.generate("e2e-test-")
```
kargakis: We run project creation before each test as opposed to before each suite to create a separate namespace for every test that runs for end users.

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const GOOD_TEMPLATE = "nodejs-mongodb-example"
+const BAD_TEMPLATE = "non-existent"
+const REGEX = `\<span.*id=\"count-value\"\>(?P<counter>\d+)\<\/span\>`
+
+var _ = Describe("Openshift on Azure end user e2e tests [SmokeTest]", func() {
+	defer GinkgoRecover()
+
+	Context("When the template exists", func() {
```
kargakis: I would expect to see this test as part of the end_user_test.go file since it's an e2e test that should work for end users.

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+		It("should deploy a template and check the visit counter increments", func() {
+			// Create the template
+			templateclient, err := templatev1client.NewForConfig(c.config)
```
kargakis: Instead of creating the template client here, add it in `testClient` similar to how we hold `ProjectV1Client`.

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+				default:
+					panic("unexpected event type " + event.Type)
+				}
+			}
+
+			r := regexp.MustCompile(REGEX)
+
+			// Pull the route ingress from the namespace
+			routeclient, err := routev1client.NewForConfig(c.config)
```
kargakis: Move inside `newTestClient`

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+				currCounter, err = strconv.Atoi(matches[1])
+				Expect(err).NotTo(HaveOccurred())
+				Expect(currCounter).Should(BeNumerically(">", prevCounter))
+				prevCounter = currCounter
+				time.Sleep(1)
+			}
+		})
+	})
+
+	Context("When the template doesn't exist", func() {
```
kargakis: The template should always exist, I think this check is redundant.

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const GOOD_TEMPLATE = "nodejs-mongodb-example"
+const BAD_TEMPLATE = "non-existent"
+const REGEX = `\<span.*id=\"count-value\"\>(?P<counter>\d+)\<\/span\>`
+
+var _ = Describe("Openshift on Azure end user e2e tests [SmokeTest]", func() {
```
kargakis: SmokeTest is really EndUser

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			host := route.Status.Ingress[0].Host
+			url := fmt.Sprintf("http://%s", host)
+
+			// Hit the ingress 3 times, each time the hit counter should increment
+			prevCounter := 0
+			currCounter := 0
+			for i := 1; i <= 3; i++ {
+				resp, err := http.Get(url)
+				Expect(err).NotTo(HaveOccurred())
```
kargakis: Some of these errors are going to happen eventually and I would like to see as build robust e2e tests from the start. I would suggest create a wrapper for http.Get that retries a couple of times on transport errors and 500s.

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+				defer resp.Body.Close()
+				contents, err := ioutil.ReadAll(resp.Body)
```
kargakis: ANother check missing from here is to ensure resp.StatusCode == http.StatusOK

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				matches := r.FindStringSubmatch(string(contents))
+				currCounter, err = strconv.Atoi(matches[1])
+				Expect(err).NotTo(HaveOccurred())
+				Expect(currCounter).Should(BeNumerically(">", prevCounter))
+				prevCounter = currCounter
+				time.Sleep(1)
```
kargakis: This is not a second but a nanosecond. You probably want `time.Sleep(time.Second))`

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+						Template: *template,
+						Secret: &corev1.LocalObjectReference{
+							Name: secret.Name,
+						},
+					},
+				})
+			Expect(err).NotTo(HaveOccurred())
+
+			// Wait for the deployments to finish
+			watcher, err := templateclient.TemplateInstances(c.namespace).Watch(
```
kargakis: Instead of a watch I would vote for polling since it is less code here and simpler to maintain. Add a `func (t *testclient) templateInstanceIsReady (bool, error)` function in conditions_test.go and wait for it here with a `wait.PollImmediate(2*time.Second, time.Minute, t.templateInstanceIsReady)`

test/e2e/smoke_test.go
```diff
@@ -0,0 +1,141 @@
...
+	templatev1 "github.com/openshift/api/template/v1"
+	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const GOOD_TEMPLATE = "nodejs-mongodb-example"
```
kargakis: I would inline these inside the test vs having them as globals.

test/e2e/end_user_test.go
pweil-: +1 this is important, it allows each test to ensure it is working with a clean slate rather than trying to reason about existing state.  It also makes clean up easier since we can just blast the namespace, assuming that the test isn't creating cluster scoped objects

Makefile
thekad: @kargakis as you can see in `hack/e2e.sh` the current behavior is maintained: an empty $SUITE variable will run all tests, this is just to override that behavior.

test/e2e/end_user_test.go
thekad: Thanks for catching this, I had moved this to `BeforeSuite` but it does make more sense to have it on `BeforeEach`

thekad: fixed in b196fa3

test/e2e/smoke_test.go
thekad: This comment got buried in the changes I pushed: @kargakis I'll ping you tomorrow to go over this, I'm a bit confused on how exactly to do that.

test/e2e/admin_test.go
```diff
@@ -16,10 +16,6 @@ import (
 	"github.com/openshift/openshift-azure/pkg/jsonpath"
 )
 
-var _ = BeforeSuite(func() {
-	c = newTestClient(*kubeconfig)
```
kargakis: This needs to be added back.

test/e2e/end_user_test.go
```diff
@@ -3,34 +3,56 @@
...
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/watch"
 )
 
+func httpWithRetry(url string, times int) (string, error) {
+	// attempt up to 3 times with linear backoff
+	for i := 1; i <= times; i++ {
+		delay := time.Duration(i*100) * time.Millisecond
+		resp, err := http.Get(url)
+		defer resp.Body.Close()
```
kargakis: resp may be nil at this point; move after the error check

test/e2e/end_user_test.go
```diff
@@ -3,34 +3,56 @@
...
+		// if the GET failed, wait a bit and try again
+		if err != nil {
+			time.Sleep(delay)
```
kargakis: Print the error

test/e2e/end_user_test.go
```diff
@@ -3,34 +3,56 @@
...
+			continue
+		}
+		// if the status code is not 200, wait a bit and try again
+		if resp.StatusCode != http.StatusOK {
+			time.Sleep(delay)
```
kargakis: Print the status

test/e2e/end_user_test.go
```diff
@@ -3,34 +3,56 @@
...
+			continue
+		}
+		// if the contents (somehow) can't be read, wait a bit and try again
+		contents, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			time.Sleep(delay)
```
kargakis: Print the error

test/e2e/end_user_test.go
```diff
@@ -3,34 +3,56 @@
...
+			continue
+		}
+		return string(contents), nil
+	}
+	// if we got to this point, all attempts failed
+	return "", errors.New("All HTTP requests failed")
```
kargakis: error messages should start with a lowercased letter

test/e2e/end_user_test.go
```diff
@@ -51,4 +73,94 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+					Template: *template,
+					Secret: &corev1.LocalObjectReference{
+						Name: secret.Name,
+					},
+				},
+			})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Wait for the deployments to finish
+		watcher, err := c.tc.TemplateInstances(c.namespace).Watch(
```
kargakis: Instead of L114-L144 you can simply do
```go
err := wait.PollImmediate(2*time.Second, time.Minute, t.templateInstanceIsReady)
```
and inside `conditions_test.go`
```go
func (t *testClient) templateInstanceIsReady() (bool, error) {
	ti, err := t.tc.TemplateInstances().Get(t.namespace, metav1.GetOptions{})
	if kerrors.IsNotFound(err) {
		return false, nil
	}
	if err != nil {
		return false, err
	}

	for _, cond := range ti.Status.Conditions {
		if cond.Type == templatev1.TemplateInstanceReady &&
			cond.Status == corev1.ConditionTrue {
			return true, nil
		} else if cond.Type == templatev1.TemplateInstanceInstantiateFailure &&
			cond.Status == corev1.ConditionTrue {
			return false, fmt.Errorf("templateinstance %q failed", t.namespace)
		}
	}

	return false, nil
}
```
You will need to use c.namespace as the name of the template instance in that case. I think it's fine to do it for now.

test/e2e/end_user_test.go
```diff
@@ -51,4 +73,94 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+
+			default:
+				panic("unexpected event type " + event.Type)
+			}
+		}
+
+		// Pull the route ingress from the namespace
+		route, err := c.rc.Routes(c.namespace).Get(NAME, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		host := route.Status.Ingress[0].Host
```
kargakis: Before accessing the host, do
```go
Expect(route.Status.Ingress).To(Equal(1))
```

test/e2e/extended_suite_test.go
```diff
@@ -16,6 +17,16 @@ var (
 	kubeconfig = flag.String("kubeconfig", "../../_data/_out/admin.kubeconfig", "Location of the kubeconfig")
 )
 
+var _ = BeforeEach(func() {
```
kargakis: Needs to be moved back to were it was (inside the end user e2e tests). In other words, there should be no diff in your PR related to this.

test/e2e/end_user_test.go
kargakis: Please revert this diff as it was before.

test/e2e/extended_suite_test.go
kargakis: The problem with doing this for every suite is that in the azure cluster reader tests we don't need to create namespaces (at least so far) so the extra creation and deletion would incur considerable slowdown to our test times.

test/e2e/admin_test.go
thekad: This got moved to `extended_suite_test.go:BeforeEach` (along with cluster creation) do you feel the client should be created only in `BeforeSuite`?

kargakis: Eventually we may want to have separate clients per namespace, doesn't seem like we gain something by doing it now.

Makefile
kargakis: ack

thekad:
One thing I feel I must point out: This new smoke test _alone_ takes an avg of 90-100 *seconds*, are we cool on running that on every build ? cc @kargakis @jim-minter

kargakis:
Yes that is fine, eventually I think we should start running all tests in a
suite in parallel but we are not there yet (will probably need to map
namespaces to clients for doing the ns cleanup appropriately)

test/e2e/end_user_test.go
```diff
@@ -51,4 +85,65 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 		// Expect(kerrors.IsForbidden(err)).To(Equal(true))
 		fmt.Printf("PDB create error: %v\n", err)
 	})
+
+	It("should deploy a template and check the visit counter increments", func() {
+		var regex = regexp.MustCompile(`\<span.*id=\"count-value\"\>(?P<counter>\d+)\<\/span\>`)
```
jim-minter: check your backslashes, I don't think you need them before `<`, `>`, `"` or `/`.
also `span.*id` is a bit dodgy - just start at `id`
also you're not using `(?P<counter>\d+)` - just do `(\d+)`

test/e2e/end_user_test.go
```diff
@@ -51,4 +85,65 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+
+		// Create the template
+		template, err := c.tc.Templates("openshift").Get(
+			"nodejs-mongodb-example", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Create a secret to hold the template params
+		secret, err := c.cc.Secrets(c.namespace).Create(&corev1.Secret{
```
jim-minter: the secret shouldn't be necessary and I don't think using it adds anything to the smoke test

test/e2e/end_user_test.go
```diff
@@ -51,4 +85,65 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+					Template: *template,
+					Secret: &corev1.LocalObjectReference{
+						Name: secret.Name,
+					},
+				},
+			})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Poll for the deployment status
+		err = wait.PollImmediate(2*time.Second, 2*time.Minute, c.templateInstanceIsReady)
```
jim-minter: 2*time.Minute is too short, you'll flake

test/e2e/end_user_test.go
```diff
@@ -51,4 +85,65 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		route, err := c.rc.Routes(c.namespace).Get(c.namespace, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(route.Status.Ingress)).To(Equal(1))
+		host := route.Status.Ingress[0].Host
+		url := fmt.Sprintf("http://%s", host)
+
+		// Hit the ingress 3 times, each time the hit counter should increment
+		prevCounter := 0
+		currCounter := 0
+		for i := 1; i <= 3; i++ {
```
jim-minter: `for i := 0; i < 3; i++` !

test/e2e/end_user_test.go
```diff
@@ -51,4 +85,65 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+			contents, err := httpWithRetry(url, 3)
+			Expect(err).NotTo(HaveOccurred())
+			matches := regex.FindStringSubmatch(contents)
```
jim-minter: expect matches notto benil

test/e2e/conditions_test.go
```diff
@@ -3,11 +3,36 @@
...
+	"fmt"
+
+	templatev1 "github.com/openshift/api/template/v1"
 	authorizationapiv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func (t *testClient) templateInstanceIsReady() (bool, error) {
```
jim-minter: should probably take the namespace and name as arguments if it's going in a different file

test/e2e/end_user_test.go
```diff
@@ -3,31 +3,65 @@
...
 	policy "k8s.io/api/policy/v1beta1"
 	_ "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+func httpWithRetry(url string, times int) (string, error) {
+	// attempt up to 3 times with linear backoff
+	for i := 1; i <= times; i++ {
```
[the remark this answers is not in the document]
jim-minter: are you finding that you need to retry?  It ought to work first time, oughtn't it?  Can the retry be removed for now?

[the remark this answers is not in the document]
kargakis: Retrying here is to weed out flakes because of transport errors.

test/e2e/end_user_test.go
kargakis: 5 or 10 minutes should be fine to use.

test/e2e/conditions_test.go
kargakis: That would not work because the fn signature here needs to be a `wait.ConditionFunc` which is a `func() (bool, error)` in order to be used as an input in `wait.PollImmediate`. I guess you could pass a closure in PollImmediate but I don't feel strong about generalizing yet. `conditions_test.go` holds all the ConditionFuncs we use in our e2es but they are not necessarily reusable across tests and I am not sure most of them are going to be used more than once, it's just a clean way to separate tests from helpers.

Makefile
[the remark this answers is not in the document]
thekad: good call

test/e2e/client_test.go
```diff
@@ -8,18 +8,24 @@ import (
...
 	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 var c *testClient
 
 type testClient struct {
+	cc        *corev1client.CoreV1Client
```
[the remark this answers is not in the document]
thekad: oh nice

test/e2e/end_user_test.go
thekad: test has been taking between 90 and 100 seconds consistently, going to use 5min here

test/e2e/conditions_test.go
thekad: I could also put the template instance reference in testClient before polling, added benefit would be I don't have to query for the template instance

kargakis: Let's keep the testClient hold just clients and not resource references. Also, you will have to query for the template instance anyway inside the condition func in order to determine whether its status is ready.

test/e2e/end_user_test.go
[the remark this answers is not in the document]
jim-minter: @kargakis I don't buy your argument, if it doesn't succeed the first time I think we should look into why.
I mean, a TCP connect is supposed to be resilient to a lot of this stuff - and we expect for an arbitrary user browsing to this app to reach it successfully first time.

kargakis:
Fwiw, we have been exposing request retries as a metric in the CI infra so
we could do something similar here but sure, can we enumerate what data we
need to expose in the test artifacts in order to debug these failures? I
don't see anything in place at the moment that would help. Fine with taking
this discussion in a follow-up issue but I really hope that such a flake is
not going to be ignored once it starts manifesting in the merge queue.

test/e2e/end_user_test.go
```diff
@@ -51,4 +52,59 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		currCounter := 0
+		httpc := &http.Client{
+			Timeout: 10 * time.Second,
+		}
+		// Hit the ingress 3 times, each time the hit counter should increment
+		for i := 0; i < 3; i++ {
+			resp, err := httpc.Get(url)
+			Expect(err).NotTo(HaveOccurred())
+			defer resp.Body.Close()
+			Expect(resp.StatusCode).Should(Equal(200))
```
kargakis: nit: instead of 200 you could use http.StatusOK

test/e2e/end_user_test.go
```diff
@@ -4,30 +4,31 @@ package e2e
...
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 	defer GinkgoRecover()
 
 	BeforeEach(func() {
 		namespace := nameGen.generate("e2e-test-")
-		// TODO: The namespace is cached in the client so this will not
```
kargakis: This is still a valid TODO

test/e2e/end_user_test.go
```diff
@@ -51,4 +52,59 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		url := fmt.Sprintf("http://%s", host)
+
+		prevCounter := 0
+		currCounter := 0
+		httpc := &http.Client{
+			Timeout: 10 * time.Second,
+		}
+		// Hit the ingress 3 times, each time the hit counter should increment
+		for i := 0; i < 3; i++ {
+			resp, err := httpc.Get(url)
```
kargakis: I am not sure at this point the application is up and running so we may still want to retry 500s.

kargakis:
Thanks @thekad 

/lgtm

Can deal with any fallout flakes as we see them, just keep a close eye in the test results.

PR #565, opened 2018-10-08
Replace azuredeploy []byte with template map[string]interface{}

### LINKED ISSUE #547: replace azuredeploy []byte with template map[string]interface{} in all function signatures
In reviewing https://github.com/openshift/openshift-azure/pull/521 @jim-minter suggested we change
all azuredeploy []byte to map[string]interface{} 

example:
	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuredeploy []byte, isUpdate bool, Deployer DeployFn) error

We have now changed the custom deploy function to the new parameter, but we should change all of them.

asalkeld:
closes #547

### DIFF
--- a/pkg/plugin/plugin_test.go
+++ b/pkg/plugin/plugin_test.go
@@ -2,7 +2,6 @@ package plugin
 
 import (
 	"context"
-	"strings"
 	"testing"
 
 	"github.com/davecgh/go-spew/spew"
@@ -31,7 +30,7 @@ func TestMerge(t *testing.T) {
 
 	// make old cluster go through plugin first
 	armTemplate := testPluginRun(p, oldCluster, nil, t)
-	if !strings.Contains(string(armTemplate), "\"type\": \"Microsoft.Network/networkSecurityGroups\"") {
+	if !hasResourceType(armTemplate, "Microsoft.Network/networkSecurityGroups") {
 		t.Fatalf("networkSecurityGroups should be applied during cluster creation")
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -128,7 +129,12 @@ func createOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 		return future.WaitForCompletionRef(ctx, deployments.Client())
 	}
 
-	err = p.CreateOrUpdate(ctx, cs, azuredeploy, oldCs != nil, deployer)
+	var azuretemplate map[string]interface{}
```
jim-minter: can we have GenerateARM return map[string]interface{} ?

asalkeld: sure, i can give that a go

pkg/plugin/plugin_test.go
```diff
@@ -65,12 +64,21 @@ func TestMerge(t *testing.T) {
...
 	armTemplate = testPluginRun(p, newCluster, oldCluster, t)
-	if strings.Contains(string(armTemplate), "\"type\": \"Microsoft.Network/networkSecurityGroups\"") {
+	if hasResourceType(armTemplate, "Microsoft.Network/networkSecurityGroups") {
 		t.Fatalf("networkSecurityGroups should not be applied during cluster upgrade")
 	}
 }
 
-func testPluginRun(p api.Plugin, newCluster *api.OpenShiftManagedCluster, oldCluster *api.OpenShiftManagedCluster, t *testing.T) (armTemplate []byte) {
+func hasResourceType(armTemplate map[string]interface{}, resType string) bool {
+	for _, res := range armTemplate["resources"].([]interface{}) {
```
jim-minter: fyi - I implemented a subset of jsonpath which can be useful for this kind of thing - take a look at our use of pkg/jsonpath

jim-minter:
rebased for you

/lgtm

PR #572, opened 2018-10-09
Move createOrUpdate in pkg/fakerp

kargakis:
I need to keep my changes small in between PRs so here I am simply moving the logic of the fake RP into its own package. It's unclear to me now whether this is going to be reused by the e2e framework but my thinking is to move code out of cmd/createorupdate/createorupdate.go, turn it into an http server, and use a separate process to do the PUT request.

/cc jim-minter

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -2,248 +2,21 @@ package main
 
 import (
 	"context"
-	"flag"
-	"fmt"
 	"io/ioutil"
 	"os"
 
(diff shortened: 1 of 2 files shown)


kargakis:
@openshift/sig-azure quite a big change that will affect a couple of PRs, ptal

mjudeikis:
I imagine you will wrap this with http server and we will do curl post from our hack scripts next?

kargakis:
Yes to wrapping, for sending the request I hope to use go.

Issue #550, opened 2018-10-03
PDB gets created during e2e test


kargakis:
```
[0mOpenshift on Azure end user e2e tests [EndUser][0m 
  [1mshould disallow PDB mutations[0m
  [37m/go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:33[0m

[91m[1m• Failure [9.035 seconds][0m
Openshift on Azure end user e2e tests [EndUser]
[90m/go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:16[0m
  [91m[1mshould disallow PDB mutations [It][0m
  [90m/go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:33[0m

  [91mExpected
      <bool>: false
  to equal
      <bool>: true[0m

  /go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:49
```

Probably a different type of error before the permissions can be figured out.
Need to identify what kind of flake it is and get a fix for it in the e2e framework

/kind test-flake
/assign

kargakis:
Turns out not to be an error. Even weirder
```
time="2018-10-03T14:49:05Z" level=info msg="<nil>"
```
Hrm, maybe racing with the sync pod?

kargakis:
sync pod completes two minutes earlier
```
time="2018-10-03T14:47:33Z" level=info msg="Sync process complete"
```

kargakis:
Indeed the PDB seems to get created :/
```
15329:I1003 14:49:05.386810       1 wrap.go:42] POST /apis/policy/v1beta1/namespaces/e2e-test-qc2rj/poddisruptionbudgets: (30.373307ms) 201 [[e2e.test/v0.0.0 (linux/amd64) kubernetes/$Format] 35.229.51.202:33662]
15359:I1003 14:49:05.561935       1 wrap.go:42] PUT /apis/policy/v1beta1/namespaces/e2e-test-qc2rj/poddisruptionbudgets/test/status: (166.785642ms) 200 [[openshift/v1.10.0+b81c8f8 (linux/amd64) kubernetes/b81c8f8/system:serviceaccount:kube-system:disruption-controller] 10.0.0.6:46620]
[...]

PR #579, opened 2018-10-10
Disable PDB api 

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/550


kargakis:
https://github.com/openshift/openshift-azure/issues/582

/retest

jim-minter:
as it stands, it looks like this allows the user to set pdbs again but disables the controller so they don't have any effect?
I'm not sure that we want that behaviour - what was the issue you ran into with disabling the PDBs in the roles not taking effect?

kargakis:
No, this should disable the PDB api in both the apiserver and controller
manager. I have an issue with the cm but the apiserver comes up and does
not recognize pdbs. The issue with the roles was that the bootstrap policy
is embedded in the apiserver and cannot be configured so the apiserver
overwrites any changes to the default roles that we have.

kargakis:
Also the e2e test has to change

PR #583, opened 2018-10-10
Transform fake RP into an http server

kargakis:
@openshift/sig-azure this change transforms our fake rp to run as an http server. In a follow-up PR I am planning on removing the http client in favor of the osa sdk. I am also thinking whether separating the client into its own process makes sense from a testing standpoint.

Part of https://jira.coreos.com/browse/AZURE-165

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"io/ioutil"
+	"net/http"
 	"os"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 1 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
 
-	// read in the external API manifest.
-	b, err := ioutil.ReadFile("_data/manifest.yaml")
+	ok := s.validate(w, req, nil)
+	if !ok {
+		return
+	}
+	defer func() {
+		// drain once we are done processing this request
+		<-s.inProgress
```
kwoodson: How does this work?  It is sending on the inProgress chan struct but I'm not sure I understand how it works.
```
inProgress: make(chan struct{}, 1)}
```

kargakis: This is a receive from the channel to drain it in order to start accepting new requests.

cmd/createorupdate/createorupdate.go
```diff
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"crypto/hmac"
```
jim-minter: is doing this needed if we're listening on localhost?

cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
+
+func main() {
+	logger := logrus.New()
+	logger.Formatter = &logrus.TextFormatter{FullTimestamp: true}
+	logger.SetLevel(logrus.DebugLevel)
+	log := logrus.NewEntry(logger)
+
+	// simulate the RP
+	entry := logrus.NewEntry(logger).WithFields(logrus.Fields{"resourceGroup": os.Getenv("RESOURCEGROUP")})
+	s := &server{log: entry, inProgress: make(chan struct{}, 1)}
```
jim-minter: probably a newServer function is warranted to hide the channel setup

cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
+	}
+	defer func() {
+		// drain once we are done processing this request
+		<-s.inProgress
+	}()
+
+	b, err := ioutil.ReadAll(req.Body)
 	if err != nil {
-		log.Fatal(err)
+		resp := "500 Internal Server Error: Failed to read request body"
```
jim-minter: fwiw these are more probably 400s (invalid request), aren't they?

cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
+
+func (s *server) ListenAndServe() {
+	http.Handle("/", s)
+	httpServer := &http.Server{Addr: "localhost:8080"}
+	s.log.Info("starting server on localhost:8080")
+	logrus.WithError(httpServer.ListenAndServe()).Warn("Server exited.")
+}
+
+func (s *server) validate(w http.ResponseWriter, r *http.Request, expected []byte) bool {
+	// TODO: Support DELETE
```
jim-minter: and also GET - currently we're implementing an asynchronous process synchronously, which is OK, but the real client would poll after the PUT completes

cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
+		resp := "500 Internal Server Error: Failed to marshal response"
+		logrus.Debug(resp, err)
+		http.Error(w, resp, http.StatusInternalServerError)
+		return
+	}
+	w.Write(res)
+}
+
+func (s *server) ListenAndServe() {
+	http.Handle("/", s)
```
jim-minter: TODO: match the request path the real RP would use

cmd/createorupdate/createorupdate.go
```diff
@@ -13,47 +16,147 @@ import (
...
+	logger.SetLevel(logrus.DebugLevel)
+	log := logrus.NewEntry(logger)
+
+	// simulate the RP
+	entry := logrus.NewEntry(logger).WithFields(logrus.Fields{"resourceGroup": os.Getenv("RESOURCEGROUP")})
+	s := &server{log: entry, inProgress: make(chan struct{}, 1)}
+	go s.ListenAndServe()
+
+	// simulate the API call to the RP
+	in, err := ioutil.ReadFile("_data/manifest.yaml")
```
jim-minter: it is almost tempting to push from here downwards out to bash.

cmd/createorupdate/createorupdate.go
kargakis: It depends on how we end up running this in CI. I am thinking that separating the client and the server into separate processes probably makes sense. I still don't know whether that is going to be two separate pods running or two containers in a pod would suffice and we can still use localhost.

kargakis:
lmk if there are any more comments, otherwise this needs a squash

/hold

cmd/createorupdate/createorupdate.go
kargakis: Going to remove until we actually use it.

Issue #581, opened 2018-10-10
logbridge CrashLoopBackOff


mjudeikis:
```
[root@master-000001 tmp]# oc get pods
NAME                        READY     STATUS             RESTARTS   AGE
api-master-000000           1/1       Running            0          1d
api-master-000001           1/1       Running            219        4m
api-master-000002           1/1       Running            5          1d
controllers-master-000000   1/1       Running            0          1d
controllers-master-000001   1/1       Running            3          4m
controllers-master-000002   1/1       Running            3          1d
etcd-master-000000          1/1       Running            0          1d
etcd-master-000001          1/1       Running            336        4m
etcd-master-000002          1/1       Running            1          1d
logbridge-compute-000000    0/1       CrashLoopBackOff   392        1d
logbridge-infra-000000      1/1       Running            1          1d
logbridge-infra-000001      1/1       Running            0          1d
logbridge-master-000000     1/1       Running            0          1d
logbridge-master-000001     1/1       Running            3          4m
logbridge-master-000002     0/1       CrashLoopBackOff   463        1d
sync-master-000000          1/1       Running            0          1d
[root@master-000001 tmp]# oc logs -f logbridge-compute-000000
panic: failed to seek to cursor "": 22

goroutine 1 [running]:
main.main()
        /data/home/jminter/go/src/github.com/openshift/openshift-azure/cmd/logbridge/main.go:194 +0x44
```

mjudeikis:
good node:
```
[root@master-000001 tmp]# cat /var/lib/logbridge/cursor
s=1bb8dfb426fb4938b17c5f996273b44a;i=e00e0;b=7bad86be50d647fa9f4f53b83c392fef;m=26955e33d0;t=577de27c72d74;x=b543b160071b151b
```
broken:
```
[root@master-000002 ~]# cat /var/lib/logbridge/cursor
```

mjudeikis:
cc: @jim-minter

mjudeikis:
+ we might want to rebuild current image:
```
panic: failed to seek to cursor "": 22
goroutine 1 [running]:
main.main()
        /data/home/jminter/go/src/github.com/openshift/openshift-azure/cmd/logbridge/main.go:194 +0x44
```

kargakis:
Ah we don't do any kind of build/promotion for logbridge.

jim-minter:
I'm at a loss to explain the panic on main.go:194.  I fear that the built image does not correspond with the checked in code :-(

[...]

PR #584, opened 2018-10-10
add cursor fix

mjudeikis:
@jim-minter  this should fix https://github.com/openshift/openshift-azure/issues/581

But I dont have deep knowledge on this implementation and journald internals. 

Idea is that we get empty `state`file and it is not validated and causes panic

### DIFF
--- a/cmd/logbridge/main.go
+++ b/cmd/logbridge/main.go
@@ -116,8 +116,23 @@ func startReader(ch chan<- []map[string]interface{}) error {
 	if err != nil {
 		return err
 	}
-
-	cursor, err := readCursor()
+	var cursor string
+	cursor, err = readCursor()
+	if err != nil {
+		return err
+	}
+	// check if cursor file is valid
+	// https://godoc.org/github.com/coreos/go-systemd/sdjournal#Journal.GetCursor
+	if len(cursor) == 0 {
+		_, err := j.Next()
+		if err != nil {
+			return err
+		}
+		cursor, err = j.GetCursor()
+		if err != nil {
+			return err
+		}
+	}
 	if err == nil {
 		err = j.SeekCursor(cursor)
 		if err != nil {


cmd/logbridge/main.go
```diff
@@ -116,8 +116,23 @@ func startReader(ch chan<- []map[string]interface{}) error {
 	if err != nil {
 		return err
 	}
-
-	cursor, err := readCursor()
+	var cursor string
+	cursor, err = readCursor()
+	if err != nil {
```
jim-minter: does not handle the startup case where the state file does not exist

cmd/logbridge/main.go
```diff
@@ -116,8 +116,23 @@ func startReader(ch chan<- []map[string]interface{}) error {
...
+		return err
+	}
+	// check if cursor file is valid
+	// https://godoc.org/github.com/coreos/go-systemd/sdjournal#Journal.GetCursor
+	if len(cursor) == 0 {
+		_, err := j.Next()
```
jim-minter: off-by-one would erroneously skip the first entry in the journal?

jim-minter:
Fundamentally, I don't understand why #581 is happening.  I'm going to close this and open a separate PR which adds protection and logging.

cmd/logbridge/main.go
mjudeikis: I was wondering, what created file in the first place?

cmd/logbridge/main.go
mjudeikis: what is the right behaviour if we don't have a cursor and we can't get with a new one?
I was refering to this from:

> GetCursor gets the cursor of the last journal entry reeferenced by the last completed Next/Previous function call. To call GetCursor, you must first have called one of the Next/Previous functions.

So if GetCursor fails, we call next and try again?

PR #586, opened 2018-10-10
Remove redundant flag from hack scripts

kargakis:
The flag was removed in https://github.com/openshift/openshift-azure/pull/572


jim-minter:
/lgtm

@kargakis any idea about the e2e failure?

PR #587, opened 2018-10-10
protect against state/cursor being empty and add logging

jim-minter:
* ensure that sync and logbridge include a git identifier with clean/dirty info
* protect against crash in #581 where it looks like cursor file is empty
* add logging to understand how crash happened


jim-minter:
@kwoodson ptal

PR #590, opened 2018-10-10
Miscellaneous fixes necessitated by addition of prod subscription

jim-minter:
* add NOGROUPTAGS option to prevent tagging groups with now/ttl (to avoid creating management subscriptions and having them accidentally cleaned up by azure-purge)
* ￼ignore env.* files (to define parameters for management subscriptions in separate env.something files)
* allow overriding of --subscription used by az commands

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -63,11 +63,14 @@ export RESOURCEGROUP=$1
 rm -rf _data
 mkdir -p _data/_out
 
-ttl=76h
-if [[ -n $RESOURCEGROUP_TTL ]]; then
-  ttl=$RESOURCEGROUP_TTL
+if [[ -z "$NOGROUPTAGS" ]]; then
+  ttl=76h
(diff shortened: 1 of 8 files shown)


hack/create.sh
```diff
@@ -63,11 +63,14 @@ export RESOURCEGROUP=$1
...
 
-ttl=76h
-if [[ -n $RESOURCEGROUP_TTL ]]; then
-  ttl=$RESOURCEGROUP_TTL
+if [[ -z "$NOGROUPTAGS" ]]; then
+  ttl=76h
+  if [[ -n "$RESOURCEGROUP_TTL" ]]; then
+    ttl=$RESOURCEGROUP_TTL
+  fi
+  GROUPTAGS=--tags now=$(date +%s) ttl=$ttl
```
kargakis: This one broke our GC, fixing in https://github.com/openshift/openshift-azure/pull/601

jim-minter: oops, thanks @kargakis

PR #592, opened 2018-10-11
Use OSA SDK for sending requests to the fake RP

kargakis:
Builds on top of https://github.com/openshift/openshift-azure/pull/583

Part of https://jira.coreos.com/browse/AZURE-165


kargakis:
@openshift/sig-azure  this depends on https://github.com/openshift/openshift-azure/pull/583 and both are ready for a review. Once we merge both of these PRs, I am planning on splitting out the client into its own binary to enable various azure tests, including tests we need to do against the real RP.

pkg/api/validate.go
```diff
@@ -148,6 +148,10 @@ func validateUpdateContainerService(cs, oldCs *OpenShiftManagedCluster, external
 		}
 		old.Properties.AgentPoolProfiles[i].Count = new.Count
 	}
+	// QUESTION: Unclear whether this is correct. We can track the state
+	// separately inside the fake rp but need to ensure first this is
+	// something that MSFT does not need.
+	old.Properties.ProvisioningState = cs.Properties.ProvisioningState
```
kargakis: Yes, this is probably wrong and users should not be able to change this field. Will remove this and make the fake RP set it after validation.

cmd/createorupdate/createorupdate.go
```diff
@@ -1,25 +1,32 @@
...
 
 type server struct {
 	// the server will not process more than a single
-	// request at all times.
+	// PUT request at all times.
 	inProgress chan struct{}
-	log        *logrus.Entry
+
+	sync.RWMutex
+	oc *v20180930preview.OpenShiftManagedCluster
```
mjudeikis: Will this not cache the previous request into the server? If I send `PUT` with `cluster1` and get succ, and later `PUT` with `cluster2` it will get `cluster1` value from cache?

mjudeikis: Otherwise I need to restart server with each new set of requests?

mjudeikis: Maybe cache should be done with something like https://github.com/golang/groupcache and add the ability to have multiple clusters/request in the same running server? We could host our fake RP outside and do more realistic testing?

kargakis: I don't think we are going to support deploying multiple clusters in our fake RP. Why would we? Anyway, it's out of scope for this PR. As far as this field is concerned, I think we can probably hold just the provisioning state.

cmd/createorupdate/createorupdate.go
```diff
@@ -45,10 +52,50 @@ func (s *server) ListenAndServe() {
...
+		return false
+	}
+
+	if r.Method == http.MethodPut {
+		select {
+		case s.inProgress <- struct{}{}:
+			// continue
+		default:
+			// did not get the lock
+			resp := "423 Locked: Processing another in-flight request"
```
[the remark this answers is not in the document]
kargakis: I don't think the real RP queues requests but @amanohar @sozercan can confirm. I defintely am not going to implement any queueing mechanism here :)

cmd/createorupdate/createorupdate.go
kargakis: Actually we also need to hold oc in order to reply to GET requests with it and the provisioning state.

kargakis:
https://github.com/openshift/openshift-azure/issues/600

/retest

jim-minter:
@kargakis I thought that until we'd simplified the loop of how we update the public api in openshift-azure -> regenerated the sdk -> revendored the sdk into openshift-azure we were just going to do basic HTTP client calls using the public api defined in openshift-azure?

kargakis:
> @kargakis I thought that until we'd simplified the loop of how we update the public api in openshift-azure -> regenerated the sdk -> revendored the sdk into openshift-azure we were just going to do basic HTTP client calls using the public api defined in openshift-azure?

I am on the fence. Doing basic HTTP calls is not great because we don't get to reuse the rest of the SDK machinery (futures, polling). Reworking the SDK to reuse our types is not great either. If we want to ensure parity between the types, we can have a reflective test. Once we figure out a home for the SDK, removing the conversions and the (now missing) reflective test should be easier than reworking any other option.

kargakis:
@jim-minter I've added two commits, one that brings the api types in field parity and another commit that is a reflective test that ensures our types stay in sync.

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -8,13 +12,15 @@ const (
 // OpenShiftManagedCluster complies with the ARM model of resource definition in
 // a JSON template.
 type OpenShiftManagedCluster struct {
+	autorest.Response `json:"-"`
+
```
asalkeld: should we be changing this? I thought that these versioned types were locked in stone.

kargakis: The only change is the addition of autorest.Response that is included in the SDK type and we were missing. The rest of the diff matches the order of our fields with that of the SDK type.

PR #593, opened 2018-10-11
Generate coverage profile during unit tests

kargakis:
@asalkeld this change should help us start collecting coverage profiles in CI tests w/o trying to open the profile in html. There is also one update that needs to happen in openshift/release.

/cc asalkeld


kargakis:
Opened https://github.com/openshift/release/pull/1896 for the release part.

PR #598, opened 2018-10-15
Prod RP requires name to match resource group

kargakis:
When I try to use the OSA sdk and the manifest we currently generate to start a cluster via the prod RP, I get the following error:
```
  containerservice.OpenShiftManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="MismatchingResourceName" Message="The provided resource name 'openshift' did not match the name in the Url 'kargakisprode2es'."
```
Seems that the prod RP requires name to match the resource group. I can start clusters successfully with this change.

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -83,7 +83,7 @@ fi
 set -x
 
 cat >_data/manifest.yaml <<EOF
-name: openshift
+name: $RESOURCEGROUP
 location: $AZURE_REGION
 properties:
   openShiftVersion: "$DEPLOY_VERSION"


kargakis:
https://github.com/openshift/openshift-azure/issues/599

/retest

jim-minter:
@amanohar is this an expected limitation?

amanohar:
@jim-minter there is no such limitation. Our tests use a different resource name and resource group name.

@kargakis can you share sample SDK code that resulted in this error? It could be a Go SDK bug in constructing the request URL. cc: @julienstroheker in case he is aware of this issue.

kargakis:
@amanohar @julienstroheker it's actually very easy to reproduce. Change the [`name`](https://github.com/openshift/openshift-azure/blob/e9c2d8a6b8b42909a8ca49df48bcaed68ec54fba/hack/create.sh#L86) field to something other than the resource group in the request we use for testing, then run `TEST_IN_PRODUCTION=true ./hack/create.sh yourresourcegroup`.

PR #601, opened 2018-10-15
Fix applying tags in resource groups

kargakis:


kargakis:
Turned into a grab bag PR but anyway :)

test/e2e/end_user_test.go
```diff
@@ -58,6 +58,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 		var regex = regexp.MustCompile(`id="count-value">(\d+)<`)
 
 		// Create the template
+		By(fmt.Sprintf("creating the template (%v)", time.Now()))
```
jim-minter: ginkgo can't be set up to log the time on by()s?

kargakis: Pretty sure it can, will be cleaned up in a follow up

Issue #602, opened 2018-10-15
Refactor azure client creation in func (u *simpleUpgrader) InitializeCluster


jim-minter:
`func (u *simpleUpgrader) InitializeCluster(` creates (a) an azure storage account client and (b) a storage client, inline.  It would be nice to pull the client creation out for a cleaner architecture, esp for unit testing.

At the moment, doing this for (a) is blocked by the fact that it uses the credentials stored in ctx, and we don't know if these can be made available in NewPlugin.

(b) is blocked by the fact that we need to discover the config storage account keys to create the storage client.  I don't have a clear idea about what we can do to make this cleaner but it'd be good to look into.

jim-minter:
@pweil- @asalkeld @kargakis @amanohar @sozercan fyi

asalkeld:
Thanks @jim-minter , I was also not a fan of creating the clients inline, but didn't see a way out of it immediately.

kargakis:
We should try to do the same thing with the k8s client too.

PR #608, opened 2018-10-16
Add unittests for drain

asalkeld:


pkg/upgrade/drain.go
```diff
@@ -20,12 +20,14 @@ import (
 )
 
 func (u *simpleUpgrader) drain(ctx context.Context, cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole, nodeName string) error {
-	kc, err := managedcluster.ClientsetFromV1Config(cs.Config.AdminKubeconfig)
-	if err != nil {
-		return err
+	var err error
+	if u.kubeclient == nil {
```
kargakis: I would prefer to see us set this during the construction of simpleUpgrader as opposed to doing it here. It keeps the logic of `drain` clearer and does not litter the code with conditionals that are helping unit testing which is not ideal.

pkg/upgrade/drain.go
```diff
@@ -20,12 +20,14 @@ import (
...
+		u.kubeclient, err = managedcluster.ClientsetFromV1Config(cs.Config.AdminKubeconfig)
```
kargakis: I know that you are not changing this but I don't think `drain` should need access to admin kubeconfig in order to perform the drain. Can you open a follow up issue to scope this down?

asalkeld: https://github.com/openshift/openshift-azure/issues/609

pkg/upgrade/drain.go
kargakis: Setting up the client does not require any sort of connection, you just need a kubeconfig and you have that.

asalkeld: from slack: "I think for now it's probably fine to do what you started doing

and we can probably clean it up as part of https://github.com/openshift/openshift-azure/issues/602

pkg/upgrade/drain_test.go
```diff
@@ -0,0 +1,114 @@
...
+				TypeMeta: metav1.TypeMeta{
+					Kind: "pod",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-pod",
+				},
+			}),
+		},
+	}
+	log.New(logrus.NewEntry(logrus.New()))
```
kargakis: Why do you need this?

pkg/upgrade/drain_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		{
+			name:     "unknown-role",
+			role:     "cant-find-this",
+			nodeName: "master-000000",
+			expectedActions: [][]string{
+				{"get", "nodes"}},
+			wantErr: true,
+			kubeclient: fake.NewSimpleClientset(&corev1.Node{
+				TypeMeta: metav1.TypeMeta{
```
kargakis: You shouldn't need to set typemeta I think.

pkg/upgrade/drain_test.go
```diff
@@ -0,0 +1,114 @@
...
+			wantErr:    false,
+			kubeclient: fake.NewSimpleClientset(),
+		},
+		{
+			name:     "unknown-role",
+			role:     "cant-find-this",
+			nodeName: "master-000000",
+			expectedActions: [][]string{
+				{"get", "nodes"}},
+			wantErr: true,
```
kargakis: Instead of a boolean, please compare the actual errors.

pkg/upgrade/drain_test.go
asalkeld: this log call causes an segfault otherwise: https://github.com/openshift/openshift-azure/blob/master/pkg/upgrade/drain.go#L32

pkg/upgrade/drain_test.go
asalkeld: ok, didn't realise that. removing..

pkg/upgrade/drain_test.go
asalkeld: will do

Issue #606, opened 2018-10-15
Possibly failed to write config blob and didn't error


jim-minter:
See https://bugzilla.redhat.com/show_bug.cgi?id=1639396

AFAICS the config blob *container* does not exist - perhaps this could have failed to create because of Azure rate limiting issues?  Can you audit the code and see if you can find a path whereby this could have occurred?

kargakis:
I think it's possible that sync can cache everything from Azure and never go live again during its lifetime. There are cases where the config blob will need to change more than once during a single version (eg. we rotate storage keys), we could require a sync restart in those cases. Today, any change in the config is a restart of the VMs so sync should always restart.

mjudeikis:
yes, so this is mainly config blob reading operation. But key reading operation created 2 more reads (list keys and get key). We could cache key and use it to get config blob. If we get Access denied at some point we retry to acquire a new key. In this case, we potentially reduce all reads by 2/3.

kargakis:
We seem to return error when either the [container](https://github.com/openshift/openshift-azure/blob/547524d316244ce67ee273fca52eda5a63507282/pkg/upgrade/initialize.go#L46) or the [blob](https://github.com/openshift/openshift-azure/blob/547524d316244ce67ee273fca52eda5a63507282/pkg/upgrade/initialize.go#L56) gets created. Without looking into the sync log we can only assume what's the actual error we are trying to solve. Caching is probably a good idea at least as long as every config update means a VM restart.

kargakis:
I don't fully understand the existing unit test that we have but it seems it can be augmented to test those code paths.

https://github.com/openshift/openshift-azure/blob/547524d316244ce67ee273fca52eda5a63507282/pkg/upgrade/initialize_test.go#L19-L54

PR #610, opened 2018-10-16
Cache storage client keys in sync

kargakis:
This PR updates the sync process to cache the storage client to avoid listing keys more than once to avoid getting ratelimited since the storage RP limits seem to be lower than other RP limits.

Closes https://github.com/openshift/openshift-azure/issues/606


kargakis:
Discussed with Jim, we agreed that we don't know whether we need to cache or not the config blob so just caching the keys for now should be enough.

mjudeikis:
lgtm, will wait for @asalkeld input

PR #612, opened 2018-10-17
Add unittest for wait.ForHTTPStatusOk

asalkeld:


pkg/util/wait/wait_test.go
```diff
@@ -0,0 +1,55 @@
...
+			wantErr:            false,
+		},
+		{
+			name:    "no response",
+			timeout: 1 * time.Second,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		i := 0
```
kargakis: Is there a reason you define the index here vs having it in the for loop?
```
for i, tt := range tests
```

pkg/util/wait/wait_test.go
```diff
@@ -0,0 +1,55 @@
...
+		responder := func(res http.ResponseWriter, req *http.Request) {
+			res.WriteHeader(tt.expectedRespStatus[i])
+			i++
+			res.Write([]byte("body"))
+		}
+		testServer := httptest.NewServer(http.HandlerFunc(responder))
+		testURL := testServer.URL
```
kargakis: nit: inline testServer.URL in `forHTTPStatusOkWithTimeout`.

pkg/util/wait/wait_test.go
kargakis: Ah sorry you are indexing based on the length of expectedRespStatus. nvmd

pkg/util/wait/wait_test.go
```diff
@@ -0,0 +1,55 @@
...
+			name:               "an error then ok",
+			expectedRespStatus: []int{500, 200},
+			timeout:            2 * time.Second,
+			wantErr:            false,
+		},
+		{
+			name:    "no response",
+			timeout: 1 * time.Second,
+			wantErr: true,
+		},
```
kargakis: I am wondering how can we structure these tests to allow unit-testing the syscall.ENETUNREACH and err.Timeout() || err.Err == io.EOF || err.Err == io.ErrUnexpectedEOF branches too. We probably need to mock `cli.Do(req)` which then would not require the current httptest setup that you have currently.

kargakis:
nice!

/lgtm

PR #613, opened 2018-10-17
real rp e2e tests - verify customer updates not possible

charlesakalugwu:
This PR involves all the work required for the resolution of https://jira.coreos.com/browse/AZURE-160


pkg/util/azureclient/helpers.go
```diff
@@ -34,3 +39,47 @@ func ExternalToSdk(in *v20180930preview.OpenShiftManagedCluster) (out *sdk.OpenS
 
 	return
 }
+
+// ResourceGroupFromManagedResourceID returns the resource group name from  a managed resource group identifier. The managed resource
+// group id for managed applications is formatted as follows `/subscriptions/(.+)/resourceGroups/(.+)`
+func ResourceGroupFromManagedResourceID(resourceID string) (string, error) {
```
kargakis: Move all these helpers inside `test/e2erp`. I'd prefer that and moving them back to this package if we ever need to use in prod code vs dumping everything here.

test/e2erp/compute.go
```diff
@@ -0,0 +1,203 @@
+package e2erp
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/openshift/openshift-azure/pkg/api"
```
kargakis: nit: split these imports as elsewhere:
stdlib

third-party

openshift-azure

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+
+		scaleSets, err := ScaleSets(ctx, logger, c.ssc, mrg)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(scaleSets).NotTo(And(BeNil(), BeEmpty()))
+		Expect(len(scaleSets)).Should(BeNumerically(">=", 3))
+
+		// TODO: get detailed error and match on them since we expect the customer to see errors with Code=ScopeLocked
+		var errs []error
+
+		logger.Info("Updating the scale set instance count")
```
kargakis: Instead of logrus we should be using the ginkgo framework for this. There is a By function that we use to print info.

test/e2erp/realrp_test.go
```diff
@@ -3,8 +3,17 @@
 package e2erp
 
 import (
+	"context"
+	"os"
+
+	"github.com/openshift/openshift-azure/pkg/log"
```
kargakis: nit: fix the order in these imports

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
 		var err error
 		Expect(err).NotTo(HaveOccurred())
 	})
+
+	It("should not be possible for customer to mutate an osa scale set", func() {
+
```
kargakis: kill this newline

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+		ctx := context.Background()
+		ctx = context.WithValue(ctx, api.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
+		ctx = context.WithValue(ctx, api.ContextKeyClientSecret, os.Getenv("AZURE_CLIENT_SECRET"))
+		ctx = context.WithValue(ctx, api.ContextKeyTenantID, os.Getenv("AZURE_TENANT_ID"))
+
+		logrus.SetLevel(log.SanitizeLogLevel("Debug"))
+		logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+		logger := logrus.WithFields(logrus.Fields{"location": c.location, "resourceGroup": c.resourceGroup})
+
+		arg := azureclient.ApplicationResourceGroup(c.resourceGroup, c.resourceGroup, c.location)
```
kargakis: arg is not very intuitive. maybe appRg, appRG or something along those lines is clearer

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+		Expect(arg).NotTo(And(BeNil(), BeEmpty()))
+		logger.Infof("application resource group is %s", arg)
+
+		// update the logger to use the application resource group
+		logger = logrus.WithFields(logrus.Fields{"location": c.location, "resourceGroup": arg})
+
+		mrg, err := azureclient.ManagedResourceGroup(ctx, arg)
```
kargakis: managedRg

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(arg).NotTo(And(BeNil(), BeEmpty()))
```
kargakis: managedRg

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+		logger.Infof("managed resource group is %s", mrg)
+
+		// update the logger to use the managed resource group
+		logger = logrus.WithFields(logrus.Fields{"location": c.location, "resourceGroup": mrg})
+
+		scaleSets, err := ScaleSets(ctx, logger, c.ssc, mrg)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(scaleSets).NotTo(And(BeNil(), BeEmpty()))
+		Expect(len(scaleSets)).Should(BeNumerically(">=", 3))
```
kargakis: Are customers able to see all of these resources? If so, then this should be == 3, right?

test/e2erp/compute.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return vms, nil
+}
+
+// UpdateScaleSetsCapacityErrors returns a slice of the errors it encounters as it attempts to increment the capacity
+// (by one) for all scale sets within a given resource group
+func UpdateScaleSetsCapacityErrors(ctx context.Context, logger *logrus.Entry, ssc azureclient.VirtualMachineScaleSetsClient, ssvmc azureclient.VirtualMachineScaleSetVMsClient, resourceGroup string) []error {
```
kargakis: Errors is probably redundant as a suffix in these helpers.

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
...
+		// TODO: get detailed error and match on them since we expect the customer to see errors with Code=ScopeLocked
+		var errs []error
+
+		logger.Info("Updating the scale set instance count")
+		errs = UpdateScaleSetsCapacityErrors(ctx, logger, c.ssc, c.ssvmc, mrg)
+		Expect(errs).NotTo(BeNil())
+		Expect(len(errs)).To(BeEquivalentTo(len(scaleSets)))
+
+		logger.Info("Updating the scale set instance type")
+		errs = UpdateScaleSetsInstanceTypeErrors(ctx, logger, c.ssc, mrg)
```
mjudeikis: I might be looking this without context, but I would expect methods like `UpdateScaleSetsInstanceType` and validate errors here. `Errors` ant the end of methods looks like unnecessary?

test/e2erp/realrp_test.go
```diff
@@ -14,4 +23,72 @@ var _ = Describe("Resource provider e2e tests [Real]", func() {
 		var err error
 		Expect(err).NotTo(HaveOccurred())
```
kargakis: While here, remove this dummy test.

charlesakalugwu: done

test/e2erp/compute.go
charlesakalugwu: I think the 'Errors' keyword in those function names are necessary to point out that those functions, instead of retrieving various azure resources, are only involved in retrieving errors.

test/e2erp/realrp_test.go
charlesakalugwu: I think the 'Errors' keyword in those function names are necessary to point out that those functions, instead of retrieving various azure resources, are only involved in retrieving errors.

test/e2erp/realrp_test.go
charlesakalugwu: Yes it could be set equal to 3 but `>= 3` allows this to work when we have more than a group of compute nodes (multiple compute node profiles) in the future. The idea is that each clusters needs at least 1 master scale set, 1 infra scale set and 1 compute scale set.

test/e2erp/realrp_test.go
charlesakalugwu: done ...changed to `appRg`

test/e2erp/realrp_test.go
charlesakalugwu: done

test/e2erp/realrp_test.go
charlesakalugwu: done

test/e2erp/realrp_test.go
charlesakalugwu: done

test/e2erp/realrp_test.go
charlesakalugwu: from the docs I only see support for a `GingkoWriter`. Is this what you are referring to?

test/e2erp/compute.go
charlesakalugwu: done

pkg/util/azureclient/helpers.go
charlesakalugwu: done

test/e2erp/realrp_test.go
charlesakalugwu: Lets leave the logrus as is and discuss potentially updating to `GingkoWriter` in the next sprint

test/e2erp/realrp_test.go
charlesakalugwu: Sorry...I don't agree with this. There's nothing wrong with having a look ahead and planning or writing code to take future situations into account as long as the present is also taken into account.

test/e2erp/realrp_test.go
```diff
@@ -3,15 +3,84 @@
...
+		errs = UpdateScaleSetsCapacityErrors(ctx, logger, c.ssc, c.ssvmc, managedRg)
+		Expect(errs).NotTo(BeNil())
+		Expect(len(errs)).To(BeEquivalentTo(len(scaleSets)))
+
+		logger.Info("Updating the scale set instance type")
+		errs = UpdateScaleSetsInstanceTypeErrors(ctx, logger, c.ssc, managedRg)
+		Expect(errs).NotTo(BeNil())
+		Expect(len(errs)).To(BeEquivalentTo(len(scaleSets)))
+
+		logger.Info("Updating the scale set SSH key")
```
kargakis: I would put each test in its own spec so we will not be maintaining a monolithic test but can be done in a follow-up.

test/e2erp/compute.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+)
+
+const (
+	fakepubkey = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7laRyN4B3YZmVrDEZLZoIuUA72pQ0DpGuZBZWykCofIfCPrFZAJgFvonKGgKJl6FGKIunkZL9Us/mV4ZPkZhBlE7uX83AAf5i9Q8FmKpotzmaxN10/1mcnEE7pFvLoSkwqrQSkrrgSm8zaJ3g91giXSbtqvSIj/vk2f05stYmLfhAwNo3Oh27ugCakCoVeuCrZkvHMaJgcYrIGCuFo6q0Pfk9rsZyriIqEa9AtiUOtViInVYdby7y71wcbl0AbbCZsTSqnSoVxm2tRkOsXV6+8X4SnwcmZbao3H+zfO1GBhQOLxJ4NQbzAa8IJh810rYARNLptgmsd4cYXVOSosTX azureuser"
+)
+
+// ResourceGroupFromManagedResourceID returns the resource group name from  a managed resource group identifier. The managed resource
+// group id for managed applications is formatted as follows `/subscriptions/(.+)/resourceGroups/(.+)`
+func ResourceGroupFromManagedResourceID(resourceID string) (string, error) {
```
jim-minter: use azure.ParseResourceID from the Go SDK

PR #619, opened 2018-10-17
Don't print the diff if the object kind is Secret

### LINKED ISSUE #504: sync process logs changes to secrets
The sync pod ensures that the actual state of OpenShift objects matches the desired state.
Any time when there is a mismatch and the sync pod updates an object, it logs the delta to stdout.
We should ensure that if the changed object is a Secret, that we do not log the secret contents to stdout.

===

And we gather sync logs in CI now. We will need to fix printing secret content in the logs before we do any testing with secret rotation.

/priority important-soon

thekad:
Fixes #504
We don't want secret values to be logged *anywhere*

### DIFF
--- a/pkg/addons/client.go
+++ b/pkg/addons/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"reflect"
+	"strings"
 
 	"github.com/go-test/deep"
 	log "github.com/sirupsen/logrus"
@@ -240,8 +241,13 @@ func needsUpdate(existing, o *unstructured.Unstructured) bool {
 	// TODO: we should have tests that monitor these diffs:
 	// 1) when a cluster is created
 	// 2) when sync is run twice back-to-back on the same cluster
-	for _, diff := range deep.Equal(*existing, *o) {
-		log.Infof("- " + diff)
+
+	// Don't show a diff if kind is Secret
+	oGroupKind := o.GroupVersionKind().GroupKind()
+	if strings.ToLower(oGroupKind.String()) != "secret" {
+		for _, diff := range deep.Equal(*existing, *o) {
+			log.Infof("- " + diff)
+		}
 	}
 
 	return true
--- a/pkg/addons/client.go
+++ b/pkg/addons/client.go
@@ -219,6 +219,7 @@ func write(dyn dynamic.ClientPool, grs []*discovery.APIGroupResources, o *unstru
 		if !needsUpdate(existing, o) {
 			return
 		}
+		printDiff(existing, o)
 
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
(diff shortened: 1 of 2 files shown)


thekad:
@kargakis @mjudeikis ptal

pkg/addons/client_test.go
```diff
@@ -25,6 +25,12 @@ func TestNeedsUpdate(t *testing.T) {
 			updated:  getObjectFromFile("testdata/service3.yaml"),
 			exp:      true,
 		},
+		{
```
kargakis: This test does not test the new code that you have added.

pkg/addons/client.go
```diff
@@ -240,8 +241,13 @@ func needsUpdate(existing, o *unstructured.Unstructured) bool {
...
 	// 2) when sync is run twice back-to-back on the same cluster
-	for _, diff := range deep.Equal(*existing, *o) {
-		log.Infof("- " + diff)
+
+	// Don't show a diff if kind is Secret
+	oGroupKind := o.GroupVersionKind().GroupKind()
+	if strings.ToLower(oGroupKind.String()) != "secret" {
+		for _, diff := range deep.Equal(*existing, *o) {
+			log.Infof("- " + diff)
+		}
```
asalkeld: do we want to  log the name of the secret, just so we know that sync recognises that it changed?

kargakis: Printed at L239

asalkeld: thanks, missed that

pkg/addons/client.go
```diff
@@ -240,8 +241,13 @@ func needsUpdate(existing, o *unstructured.Unstructured) bool {
 	// TODO: we should have tests that monitor these diffs:
 	// 1) when a cluster is created
 	// 2) when sync is run twice back-to-back on the same cluster
-	for _, diff := range deep.Equal(*existing, *o) {
-		log.Infof("- " + diff)
+
+	// Don't show a diff if kind is Secret
+	oGroupKind := o.GroupVersionKind().GroupKind()
+	if strings.ToLower(oGroupKind.String()) != "secret" {
```
kargakis: The simplest thing to do here in order to test this is to move this part into a  `printDiff(existing, o) bool` function that returns true when something is printed and false otherwise, then write tests for that function.

thekad: thanks @kargakis you are right

pkg/addons/client_test.go
```diff
@@ -12,25 +12,38 @@ func TestNeedsUpdate(t *testing.T) {
...
+			exp:      true,  // upgrade expected
+			diff:     false, // no diff expected
 		},
 	}
 
 	for _, test := range tests {
 		if got := needsUpdate(test.existing, test.updated); got != test.exp {
 			t.Errorf("%s: expected update %t, got %t", test.name, test.exp, got)
 		}
+		if got := printDiff(test.existing, test.updated); got != test.diff {
```
kargakis: This unit test is specifically about needsUpdate (`TestNeedsUpdate`). You will have to create your own test function (`TestPrintDiff`).

pkg/addons/client_test.go
```diff
@@ -6,31 +6,48 @@ import (
...
-	for _, test := range tests {
+func TestNeedsUpdate(t *testing.T) {
+	for _, test := range clientTests {
 		if got := needsUpdate(test.existing, test.updated); got != test.exp {
 			t.Errorf("%s: expected update %t, got %t", test.name, test.exp, got)
 		}
 	}
+}
 
+func TestShouldPrintDiff(t *testing.T) {
```
kargakis: nit: unit tests are usually named as `Test` followed by the name of the function that is unit-tested in camel-case. In this case `TestPrintDiff`.

Issue #595, opened 2018-10-12
Rewrite hack/dns.sh in Go


kargakis:
https://github.com/openshift/openshift-azure/pull/594 creates a new e2e suite for running against the RP (either our fake or the production RP). In the case of the real RP, all the DNS configuration should be handled by MSFT, in case of the fake RP we still need to setup the DNS zone and records by ourselves.

This means that we need to shell out to `hack/dns.sh` after the suite has installed a cluster which is not ideal.

mjudeikis:
I would like try this one @kargakis if you ok with it? Do you imagine this as package in our utils folder which we can call from e2e fake RP?

kargakis:
Go ahead @mjudeikis ! I started writting some code locally and had added it in `pkg/util/azureclient/dns` but never got a chance to push it forward.
```go
package dns

import (
	"context"
	"fmt"

	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
)

func int64p(n int) *int64 {
	n64 := int64(n)
	return &n64
}

func CreateZone(ctx context.Context, subscriptionID, resourceGroup, dnsResourceGroup, dnsDomain string) error {
	zoneName := fmt.Sprintf("%s.%s", resourceGroup, dnsDomain)
	if _, err := dns.NewZonesClient(subscriptionID).CreateOrUpdate(ctx, dnsResourceGroup, zoneName, dns.Zone{}, "", ""); err != nil {
		return err
	}

	fqdn := fmt.Sprintf("%s.", zoneName)
	soa := dns.RecordSet{
		TTL:  int64p(3600),
		Fqdn: &fqdn,
		RecordSetProperties: &dns.RecordSetProperties{
			SoaRecord: &dns.SoaRecord{
				RefreshTime: int64p(60),
				RetryTime:   int64p(60),
				ExpireTime:  int64p(60),
				MinimumTTL:  int64p(60),
			},
		},
	}
	if _, err := dns.NewRecordSetsClient(subscriptionID).CreateOrUpdate(ctx, dnsResourceGroup, zoneName, "@", dns.SOA, soa, "", ""); err != nil {
		return err
	}

	ns := dns.RecordSet{
		TTL:  int64p(60),
		Fqdn: &fqdn,
	}
	_, err := dns.NewRecordSetsClient(subscriptionID).CreateOrUpdate(ctx, dnsResourceGroup, zoneName, "@", dns.NS, ns, "", "")
	return err
}
```

[...]

PR #621, opened 2018-10-18
move dns.sh to go

mjudeikis:
very work in progress. At the moment it dos not work because of cname creation. Dont know why and where is the issue. 

Fixes https://github.com/openshift/openshift-azure/issues/595


mjudeikis:
/cc asalkeld 
I think we took 2 different approaches. Will send an email with what I was thinking behind my implementation and we can align :)

mjudeikis:
Need to wait until fakeRD delete code lands to finish

pkg/util/azureclient/dns/dns.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+
+	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
+)
```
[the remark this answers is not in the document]
mjudeikis: moved

pkg/fakerp/dns.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		nsServerList = append(nsServerList, rec)
+	}
+
+	// update default SOA record
+	rsc := azureclient.NewRecordSetsClient(subscriptionID, authorizer, config.AcceptLanguages)
+	soa := dns.RecordSet{
+		Etag: zone.Etag,
+		RecordSetProperties: &dns.RecordSetProperties{
+			TTL: int64p(3600),
```
asalkeld: import "github.com/Azure/go-autorest/autorest/to"

to.Int64Str()

kargakis:
/uncc

let me know when you need a review here

jim-minter:
@mjudeikis bump

mjudeikis:
This is almost ready, need few hours to get back to this. Will do this today.

mjudeikis:
this does not yet work for some strange reasons. Records looks identical, but SDK one is not being served by DNS server

mjudeikis:
@kargakis @charlesakalugwu ready to review. Functions are quite big, not they are very unique and splitting does not make a lot of sense. Let me know WDYT

pkg/fakerp/fakerp.go
```diff
@@ -54,6 +54,13 @@ func CreateOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 
 		log.Info("merge old and new config")
 		p.MergeConfig(ctx, cs, oldCs)
+	} else {
```
kargakis: We should probably create the DNS records only after the creation is complete.

pkg/fakerp/server.go
```diff
@@ -101,6 +101,19 @@ func (s *Server) validate(w http.ResponseWriter, r *http.Request) bool {
 }
 
 func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
+
```
kargakis: kill this newline

pkg/fakerp/server.go
```diff
@@ -101,6 +101,19 @@ func (s *Server) validate(w http.ResponseWriter, r *http.Request) bool {
...
+	config := &api.PluginConfig{
+		AcceptLanguages: []string{"en-us"},
```
kargakis: Not necessary now but in the future we can keep this inside a `map[resourceGroup]PluginConfig; type resourceGroup string` so that it can stay in sync between PUT and DELETE.

pkg/fakerp/fakerp.go
mjudeikis: I got the impression that RP creates it before. Need to check

pkg/fakerp/server.go
```diff
@@ -109,13 +122,18 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 		http.Error(w, resp, http.StatusInternalServerError)
 		return
 	}
+
+	// delete dns records
+	err = DeleteOCPDNS(ctx, s.conf.SubscriptionID, s.conf.ResourceGroup, s.conf.DnsResourceGroup, s.conf.DnsDomain, config)
+	if err != nil {
+		s.log.Fatal(err)
```
kargakis: Can't be `s.log.Fatal(err)` but similar to the rest of the error handling we do in the server.

pkg/fakerp/server.go
```diff
@@ -101,6 +101,19 @@ func (s *Server) validate(w http.ResponseWriter, r *http.Request) bool {
 }
 
 func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
+
+	config := &api.PluginConfig{
+		AcceptLanguages: []string{"en-us"},
+	}
+
+	// simulate Context with property bag
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
```
kargakis: Why the timeout change here?

pkg/util/azureclient/dns.go
```diff
@@ -0,0 +1,64 @@
+package azureclient
```
kargakis: I wouldn't add the mocked clients here if we are going to use them only in the fake rp.

pkg/fakerp/dns.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// update main NS records
+	ns = dns.RecordSet{
+		RecordSetProperties: &dns.RecordSetProperties{
+			TTL:       to.Int64Ptr(60),
+			NsRecords: &nsServerList,
+		},
+	}
+	_, err = rsc.CreateOrUpdate(ctx, dnsResourceGroup, dnsDomain, resourceGroup, dns.NS, ns, "", "")
+	if err != nil {
```
kargakis: nit: `return err`

pkg/fakerp/dns.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+
+	if err := future.WaitForCompletionRef(ctx, zc.Client()); err != nil {
+		return err
+	}
+
+	// delete main zone NS record
+	_, err = rsc.Delete(ctx, dnsResourceGroup, dnsDomain+"-test", resourceGroup, dns.NS, "")
+	if err != nil {
```
kargakis: nit: `return err`

pkg/fakerp/server.go
mjudeikis: I used the same context for both objects now. We can reduce it to 30 min if there was a reason for it.

pkg/fakerp/fakerp.go
kargakis: If that's the case, then fine.

pkg/fakerp/server.go
kargakis: The reason is that 1 hour for a DELETE is ridiculously long. If it starts taking longer than 30m, we need to know. Even 30m is a lot so we may want to reduce but I wouldn't do that now :)

pkg/util/azureclient/dns.go
kargakis: We are getting mocked clients for free soon via the SDK, see https://github.com/openshift/openshift-azure/issues/725

The more we add now, the harder it will be to move away in the future.

mjudeikis: dam! good dam! :D

pkg/fakerp/server.go
```diff
@@ -170,7 +190,7 @@ func (s *Server) handlePut(w http.ResponseWriter, req *http.Request) {
 	s.write(oc)
 
 	// simulate Context with property bag
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
```
mjudeikis: aligned with create context

kargakis: Please revert back to 1h, this is also responsible for upgrades.

PR #622, opened 2018-10-18
Setup the clients in NewSimpleUpgrader

### LINKED ISSUE #602: Refactor azure client creation in func (u *simpleUpgrader) InitializeCluster

asalkeld:
- there is an API change to NewPlugin
- This saves recreating the client all over the place
- makes testing easier

// now NewPlugin take a context and a clusterConfig so we can create all the clients in one go
p := plugin.NewPlugin(ctx, entry, config, cs)

closes #602


jim-minter:
@asalkeld I do not understand how this can work - a) have you been able to check with MSFT that this is acceptable, and b) how can we ListKeys on a SA that doesn't exist yet?  It isn't created until the Deploy is done.

asalkeld:
> @asalkeld I do not understand how this can work - a) have you been able to check with MSFT that this is acceptable, and b) how can we ListKeys on a SA that doesn't exist yet? It isn't created until the Deploy is done.

@jim-minter I'll post a new one now that takes a different approach.

pkg/upgrade/types.go
```diff
@@ -27,10 +28,12 @@ type Upgrader interface {
...
-	pluginConfig   api.PluginConfig
-	accountsClient azureclient.AccountsClient
-	storageClient  storage.Client
-	kubeclient     kubernetes.Interface
+	pluginConfig                    api.PluginConfig
+	accountsClient                  azureclient.AccountsClient
+	storageClient                   storage.Client
+	virtualMachineScaleSetVMsClient azureclient.VirtualMachineScaleSetVMsClient
+	virtualMachineScaleSetsClient   azureclient.VirtualMachineScaleSetsClient
+	adminKubeClient                 kubernetes.Interface
```
kargakis: `s/adminKubeClient/kubeClient`

We have an issue about stop using admin kubeconfig during upgrades, no need to name the client field after it tho here.

pkg/upgrade/types.go
```diff
@@ -27,10 +28,12 @@ type Upgrader interface {
...
 
 type simpleUpgrader struct {
-	pluginConfig   api.PluginConfig
-	accountsClient azureclient.AccountsClient
-	storageClient  storage.Client
-	kubeclient     kubernetes.Interface
+	pluginConfig                    api.PluginConfig
+	accountsClient                  azureclient.AccountsClient
+	storageClient                   storage.Client
+	virtualMachineScaleSetVMsClient azureclient.VirtualMachineScaleSetVMsClient
```
kargakis: vmc

pkg/upgrade/types.go
```diff
@@ -27,10 +28,12 @@ type Upgrader interface {
...
+	virtualMachineScaleSetsClient   azureclient.VirtualMachineScaleSetsClient
```
kargakis: ssc

pkg/upgrade/initialize.go
```diff
@@ -6,25 +6,15 @@ import (
...
-
-		u.accountsClient = azureclient.NewAccountsClient(cs.Properties.AzProfile.SubscriptionID, authorizer, u.pluginConfig.AcceptLanguages)
-	}
 	if u.storageClient == nil {
 		keys, err := u.accountsClient.ListKeys(ctx, cs.Properties.AzProfile.ResourceGroup, cs.Config.ConfigStorageAccount)
 		if err != nil {
 			return err
 		}
-
 		u.storageClient, err = storage.NewClient(cs.Config.ConfigStorageAccount, *(*keys.Keys)[0].Value, storage.DefaultBaseURL, storage.DefaultAPIVersion, true)
```
kargakis: Shouldn't you be moving the storage client setup out of InitializeCluster too?

kargakis:
> closes #602

I don't think we can close #602 yet, after this PR gets merged we still assume that NewPlugin contains the azure credentials to setup the storage accounts, right?

pkg/upgrade/types.go
asalkeld: ok

pkg/upgrade/initialize.go
asalkeld: the sa is created in the deploy, and this is the only place it is used so it kinda makes sense to me.

asalkeld:
> after this PR gets merged we still assume that NewPlugin contains the azure credentials to setup the storage  accounts, right?

OK, i'll remove the closes.

pkg/util/fixtures/fixtures.go
```diff
@@ -1,6 +1,40 @@
...
 
-import "github.com/openshift/openshift-azure/pkg/api"
+import (
+	"k8s.io/client-go/tools/clientcmd/api/v1"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+// NewTestOpenShiftClusterConfig sets up required Kubeconfigs
+func NewTestOpenShiftClusterConfig() *api.Config {
```
kargakis: Where are you using this?

pkg/upgrade/initialize.go
kargakis: Makes sense. Maybe it would be clearer if we could separate between create and update in createClients.

kargakis:
needs a rebase

pkg/upgrade/deploy.go
```diff
@@ -17,6 +22,11 @@ func (u *simpleUpgrader) Deploy(ctx context.Context, cs *api.OpenShiftManagedClu
 		return err
 	}
 
+	err = managedcluster.WaitForHealthz(ctx, cs.Config.AdminKubeconfig)
```
kargakis: All of these health checks seems to be all over the place. We should probably consolidate this api server healthcheck, the node healthchecks below, and the infra services healthcheck inside the healthcheck plugin function.

pkg/upgrade/ready.go
```diff
@@ -119,13 +102,16 @@ func (u *simpleUpgrader) waitForNodes(ctx context.Context, cs *api.OpenShiftMana
 	return nil
 }
 
-// WaitForInfraServices verifies daemonsets, statefulsets
-func WaitForInfraServices(ctx context.Context, kc *kubernetes.Clientset) error {
+func (u *simpleUpgrader) WaitForInfraServices(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+	err := managedcluster.WaitForHealthz(ctx, cs.Config.AdminKubeconfig)
```
kargakis: Seems redundant to do this twice (once here and once in `Deploy`.

pkg/upgrade/deploy.go
kargakis: Separate issue is fine if it makes sense to do this.

kargakis: https://github.com/openshift/openshift-azure/issues/675

Issue #618, opened 2018-10-17
Rewrite hack/aad.sh in Go


kargakis:
Similar to https://github.com/openshift/openshift-azure/issues/595, the more of hack/create.sh we move out of bash into Go, the better it's going to be for our e2e framework. Once we rewrite hack/aad.sh (and hack/dns.sh) it should be very easy to get rid of hack/create.sh altogether.

/kind cleanup

asalkeld:
@kargakis unless you have started on this, I can do it.

mjudeikis:
@asalkeld I started #595 but got stuck. https://github.com/openshift/openshift-azure/pull/621 
Pushed so you know where I am :)

PR #623, opened 2018-10-19
Rewrite hack/aad.sh in go

asalkeld:
This works for me now, although note it now requires we set AZURE_USERNAME and AZURE_PASSWORD.

Fixes https://github.com/openshift/openshift-azure/issues/618

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 	"syscall"
 	"time"
@@ -29,11 +30,10 @@ import (
 )
 
 var (
-	method   = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
-	useProd  = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
-	manifest = flag.String("manifest", "_data/manifest.yaml", "Manifest to use for the initial request.")
-	update   = flag.String("update", "", "If provided, use this manifest to make a follow-up request after the initial request succeeds.")
-	cleanup  = flag.Bool("rm", false, "Delete the cluster once all other requests have completed successfully.")
+	method  = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
+	useProd = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
+	update  = flag.String("update", "", "If provided, use this manifest to make a follow-up request after the initial request succeeds.")
+	cleanup = flag.Bool("rm", false, "Delete the cluster once all other requests have completed successfully.")
 
 	// timeouts
 	rmTimeout     = flag.Duration("rm-timeout", 20*time.Minute, "Timeout of the cleanup request")
@@ -46,6 +46,10 @@ var (
 
 	artifactDir        = flag.String("artifact-dir", "", "Directory to place artifacts before a cluster is deleted.")
 	artifactKubeconfig = flag.String("artifact-kubeconfig", "", "Path to kubeconfig to use for gathering artifacts.")
+)
+
+const (
+	outputDirectory = "_data"
 )
 
 func validate() error {
@@ -92,17 +96,13 @@ func delete(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClus
 	return nil
 }
 
-func createOrUpdate(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, resourceGroup, manifest string) error {
+func createOrUpdate(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, resourceGroup, manifestTemplate, manifestFile string) error {
 	log.Info("creating/updating cluster")
-	in, err := ioutil.ReadFile(manifest)
+	oc, err := fakerp.GenerateManifest(manifestTemplate)
 	if err != nil {
 		return err
 	}
-	var oc sdk.OpenShiftManagedCluster
-	if err := yaml.Unmarshal(in, &oc); err != nil {
-		return err
-	}
-	future, err := rpc.CreateOrUpdate(ctx, resourceGroup, resourceGroup, oc)
+	future, err := rpc.CreateOrUpdate(ctx, resourceGroup, resourceGroup, *oc)
 	if err != nil {
 		return err
 	}
@@ -118,7 +118,7 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftMan
 		return err
 	}
 	log.Info("created/updated cluster")
-	return ioutil.WriteFile(manifest, out, 0666)
+	return ioutil.WriteFile(manifestFile, out, 0666)
 }
 
 func execCommand(c string) error {
@@ -169,8 +169,9 @@ func createResourceGroup(conf *fakerp.Config) (bool, error) {
 
 func execute(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, conf *fakerp.Config) error {
 	// simulate the API call to the RP
+	manifestFile := path.Join(outputDirectory, "manifest.yaml")
 	if err := wait.PollImmediate(time.Second, 1*time.Hour, func() (bool, error) {
-		if err := createOrUpdate(ctx, log, rpc, conf.ResourceGroup, *manifest); err != nil {
+		if err := createOrUpdate(ctx, log, rpc, conf.ResourceGroup, conf.Manifest, manifestFile); err != nil {
 			if autoRestErr, ok := err.(autorest.DetailedError); ok {
 				if urlErr, ok := autoRestErr.Original.(*url.Error); ok {
 					if netErr, ok := urlErr.Err.(*net.OpError); ok {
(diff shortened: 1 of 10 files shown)


asalkeld:
> @asalkeld what's the status on this?

I'll get back to it right after the backup work

asalkeld:
@jim-minter ptal

pkg/util/azureclient/graphrbac.go
```diff
@@ -0,0 +1,94 @@
...
+
+const (
+	// https://developer.microsoft.com/en-us/graph/docs/api-reference/beta/api/application_list
+	// To list and patch AAD applications, this code needs to have the clientID
+	// of an application with the following permissions:
+	// API: Windows Azure Active Directory
+	//   Delegated permissions:
+	//      Sign in and read user profile
+	//      Access the directory as the signed-in user
+	clientID = "5935b8e2-3915-409c-bfb2-865b7a9291e0"
```
asalkeld: @jim-minter is this correct to have it hardcoded? I wasn't sure if it should be AZURE_CLIENT_ID

mjudeikis: This should be corresponding to "permission id" so I think its fine.

mjudeikis:
@asalkeld can you make sure env.example is updated with new required values/

pkg/fakerp/manifest.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+
+	sdk "github.com/openshift/openshift-azure/pkg/util/azureclient/osa-go-sdk/services/containerservice/mgmt/2018-09-30-preview/containerservice"
+)
+
+func readEnv() map[string]string {
```
kargakis: Use kelsheyhightower/envconfig?

pkg/fakerp/manifest.go
```diff
@@ -0,0 +1,46 @@
...
+		env[pair[0]] = pair[1]
+	}
+	return env
+}
+
+// GenerateManifest returns the input manifest using the envirionment
+func GenerateManifest() (*sdk.OpenShiftManagedCluster, error) {
+	manifestFile := os.Getenv("MANIFEST")
+	if len(manifestFile) == 0 {
+		manifestFile = "test/manifests/normal/create.yaml"
```
kargakis: pass the default manifest in GenerateManifest as an arg

kargakis: Actually, do the defaulting in fakerp.Config and pass a single string here that is going to be read, parsed, and generated

pkg/fakerp/manifest.go
```diff
@@ -0,0 +1,46 @@
...
+	err = t.Execute(b, struct{ Env map[string]string }{Env: readEnv()})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed templating the manifest")
+	}
+
+	var oc sdk.OpenShiftManagedCluster
+	if err = yaml.Unmarshal(b.Bytes(), &oc); err != nil {
+		return nil, err
+	}
+	return &oc, err
```
kargakis: nit: turn oc to a pointer and here return nil explicitly
```
return oc, nil
```

pkg/fakerp/manifest.go
kargakis: nevermind, seems it's useful for the generation stuff

cmd/createorupdate/createorupdate.go
```diff
@@ -93,16 +94,38 @@ func delete(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClus
 }
 
 func createOrUpdate(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, resourceGroup, manifest string) error {
+	aadClientID := os.Getenv("AZURE_AAD_CLIENT_ID")
+	clientID := os.Getenv("AZURE_CLIENT_ID")
+	aadClientSecret := os.Getenv("AZURE_CLIENT_SECRET")
+	if len(aadClientID) > 0 && aadClientID != clientID {
```
kargakis: We should be running the aad stuff only on creates. This can also be an update.

cmd/createorupdate/createorupdate.go
```diff
@@ -93,16 +94,38 @@ func delete(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClus
...
+		log.Info("updating the aad application")
+		if len(os.Getenv("AZURE_USERNAME")) == 0 || len(os.Getenv("AZURE_PASSWORD")) == 0 {
```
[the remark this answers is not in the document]
kargakis: I would suggest moving all aad-related stuff into its own function and call it in main after `if cleanup` and before the first `createorupdate`. The challenge there is to determine whether this is a create or update. You can achieve that either by trying to read `containerservice.yaml` or by doing a dummy check against the resourcegroup (if not exists, it's a create). The latter is probably a more realistic check since the client should not have access to `containerservice.yaml`.

cmd/aad/aad.go
```diff
@@ -0,0 +1,234 @@
+package main
```
kargakis: Similarly to @mjudeikis's work, I would just add the necessary pieces for createorupdate in this PR and maybe have a follow-up PR to add the aad binary?

cmd/createorupdate/createorupdate.go
```diff
@@ -93,16 +94,38 @@ func delete(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClus
 }
 
 func createOrUpdate(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, resourceGroup, manifest string) error {
+	aadClientID := os.Getenv("AZURE_AAD_CLIENT_ID")
```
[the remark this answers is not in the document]
asalkeld: done, moved to main()

cmd/createorupdate/createorupdate.go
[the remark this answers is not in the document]
asalkeld: aad seems to authenticate differently. this works today as we use "az" which is a logged in user.
I struggled with getting this working and got some help from @jim-minter to get the initial login working.

pkg/fakerp/manifest.go
asalkeld: done

cmd/aad/aad.go
asalkeld: no problem

pkg/cmd/aad/aad.go
```diff
@@ -0,0 +1,68 @@
...
+	newPc := []graphrbac.PasswordCredential{
+		{
+			EndDate:   &timeend,
+			StartDate: &timestart,
+			KeyID:     to.StringPtr(azureAadClientSecretID),
+			Value:     to.StringPtr(azureAadClientSecretValue),
+		},
+	}
+
+	// delete the old one(s) by zero'ing out the value
```
asalkeld: I am going to remove this, the az client does not do it.

pkg/cmd/aad/aad.go
```diff
@@ -0,0 +1,59 @@
+package aad
```
kargakis: This is a function called by the RP so I would move it into pkg/fakerp

hack/create.sh
```diff
@@ -51,23 +51,8 @@ export RESOURCEGROUP=$1
 rm -rf _data
 mkdir -p _data/_out
 
-# if AZURE_CLIENT_ID is used as AZURE_AAD_CLIENT_ID, script will reset global team account!
-set +x
```
kargakis: Can you also remove this from `test/e2e.sh`?

asalkeld: oops, I missed that, thanks.

pkg/cmd/aad/aad.go
kargakis: I still don't think the separate package is worth it. cmd/aad/aad is not production code so the extra package is mere cognitive overhead.

pkg/fakerp/config.go
```diff
@@ -16,18 +16,23 @@ var supportedRegions = []string{
 }
 
 type Config struct {
-	SubscriptionID   string `envconfig:"AZURE_SUBSCRIPTION_ID" required:"true"`
-	TenantID         string `envconfig:"AZURE_TENANT_ID" required:"true"`
-	ClientID         string `envconfig:"AZURE_CLIENT_ID" required:"true"`
-	ClientSecret     string `envconfig:"AZURE_CLIENT_SECRET" required:"true"`
-	AADClientID      string `envconfig:"AZURE_AAD_CLIENT_ID"`
+	Username        string `envconfig:"AZURE_USERNAME"`
```
kargakis: We are not setting these in CI today, does this change mean we need to start doing that?

kargakis: Nevermind, we won't need them in CI

https://github.com/openshift/release/blob/859b37e14a88df9f88cf86b7af112ed6eee1c86b/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L180

test/manifests/normal/create.yaml
```diff
@@ -1,16 +1,16 @@
-name: $RESOURCEGROUP
-location: $AZURE_REGION
+name: "{{ .Env.RESOURCEGROUP }}"
```
jim-minter: @kargakis @asalkeld I don't like the fact that this has changed to a Go template, I wish I'd seen this before it merged :(

jim-minter: @kargakis @asalkeld I get the feeling there is some over-engineering going on here and we may need to refocus.

kargakis: @jim-minter what's the difference between this and the previous form? If we are only going to use these manifests in our dev and CI environments then what's the point?

kargakis: The benefit with keeping this in Go (among other changes) is less dependance on the host (eg. for replacing the envs in the manifest before, we had to install `gettext` in the test base image and get to use `envsubst`. Just an example. Others include the hack/dns and hack/aad rewrites to remove our dependance on `az`.

PR #630, opened 2018-10-19
Automate splitting imports into three different groups

kargakis:
@charlesakalugwu the automation we were talking about re import groups


charlesakalugwu:
Ah this is nice :D

PR #631, opened 2018-10-22
Change test template

### LINKED ISSUE #625: Job exits successfully but ginkgo test seems to have failed
```
• Failure [118.518 seconds]
Openshift on Azure end user e2e tests [EndUser]
/go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:23
  should deploy a template and check the visit counter increments [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:54

  Expected
      <int>: 1
  to be >
      <int>: 1

  /go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:107
```
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/608/pull-ci-openshift-openshift-azure-master-e2e-azure/340
/kind bug
/kind test-flake

thekad:
Fixes #625
Change the template used, mongodb is playing the eventually consistent game. No need to change anything else, the web page shown has the same structure and the counter increments.

### DIFF
--- a/test/e2e/end_user_test.go
+++ b/test/e2e/end_user_test.go
@@ -52,7 +52,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 	})
 
 	It("should deploy a template and check the visit counter increments", func() {
-		const TPL = "nodejs-mongodb-example"
+		const TPL = "cakephp-mysql-example"
 		var regex = regexp.MustCompile(`id="count-value">(\d+)<`)
 
 		// Create the template


kargakis:
Any idea why the job would exit successfully when the test would fail?

kargakis:
> Any idea why the job would exit successfully when the test would fail?

We do not error out in case of an error in a pipe so that's probably it. We should `set -o pipefail`. In any case, I don't think we need to be gathering ginkgo logs in the artifacts anymore so `tee` can probably be removed in a follow-up.

thekad:
Probably because of the tee at the end. The tee succeeds regardless.

PR #633, opened 2018-10-22
Support DELETE in the fake RP

kargakis:
/cc @mjudeikis @asalkeld

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -3,11 +3,16 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"flag"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path/filepath"
+	"strings"
 	"sync"
 
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -92,6 +102,45 @@ func (s *server) validate(w http.ResponseWriter, r *http.Request) bool {
...
+func (s *server) handleDelete(w http.ResponseWriter, req *http.Request) {
+	authorizer, err := azureclient.NewAuthorizer(os.Getenv("AZURE_CLIENT_ID"), os.Getenv("AZURE_CLIENT_SECRET"), os.Getenv("AZURE_TENANT_ID"))
+	if err != nil {
+		resp := "500 Internal Error: Failed to determine request credentials"
+		s.log.Debug(resp, err)
+		http.Error(w, resp, http.StatusInternalServerError)
+		return
+	}
+	subID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+	gc := resources.NewGroupsClient(subID)
```
mjudeikis: Why we dont move this to azureclient as all other? any particular reason?

kargakis: No particular reason, I am just not sure it's worth it since it's not going to be used by prod code and I don't know whether we want to spend time unit-testing the fake rp.

cmd/createorupdate/createorupdate.go
```diff
@@ -186,10 +240,56 @@ func (s *server) reply(w http.ResponseWriter, req *http.Request) {
...
 func main() {
+	flag.Parse()
+	if err := validate(); err != nil {
+		logrus.Fatal(err)
+	}
+
 	// simulate the RP
-	s := newServer(os.Getenv("RESOURCEGROUP"))
-	go s.ListenAndServe()
+	if !*useProd {
```
mjudeikis: this looks little bit unnecessary at this point in time, considering we set address to fakeRpAddr below?

kargakis: No, this is needed to start the fake RP. The address below is set in the client.

cmd/createorupdate/createorupdate.go
mjudeikis: true... dont think we will be mocking this one any time soon...

Issue #636, opened 2018-10-23
upgrade tests for master are broken


kargakis:
```
time="2018-10-22T21:58:47Z" level=debug msg="400 Bad Request: Failed to apply requestinvalid change [Properties.ProvisioningState: Succeeded != ]" resourceGroup=ci-op-w1rbqi9p-3b9ab
```
https://openshift-gce-devel.appspot.com/builds/origin-ci-test/logs/branch-ci-openshift-openshift-azure-master-e2e-azure-upgrade/
/assign
/kind bug

kargakis:
Tests started failing after https://github.com/openshift/openshift-azure/pull/592 got merged.

kargakis:
The problem is that the provisioningState is included in the reply from the fake RP so it's saved in the manifest during create and the same object is reused during the update so we get the invalid change error.

PR #637, opened 2018-10-23
Delete provisioningState before an update

kargakis:
We started using provisioningState in the fake RP to ensure the OSA SDK is working properly when querying the RP (https://github.com/openshift/openshift-azure/pull/592) but since we now write it down in manifest.yaml, upgrade.sh picks it up and fails to validate.

Fixes https://github.com/openshift/openshift-azure/issues/636


kargakis:
https://github.com/openshift/openshift-azure/issues/643

/retest

Issue #605, opened 2018-10-15
Occasionally "connection refused" using hack/create.sh


akrzos:
I have seen this twice now with the hack/create.sh script:

```
+ go run cmd/createorupdate/createorupdate.go
INFO[2018-10-15T15:24:50-04:00] starting server on localhost:8080             resourceGroup=perftest15node-2
FATA[0000] Put http://localhost:8080: dial tcp [::1]:8080: connect: connection refused 
exit status 1
```

kargakis:
/kind test-flake

We should be retrying the request until it succeeds since this is the fake RP we are running against and there is no reason to fail.

PR #641, opened 2018-10-23
Allow passing resource group as an arg in upgrade.sh

kargakis:


kargakis:
https://github.com/openshift/openshift-azure/issues/605

/retest

PR #646, opened 2018-10-23
defer cleanup of the cluster if it was requested

### LINKED ISSUE #644: Ensure cluster cleanup is run unconditionally at the end of createorupdate
by deferring it

charlesakalugwu:
closes #644

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -356,6 +356,15 @@ func main() {
 		return
 	}
 
+	// if a cleanup is requested, do it unconditionally at the end
+	if *cleanup {
+		defer func() {
+			if err := delete(ctx, log, rpc); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -349,6 +349,15 @@ func main() {
 	}
 	rpc.Authorizer = authorizer
 
+	defer func() {
```
kargakis: For completeness, this can be moved below the `strings.ToUpper(*method) == http.MethodDelete` check so that the defer statement will never run when -request=DELETE.

kargakis: Also, this can be turned to:
```go
	if *cleanup {
		defer func() {
			if err := delete(ctx, log, rpc); err != nil {
				log.Fatal(err)
			}
		}()
	}
```

Issue #643, opened 2018-10-23
timed out waiting for templateinstance to become ready


kargakis:
```
STEP: creating the template (2018-10-23 13:25:20.924564033 +0000 UTC m=+0.413895803)
STEP: waiting for the template instance to turn ready (2018-10-23 13:25:21.030584683 +0000 UTC m=+0.519916606)

• Failure [306.693 seconds]
Openshift on Azure end user e2e tests [EndUser]
/go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:23
  should deploy a template and check the visit counter increments [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:54

  Expected error:
      <*errors.errorString | 0xc4200c16e0>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
  not to have occurred

  /go/src/github.com/openshift/openshift-azure/test/e2e/end_user_test.go:79
```
/kind test-flake
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/637/pull-ci-openshift-openshift-azure-master-e2e-azure/382

kargakis:
Seems like the PHP app is not starting properly

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/649/pull-ci-openshift-openshift-azure-master-e2e-azure/400

kargakis:
Two warning events in the test namespace
```
count: 4
eventTime: null
firstTimestamp: 2018-10-25T17:09:48Z
involvedObject:
  apiVersion: v1
  fieldPath: spec.containers{mysql}
  kind: Pod
  name: mysql-1-rlpzg
  namespace: e2e-test-8lsp5
  resourceVersion: "4211"
  uid: b2d85c5b-d878-11e8-8434-000d3a4eb2b2
lastTimestamp: 2018-10-25T17:10:18Z
message: |
  Readiness probe failed: sh: cannot set terminal process group (-1): Inappropriate ioctl for device
  sh: no job control in this shell
  ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1' (111)
metadata:
  creationTimestamp: 2018-10-25T17:09:48Z
  name: mysql-1-rlpzg.1560e91e419e6574
  namespace: e2e-test-8lsp5
  resourceVersion: "4461"
[...]

PR #649, opened 2018-10-24
Get debug info from end-user e2e tests

kargakis:
Should help debug https://github.com/openshift/openshift-azure/issues/643

/cc thekad

### DIFF
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -3,13 +3,18 @@
 package e2e
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"time"
 
+	"github.com/ghodss/yaml"
 	project "github.com/openshift/api/project/v1"
 	projectclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
 	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
+	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


test/e2e/client_test.go
```diff
@@ -106,3 +115,58 @@ func (t *testClient) cleanupProject(timeout time.Duration) error {
...
 	return nil
 }
+
+func (t *testClient) dumpInfo() error {
+	// gather events
+	eventList, err := t.kc.CoreV1().Events(t.namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	eventBuf := bytes.NewBuffer(nil)
```
charlesakalugwu: whats the difference/reason in doing it this way vs doing it like `var eventBuf bytes.Buffer` or `eventBuf := new(bytes.Buffer)`?

test/e2e/client_test.go
```diff
@@ -106,3 +115,58 @@ func (t *testClient) cleanupProject(timeout time.Duration) error {
...
+		}
+		if _, err := podBuf.Write([]byte("\n")); err != nil {
+			return err
+		}
+	}
+
+	if t.artifactDir != "" {
+		if err := ioutil.WriteFile(filepath.Join(t.artifactDir, fmt.Sprintf("events-%s.yaml", t.namespace)), eventBuf.Bytes(), 0777); err != nil {
+			logrus.Warn(err)
+			fmt.Println(eventBuf.String())
```
charlesakalugwu: why not `logrus.Info(eventBuf.String())`?

test/e2e/client_test.go
kargakis: There isn't any difference.

test/e2e/client_test.go
kargakis: I did a bare print here mainly because at this point we don't really care about the logging context and the yaml is going to be better formatted in the log.

Issue #615, opened 2018-10-17
Scaleup with hack/upgrade.sh seems to only work one compute node at a time


akrzos:
I am unsure if this a subscription limitation or if larger instances just take longer to provision but I was unable to scale up from a core cluster (3 masters, 2 infras, 2 computes) to a larger cluster (3 masters, 2 infras, 10 computes) without hitting `400 Bad Request: Failed to apply requestFuture#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded  resourceGroup=perftest15node-2`

Subsequently, I have been able to slowly scale up the cluster one compute at a time however we should be able to scale up in bulk.

akrzos:
After the "large" scale up of 2 to 10 (Add 8 more compute nodes) failed, I attempted to scale up by 2 compute nodes and was still met with the same problem:

```
[akrzos@fedora28 openshift-azure]$ time ./hack/upgrade.sh 
...
+ go run cmd/createorupdate/createorupdate.go
INFO[2018-10-17T08:15:01-04:00] starting server on localhost:8080             resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] convert to internal                           resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] enrich                                        resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] read old config                               resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] merge old and new config                      resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] merging internal data models                  resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] validating internal data models               resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] generating configs                            resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] persist config                                resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] generating arm templates                      resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] write helpers                                 resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:01-04:00] starting update                               resourceGroup=perftest15node-2
INFO[2018-10-17T08:15:02-04:00] applying arm template deployment              resourceGroup=perftest15node-2
[...]

PR #653, opened 2018-10-24
Get result from applied arm template

kargakis:
/hold

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -59,6 +59,11 @@ func Fatalf(format string, args ...interface{}) {
 	logger.log.Fatalf(format, args...)
 }
 
+// Warn logs at warn level
+func Warn(args ...interface{}) {
+	logger.log.Warn(args...)
+}
+
 // WithFields adds a map of fields to the Entry.
 func WithFields(fields logrus.Fields) *logrus.Entry {
 	return logger.log.WithFields(fields)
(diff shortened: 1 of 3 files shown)


pkg/util/azureclient/resources.go
```diff
@@ -11,6 +11,7 @@ import (
 type DeploymentsClient interface {
 	CreateOrUpdate(ctx context.Context, resourceGroupName string, deploymentName string, parameters resources.Deployment) (result resources.DeploymentsCreateOrUpdateFuture, err error)
 	Client
+	DeploymentClient() resources.DeploymentsClient
```
kargakis: @mjudeikis @asalkeld this one looks like a limitation of our sdk interfaces.

mjudeikis: it is. But this way we can mock any interfaces with mock data for unti tests

pkg/log/log.go
```diff
@@ -59,6 +59,11 @@ func Fatalf(format string, args ...interface{}) {
 	logger.log.Fatalf(format, args...)
 }
 
+// Warn logs at warn level
+func Warn(format string, args ...interface{}) {
```
kargakis: Honestly I am sick of maintaining all these shims already

kargakis:
> @kargakis remove hold?

I would like to see https://github.com/openshift/openshift-azure/issues/615#issuecomment-433337154 answered first.

pkg/log/log.go
charlesakalugwu: this should be `Warnf` ...however we probably want both versions `Warn` and `Warnf`

kargakis: Yes, we want both. I should get rid of the formatting here since the correct use is Warn. I am adding Warnf in a different PR.

PR #656, opened 2018-10-24
end-to-end tests should honor KUBECONFIG environment variable

### LINKED ISSUE #577: end-to-end tests should honor KUBECONFIG environment variable
I think it makes more sense for the e2es to read the KUBECONFIG environment variable instead of a -kubeconfig flag - this is pretty standard in the ecosystem, isn't it?  @kargakis are you +1 on this?

charlesakalugwu:
closes #577

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -13,31 +13,40 @@ fi
 
 if [[ -z "$SUITE" || "$SUITE" == "enduser" ]]; then
   echo "Running end user e2e tests"
-  # Login as osadmin to simulate a regular user
+  # Login as enduser to simulate a regular user
   password=$(awk '/^  endUserPasswd:/ { print $2 }' <_data/containerservice.yaml)
   fqdn=$(awk '/^  fqdn:/ { print $2 }' <_data/containerservice.yaml)
-  export KUBECONFIG=_data/_out/enduser.kubeconfig
+  export KUBECONFIG=$(pwd)/_data/_out/enduser.kubeconfig
+  # oc login is going to create the enduser.kubeconfig for us
   oc login $fqdn --username enduser --password $password --insecure-skip-tls-verify=true
-  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[EndUser\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e -kubeconfig ../../_data/_out/enduser.kubeconfig "${ARTIFACT_FLAG:-}"
+  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[EndUser\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e "${ARTIFACT_FLAG:-}"
   oc logout
 fi
 
 if [[ -z "$SUITE" || "$SUITE" == "clusterreader" ]]; then
   echo "Running azure cluster reader e2e tests"
   (awk '/^  azureClusterReaderKubeconfig:/ { print $2 }' <_data/containerservice.yaml | base64 -d) > _data/_out/azure-cluster-reader.kubeconfig
-  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[AzureClusterReader\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e -kubeconfig ../../_data/_out/azure-cluster-reader.kubeconfig
+
+  export KUBECONFIG=$(pwd)/_data/_out/azure-cluster-reader.kubeconfig
+  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[AzureClusterReader\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e
 fi
 
 if [[ -z "$SUITE" || "$SUITE" == "customer-cluster-admin" ]]; then
   echo "Running azure customer-cluster-admin tests"
   fqdn=$(awk '/^  fqdn:/ { print $2 }' <_data/containerservice.yaml)
-  export KUBECONFIG_CUSTOMER_ADMIN=_data/_out/customer-cluster-admin.kubeconfig
-  export KUBECONFIG_CUSTOMER_READER=_data/_out/customer-cluster-reader.kubeconfig
-  export KUBECONFIG_END_USER=_data/_out/enduser.kubeconfig
 
-  # oc login is going to create the kubeconfig with the proper permissions
-  KUBECONFIG=$KUBECONFIG_CUSTOMER_ADMIN oc login $fqdn --username customer-cluster-admin --password "$(awk '/^  customerAdminPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  KUBECONFIG=$KUBECONFIG_CUSTOMER_READER oc login $fqdn --username customer-cluster-reader --password "$(awk '/^  customerReaderPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  KUBECONFIG=$KUBECONFIG_END_USER oc login $fqdn --username enduser --password "$(awk '/^  endUserPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[CustomerAdmin\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e -kubeconfig ../../_data/_out/enduser.kubeconfig "${ARTIFACT_FLAG:-}"
+  export KUBECONFIG=$(pwd)/_data/_out/customer-cluster-admin.kubeconfig
+  oc login $fqdn --username customer-cluster-admin --password "$(awk '/^  customerAdminPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) customer-cluster-admin
+
+  export KUBECONFIG=$(pwd)/_data/_out/customer-cluster-reader.kubeconfig
+  oc login $fqdn --username customer-cluster-reader --password "$(awk '/^  customerReaderPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) customer-cluster-reader
+
+  export KUBECONFIG=$(pwd)/_data/_out/customer-cluster-enduser.kubeconfig
+  oc login $fqdn --username enduser --password "$(awk '/^  endUserPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) enduser
+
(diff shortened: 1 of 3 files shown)


hack/e2e.sh
```diff
@@ -15,15 +15,16 @@ if [[ -z "$SUITE" || "$SUITE" == "enduser" ]]; then
   # Login as osadmin to simulate a regular user
   password=$(awk '/^  adminPasswd:/ { print $2 }' <_data/containerservice.yaml)
   fqdn=$(awk '/^  fqdn:/ { print $2 }' <_data/containerservice.yaml)
-  export KUBECONFIG=_data/_out/osadmin.kubeconfig
```
kargakis: This change is going to break the test because `oc login` will not generate the kubeconfig in this path.

test/e2e/client_test.go
```diff
@@ -22,6 +22,13 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+type Config struct {
+	// Directory to place artifacts when a test fails
+	ArtifactDir string `envconfig:"ARTIFACTS" required:"false"`
```
kargakis: ARTIFACT_DIR is used in our tests, see [here](https://github.com/openshift/release/blob/3cbd3516ac42e15ede15052ce55201afe1aa0079/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L76).

hack/e2e.sh
kargakis: Maybe make the path absolute by using `pwd`.

test/e2e/extended_suite_test.go
```diff
@@ -7,18 +7,25 @@ import (
...
 )
 
 var _ = BeforeSuite(func() {
-	c = newTestClient(*kubeconfig, *artifactDir)
+	err := envconfig.Process("", &testConfig)
+	if err != nil {
+		panic(err)
+	}
+	if testConfig.ArtifactDir == "" {
+		testConfig.ArtifactDir = *flag.String("artifact-dir", "", "Directory to place artifacts when a test fails")
```
kargakis: You are missing `flag.Parse` but I am wondering what's the added value of maintaining two different ways of doing the same thing. I am not in love with environment variables because we need to be documenting them. Maybe revert the move of artifact-dir to an env since the issue was only about KUBECONFIG? I don't feel strong but I definitely don't want two separate ways to do the same thing.

kargakis:
needs a rebase

hack/e2e.sh
```diff
@@ -13,24 +13,28 @@ fi
...
   echo "Running end user e2e tests"
-  # Login as osadmin to simulate a regular user
+  # Login as enduser to simulate a regular user
   password=$(awk '/^  endUserPasswd:/ { print $2 }' <_data/containerservice.yaml)
   fqdn=$(awk '/^  fqdn:/ { print $2 }' <_data/containerservice.yaml)
   export KUBECONFIG=_data/_out/enduser.kubeconfig
+  # oc login is going to create the enduser.kubeconfig for us
   oc login $fqdn --username enduser --password $password --insecure-skip-tls-verify=true
-  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[EndUser\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e -kubeconfig ../../_data/_out/enduser.kubeconfig "${ARTIFACT_FLAG:-}"
+  export KUBECONFIG_END_USER=../../_data/_out/enduser.kubeconfig
```
kargakis: Can you revert to the use of KUBECONFIG for end user and azure cluster reader? Is there a reason we are ditching KUBECONFIG altogether for these types of tests?

test/e2e/extended_suite_test.go
```diff
@@ -5,37 +5,39 @@ package e2e
...
-		creader = newTestClient("../../_data/_out/customer-cluster-reader.kubeconfig", *artifactDir)
-		cadmin = newTestClient("../../_data/_out/customer-cluster-admin.kubeconfig", *artifactDir)
+	suiteSummary := config.GinkgoConfig.FocusString
+	if inFocus(suiteSummary, "CustomerAdmin") {
+		c = newTestClient(testConfig.EndUserKubeConfig, testConfig.ArtifactDir)
+		creader = newTestClient(testConfig.CustomerClusterReaderKubeConfig, testConfig.ArtifactDir)
+		cadmin = newTestClient(testConfig.CustomerClusterAdminKubeConfig, testConfig.ArtifactDir)
+		return
+	}
+	if inFocus(suiteSummary, "AzureClusterReader") {
```
kargakis: I would prefer we still honor KUBECONFIG for these and not have separate clients/kubeconfig entries.

hack/e2e.sh
charlesakalugwu: I would like to have the `KUBECONFIG_`  usage consistent across the different types of tests. CustomerAdmin tests also require an enduser kubeconfig. Since CustomerAdmin requires 3 clients with 3 different kubeconfigs, those kubeconfigs can only be supplied for each client using the `KUBECONFIG_` variables. 

tldr; I didn't want to use `KUBECONFIG_END_USER` for an end user kubeconfig for `CustomerAdmin` tests and then go back to `KUBECONFIG` also for an end user kubeconfig in `EndUser` tests.

test/e2e/extended_suite_test.go
charlesakalugwu: This is only possible if we have only one client per test category. So we might want to break the customer admin tests into smaller categories.

kargakis: The customer admin tests are a special case I think and we will not be able to get rid of the end user kubeconfig from them (unless we use auth delegation which is not ideal). The rest of the tests (end user, azure reader) do not need a second kubeconfig so it makes sense for them to honor KUBECONFIG (we already run all these suites separately in hack/e2e.sh).

kargakis: @jim-minter @mjudeikis any feelings here?

mjudeikis: I think this is needed, but this looks very complicated. I feel we need to have smaller chunks where we can and honour `KUBECONFIG` and keep `KUBECONFIG_` to minimal usage. Maybe we need to rethink little bit our test structure?

kargakis: Not sure it's necessary to do it here and I cannot see how we get around the fact that we need multiple kubeconfigs in a single test run. FWIW, our approach is better than what has been used in the origin e2e tests historically (a single bash script that jumps around contexts via oc login).

charlesakalugwu: @kargakis @mjudeikis what do you think about the redesign?

hack/e2e.sh
```diff
@@ -13,31 +13,38 @@ fi
...
-  export KUBECONFIG_CUSTOMER_READER=_data/_out/customer-cluster-reader.kubeconfig
-  export KUBECONFIG_END_USER=_data/_out/enduser.kubeconfig
-
-  # oc login is going to create the kubeconfig with the proper permissions
-  KUBECONFIG=$KUBECONFIG_CUSTOMER_ADMIN oc login $fqdn --username customer-cluster-admin --password "$(awk '/^  customerAdminPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  KUBECONFIG=$KUBECONFIG_CUSTOMER_READER oc login $fqdn --username customer-cluster-reader --password "$(awk '/^  customerReaderPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  KUBECONFIG=$KUBECONFIG_END_USER oc login $fqdn --username enduser --password "$(awk '/^  endUserPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
-  go test ./test/e2e -test.v -ginkgo.v -ginkgo.focus="\[CustomerAdmin\]" -ginkgo.noColor -ginkgo.randomizeAllSpecs -tags e2e -kubeconfig ../../_data/_out/enduser.kubeconfig "${ARTIFACT_FLAG:-}"
+
+  export KUBECONFIG=_data/_out/customer-cluster-admin.kubeconfig
```
kargakis: Honestly I don't know how running these tests in production is going to look like. I like the approach of trimming our input to the tests down to a single kubeconfig but I would prefer we mimic how we are going to run these in prod and I feel we are going to end up using separate kubeconfigs. @kwoodson thoughts?

test/e2e/client_test.go
```diff
@@ -23,6 +24,14 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+type Config struct {
+	// Directory to place artifacts when a test fails
+	ArtifactDir string `envconfig:"ARTIFACT_DIR" required:"false"`
```
kargakis: nit: remove required

test/e2e/client_test.go
```diff
@@ -23,6 +24,14 @@ import (
...
+
+	// Location of the kubeconfig
+	KubeConfig string `envconfig:"KUBECONFIG" default:"../../_data/_out/admin.kubeconfig"`
```
kargakis: I would prefer we not default to anything and have this required:true

kargakis:
We may or may not need to figure out here how to run these tests in production. Other than that, lgtm.

hack/e2e.sh
kwoodson: I like the idea of a single kubeconfig but I'm not sure how it is possible.  The admin test requires separate kubeconfigs.  We specifically need to test a create from an enduser and verify the permissions are then applied correctly through the customer-admin kubeconfig.  Maybe we should have a meeting to discuss what prod tests and our e2e test strategy.

kargakis: The only reason I can think of is that these kubeconfig are going to be generated separately so in prod we will end up needing to run them separately as opposed to merging them into one.

hack/e2e.sh
```diff
@@ -13,31 +13,36 @@ fi
...
+  oc login $fqdn --username customer-cluster-admin --password "$(awk '/^  customerAdminPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) customer-cluster-admin
+  oc login $fqdn --username customer-cluster-reader --password "$(awk '/^  customerReaderPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) customer-cluster-reader
+  oc login $fqdn --username enduser --password "$(awk '/^  endUserPasswd:/{ print $2 }' <_data/containerservice.yaml)" --insecure-skip-tls-verify=true
+  oc config rename-context $(oc config current-context) enduser
+
+  export KUBECONFIG=../../_data/_out/customer-cluster-admin.kubeconfig
```
mjudeikis: L38 and L46 is little bit confusing. Im ok with this, but we should think how to simplify this and not to have these walking up and down the directories :/

test/e2e/extended_suite_test.go
mjudeikis: I like this more. maybe just rename kubeconfig from `customer-cluster-admin.kubeconfig` to something saying that now it was all context inside. as its not just cluster-admin now,

hack/e2e.sh
charlesakalugwu: @kargakis makes sense.

kargakis:
This makes our test suite depend on e2e.sh. I think we should figure out how we want to run these tests in prod first, before proceeding any further here.

/hold

mjudeikis:
Close in favor of recent @jim-minter PR?

Issue #654, opened 2018-10-24
30 minutes do not seem to be enough for an upgrade


kargakis:
65, 66, and 67 builds fail with the exact same error

https://openshift-gce-devel.appspot.com/builds/origin-ci-test/logs/branch-ci-openshift-openshift-azure-master-e2e-azure-upgrade/

/kind test-flake
/kind bug

PR #660, opened 2018-10-25
fakerp timeout extended to 1 hour

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/654

Empirically our upgrade test takes from 35m up to 40-45m.

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -173,7 +173,7 @@ func (s *server) handlePut(w http.ResponseWriter, req *http.Request) {
 	s.write(oc)
 
 	// simulate Context with property bag
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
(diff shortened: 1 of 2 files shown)


kargakis:
https://github.com/openshift/openshift-azure/issues/661

/test upgrade

cmd/createorupdate/createorupdate.go
```diff
@@ -173,7 +173,7 @@ func (s *server) handlePut(w http.ResponseWriter, req *http.Request) {
 	s.write(oc)
 
 	// simulate Context with property bag
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
```
kargakis: @amanohar it would be useful to know the timeout used by the RP so we can use the same here. This is the timeout of the context passed to the plugins.

Issue #560, opened 2018-10-05
Cannot query api server - connection timed out


charlesakalugwu:
Failure to complete setup of e2e tests

```
+ az group create -n ci-op-s6k78vy0-3b9ab -l eastus --tags now=1538581995 ttl=76h
+ set +x
+ cat
+ go generate ./...
bindata.go
bindata.go
+ go run cmd/createorupdate/createorupdate.go -loglevel=debug
time="2018-10-03T15:53:29Z" level=info msg="convert to internal" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:29Z" level=info msg=enrich resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:29Z" level=info msg="validating internal data models" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:29Z" level=info msg="generating configs" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:38Z" level=info msg="persist config" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:38Z" level=info msg="generating arm templates" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:38Z" level=info msg="write helpers" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:38Z" level=info msg="applying arm template deployment" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:53:40Z" level=info msg="waiting for arm template deployment to complete" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:58:40Z" level=info msg="initializing cluster" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:58:41Z" level=info msg="starting health check" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T15:59:47Z" level=info msg="waiting for master-000000 to be ready" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T16:03:55Z" level=info msg="waiting for master-000001 to be ready" resourceGroup=ci-op-s6k78vy0-3b9ab
time="2018-10-03T16:03:57Z" level=info msg="waiting for master-000002 to be ready" resourceGroup=ci-op-s6k78vy0-3b9ab
[...]

PR #670, opened 2018-10-30
Create a new storage blob to track vm hashes

kargakis:
Create a storage blob in azure to track vm hashes since we cannot tag VMs individually.

Part of AZURE-199

/cc jim-minter asalkeld

### DIFF
--- a/pkg/cluster/deploy.go
+++ b/pkg/cluster/deploy.go
@@ -1,9 +1,17 @@
 package cluster
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/jsonpath"
+	"github.com/openshift/openshift-azure/pkg/log"
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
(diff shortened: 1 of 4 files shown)


kargakis:

/retest

pkg/arm/arm.go
```diff
@@ -77,10 +83,133 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
...
+			newTags := make(map[string]interface{})
+			if tags, ok := resource["tags"].(map[string]interface{}); ok {
+				for k, v := range tags {
+					newTags[k] = v
+				}
+			}
+			newTags[key] = value
+			resource["tags"] = newTags
+			added = true
+			break out
```
asalkeld: these break labels are really not intuitive, I had to go read https://golang.org/ref/spec#Break_statements to figure out it goes to the next statement out of the loop - weird.

mjudeikis: Agree :)

pkg/arm/arm.go
```diff
@@ -77,10 +83,133 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
...
+			if !ok {
+				continue
+			}
+
+			if !isScaleSet(resource) {
+				continue
+			}
+
+			// cleanup capacity
+			deleteField(resource, "sku", "capacity")
```
mjudeikis: Just for my info, why we cleaning this one?

pkg/arm/arm.go
kargakis: This can also be a return, will update

pkg/arm/arm.go
kargakis: So that scaling up/down will not cause unnecessary VM rotations.

asalkeld:
this looks fine to me, just waiting on jim's opinion.

kargakis:
/hold

After some testing, this approach seems useless. When a tag gets applied in a scale set, the underlying VMs get their tags updated automatically. :/

asalkeld:
> /hold
> 
> After some testing, this approach seems useless. When a tag gets applied in a scale set, the underlying VMs get their tags updated automatically. :/

That sucks :-(

kargakis:
Keeping the tag in the scale set is probably something that we still need to do. We need to use it when comparing with the VM tags, just need to track VM tags elsewhere and not on the VMs.

/hold cancel

kargakis:
Updated to add the scale set hash directly in the storage blob as discussed with @jim-minter, should be ready for a review

jim-minter:
Given that #683 builds on this, and I've done a detailed review there, I would be inclined to close this in favour of getting #683 through in one go.

/hold

PR #673, opened 2018-10-30
remove extraneous functions from external Plugin interface

jim-minter:


asalkeld:
looks fine to me, does MSFT need to confirm?

PR #674, opened 2018-10-31
Return a structured error in the plugin api so the step that failed is obvious

### LINKED ISSUE #672: Structured errors for plugin CreateOrUpdate workflow
As an RP engineer I would like to better report at what point a long-running plugin CreateOrUpdate execution failed.  Adding structured wrapped errors with a limited number of different error codes would help here.

Create something like the following in pkg/api/plugin.go:

```
type PluginStep string

const (
	PluginStepFoo PluginStep = "foo"
	...
)

type PluginError struct {
	Err error
	Step PluginStep
}

var _ error := &PluginError{}

func (pe *PluginError) Error() string {
	return string(pe.Step)+": "+pe.Err.Error()
}
```

The intention in the above API is we will not nest PluginErrors, even though it is feasible in principle.

At least each of the following should probably be differentiated steps:
* Update (pre-deploy drain)
* Deploy
* InitializeCluster
* Update (post-deploy readiness)
* Update (master VM rotation)
* Update (infra VM rotation)
* Update (compute VM rotation)
* Health check (wait for healthz)
* Health check (wait for nodes)
* Health check (daemonsets)
* Health check (deployments)
* Health check (console)

Find a maintainable way to implement the above and implement.  Potentially, judicious changing of functions associated with "steps" such that their plugin-internal interfaces return `*PluginError` instead of `error` might be an option.

asalkeld:
closes #672

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -14,6 +14,35 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+type PluginStep string
+
+const (
+	PluginStepDeploy                     PluginStep = "Deploy"
+	PluginStepInitialize                 PluginStep = "InitializeCluster"
+	PluginStepClientCreation             PluginStep = "ClientCreation"
+	PluginStepDrain                      PluginStep = "Drain"
+	PluginStepUpdateMasterVMRotation     PluginStep = "UpdateMasterVMRotation"
+	PluginStepUpdateInfraVMRotation      PluginStep = "UpdateInfraVMRotation"
+	PluginStepUpdateComputeVMRotation    PluginStep = "UpdateComputeVMRotation"
+	PluginStepWaitForWaitForOpenShiftAPI PluginStep = "WaitForOpenShiftAPI"
+	PluginStepWaitForNodes               PluginStep = "WaitForNodes"
+	PluginStepWaitForConsoleHealth       PluginStep = "WaitForConsoleHealth"
+	PluginStepWaitForInfraDaemonSets     PluginStep = "WaitForInfraDaemonSets"
+	PluginStepWaitForInfraDeployments    PluginStep = "WaitForInfraDeployments"
+)
+
+// PluginError error returned by CreateOrUpdate to specify the step that failed.
+type PluginError struct {
+	Err  error
+	Step PluginStep
+}
+
+var _ error = &PluginError{}
+
+func (pe *PluginError) Error() string {
+	return string(pe.Step) + ": " + pe.Err.Error()
+}
+
 // DeployFn makes it possible to plug in different logic to the deploy.
 // The implementor must initiate a deployment of the given template using
 // mode resources.Incremental and wait for it to complete.
--- a/pkg/upgrade/deploy.go
+++ b/pkg/upgrade/deploy.go
@@ -10,23 +10,23 @@ import (
 func (u *simpleUpgrader) Deploy(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error {
 	err := u.createClients(ctx, cs)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepClientCreation}
 	}
 	err = deployFn(ctx, azuretemplate)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepDeploy}
 	}
-
 	err = u.InitializeCluster(ctx, cs)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepInitialize}
 	}
-
 	err = managedcluster.WaitForHealthz(ctx, cs.Config.AdminKubeconfig)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepWaitForWaitForOpenShiftAPI}
 	}
-
-	// ensure that all nodes are ready
-	return u.waitForNodes(ctx, cs)
+	err = u.waitForNodes(ctx, cs)
+	if err != nil {
+		return &api.PluginError{Err: err, Step: api.PluginStepWaitForNodes}
+	}
+	return nil
 }
--- a/pkg/upgrade/update.go
+++ b/pkg/upgrade/update.go
@@ -19,42 +19,71 @@ func (u *simpleUpgrader) Update(ctx context.Context, cs *api.OpenShiftManagedClu
 	if err != nil {
 		return err
 	}
-	// Deploy() may change the number of VMs.  If we can see that any VMs are
+	// deployFn() may change the number of VMs.  If we can see that any VMs are
 	// about to be deleted, drain them first.  Record which VMs are visible now
-	// so that we can detect newly created VMs and wait for them to become
-	// ready.
+	// so that we can detect newly created VMs and wait for them to become ready.
+	vmsBefore, err := u.getNodesAndDrain(ctx, cs)
+	if err != nil {
+		return err
+	}
+	err = deployFn(ctx, azuretemplate)
+	if err != nil {
+		return err
+	}
+	err = u.InitializeCluster(ctx, cs)
+	if err != nil {
+		return err
+	}
+	err = u.waitForNewNodes(ctx, cs, vmsBefore)
+	if err != nil {
+		return err
+	}
 
+	// For PP day 1, scale is permitted but not any other sort of update.  When
+	// we enable configuration changes and/or upgrades, uncomment this code.  At
+	// the same time, current thinking is that we will add a hash-based
+	// mechanism to avoid unnecessary VM rotations as well.
+	if os.Getenv("RUNNING_UNDER_TEST") != "" {
+		err := u.updateInPlace(ctx, cs, api.AgentPoolProfileRoleMaster)
+		if err != nil {
+			return err
+		}
+		err = u.updatePlusOne(ctx, cs, api.AgentPoolProfileRoleInfra)
+		if err != nil {
+			return err
+		}
+		err = u.updatePlusOne(ctx, cs, api.AgentPoolProfileRoleCompute)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (u *simpleUpgrader) getNodesAndDrain(ctx context.Context, cs *api.OpenShiftManagedCluster) (map[string]struct{}, error) {
 	vmsBefore := map[string]struct{}{}
 
 	for _, agent := range cs.Properties.AgentPoolProfiles {
 		vms, err := listVMs(ctx, cs, u.vmc, agent.Role)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		for i, vm := range vms {
 			if i < agent.Count {
 				vmsBefore[*vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName] = struct{}{}
-
 			} else {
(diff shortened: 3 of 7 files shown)


asalkeld:
the above is a repeatable error, don't retest, it needs a fix.

asalkeld:
This is really odd:

	_, err = fakerp.CreateOrUpdate(ctx, oc, s.log, config)
	if err != nil {
		s.writeState(v20180930preview.Failed)
		resp := "400 Bad Request: Failed to apply request"
		s.log.Error("error is:", err)
it is printing out:
INFO[2018-10-31T16:03:05+10:00] starting health check                         resourceGroup=angus-errors
INFO[2018-10-31T16:03:05+10:00] checking console health                       resourceGroup=angus-errors
INFO[2018-10-31T16:03:06+10:00] OK                                            resourceGroup=angus-errors
ERRO[2018-10-31T16:03:06+10:00] error is:<nil>                                resourceGroup=angus-errors
DEBU[2018-10-31T16:03:06+10:00] 400 Bad Request: Failed to apply request<nil>  resourceGroup=angus-errors

How can err != nil and still get "error is:<nil> "?

pkg/plugin/plugin.go
```diff
@@ -119,15 +123,16 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
...
 	}
-	if err != nil {
+	if err != (*api.PluginError)(nil) {
 		return err
 	}
-
-	// Wait for infrastructure services to be healthy
 	err = p.clusterUpgrader.WaitForInfraServices(ctx, cs)
-	if err != nil {
+	if err != (*api.PluginError)(nil) {
```
kargakis: Are these type casts necessary if the clusterUpgrader functions return errors instead of PluginErrors? Either way the plugin caller needs to cast `error` to `PluginError` so this seems unnecessary.

pkg/api/plugin.go
```diff
@@ -14,6 +14,34 @@ const (
 	ContextKeyTenantID     ContextKey = "TenantID"
 )
 
+type PluginStep string
+
+const (
+	PluginStepDeploy                  PluginStep = "Deploy"
+	PluginStepInitialize              PluginStep = "InitializeCluster"
+	PluginStepUpdateDrain             PluginStep = "UpdateDrain"
```
kargakis: Why not simply Drain?

pkg/api/plugin.go
```diff
@@ -14,6 +14,34 @@ const (
...
+	PluginStepUpdateMasterVMRotation  PluginStep = "UpdateMasterVMRotation"
+	PluginStepUpdateInfraVMRotation   PluginStep = "UpdateInfraVMRotation"
+	PluginStepUpdateComputeVMRotation PluginStep = "UpdateComputeVMRotation"
+	PluginStepWaitForHealthz          PluginStep = "WaitForHealthz"
+	PluginStepWaitForNodes            PluginStep = "WaitForNodes"
+	PluginStepHealthCheck             PluginStep = "HealthCheck"
+	PluginStepWaitForInfraDaemonSets  PluginStep = "WaitForInfraDaemonSets"
```
kargakis: Not sure it's worth separating between DaemonSets and Deployments. `WaitForInfraServices` is probably more appropriate.

pkg/api/plugin.go
```diff
@@ -14,6 +14,34 @@ const (
...
+const (
+	PluginStepDeploy                  PluginStep = "Deploy"
+	PluginStepInitialize              PluginStep = "InitializeCluster"
+	PluginStepUpdateDrain             PluginStep = "UpdateDrain"
+	PluginStepUpdateMasterVMRotation  PluginStep = "UpdateMasterVMRotation"
+	PluginStepUpdateInfraVMRotation   PluginStep = "UpdateInfraVMRotation"
+	PluginStepUpdateComputeVMRotation PluginStep = "UpdateComputeVMRotation"
+	PluginStepWaitForHealthz          PluginStep = "WaitForHealthz"
+	PluginStepWaitForNodes            PluginStep = "WaitForNodes"
+	PluginStepHealthCheck             PluginStep = "HealthCheck"
```
kargakis: This is confusing when we also have WaitForHealthz. `ConsoleHealthz` or something along those lines.

pkg/api/plugin.go
```diff
@@ -14,6 +14,34 @@ const (
...
+type PluginStep string
+
+const (
+	PluginStepDeploy                  PluginStep = "Deploy"
+	PluginStepInitialize              PluginStep = "InitializeCluster"
+	PluginStepUpdateDrain             PluginStep = "UpdateDrain"
+	PluginStepUpdateMasterVMRotation  PluginStep = "UpdateMasterVMRotation"
+	PluginStepUpdateInfraVMRotation   PluginStep = "UpdateInfraVMRotation"
+	PluginStepUpdateComputeVMRotation PluginStep = "UpdateComputeVMRotation"
+	PluginStepWaitForHealthz          PluginStep = "WaitForHealthz"
```
kargakis: Too vague I think, should be clearer that we wait for the Openshift API server.

pkg/upgrade/update.go
```diff
@@ -12,49 +12,83 @@ import (
...
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
-func (u *simpleUpgrader) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error {
+func (u *simpleUpgrader) Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) *api.PluginError {
 	err := u.createClients(ctx, cs)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepUpdateDrain}
```
kargakis: This is not a drain error.

pkg/upgrade/deploy.go
```diff
@@ -7,26 +7,26 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
-func (u *simpleUpgrader) Deploy(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error {
+func (u *simpleUpgrader) Deploy(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) *api.PluginError {
 	err := u.createClients(ctx, cs)
 	if err != nil {
-		return err
+		return &api.PluginError{Err: err, Step: api.PluginStepDeploy}
```
kargakis: Seems we need a separate error for client setup.

kargakis:
> How can err != nil and still get "error is:<nil> "?

https://golang.org/doc/faq#nil_error

pkg/plugin/plugin.go
kargakis: See also https://golang.org/doc/faq#nil_error

pkg/api/plugin.go
```diff
@@ -14,6 +14,34 @@ const (
...
+	PluginStepWaitForNodes            PluginStep = "WaitForNodes"
+	PluginStepHealthCheck             PluginStep = "HealthCheck"
+	PluginStepWaitForInfraDaemonSets  PluginStep = "WaitForInfraDaemonSets"
+	PluginStepWaitForInfraDeployments PluginStep = "WaitForInfraDeployments"
+)
+
+// PluginError error returned by CreateOrUpdate to specify the step that failed.
+type PluginError struct {
```
pweil-: Thoughts on making this an interface and introducing  functions to ensure that a `PluginError` is always well formed?  Suggestions:

* error cannot be nil
* step cannot be empty
* constructor could return nil if error is nil and panic if step is empty (so that the last call of methods could be `return NewPluginError(err, step)`)

pkg/plugin/plugin.go
asalkeld: OK, agreed. I'll always return error instead of PluginError

pkg/api/plugin.go
asalkeld: see: https://github.com/openshift/openshift-azure/issues/672#issuecomment-434546004

pkg/api/plugin.go
asalkeld: ack, I'll fix

pkg/api/plugin.go
asalkeld: I think the idea is that "error" is the interface that we use. My first time making extended errors, so I have not done the right thing here. I should be returning "error" everywhere and making sure to explicitly return "nil" where possible.

pkg/plugin/plugin.go
```diff
@@ -120,5 +120,9 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
 	}
 
 	log.Info("starting health check")
-	return p.clusterUpgrader.HealthCheck(ctx, cs)
+	err = p.clusterUpgrader.HealthCheck(ctx, cs)
+	if err != nil {
+		return &api.PluginError{Err: err, Step: api.PluginStepWaitForConsoleHealth}
```
kargakis: It seems you have updated all simpleUpgrader methods apart from HealthCheck to return PluginError. Can we return this error from inside HealthCheck to stay consistent with the rest?

pkg/api/plugin.go
pweil-: things can implement [multiple interfaces](https://play.golang.org/p/IQ60H3YN41q).  Shouldn't be an issue returning `error` in signatures, even from a `NewPluginError` constructor.  

I guess I should note that Kube just uses a struct for their structured validation errors (in apimachinery utils).  But it does seem like we want to police the structure so we do not end up with uncategorized errors or a `PluginError` with a nil error.  If folk agree it can be done with code review - just want to make sure we're deciding.

pweil-: For the record, I'm fine with this as-is.  Most of the time we want to check nil errors anyway unless we're at the end of a method

pkg/plugin/plugin.go
asalkeld: done

pkg/upgrade/update.go
```diff
@@ -12,49 +12,83 @@ import (
...
+	}
+	err = deployFn(ctx, azuretemplate)
+	if err != nil {
+		return &api.PluginError{Err: err, Step: api.PluginStepDeploy}
+	}
+	err = u.InitializeCluster(ctx, cs)
+	if err != nil {
+		return &api.PluginError{Err: err, Step: api.PluginStepInitialize}
+	}
+	err = managedcluster.WaitForHealthz(ctx, cs.Config.AdminKubeconfig)
```
kargakis: We probably need an issue to enable updates of unresponsive clusters. @jim-minter thoughts?

kargakis:
/lgtm
/hold

waiting for the upgrade test to complete

kargakis:
upgrade test is green

/hold cancel

pkg/upgrade/update.go
```diff
@@ -12,49 +12,83 @@ import (
...
+	err = u.waitForNewNodes(ctx, cs, vmsBefore)
+	if err != nil {
+		return &api.PluginError{Err: err, Step: api.PluginStepWaitForNodes}
+	}
 
+	// For PP day 1, scale is permitted but not any other sort of update.  When
+	// we enable configuration changes and/or upgrades, uncomment this code.  At
+	// the same time, current thinking is that we will add a hash-based
+	// mechanism to avoid unnecessary VM rotations as well.
+	if os.Getenv("RUNNING_UNDER_TEST") != "" {
```
amanohar: @jim-minter @asalkeld as requested here: https://github.com/openshift/openshift-azure/issues/672#issuecomment-434546004 can we more granularity in the stage of error as well?

PR #676, opened 2018-10-31
Opening up more instance types

### LINKED ISSUE #614: Support more vm sizes
Currently only Standard_D2s_v3 and Standard_D4s_v3 are the only deploy-able vm sizes (without a workaround or modifying the code).

thekad:
Fixes #614 

Previously we were only allowing the smallest instance types to be
created. There really is no reason to not allow more instance types, so
these should be available.
We get validation "for free" in `pkg/api/validate.go` with these changes

I didn't change Kube/System reserved resources but maybe I should, open to suggestions on what values make sense


pkg/api/default.go
```diff
@@ -28,6 +28,214 @@ var DefaultVMSizeKubeArguments = map[VMSize]map[AgentPoolProfileRole]map[Reserve
 			SystemReserved: "cpu=500m,memory=512Mi",
 		},
 	},
+	StandardD8sV3: {
```
kwoodson: I think we need to rework this.

each instance size has different mem, disk, cpu requirements.  We should probably make this easier by defaulting to a percentage of cpu and memory for instance types and also allow someone the ability to specify it.  That would simplify this a bit.

Thoughts?

charlesakalugwu: I like the idea of defaulting to a percentage of the available resources. +1

PR #677, opened 2018-10-31
Adding another e2e test

### LINKED ISSUE #645: E2e test for persistent storage
Either [this](https://github.com/openshift/openshift-azure/blob/eead2c49802df1923e4023dbbe7e3a4653586d5f/test/e2e/end_user_test.go#L54) test should be enhanced to use persistent storage in the db, then do a pod db deletion, wait for the new pod to come up and ensure the data in the storage are the same OR add a separate test. Potentially, the existing test will get really slow and flaky so we may want to think about adding a separate test and use an exec probe to add data in its storage then ensure the data gets persisted.

/area tests

thekad:
Fixes #645 

This should:
* Deploy a template with persistent storage
* Get the ingress routes from the deployment
* Hit the routes several times, ensuring the visit counter increments
* Kill the database pod
* Wait for the pod to come back
* Hit the route again a few times

### DIFF
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -6,7 +6,10 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"time"
 
 	"github.com/ghodss/yaml"
@@ -212,4 +215,45 @@ func (t *testClient) instantiateTemplate(tpl string) error {
 
 	// Return after waiting for instance to complete
 	return wait.PollImmediate(2*time.Second, 10*time.Minute, c.templateInstanceIsReady)
+}
+
+func (t *testClient) loopHTTPGet(url string, regex *regexp.Regexp, times int) func() error {
+
+	httpc := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+	var prevCounter, currCounter int
+
+	return func() error {
+		for i := 0; i < times; i++ {
+			resp, err := httpc.Get(url)
+			if err != nil {
+				return err
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != http.StatusOK {
+				return fmt.Errorf("unexpected http error returned: %d", resp.StatusCode)
+			}
+
+			contents, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				return err
+			}
+			matches := regex.FindStringSubmatch(string(contents))
+			if matches == nil {
+				return fmt.Errorf("no matches found for %s", regex)
+			}
+
+			currCounter, err = strconv.Atoi(matches[1])
+			if err != nil {
+				return err
+			}
+			if currCounter <= prevCounter {
+				return fmt.Errorf("visit counter didn't increment: %d should be > than %d", currCounter, prevCounter)
+			}
+			prevCounter = currCounter
+			time.Sleep(time.Second)
+		}
+		return nil
+	}
 }
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/ghodss/yaml"
 	project "github.com/openshift/api/project/v1"
+	templatev1 "github.com/openshift/api/template/v1"
+	appsv1 "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	projectclient "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
 	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	templatev1client "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
@@ -26,6 +28,7 @@ import (
 var c, cadmin, creader *testClient
 
 type testClient struct {
+	ac        *appsv1.AppsV1Client
 	kc        *kubernetes.Clientset
 	pc        *projectclient.ProjectV1Client
 	rc        *routev1client.RouteV1Client
@@ -83,7 +86,13 @@ func newTestClient(kubeconfig, artifactDir string) *testClient {
 		panic(err)
 	}
 
+	ac, err := appsv1.NewForConfig(config)
+	if err != nil {
+		panic(err)
+	}
+
 	return &testClient{
+		ac:          ac,
 		kc:          kc,
 		pc:          pc,
 		rc:          rc,
@@ -177,4 +186,30 @@ func (t *testClient) dumpInfo() error {
 		fmt.Println(podBuf.String())
 	}
 	return nil
+}
+
+func (t *testClient) instantiateTemplate(tpl string) error {
+	// Create the template
+	template, err := t.tc.Templates("openshift").Get(
+		tpl, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Instantiate the template
+	_, err = t.tc.TemplateInstances(c.namespace).Create(
+		&templatev1.TemplateInstance{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: c.namespace,
+			},
+			Spec: templatev1.TemplateInstanceSpec{
+				Template: *template,
+			},
+		})
+	if err != nil {
+		return err
+	}
+
+	// Return after waiting for instance to complete
+	return wait.PollImmediate(2*time.Second, 10*time.Minute, c.templateInstanceIsReady)
 }
--- a/test/e2e/client_test.go
+++ b/test/e2e/client_test.go
@@ -6,10 +6,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"path/filepath"
-	"regexp"
-	"strconv"
 	"time"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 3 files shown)


test/e2e/end_user_test.go
```diff
@@ -23,6 +23,92 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+var _ = Describe("Openshift on Azure end user tests [AdvancedEndUser]", func() {
```
kargakis: Why is this advanced? It should be something end-users should be able to run. I don't see any gain for the separate suite.

hack/e2e.sh
```diff
@@ -22,6 +22,19 @@ if [[ -z "$SUITE" || "$SUITE" == "enduser" ]]; then
   oc logout
 fi
 
+# These don't get run automatically
+if [[ "$SUITE" == "advanced-e2e" ]]; then
```
kargakis: Please remove the additional suite, this is an end user test.

thekad:
The idea was to add another suite that is not automatically run every time.
By adding another suite with a different [tag] I can call go test with a
different ginkgo.focus parameter, while keeping the [EndUser] test as it is
(other than changing what instance gets instantiated)

thekad:
To be clear: I am not married to the name, so whatever name/id I am ok
with. Just picked that one because I couldn't think of anything else.

kargakis:
Why wouldn't this test run automatically? We should not be adding new test
suites without a very good reason because it increases complexity (when
does the new suite run? why do we need it on top of the existing suites?)

test/e2e/end_user_test.go
```diff
@@ -163,4 +128,72 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		}
+		pods, err := c.kc.CoreV1().Pods(c.namespace).List(listOptions)
+		Expect(err).NotTo(HaveOccurred())
+		pgsqlPod := pods.Items[0]
+		By(fmt.Sprintf("deleting the database pod (%v)", time.Now()))
+		err = c.kc.CoreV1().Pods(c.namespace).Delete(pgsqlPod.Name, &metav1.DeleteOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Pod should be re-created automatically
+		time.Sleep(5 * time.Second)
```
kargakis: You shouldn't be sleeping here and rely only on PollImmediate

test/e2e/client_test.go
```diff
@@ -178,3 +180,41 @@ func (t *testClient) dumpInfo() error {
 	}
 	return nil
 }
+
+func (t *testClient) instantiateTemplate(tpl string) (*routev1.Route, error) {
+
```
kargakis: kill this newline

kargakis:
a couple of comments, lgtm overall

test/e2e/end_user_test.go
thekad: @kargakis it appears to take a bit for the pod to go from Ready -> Deleting so the test flakes some times because the polling returns true before the pod is deleted and the http GET failes later on

kargakis: deleting the pod and waiting an arbitrary amount of time is going to be flaky. A more robust solution would be to scale the deployment/statefulset to zero, wait for zero pods to exist, then scale back to one, wait for one pod to exist.

thekad:
@mjudeikis @kwoodson @kargakis ptal

kwoodson:
@thekad, looks like we need an import fix:
```
You have go import errors in the below files, please run goimports -e -w -local=github.com/openshift/openshift-azure pkg cmd test test/e2e/end_user_test.go
```

test/e2e/client_test.go
```diff
@@ -178,3 +188,40 @@ func (t *testClient) dumpInfo() error {
 	}
 	return nil
 }
+
+func (t *testClient) instantiateTemplate(tpl string) (*routev1.Route, error) {
```
mjudeikis: I dont think so but returning route from template initiation function does not look right. First, if there is more than 1 route? If there are no routes at all? If the route is named differently (L221).

test/e2e/end_user_test.go
```diff
@@ -163,4 +131,90 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		}
+
+		// Find the database deploymentconfig and scale to 0
+		var dcName = "postgresql"
+		for i := range []int{0, 1} {
+			By(fmt.Sprintf("searching for the database deploymentconfig (%v)", time.Now()))
+			dbDeploymentConfig, err := c.ac.DeploymentConfigs(c.namespace).Get(dcName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(dbDeploymentConfig).NotTo(BeNil())
+			By(fmt.Sprintf("scaling the database deploymentconfig to %d (%v)", i, time.Now()))
```
mjudeikis: ni: add new lines before separate logical actions get-> check-> /n -> get -> check

test/e2e/client_test.go
kargakis: Agreed it's a weird signature. I wouldn't add this as a method in testClient but would probably keep it as a function as close to the test that is using it as possible.

test/e2e/client_test.go
```diff
@@ -178,3 +187,34 @@ func (t *testClient) dumpInfo() error {
...
+			Spec: templatev1.TemplateInstanceSpec{
+				Template: *template,
+			},
+		})
+	if err != nil {
+		return err
+	}
+
+	// Poll for the deployment status
+	err = wait.PollImmediate(2*time.Second, 10*time.Minute, c.templateInstanceIsReady)
```
jim-minter: nit: you can shortcut everything below to `return wait.PollImmediate(2*time.Second, 10*time.Minute, c.templateInstanceIsReady)`

test/e2e/end_user_test.go
```diff
@@ -59,63 +62,32 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
-			time.Sleep(time.Second)
-		}
+		By(fmt.Sprintf("hitting the route and checking the contents (%v)", time.Now()))
+		resp, err := httpc.Get(url)
+		Expect(err).NotTo(HaveOccurred())
+		defer resp.Body.Close()
+		Expect(resp.StatusCode).Should(Equal(http.StatusOK))
+		contents, err := ioutil.ReadAll(resp.Body)
+		Expect(err).NotTo(HaveOccurred())
+		found := strings.Contains(string(contents), "Welcome to your static nginx application on OpenShift")
```
jim-minter: nit: or `Expect(contents).Should(ContainSubstring("Welcome to your static nginx application on OpenShift"))`

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+			currCounter, err = strconv.Atoi(matches[1])
+			Expect(err).NotTo(HaveOccurred())
+			Expect(currCounter).Should(BeNumerically(">", prevCounter))
+			prevCounter = currCounter
+			time.Sleep(time.Second)
+		}
+
+		// Find the database deploymentconfig and scale to 0
+		var dcName = "postgresql"
+		for i := range []int{0, 1} {
```
jim-minter: `for i := range []int32{0, 1}` better?

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+			dbDeploymentConfig, err := c.ac.DeploymentConfigs(c.namespace).Get(dcName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(dbDeploymentConfig).NotTo(BeNil())
+			By(fmt.Sprintf("scaling the database deploymentconfig to %d (%v)", i, time.Now()))
+			dbDeploymentConfig.Spec.Replicas = int32(i)
+			updated, err := c.ac.DeploymentConfigs(c.namespace).Update(dbDeploymentConfig)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updated).NotTo(BeNil())
+			By(fmt.Sprintf("waiting for database deploymentconfig to reflect %d replicas (%v)", i, time.Now()))
+			waitErr := wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
```
jim-minter: feel like this should be a helper fn somewhere

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
 		fmt.Println(result.Error().Error())
 		Expect(result.Error().Error()).To(ContainSubstring("pods \"sync-master-000000\" is forbidden: User \"enduser\" cannot get pods/log in the namespace \"kube-system\""))
 	})
+
+	It("should deploy a template with persistent storage and test failure modes", func() {
+		tpl := "django-psql-persistent"
+		By(fmt.Sprintf("instantiating the template and getting the route (%v)", time.Now()))
```
jim-minter: feel like this should probably be a helper fn somewhere?

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		host := route.Status.Ingress[0].Host
+		url := fmt.Sprintf("http://%s", host)
+
+		// Curl the endpoint and search for a string
+		httpc := &http.Client{
+			Timeout: 10 * time.Second,
+		}
+		var regex = regexp.MustCompile(`Page views:\s*(\d+)`)
+		By(fmt.Sprintf("hitting the route 3 times, expecting counter to increment (%v)", time.Now()))
+		var prevCounter, currCounter int
```
jim-minter: scope currCounter into the loop

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+		// make sure only 1 ingress point is returned
+		Expect(len(route.Status.Ingress)).To(Equal(1))
+		host := route.Status.Ingress[0].Host
+		url := fmt.Sprintf("http://%s", host)
+
+		// Curl the endpoint and search for a string
+		httpc := &http.Client{
+			Timeout: 10 * time.Second,
+		}
+		var regex = regexp.MustCompile(`Page views:\s*(\d+)`)
```
jim-minter: nit: I'd prefer it stylistically if you did `foo :=` consistently instead of `var foo =`

test/e2e/end_user_test.go
```diff
@@ -163,4 +135,99 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser]", func() {
...
+			Expect(waitErr).NotTo(HaveOccurred())
+		}
+
+		// wait for the ingress to return 200 in case healthcheck failed when database got recreated
+		By(fmt.Sprintf("making sure the ingress is healthy (%v)", time.Now()))
+		var rt http.RoundTripper
+		waitErr := waitutil.ForHTTPStatusOk(context.Background(), rt, url)
+		Expect(waitErr).NotTo(HaveOccurred())
+
+		By(fmt.Sprintf("hitting the route again, expecting counter to increment from last=%d (%v)", currCounter, time.Now()))
```
jim-minter: would prefer this code not to be repeated.  You could define a closure (nested function) if you wanted and call it pre- and post- db restart.

test/e2e/end_user_test.go
thekad: I actually tried that before, but weirdly: https://play.golang.org/p/ar68WXvytUH

PR #679, opened 2018-10-31
Remove local htpasswd auth config when not testing.

### LINKED ISSUE #669: lack RUNNING_UNDER_TEST=true breaks local create
Lack of `RUNNING_UNDER_TEST=true` makes cluster creation to fail

kwoodson:
Fixes #669 

Remove local httpasswd auth provider when not `RUNNING_UNDER_TEST=true`


kargakis:
/lgtm
/hold

@jim-minter thoughts?

mjudeikis:
@jim-minter background. We had a long discussion in IRC and came to the conclusion that we don't even use now this in PROD, and it will be gone either way in GA. So remove it in prod deployments at all. For this, we have kube-config if needed and tests are using AAD users you created. 

And it will look cleaner in UI with one identity provider :)

p.s. its broken in prod now too

kargakis:
Based on scrum discussion, we can revert if @jim-minter has something else in mind

/hold cancel

mjudeikis:
merge, fix fallout later :) I prefer working code over broken one :D

Issue #642, opened 2018-10-23
pkg/fakerp hardcodes the location of _data


kargakis:
We should probably make it configurable.

/kind cleanup

kargakis:
@charlesakalugwu actually I don't have a strong reason for making it configurable but I thought it would be useful. Let's do it only when the need rises (I suspect it won't).

PR #681, opened 2018-11-01
WIP backup/recovery - not fully tested

asalkeld:
I have only tested bits:
- the backup download container is working nicely to retrieve the snapshot
- I am using quay.io/asalkeld/getbackup:latest (I dont' have permission to quay.io/openshift-azure/)
- next step is to test etcdctl snapshot restore
- then test end-to-end

Note: 
- the backup is not working as I would expect. It *is* running every 10 minutes but it keeps writing over blob "etcd/backup" (you can see the timestamp getting updated)

/hold


Dockerfile.getbackup
```diff
@@ -0,0 +1,6 @@
+FROM centos:7
+RUN yum -y update
+COPY getbackup .
+ENTRYPOINT [ "/getbackup" ]
+# example:
+# docker run -it -v /etc/origin/cloudprovider/:/_data/_out -v /var/lib/etcd:/out:z quay.io/asalkeld/getbackup:dbbeeaaa --blobname=etcd/backup --destination=/out/backup
```
kargakis: Add this somewhere in the readme instead of here

kargakis: Or remove altogether

cmd/getbackup/getbackup.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/openshift-azure/pkg/log"
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	azureclientstorage "github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
+	"github.com/openshift/openshift-azure/pkg/util/nodeconf"
+)
+
+var (
+	logLevel    = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
```
kargakis: Keep the help messages across all flags consistent: start with a capital letter.

cmd/getbackup/getbackup.go
```diff
@@ -0,0 +1,164 @@
...
+	interval    = flag.Duration("interval", 5*time.Second, "How often to retry the download.")
+	blobName    = flag.String("blobname", "", "name of the blog to download")
```
kargakis: s/blog/blob/

pkg/arm/arm.go
```diff
@@ -54,12 +54,15 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
 	if err != nil {
 		return nil, err
 	}
+	log.Infof("*** IsRecovery: %v, BackupBlobName: %v", len(g.pluginConfig.RecoverEtcdFromBackup) > 0, g.pluginConfig.RecoverEtcdFromBackup)
```
kargakis: remove

pkg/fakerp/fakerp.go
```diff
@@ -94,7 +92,7 @@ func CreateOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 
 	// write out development files
 	log.Info("write helpers")
-	err = writeHelpers(cs, azuretemplate)
```
kargakis: revert this change since it's only used in here.

pkg/util/nodeconf/nodeconf.go
```diff
@@ -0,0 +1,65 @@
...
+	b, err = yaml.Marshal(c.Config.AdminKubeconfig)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile("_data/_out/admin.kubeconfig", b, 0600)
+}
+
+// GetAzureConf returns the azure credentials to be used by sync and getbackup
+func GetAzureConf() (map[string]string, error) {
+	b, err := ioutil.ReadFile("_data/_out/azure.conf")
```
kargakis: If we want this in a shared library, I'd probably make _data configurable.

cmd/getbackup/getbackup.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+	defer rc.Close()
+	logrus.Print("read blob")
+
+	b, err := ioutil.ReadAll(rc)
+	if err != nil {
+		return err
+	}
+	logrus.Print("creating ", destPath)
```
kargakis: prefer using string formatting, also is consistent with what you do elsewhere

cmd/getbackup/getbackup.go
```diff
@@ -0,0 +1,164 @@
...
+	df, err := os.Create(destPath)
+	if err != nil {
+		return err
+	}
+	defer df.Close()
+	_, err = df.Write(b)
+	logrus.Print("Write returned ", err)
```
kargakis: unnecessary, the error is handled already in the caller

cmd/getbackup/getbackup.go
```diff
@@ -0,0 +1,164 @@
...
+	return err
+}
+
+func main() {
+	flag.Parse()
+	logrus.SetLevel(log.SanitizeLogLevel(*logLevel))
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	logrus.Printf("getbackup pod starting, git commit %s", gitCommit)
```
kargakis: this app shouldn't really care whether it runs in a pod or not, also you seem to run it as a bare container.

pkg/arm/data/master-startup.sh
```diff
@@ -659,6 +659,16 @@ sed -i -re "s#( *server: ).*#\1https://$(hostname)#" /etc/origin/node/node.kubec
 # daemonset will not fail and set the master node labels correctly.
 cp /etc/origin/node/node.kubeconfig /etc/origin/node/bootstrap.kubeconfig
 
+{{- if $.Extra.IsRecovery }}
+tempBackDir=$(mktemp -d -t etcd-backup)
+docker run -v /etc/origin/cloudprovider/:/_data/_out -v $tempBackDir:/out:z {{ .Config.Images.GetBackup }} --blobname=$.Extra.BackupBlobName --destination=/out/backup.db
```
kargakis: @aaronlevy fyi, this is a proposed solution for how we are going to restore from backups. Is this something handled during Ignition in the 4.x installer today? If not, how are you thinking of doing restores?

pkg/addons/data/ConfigMap/openshift-etcd/backup-config.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
```
kargakis: Do we need to keep `pkg/addons/data/EtcdBackup.etcd.database.coreos.com/openshift-etcd/etcd-backup.yaml` with this?

pkg/addons/data/CronJob.batch/openshift-etcd/etcd-backup-cleaner.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: etcd-backup-cleaner
+  namespace: openshift-etcd
+spec:
+  # todo: move to 1 day after testing
+  schedule: "0 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: etcd-operator
```
kargakis: The etcd-operator has elevated access where this cronjob only needs to get and delete an etcdbackup. Needs its own service account, role, and rolebinding.

pkg/addons/data/CronJob.batch/openshift-etcd/etcd-backup.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: etcd-backup
+  namespace: openshift-etcd
+spec:
+  # todo: move to 1 hour after testing
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: etcd-operator
```
kargakis: The etcd-operator has elevated access where this cronjob only needs to create an etcdbackup. Needs its own service account, role, and rolebinding.

pkg/addons/data/ConfigMap/openshift-etcd/backup-config.yaml
kargakis: Nevermind, I see you remove this in this PR.

pkg/config/images.go
```diff
@@ -97,6 +98,7 @@ func (g *simpleGenerator) selectContainerImagesOrigin(cs *api.OpenShiftManagedCl
 		c.Images.RegistryConsole = "docker.io/cockpit/kubernetes:latest"
 		c.Images.Sync = "quay.io/openshift-on-azure/sync:v3.10"
 		c.Images.LogBridge = "quay.io/openshift-on-azure/logbridge:latest"
+		c.Images.GetBackup = "quay.io/openshift-on-azure/getbackup:latest"
```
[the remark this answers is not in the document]
asalkeld: all it does is download a blob to a local file

Dockerfile.getbackup
asalkeld: done

cmd/getbackup/getbackup.go
asalkeld: thanks

cmd/getbackup/getbackup.go
asalkeld: hehe,. oops

pkg/fakerp/fakerp.go
asalkeld: ok, i'll revert. My thought was to have the read/write next to each other for maintainability.

cmd/getbackup/getbackup.go
asalkeld: done

cmd/getbackup/getbackup.go
asalkeld: yeah, debug artifact

cmd/getbackup/getbackup.go
asalkeld: ack

asalkeld:
so after a bit of testing, I am going to have to solve the hostname problem. (when you delete the vms and update the hostnames just keep incrementing). So my plan is to make another little binary that runs in a container that lists the vms in the scale set and does the following
(in master-startup.sh before any use of $(hostname))

# call the container: "what-should-my-hostname-be" $(hostname)
# it lists the vms in the scale set (and waits until there are 3) and sorts them by name
# it then returns what this host should be called
# eg:
# master-03 -> master-00
# master-07 -> master-01
# master-23 -> master-02
we can then set the hostname - getting stable hostnames on the masters

cc @jim-minter

pkg/fakerp/fakerp.go
kargakis: Agreed, they are all over the place, we may want to revive https://github.com/openshift/openshift-azure/issues/642

kargakis:
@asalkeld can this be closed in favor of https://github.com/openshift/openshift-azure/pull/713?

asalkeld:
> @asalkeld can this be closed in favor of #713?

sure, i'll open a new one with just the recovery part once the backup is merged

PR #686, opened 2018-11-01
createorupdate: env checks and random region selection

kargakis:
Collapsing some parts of our bash scripts in createorupdate and enabling random region selection to ensure in our tests we can deploy clusters in every supported region.

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -118,7 +118,7 @@ func (s *server) handleDelete(w http.ResponseWriter, req *http.Request) {
 	authorizer, err := azureclient.NewAuthorizer(conf.ClientID, conf.ClientSecret, conf.TenantID)
 	if err != nil {
 		resp := "500 Internal Error: Failed to determine request credentials"
-		s.log.Debug(resp, err)
+		s.log.Debugf("%s: %v", resp, err)
 		http.Error(w, resp, http.StatusInternalServerError)
 		return
 	}
@@ -134,20 +134,20 @@ func (s *server) handleDelete(w http.ResponseWriter, req *http.Request) {
 	future, err := gc.Delete(ctx, resourceGroup)
 	if err != nil {
 		resp := "500 Internal Error: Failed to delete resource group"
-		s.log.Debug(resp, err)
+		s.log.Debugf("%s: %v", resp, err)
 		http.Error(w, resp, http.StatusInternalServerError)
 		return
 	}
 	if err := future.WaitForCompletionRef(ctx, gc.Client); err != nil {
 		resp := "500 Internal Error: Failed to wait for resource group deletion"
-		s.log.Debug(resp, err)
+		s.log.Debugf("%s: %v", resp, err)
 		http.Error(w, resp, http.StatusInternalServerError)
 		return
 	}
 	resp, err := future.Result(gc)
 	if err != nil {
 		resp := "500 Internal Error: Failed to get resource group deletion response"
-		s.log.Debug(resp, err)
+		s.log.Debugf("%s: %v", resp, err)
 		http.Error(w, resp, http.StatusInternalServerError)
 		return
 	}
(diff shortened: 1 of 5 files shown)


kargakis:
@asalkeld @mjudeikis this one should be of interest to you.

cmd/createorupdate/createorupdate.go
```diff
@@ -263,6 +267,44 @@ func (s *server) reply(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	var supported bool
+	for _, region := range supportedRegions {
+		if c.Region == region {
+			supported = true
+		}
+	}
+	if !supported {
+		return fmt.Errorf("unsupported region: %q - supported regions are eastus, westeurope, and australiasoutheast", c.Region)
```
charlesakalugwu: use `supportedRegions` instead of hardcoding the regions in the error message

cmd/createorupdate/createorupdate.go
```diff
@@ -367,20 +409,61 @@ func execCommand(c string) error {
...
+	if err != nil {
+		return err
+	}
+	gc := resources.NewGroupsClient(conf.SubscriptionID)
+	gc.Authorizer = authorizer
+
+	var tags map[string]*string
+	if !conf.NoGroupTags {
+		tags = make(map[string]*string)
+		ttl, now := "76h", fmt.Sprintf("%d", time.Now().Unix())
```
charlesakalugwu: this seems fine

cmd/createorupdate/createorupdate.go
```diff
@@ -263,6 +267,44 @@ func (s *server) reply(w http.ResponseWriter, req *http.Request) {
...
+	ResourceGroupTTL string `envconfig:"RESOURCEGROUP_TTL"`
+}
+
+func (c *config) init() error {
+	supportedRegions := []string{"eastus", "westeurope", "australiasoutheast"}
+	if c.Region == "" {
+		// Randomly assign a supported region
+		rand.Seed(time.Now().UTC().UnixNano())
+		c.Region = supportedRegions[rand.Intn(3)]
+		logrus.Infof("Using randomly selected region %q", c.Region)
```
asalkeld: you could also just default to eastus

cmd/createorupdate/createorupdate.go
```diff
@@ -263,6 +267,44 @@ func (s *server) reply(w http.ResponseWriter, req *http.Request) {
...
+	Region           string `envconfig:"AZURE_REGION"`
+	DnsDomain        string `envconfig:"DNS_DOMAIN" required:"true"`
+	DnsResourceGroup string `envconfig:"DNS_RESOURCEGROUP" required:"true"`
+	ResourceGroup    string `envconfig:"RESOURCEGROUP" required:"true"`
+
+	NoGroupTags      bool   `envconfig:"NOGROUPTAGS"`
+	ResourceGroupTTL string `envconfig:"RESOURCEGROUP_TTL"`
+}
+
+func (c *config) init() error {
```
asalkeld: perhaps changes this to validateRegion or initRegion?

cmd/createorupdate/createorupdate.go
kargakis: The point of using random regions is to have our tests ensure we can deploy in every supported region. We could have more scoped tests that always deploy in a specific region but it's more unnecessary work at this point.

cmd/createorupdate/createorupdate.go
```diff
@@ -367,20 +409,61 @@ func execCommand(c string) error {
 	return nil
 }
 
+func setup() error {
```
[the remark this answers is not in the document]
kargakis: I kept this name generic on purpose since I expect the aad setup (application update) to happen here once we move away from hack/aad.sh.

cmd/createorupdate/createorupdate.go
kargakis: This function actually does two things, it both validates and defaults things in the config.

cmd/createorupdate/createorupdate.go
kargakis: done

cmd/createorupdate/createorupdate.go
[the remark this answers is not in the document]
kargakis: Hrm, probably a more appropriate place for setting up aad is inside the fake rp instead of here client-side.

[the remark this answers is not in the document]
kargakis: renamed to `createResourceGroup`

kargakis:
@asalkeld @charlesakalugwu any more comments here?

PR #693, opened 2018-11-05
[WIP] Customer admin operator.

kwoodson:


controllers/customer-admin-controller/Gopkg.toml
```diff
@@ -0,0 +1,54 @@
+# Force dep to vendor the code generators, which aren't imported just used at dev time.
```
kargakis: Why a totally separate place to manage dependencies? I thought we had agreed we were going to manage deps in a single place.

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/openshift/openshift-azure/controllers/customer-admin-controller/pkg/common"
+)
+
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
```
kargakis: Remove

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if !common.RestrictedNamespace(request.Name) {
+		// move to appropriate place
+		// If we are namespace we need to reconcile 3 rolebindings:
+		// - customer-admin
+		// - customer-project-admin
+		// - customer-view
+		log.Printf("Reconciling Namespace %s/%s\n", request.Namespace, request.Name)
```
kargakis: You don't need \n with log. Also, should we start using our own log package or at least logrus? I'd go with logrus if we don't want to pull any deps from our repo. I am fine with switching to logrus in a follow-up.

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+
+		// check for existence of namespace
+		ns := &corev1.Namespace{}
+		err := r.client.Get(context.TODO(), request.NamespacedName, ns)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				log.Printf("error: %v", err)
+				return reconcile.Result{}, nil
+			}
```
kargakis: Do you want to continue when err != nil and != 404?

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+			Namespace: request.Name,
+		}
+		for _, rbname := range common.RoleBindingNames {
+			nrb := common.CustomerRoleBindingMap()(rbname)
+			nrb.Namespace = request.Name
+			nn.Name = rbname
+			rb := &rbacv1.RoleBinding{}
+			err := r.client.Get(context.TODO(), nn, rb)
+			if err != nil {
+				if kerrors.IsNotFound(err) {
```
kargakis: It would be clearer (and less arrowized) to handle IsNotFound out of err != nil

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+					continue
+				}
+
+				return reconcile.Result{}, err
+			}
+			// check that our rolebinding is correct by seeing if the roleref
+			// and subjects are in desired state
+			missingSubjects := common.MissingSubjectsFromRoleBinding(&nrb, rb)
+			if len(missingSubjects) == 0 {
+				log.Printf("Rolebinding %v looks ok!\n", rbname)
```
kargakis: no need for the newline

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,137 @@
...
+				continue // rolebinding looks ok
+			}
+
+			// append missing subjects
+			rb.Subjects = append(rb.Subjects, missingSubjects...)
+			log.Printf("Updating rolebinging[%v]", rbname)
+			err = r.client.Update(context.TODO(), rb)
+			if err != nil {
+				log.Printf("Error updating rolebinding: %v\n", err)
```
kargakis: no need for the newline

controllers/customer-admin-controller/.gitignore
```diff
@@ -0,0 +1,77 @@
+# Temporary Build Files
```
kargakis: Since this is an image just like sync, I would try to follow the current structure we use in the repo, that is the entrypoint of the binary is found under `cmd` and its libraries under `pkg`.

controllers/customer-admin-controller/build/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.6
```
mjudeikis: s/alpine/centos/

controllers/customer-admin-controller/Gopkg.toml
kwoodson: This will definitely get placed under the glide file.  I just plopped my code here.  I didn't want to refactor everything if this wasn't the final resting spot.

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: I didn't know about the `\n`, thanks.  I'm fine with logrus.

controllers/customer-admin-controller/.gitignore
kwoodson: Ok, i'll attempt to move it under our current pattern of sync.

controllers/customer-admin-controller/build/Dockerfile
kwoodson: I haven't actually built this but that definitely makes sense.

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: Will clean up.

controllers/customer-admin-controller/.gitignore
kwoodson: I attempted to do this like the sync pod but ran into some errors while attempting to merge vendor directories.  The operator-sdk relies on newer libraries than we are allowing in our glide.yaml file.  I think it relies on 1.11.2 libraries and we are locking to 1.10.6.

I probably need more assistance if we attempt to place this controller in our code base.

jim-minter:
This looks really interesting @kwoodson, excellent work and thanks for taking the initiative!

sigs.k8s.io/controller-runtime > v0.1.1 rely on the 1.11 client libraries, which we can't move to until MSFT is ready.
With sigs.k8s.io/controller-runtime == v0.1.1, it at least compiles for me without the nested vendor dir.

Please can you get rid of the nested vendor directory and the Gopkg.*, add 
```
- package: sigs.k8s.io/controller-runtime
  version: v0.1.1
```
to glide.yaml, then do a glide up -v.

Commit that as a separate vendor commit, then make a second commit with everything else and kargakis' changes, on top.

Then would like to take a closer look at this.

controllers/customer-admin-controller/pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: updated to logrus.  also removed teh `\n`

controllers/customer-admin-controller/Gopkg.toml
kwoodson: I've taken @jim-minter's advice and removed the controllers `vendor` directory and `gopkg` files.

kwoodson:
@jim-minter,  I have cleaned up the code a bit.  The original code was under a single package.  I had a chat with a team member and we decided multiple controllers to manage the reconciling might make more sense.  One for rolebindings and one for namespaces.  Since the reconciling process does very similar work I created a common pkg to share the functionality between the rolebinding and namespace controllers.

If we merge those pkgs back into a single package it would shrink the code a bit more. 3 pkg down to a single pkg.  I would then need to come up with a way to reduce duplication in the logic.  The single pkg version can be seen here https://gist.github.com/kwoodson/d80862a2d777e15835d99081e93c1398.  This probably makes more sense being in a larger repository with shared code.

As stated above, most of this code is boilerplate and the two import functions are the reconcile function bits in the pkgs.

Thanks!

kwoodson:
Fixed import statements with `goimports`.

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -75,12 +69,13 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 		// - customer-admin
 		// - customer-project-admin
 		// - customer-view
-		log.Printf("Reconciling Namespace %s/%s\n", request.Namespace, request.Name)
+		log.Printf("Reconciling Namespace %s/%s", request.Namespace, request.Name)
```
mjudeikis: move to logrus as all our code?

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
+package common
```
mjudeikis: name does not sound right :/ Maybe we should move all this operator to subpackage like:
```
pkg/operators/rbac/common
pkg/operators/rbac/controllernamespaces
...
``` 

@kargakis @jim-minter wdyt?

pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: Yeah, I think during my rebase-rework I didn't include these.  Thanks, will fix.

pkg/common/common.go
kwoodson: I agree with the common being out of place.  A self containing option would be prefered.  I'll defer to the more seasoned go veterans for pkg help.

kwoodson:
I tested this code in cluster from an image and everything looks to be working.
```
time="2018-11-08T21:42:00Z" level=info msg="Reconciling Namespace /test3"
time="2018-11-08T21:42:00Z" level=info msg="Rolebinding customer-admin looks ok!"
time="2018-11-08T21:42:00Z" level=info msg="Rolebinding customer-admin-project looks ok!"
time="2018-11-08T21:42:00Z" level=info msg="Rolebinding customer-view looks ok!"
```

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if !common.RestrictedNamespace(request.Name) {
+		// move to appropriate place
+		// If we are namespace we need to reconcile 3 rolebindings:
+		// - customer-admin
+		// - customer-project-admin
+		// - customer-view
+		logrus.Printf("Reconciling Namespace %s/%s", request.Namespace, request.Name)
```
jim-minter: request.Namespace will presumably be "" for a namespace.  Dial down your log verbosity.

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+
+		// check for existence of namespace
+		ns := &corev1.Namespace{}
+		err := r.client.Get(context.TODO(), request.NamespacedName, ns)
+		if err != nil {
+			// namespace is not found
+			if kerrors.IsNotFound(err) {
+				logrus.Printf("error: %v", err)
```
jim-minter: this case is presumably legitimate if a namespace once existed and now it doesn't.  I don't think it's an error so don't print that it is.  The `return` is correct.

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+				return reconcile.Result{}, nil
+			}
```
jim-minter: If !kerrors.IsNotFound, you probably need to return err, right?

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+			return reconcile.Result{}, nil
+		}
+
+		nn := types.NamespacedName{
+			Namespace: request.Name,
+		}
+		for _, rbname := range common.RoleBindingNames {
+			nrb := common.CustomerRoleBindingMap()(rbname)
+			nrb.Namespace = request.Name
+			nn.Name = rbname
```
jim-minter: scope nn inwards.  Do
```golang
	nn := types.NamespacedName{
		Namespace: request.Name,
		Name: rbname,
	}
```
here and remove the `nn :=` outside the range

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+		// move to appropriate place
+		// If we are namespace we need to reconcile 3 rolebindings:
+		// - customer-admin
+		// - customer-project-admin
+		// - customer-view
+		logrus.Printf("Reconciling Namespace %s/%s", request.Namespace, request.Name)
+
+		// check for existence of namespace
+		ns := &corev1.Namespace{}
+		err := r.client.Get(context.TODO(), request.NamespacedName, ns)
```
jim-minter: do a single `ctx := context.Background()` at the top and use that throughout instead of `context.TODO()`

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+					if err != nil {
+						logrus.Printf("Error Creating: %v", err)
+						return reconcile.Result{}, err
+					}
+					continue
+				}
+
+				return reconcile.Result{}, err
+			}
+			// check that our rolebinding is correct by seeing if the roleref
```
jim-minter: I feel like I need to read up, but I suspect that the way this needs to work is:
1) the namespace controller is responsible solely for creating rolebindings when they don't exist in a namespace.  It can create a rolebinding correctly to save the rolebinding controller some effort, but it only really cares that the rolebinding exists to bootstrap the process.
2) the rolebinding controller is responsible for catching changes to rolebindings it watches and reverting them if it doesn't like them.  Also it can try to catch deletions and if it sees one recreate the object.  If it misses a deletion then (1) will deal with it on a relist.

Given the above:
* it is inevitable that there are situations in which (1) and (2) could both race to create a rolebinding at the same time.  So they need to be tolerant of failing to create because they lost the race.
* there is no need for the namespace controller to worry about changes to rolebindings, those will always be picked up by the rolebinding controller.

Thus, all the code below is not necessary, and the above can be simplified to get rid of some of the tabs.

pkg/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,132 @@
...
+		nn := types.NamespacedName{
+			Namespace: request.Name,
+		}
+		for _, rbname := range common.RoleBindingNames {
+			nrb := common.CustomerRoleBindingMap()(rbname)
+			nrb.Namespace = request.Name
+			nn.Name = rbname
+			rb := &rbacv1.RoleBinding{}
+			err := r.client.Get(context.TODO(), nn, rb)
+			if err != nil {
```
jim-minter: see comment below, but in brief: if it exists, `return reconcile.Result{}, nil`.  If it doesn't, try to create.  If your create succeded, or if it failed because the object already exists (lost the race) `return reconcile.Result{}, nil`.  Otherwise, `return reconcile.Result{}, err`.

pkg/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,122 @@
...
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile receives events for rolebindings
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileRolebinding) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if !common.RestrictedNamespace(request.Name) {
```
jim-minter: carry through the same changes that were requested on the other controller.

pkg/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,122 @@
...
+			nrb.Namespace = request.Namespace
+			rb := &rbacv1.RoleBinding{}
+			//check for rolebinding existence
+			err = r.client.Get(context.TODO(), request.NamespacedName, rb)
+			if err != nil {
+				if kerrors.IsNotFound(err) {
+					// we didn't find our rolebinding, create it!
+					logrus.Printf("Creating rolebinding: %#v", nrb)
+					err = r.client.Create(context.TODO(), &nrb)
+					if err != nil {
```
jim-minter: again, be tolerant here of losing the race.

pkg/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,122 @@
...
+						logrus.Printf("Error Creating: %v", err)
+						return reconcile.Result{}, err
+					}
+				}
+			}
+			// check that our rolebinding is correct by seeing if the roleref
```
jim-minter: I don't love this approach.  I'd sooner see you do something similar to what the sync pod does.  Construct the object as you want to see it (I guess this is nrb, right?).  If it doesn't closely match what actually exists (think some sort of semi-tolerant DeepEqual), then call Update.

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
...
+		if subject == sub.Name {
+			return true
+		}
+	}
+	return false
+}
+
+// CustomerRoleBindingMap returns a rolebinding
+func CustomerRoleBindingMap() func(string) rbacv1.RoleBinding {
+	rbMap := map[string]rbacv1.RoleBinding{
```
jim-minter: whoa, just rename rbMap to CustomerRoleBindingMap and make it a top level exported var.

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
...
+		"customer-admin": {
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "customer-admin",
```
jim-minter: should name these so they don't conflict with anything a customer might want to use ("zzz-owned-by-osa-customer-admin" - you know the drift)

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
...
+func RestrictedNamespace(name string) bool {
+	return rxRestrictedNamespace.MatchString(name)
+}
+
+// CustomerRoleBinding uses a regexp to match rolebindings we manage
+func CustomerRoleBinding(name string) bool {
+	return rxCustomerRoleBinding.MatchString(name)
+}
+
+// verifyCustomerRoleBinding takes the existing rolebinding and the proposed
```
jim-minter: Hopefully this function and the next will go away.

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
...
+	"regexp"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+
+	// RoleBindingNames holds the customer rolebindings
+	RoleBindingNames = []string{
```
jim-minter: if you did `RoleBindingNames = map[string]struct{}{"customer-admin": {}, ...}` you'd be able to iterate it and do a O(1) lookup - ie not need rxCustomerRoleBinding or CustomerRoleBinding().

pkg/common/common.go
```diff
@@ -0,0 +1,119 @@
...
+		"customer-admin",
+		"customer-admin-project",
+		"customer-view",
+	}
+
+	rxCustomerRoleBinding = regexp.MustCompile(`(?i)^` +
+		`customer-(admin|view|admin-project)$`)
+
+	rxRestrictedNamespace = regexp.MustCompile(`(?i)^` +
```
jim-minter: This is subtly wrong in multiple ways.  Maybe write it out the long way without a regexp? ;-)

cmd/customer-admin-controller/customer-admin-controller.go
```diff
@@ -0,0 +1,63 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+
+	"github.com/openshift/openshift-azure/pkg/controllernamespace"
+	"github.com/openshift/openshift-azure/pkg/controllerrolebinding"
+)
+
+func printVersion() {
```
jim-minter: I'm far more interested in printing gitCommit like all the other tools do.  Can you just do that instead?

Makefile
```diff
@@ -1,13 +1,13 @@
 COMMIT=$(shell git rev-parse --short HEAD)$(shell [[ $$(git status --porcelain --ignored) = "" ]] && echo -clean || echo -dirty)
 
 # all is the default target to build everything
-all: clean build sync e2e-bin logbridge
+all: clean build sync e2e-bin logbridge customer-admin-controller
```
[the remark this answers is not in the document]
kwoodson: I like the idea of having a single image for multiple controllers.  This would simplify the follow on cases.

pkg/addons/data/Deployment.apps/openshift-infra/customer-admin-reconciler.yaml
```diff
@@ -1,60 +0,0 @@
...
-        - -c
-        image: openshift/observe
-        name: reconciler
-        volumeMounts:
-        - name: reconciler-rolebindings
-          mountPath: /tmp/rolebindings
-      volumes:
-      - configMap:
-          defaultMode: 0660
-          name: reconciler-rolebindings
```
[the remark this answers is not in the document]
kwoodson: Since these were implemented in common.go these aren't needed.

kwoodson:
@jim-minter, thanks for the review.  I'll update with your suggested changes.

Do you think it would be worth packing the namespace and rolebinding into a single controller?  Instead of 2x the files it would be a single controller?

pkg/addons/data/Deployment.apps/openshift-infra/customer-admin-reconciler.yaml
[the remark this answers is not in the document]
jim-minter: I mean, I think the reconciler-rolebindings configmap should be deleted as part of this PR as well?

jim-minter:
> Do you think it would be worth packing the namespace and rolebinding into a single controller? Instead of 2x the files it would be a single controller?

I think it needs to be 2 controllers, and they probably should be in 1 package.  If you look at the origin code for ensuring the builder/deployer/etc. service accounts are created when a new project is created, it's a similar pattern.  One controller is watching namespaces, the other service accounts (in our case, rolebindings).

pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: will remove print.

pkg/controllernamespace/namespace/namespace_controller.go
kwoodson: will add return for `err`.

kwoodson:
This fails e2e because we are replacing an item in the sync pod:
```
time="2018-11-12T15:58:38Z" level=info msg="checking deployment openshift-infra/customer-admin-reconciler" resourceGroup=ci-op-yxvnf8zb-8474e
time="2018-11-12T16:20:06Z" level=fatal msg="containerservice.OpenShiftManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: context deadline exceeded" resourceGroup=ci-op-yxvnf8zb-8474e
exit status 1
```

cmd/customer-admin-controller/customer-admin-controller.go
kwoodson: Added commit like the sync pod.

jim-minter:
> This fails e2e because we are replacing an item in the sync pod:

Looks to me like the deployment with the controller is never becoming ready?  Is that because azure-controllers wasn't created in quay? Have created it, let's see if that helps

/retest

pkg/config/images.go
```diff
@@ -104,6 +104,7 @@ func (g *simpleGenerator) selectContainerImagesOrigin(cs *api.OpenShiftManagedCl
 
 		c.Images.Sync = "quay.io/openshift-on-azure/sync:v3.11"
 		c.Images.LogBridge = "quay.io/openshift-on-azure/logbridge:latest"
+		c.Images.CustomerAdminController = "quay.io/openshift-on-azure/azure-controllers:v3.10"
```
jim-minter: change to v3.11

jim-minter:
Probably that the new controller image isn't correctly integrated in our CI - presumably it needs an env var override to be able to select the image as part of the CI run and the CI needs to be configured to do it.  @mjudeikis can you help out here?

pkg/config/images.go
```diff
@@ -149,6 +150,7 @@ func (g *simpleGenerator) selectContainerImagesOSA(cs *api.OpenShiftManagedClust
...
```
jim-minter: and this one

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Reconcile receives a namespace event and reconciles rolebindings for
+// customer roles
+// Note:
+// The Controller will requeue the Request to be processed again if the
+// returned error is non-nil or Result.Requeue is true, otherwise upon
+// completion it will remove the work from the queue.
+func (r *reconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	for _, forbiddenName := range []string{"default", "kube", "openshift", "kubernetes"} {
```
jim-minter: Still wrong! Copy the logic at https://github.com/openshift/origin/blob/master/pkg/project/apiserver/registry/projectrequest/delegated/delegated.go#L100-L126 and add `default` for good measure.

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,126 @@
...
+	for rbname := range rbs.RoleBindingNames {
+		nn := types.NamespacedName{
+			Namespace: request.Name,
+			Name:      rbname,
+		}
+		nrb := rbs.RoleBindingNames[rbname]
+		nrb.Namespace = request.Name
+		nn.Name = rbname
+		rb := &rbacv1.RoleBinding{}
+		err := r.client.Get(ctx, nn, rb)
```
jim-minter: Note: in the future dropping the Get may be a good optimisation as it will halve the number of API calls.  Can just do a Create and ignore AlreadyExists errors.

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
```
jim-minter: note: the following structure will clarify your code by reducing nesting and enabling you to remove the `continue`s below:
```golang
	switch {
	case err != nil && kerrors.IsNotFound(err):
		// ...
	case err != nil:
	}
```

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,123 @@
...
+		nn := types.NamespacedName{
+			Name: request.Namespace,
+		}
+		err := r.client.Get(ctx, nn, ns)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				// namespace isnotfound, nothing to do!
+				return reconcile.Result{}, nil
+			}
+			return reconcile.Result{}, nil
```
jim-minter: `, err`

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,123 @@
...
+		rb := rbacv1.RoleBinding{}
+		//check for rolebinding existence
+		err = r.client.Get(ctx, request.NamespacedName, &rb)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				// we didn't find our rolebinding, create it!
+				logrus.Printf("RBC: Creating rolebinding[%v] in namespace[%v]", nrb.Name, nrb.Namespace)
+				err = r.client.Create(ctx, &nrb)
+				if err != nil {
+					logrus.Printf("RBC: Error Creating rolebinding[%v]: %v", nrb.Name, err)
```
jim-minter: you can lose the race here as well

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
```diff
@@ -0,0 +1,123 @@
...
+			return reconcile.Result{}, err
+		}
+		// check that our rolebinding matches roleref and subjects
+		if rbs.RoleBindingAsExpected(nrb, rb) {
+			logrus.Printf("RBC: Rolebinding[%v] looks ok!", request.Name)
+			return reconcile.Result{}, nil
+		}
+
+		logrus.Printf("RBC: Updating rolebinging[%v] in namespace[%v]\n", request.Name, request.Namespace)
+		err = r.client.Update(context.TODO(), &nrb)
```
jim-minter: you have a ctx above

pkg/controllers/customeradmin/rolebindings/rolebindings.go
```diff
@@ -0,0 +1,97 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	// RoleBindingNames holds the customer-admin desired rolebindings
+	RoleBindingNames = map[string]rbacv1.RoleBinding{
+		"customer-admin": {
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "customer-admin",
```
jim-minter: still need to name these in a way that is unlikely to collide

pkg/controllers/customeradmin/rolebindings/rolebindings.go
```diff
@@ -0,0 +1,97 @@
...
+			},
+			Subjects: []rbacv1.Subject{
+				{
+					Kind: "Group",
+					Name: "customer-admins",
```
jim-minter: and the group names, really

pkg/controllers/customeradmin/rolebindings/rolebindings.go
```diff
@@ -0,0 +1,97 @@
...
+				},
+			},
+			RoleRef: rbacv1.RoleRef{
+				Kind: "ClusterRole",
+				Name: "view",
+			},
+		},
+	}
+
+	// RoleBindingNames holds the customer rolebindings
```
jim-minter: delete

pkg/controllers/customeradmin/rolebindings/rolebindings.go
```diff
@@ -0,0 +1,97 @@
...
+		`default|` +
+		`kubernetes)-?`)
+	*/
+)
+
+// RoleBindingAsExpected takes the existing rolebinding and the proposed
+// and verifies that the subjects that are in the proposed exist in the
+// existing rolebinding
+func RoleBindingAsExpected(incoming, existingRoleBinding rbacv1.RoleBinding) bool {
+	if incoming.RoleRef.Name == existingRoleBinding.RoleRef.Name &&
```
jim-minter: do `incoming.RoleRef == existingRoleBinding.RoleRef` so you test all fields (you were missing APIGroup).
Also, do a `reflect.DeepEqual(incoming.Subjects, existingRoleBinding.Subjects)` (similar problems).

pkg/common/common.go
kwoodson: Added `owned-by-osa-` prefix.

pkg/config/images.go
kwoodson: updated.

pkg/config/images.go
kwoodson: updated.

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
kwoodson: Added the logic from origin + default.

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
kwoodson: This becomes much more efficient and clean this way.  Updated.

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
kwoodson: updated.

pkg/controllers/customeradmin/rolebindings/rolebindings.go
kwoodson: updated to `owned-by-osa-`

pkg/controllers/customeradmin/rolebindings/rolebindings.go
kwoodson: updated with prefix `owned-by-sa-`

pkg/controllers/customeradmin/rolebindings/rolebindings.go
kwoodson: I did not know about `reflect.DeepEqual`.  I wondered if comparing objects would be shallow or pointer address.  Thanks!

pkg/controllers/customeradmin/rolebindings/rolebindings.go
kwoodson: removed.

kwoodson: removed.

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
kwoodson: fixed.

pkg/controllers/customeradmin/controllerrolebinding/rolebinding/rolebinding_controller.go
kwoodson: fixed.[

pkg/addons/data/Deployment.apps/openshift-infra/customer-admin-reconciler.yaml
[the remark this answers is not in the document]
kwoodson: removed.

pkg/controllers/customeradmin/controllernamespace/namespace/namespace_controller.go
kwoodson: Updated to use `switch`.

kwoodson:
Unfamiliar with this one:
```
--- FAIL: TestUnmarshall (0.10s)
	types_test.go:263: json.Unmarshal returned unexpected result
		api.OpenShiftManagedCluster{Plan:(*api.ResourcePurchasePlan)(0xc4200d90c0), Properties:(*api.Properties)(0xc420214bd0), ID:"ID", Name:"Name", Type:"Type", Location:"Location", Tags:map[string]string{"Tags.key":"Tags.val"}, Config:(*api.Config)(0xc420392500)}
FAIL
```

kwoodson:
Looks like types_test.go is having an unmarshal exception.
```
	types_test.go:263: json.Unmarshal returned unexpected result
		api.OpenShiftManagedCluster{Plan:(*api.ResourcePurchasePlan)(0xc42001f700), Properties:(*api.Properties)(0xc4201aac60), ID:"ID", Name:"Name", Type:"Type", Location:"Location", Tags:map[string]string{"Tags.key":"Tags.val"}, Config:(*api.Config)(0xc420324500)}
```

kwoodson:
@jim-minter this is occuring because the sync image is running from master which has the reconciler.  This pull request renamed the reconciler's serviceaccount to customer-admin-controller and therefore this will fail.  

If the PR runs the sync image from the current PR then I believe it would work fine.
```
time="2018-11-14T01:53:21Z" level=info msg="checking deployment openshift-infra/customer-admin-reconciler" resourceGroup=ci-op-0dbs4717-8474e
time="2018-11-14T02:09:05Z" level=fatal msg="containerservice.OpenShiftManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: context deadline exceeded" resourceGroup=ci-op-0dbs4717-8474e
```

kwoodson:
Added image to the types_test.go for the marshal/unmarshal test.

kwoodson:
types_test.go requires ordered data.  I have added and updated the order of the azureControllers field.

kwoodson:
required a `go generate`

kwoodson:
I have added the corresponding PR for image building https://github.com/openshift/release/pull/2150

jim-minter:
@kwoodson please squash this

jim-minter:
(ideally it would be preferable if both controllers were in the same package but it doesn't matter too much)
will aim to get this merged after angus' etcd work has gone in.

kwoodson:
squashed, thanks @jim-minter

kwoodson:
Fix for e2e tests.  ready.go was still looking for reconciler.  Renamed to controller.  Expecting this to pass.

kwoodson:
e2e fails due to the image not existing in quay:
```
  Normal   Pulling           2m (x2 over 2m)   kubelet, master-000000  pulling image "quay.io/openshift-on-azure/azure-controllers:v3.11"
  Warning  Failed            2m (x2 over 2m)   kubelet, master-000000  Failed to pull image "quay.io/openshift-on-azure/azure-controllers:v3.11": rpc error: code = Unknown desc = Tag v3.11 not found in repository quay.io/openshift-on-azure/azure-controllers
  Warning  Failed            2m (x2 over 2m)   kubelet, master-000000  Error: ErrImagePull
  Normal   SandboxChanged    2m (x7 over 2m)   kubelet, master-000000  Pod sandbox changed, it will be killed and re-created.
  Normal   BackOff           2m (x6 over 2m)   kubelet, master-000000  Back-off pulling image "quay.io/openshift-on-azure/azure-controllers:v3.11"
  Warning  Failed            2m (x6 over 2m)   kubelet, master-000000  Error: ImagePullBackOff
```

jim-minter:
https://github.com/openshift/release/pull/2177 merged

jim-minter:
@mjudeikis am removing management-infra from 3.11 with this PR as well - not applicable for OSA.

jim-minter:
@asalkeld am fixing up some confusion around underscores in etcdbackup env vars with this PR - please take note.  Standardising on etcdbackup, not etcd-backup (it was almost 50/50).

PR #678, opened 2018-10-31
test struct tags of the api types

charlesakalugwu:
part of #386

PR #699, opened 2018-11-06
Added a recursive struct value population function

### LINKED ISSUE #386: add `json` tags to Config struct members

charlesakalugwu:
useful for creating dynamic fixtures of complex types

closes #386 

ptal @jim-minter


kargakis:
I don't like the fact that we create utility packages that are going to be used only in testing. Feels like an anti-pattern to me. Similar worries with https://github.com/openshift/openshift-azure/pull/678. Can we do something about it?

pkg/util/fixtures/populate/populate_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"crypto/rsa"
+	"crypto/x509"
+	"fmt"
+	"strconv"
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestPopulate(t *testing.T) {
```
jim-minter: all of this can go away - replace it with tests which walk `api.OpenShiftManagedCluster` (for both the internal and external api), yaml marshal the result and compare it against a known good value.

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+	"math/big"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+var sample2048Key *rsa.PrivateKey
+
+func init() {
+	sample2048Key = &rsa.PrivateKey{
```
jim-minter: shorter:
```golang
var err error
dummyKey, err = x509.ParsePKCS1PrivateKey([]byte{
	0x30, 0x25, 0x02, 0x01, 0x00, 0x02, 0x03, 0x00, 0xa2, 0x5f, 0x02, 0x03,
	0x01, 0x00, 0x01, 0x02, 0x02, 0x43, 0xf1, 0x02, 0x02, 0x00, 0xc5, 0x02,
	0x02, 0x00, 0xd3, 0x02, 0x02, 0x00, 0x91, 0x02, 0x02, 0x00, 0xad, 0x02,
	0x02, 0x00, 0xb7,
})
if err != nil {
	panic(err)
}
```

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+func Walk(v interface{}) {
+	val := reflect.ValueOf(v)
+	if val.Kind() != reflect.Ptr {
+		panic("argument is not a pointer to a value")
+	}
+	walk(val, "")
+}
+
+// walk fills in the complete structure of a complex value v using path as the root of the labelling.
+func walk(v reflect.Value, path string) {
```
jim-minter: right at the top of this function, do:

```golang
// special cases
switch v.Interface().(type) {
case []byte:
	v.SetBytes([]byte(path))
	return

case *rsa.PrivateKey:
	// use a dummy value because the zero value cannot be marshalled
	v.Set(reflect.ValueOf(dummyKey))
	return

case []api.IdentityProvider:
	// override all the supported provider kinds here
	v.Set(reflect.ValueOf([]api.IdentityProvider{{Provider: &api.AADIdentityProvider{}}}))
}
```

the `case []api.IdentityProvider` is needed to fully walk an api.OpenShiftManagedCluster.

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+		// do not go on with the recursion if it isn't one of the core openshift-azure types
+		if !strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/") ||
+			strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/vendor/") {
+			return
+		}
+		for i := 0; i < v.NumField(); i++ {
+			field := v.Field(i)
+			if field.IsValid() {
+				newpath := extendPath(path, v.Type().Field(i).Name, v.Kind())
+				switch field.Interface().(type) {
```
jim-minter: moved above, remove here

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+		walk(v.Elem(), path)
+	case reflect.Struct:
+		// do not go on with the recursion if it isn't one of the core openshift-azure types
+		if !strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/") ||
+			strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/vendor/") {
+			return
+		}
+		for i := 0; i < v.NumField(); i++ {
+			field := v.Field(i)
+			if field.IsValid() {
```
jim-minter: don't believe this check is needed

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+				case *rsa.PrivateKey:
+					field.Set(reflect.ValueOf(sample2048Key))
+				}
+				walk(field, newpath)
+			}
+		}
+	case reflect.Array, reflect.Slice:
+		// if the array/slice has length 0 create a new slice of length 1
+		if v.Len() == 0 {
+			v.Set(reflect.MakeSlice(v.Type(), 3, 3))
```
jim-minter: the following avoids a very edge case on a zero length array:

```golang
case reflect.Slice:
	if v.Len() == 0 {
		v.Set(reflect.MakeSlice(v.Type(), 1, 1))
	}
	fallthrough

case reflect.Array:
	for i := 0; i < v.Len(); i++ {
...
```

prefer slice length of 1 instead of 3

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		for i := 0; i < v.Len(); i++ {
+			field := v.Index(i)
+			newpath := extendPath(path, strconv.Itoa(i), v.Kind())
+			walk(field, newpath)
+		}
+	case reflect.Map:
+		if v.IsNil() {
+			// only map keys of string type are supported
+			switch v.Type().Key().Kind() {
```
jim-minter: just
```golang
if v.Type().Key().Kind() != reflect.String ||
	v.Type().Elem().Kind() != reflect.String {
	return
}
```

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+			default:
+				return
+			}
+			// only map values of string type are supported
+			switch v.Type().Elem().Kind() {
+			case reflect.String:
+			default:
+				return
+			}
+			mapEntries := 3
```
jim-minter: all wordy - condense to
```golang
if v.IsNil() {
	v.Set(reflect.MakeMap(v.Type()))
}
v.SetMapIndex(reflect.ValueOf(path+".key"), reflect.ValueOf(path+".val"))
```

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+				index := strconv.Itoa(i + 1)
+				key := fmt.Sprintf("%s.k%s", path, index)
+				val := fmt.Sprintf("v%s", index)
+				rkey := reflect.ValueOf(&key).Elem()
+				rval := reflect.ValueOf(&val).Elem()
+				v.SetMapIndex(rkey, rval)
+			}
+		}
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		v.SetInt(3)
```
jim-minter: suggest 1s instead of 3s ;)

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		v.SetUint(3)
+	case reflect.Float32, reflect.Float64:
+		v.SetFloat(3.0)
+	case reflect.Bool:
+		v.SetBool(true)
+	case reflect.String:
+		v.SetString(path)
+	}
```
jim-minter: add
```golang
case reflect.Interface:
	walk(v.Elem(), path)

default:
	panic("unimplemented: " + v.Kind().String())
```

pkg/util/fixtures/populate/populate.go
```diff
@@ -0,0 +1,145 @@
...
+	case reflect.Struct:
+		return fmt.Sprintf("%s.%s", path, extension)
+	case reflect.Slice, reflect.Array:
+		return fmt.Sprintf("%s[%s]", path, extension)
+	}
+	return ""
+}
+
+// fromBase10 converts a number string in base 10 to a big.Int pointer
+func fromBase10(base10 string) *big.Int {
```
jim-minter: delete

jim-minter:
perhaps test/util/populate might be a better package location? @kargakis wdyt?

test/util/populate/certs.go
```diff
@@ -0,0 +1,25 @@
+package populate
+
+import (
+	"crypto/rsa"
+	"crypto/x509"
+)
+
+var dummyPrivateKey *rsa.PrivateKey
```
charlesakalugwu: @jim-minter do you have a similar trick for a `*x509.Certificate`. Right now, we cannot unmarshall `Certificates` which were marshalled from the zero-value.

pkg/api/2018-09-30-preview/api/types_test.go
```diff
@@ -1,12 +1,16 @@
-package api
+package api_test
```
charlesakalugwu: @jim-minter ran into that issue we discussed at the beginning of my time on the team....the cycles which can arise in testing and how `_test` packages help. Please feel free to suggest something different.

test/util/populate/certs.go
jim-minter: I don't immediately have a similar trick - where are you unmarshalling and why?

pkg/api/2018-09-30-preview/api/types_test.go
jim-minter: call it `package api` and remove the import `. "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"` below, then it should be fine.

charlesakalugwu: if I do that, I get the following error
*external*
```
# github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api
import cycle not allowed in test
package github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api (test)
	imports github.com/openshift/openshift-azure/test/util/populate
	imports github.com/openshift/openshift-azure/pkg/api
	imports github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api

FAIL	github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api [setup failed]
```

*internal*
```
/usr/local/go/bin/go test -c -i -o /tmp/___types_test_go__1_ github.com/openshift/openshift-azure/pkg/api #gosetup
# github.com/openshift/openshift-azure/pkg/api
import cycle not allowed in test
package github.com/openshift/openshift-azure/pkg/api (test)
	imports github.com/openshift/openshift-azure/test/util/populate
	imports github.com/openshift/openshift-azure/pkg/api

FAIL	github.com/openshift/openshift-azure/pkg/api [setup failed]
```

test/util/populate/certs.go
charlesakalugwu: We need to unmarshall a curated yaml into corresponding api structs (internal/external) and then compare those unmarshalled structs with structs which are populated using `populate.Walk`. The converse operation, marshalling a struct populated with `populate.Walk` and comparing the yaml  to a golden yaml, works.

charlesakalugwu: This is exactly where the *x509.Certificate* unmarshalling is needed. That test is currently skipped: https://github.com/openshift/openshift-azure/pull/699/files#diff-de120f39e8d8448b9aa430e66c997a97R383

pkg/api/types_test.go
```diff
@@ -198,3 +364,33 @@ func TestJSONTags(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if !bytes.Equal(b, marshalledYaml) {
+		t.Errorf("yaml.Marshall returned unexpected result\n%s\n%s\n", string(b), marshalledYaml)
+	}
+}
+
+func TestUnmarshallYaml(t *testing.T) {
+	t.Skip("skipping this test until we are able to unmarshall x509.Certificate")
```
charlesakalugwu: @jim-minter This is the test which requires unmarshalling the internal api struct from a curated yaml. `api.Config` contains `x509.Certificate`s and that is what breaks the unmarshall. This test is currently skipped:

PR #700, opened 2018-11-06
Move OSA request manifest out of hack/create.sh

kargakis:
I am moving the manifest out of hack/create.sh into `test/manifests` where we can place more manifests to test different flows (eg. custom vnet, different types of upgrades, etc.). Then, in order to actually write a test that would make use of a different manifest rather than the default we would place the manifest inside `test/testdata/{somename}/create.yaml` and set `MANIFEST=test/testdata/{somename}/create.yaml` in the prow job to make use of that request during cluster deployment. More work needs to be done to enable upgrades, will do in a follow-up.

/cc @kwoodson @jim-minter @cjryan @mjudeikis

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -59,44 +59,13 @@ else
     AZURE_AAD_CLIENT_ID=$AZURE_CLIENT_ID
     AZURE_AAD_CLIENT_SECRET=$AZURE_CLIENT_SECRET
 fi
+export AZURE_AAD_CLIENT_SECRET
 set -x
 
-cat >_data/manifest.yaml <<EOF
-name: $RESOURCEGROUP
-location: $AZURE_REGION
-properties:
-  openShiftVersion: "$DEPLOY_VERSION"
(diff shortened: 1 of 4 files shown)


hack/create.sh
```diff
@@ -59,44 +59,19 @@ else
...
-EOF
+if [[ -z "$MANIFEST" ]]; then
+    MANIFEST="test/manifests/normal/create.yaml"
+fi
+# Execute manifest creation in a separate script to
+# evaluate the manifest bash variables.
+echo 'cat <<EOF' >  temp.sh
+cat "$MANIFEST"  >> temp.sh
+echo 'EOF'       >> temp.sh
+bash temp.sh > _data/manifest.yaml
```
kwoodson: Would `envsubst` do the equivalent here?
```
cat $MANIFEST | envsubst > _data/manifest.yaml
```

asalkeld: that does look a lot neater.

mjudeikis:
Like it :)
lgtm

kargakis:
https://github.com/openshift/openshift-azure/issues/643

/retest

PR #701, opened 2018-11-06
add additional VM flavours and increase maximum number of compute nodes

jim-minter:
Allow master and infra nodes to be Standard_D{2-64}s_v3.
Allow a range of additional values for compute nodes.
Increase maximum number of compute nodes to 20.
See code for rationales.

### DIFF
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -27,6 +27,88 @@ var (
 
 	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
 	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
+
+	validMasterAndInfraVMSizes = map[VMSize]struct{}{
+		// Rationale here is: a highly limited set of modern general purpose
+		// offerings which we can reason about and test for now.
+
+		// GENERAL PURPOSE VMS
+
+		StandardD2sV3:  {}, // TODO: probably should only be enabled for test
+		StandardD4sV3:  {},
+		StandardD8sV3:  {},
+		StandardD16sV3: {},
+		StandardD32sV3: {},
+		StandardD64sV3: {},
+
+		// TODO: consider enabling storage optimized (Ls) series for masters and
+		// moving the etcd onto the VM SSD storage.
+
+		// TODO: enable vertical scaling of existing OSA clusters.
+	}
+
+	validComputeVMSizes = map[VMSize]struct{}{
+		// Rationale here is: modern offerings per (general purpose /
+		// memory-optimized / compute-optimized / storage-optimized) family,
+		// with at least 16GiB RAM, 32GiB SSD, 8 data disks, premium storage
+		// support.  GPU and HPC use cases are probably blocked for now on
+		// drivers / multiple agent pool support.
+
+		// GENERAL PURPOSE VMS
+
+		// Skiping StandardB* (burstable) VMs for now as they could be hard to
+		// reason about performance-wise.
+
+		StandardD2sV3:  {}, // TODO: probably should only be enabled for test
+		StandardD4sV3:  {},
+		StandardD8sV3:  {},
+		StandardD16sV3: {},
+		StandardD32sV3: {},
+		StandardD64sV3: {},
+
+		StandardDS4V2: {},
+		StandardDS5V2: {},
+
+		// COMPUTE OPTIMIZED VMS
+
+		StandardF8sV2:  {},
+		StandardF16sV2: {},
+		StandardF32sV2: {},
+		StandardF64sV2: {},
+		StandardF72sV2: {},
+
+		StandardF8s:  {},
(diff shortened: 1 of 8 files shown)


jim-minter:
@amanohar @akrzos @kwoodson @thekad @pweil- @kargakis ptal

jim-minter:
I'm somewhat sympathetic to the blacklist idea but it has several complexities.  With this PR I'm proposing to get at least some additional variety out to customers as a possibly interim step.

Complexities include:
* I definitely prefer the whitelist approach for masters and infras.
* I think the blacklist would necessarily be longer than the whitelist right now, and (even) harder to define (I suspect there may be quite a few old/non-documented Azure VM types, for example; also I think it is appropriate to prevent small VMs/non-premium storage etc. off the bat and it's hard to discover all of them).  We also still have to blacklist GPU/burstable nodes, etc.
* We have a little logic around setting k8s parameters according to machine parameters (e.g. RAM) right now, but it could grow, and that is harder to deal with in the blacklist case.
* Blacklist is more challenging long term wrt supportability as well so it'd be likely we'd need to revert a blacklist change before GA, which also makes me sceptical.

pweil-:
> I'm somewhat sympathetic to the blacklist idea but it has several complexities.....

Thanks for listing this out on the PR, I think that will help the discussion.  I agree with the points mentioned and that keeping a firm grasp on supportability is a good thing.  There is a good chance we'd have to maintain the blacklist via bug reporting by customers which is a bad experience.

pkg/config/derive.go
```diff
@@ -12,24 +13,55 @@ type derived struct{}
...
+
+	return "", fmt.Errorf("role %s not found", role)
 }
 
-func (derived) KubeReserved(cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) string {
-	for _, pool := range cs.Properties.AgentPoolProfiles {
-		if pool.Role != role {
+func (derived) KubeReserved(cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) (string, error) {
+	if role == api.AgentPoolProfileRoleMaster {
+		return "", fmt.Errorf("kubereserved not defined for role %s", role)
```
thekad: nit: this error message sounds like you aren't setting kubereserved for the master role, when it should convey the opposite, change to something like "cannot set kubereserved for role %s" or similar

pkg/api/validate.go
```diff
@@ -27,6 +27,88 @@ var (
 
 	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
 	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
+
+	validMasterAndInfraVMSizes = map[VMSize]struct{}{
```
amanohar: The value in the map is a `struct` because this might be more than just a `bool` in the future? Assuming struct is used to store where the size should only be used for testing?

amanohar:
@pweil- I am not on the email thread mentioned. Can you add me or add your ideas to the email I sent yesterday?

Regarding:
> if we should consider making this an external configurable?

If whitelisting the way to go we should take steps to make it easier to enable VM sizes for OSA. One way can be to allow the configuration in the RP so that it doesn't require a vendor for the plugin. Further, we do plan to do separate release of a configuration only (this way rollout can be faster) but this is not in place today.

@jim-minter regarding:
>I definitely prefer the whitelist approach for masters and infras.
I agree with this approach.

I didn't fully understand the need to revert before GA if blacklist were supported.
>Blacklist is more challenging long term wrt supportability as well so it'd be likely we'd need to revert a blacklist change before GA, which also makes me sceptical.

pweil-:
> @pweil- I am not on the email thread mentioned. Can you add me or add your ideas to the email I sent yesterday?

That comment was referencing your reply about whitelisting yesterday.  I wanted to make sure it was addressed.  Specifically I wanted to make sure we discussed this comment before this PR was approved.


>This approach is restrictive because any additions/deletions to VM sizes will always require a service rollout

>Alternative approach: RP ships with a list of not allowed VM sizes. This allows flexibility of faster support of new VM sizes. A con of this approach is that not all VM sizes can be fully vetted before being supported/not supported.

pkg/api/validate.go
jim-minter: map[foo]struct{} is a compact go idiom to indicate a set.  I don't have any immediate intention to add members to the struct.

pkg/config/derive.go
jim-minter: if this error appears, it is indicative of a programming error, not an end-user one.

jim-minter:
anubhuti indicated lgtm to me via slack and kargakis approved above - marking lgtm here manually.

pkg/api/validate.go
amanohar: Thanks, understood from standpoint of go idiom. My question was if there is a reason to use a map instead of a list/array of allowed sizes (and I was guessing the purpose using the bool example).

PR #706, opened 2018-11-07
add additional VM flavours and increase maximum number of compute nodes

jim-minter:
backport of #701


jim-minter:
merging as #701 was acceptable

PR #712, opened 2018-11-09
Return a PluginError instead of a plain error

### LINKED ISSUE #694: CreateOrUpdate should return type 'PluginError' instead of 'error'
https://github.com/openshift/openshift-azure/blob/master/pkg/plugin/plugin.go#L103 should return type 'PluginError' instead of 'error'. This is needed so that RP/service: Can expect error to be in an opinionated format and custom passing will be done on `Err` field of 'PluginError' by the RP.

cc: @asalkeld @jim-minter

thekad:
Closes #694 

Simple change to return an opinionated structure instead of a plain error

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -28,6 +28,7 @@ const (
 	PluginStepWaitForNodes               PluginStep = "WaitForNodes"
 	PluginStepWaitForConsoleHealth       PluginStep = "WaitForConsoleHealth"
 	PluginStepWaitForInfraDaemonSets     PluginStep = "WaitForInfraDaemonSets"
+	PluginStepWaitForInfraStatefulSets   PluginStep = "WaitForInfraStatefulSets"
 	PluginStepWaitForInfraDeployments    PluginStep = "WaitForInfraDeployments"
 )
 
@@ -92,5 +93,5 @@ type Plugin interface {
 
 	// CreateOrUpdate either deploys or runs the update depending on the isUpdate argument
 	// this will call the deployer.
-	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, isUpdate bool, deployer DeployFn) error
+	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, isUpdate bool, deployer DeployFn) *PluginError
 }
--- a/pkg/cluster/initialize.go
+++ b/pkg/cluster/initialize.go
@@ -9,7 +9,7 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
 )
 
-func (u *simpleUpgrader) Initialize(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+func (u *simpleUpgrader) initialize(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
 	if u.storageClient == nil {
 		keys, err := u.accountsClient.ListKeys(ctx, cs.Properties.AzProfile.ResourceGroup, cs.Config.ConfigStorageAccount)
 		if err != nil {
--- a/pkg/cluster/types.go
+++ b/pkg/cluster/types.go
@@ -21,11 +21,10 @@ import (
 
 // Upgrader is the public interface to the upgrade module used by the plugin.
 type Upgrader interface {
-	Initialize(ctx context.Context, cs *api.OpenShiftManagedCluster) error
-	Deploy(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error
-	Update(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error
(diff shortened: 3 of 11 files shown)


pkg/fakerp/fakerp.go
```diff
@@ -145,9 +145,9 @@ func CreateOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 		return nil
 	}
 
-	err = p.CreateOrUpdate(ctx, cs, azuretemplate, oldCs != nil, deployer)
-	if err != nil {
-		return nil, err
+	perr := p.CreateOrUpdate(ctx, cs, azuretemplate, oldCs != nil, deployer)
```
thekad: I feel like the fakerp should also return a `PluginError` here, thoughts?

pkg/plugin/plugin.go
```diff
@@ -110,15 +110,16 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
...
 	if err != nil {
-		return err
+		return api.PluginError{Err: err, Step: api.PluginStepDeploy}
 	}
 
 	// Wait for infrastructure services to be healthy
 	err = p.clusterUpgrader.WaitForInfraServices(ctx, cs)
 	if err != nil {
-		return err
+		return api.PluginError{Err: err, Step: api.PluginStepWaitForInfraDeployments}
```
thekad: not sure about this step here, there are several `WaitForInfra` steps...

pkg/plugin/plugin.go
```diff
@@ -110,15 +110,16 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
 		err = p.clusterUpgrader.Deploy(ctx, cs, azuretemplate, deployFn)
 	}
 	if err != nil {
-		return err
+		return api.PluginError{Err: err, Step: api.PluginStepDeploy}
```
thekad: should I return a diff step on deploy vs update?

asalkeld:
@thekad I think you only need to change the return type in plugin and in the upgrader.
https://github.com/openshift/openshift-azure/blob/master/pkg/cluster/types.go#L23-L28
you shouldn't have to add new PluginErrors as the upgrader already returns the required errors.

Although, IMHO returning `error` is the correct approach.

pkg/api/plugin.go
```diff
@@ -92,5 +92,5 @@ type Plugin interface {
 
 	// CreateOrUpdate either deploys or runs the update depending on the isUpdate argument
 	// this will call the deployer.
-	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, isUpdate bool, deployer DeployFn) error
+	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, isUpdate bool, deployer DeployFn) PluginError
```
jim-minter: this should be `*PluginError`, so that it can also be nil.

pkg/fakerp/fakerp.go
jim-minter: once CreateOrUpdate returns a `*PluginError`, this entire hunk can be reverted

pkg/plugin/plugin.go
jim-minter: this isn't right.  Update and Deploy already effectively return *PluginError, it's just that their signatures are wrong.  Change their signatures.  Ditto WaitForInfraServices, etc.

pkg/cluster/types.go
```diff
@@ -22,10 +22,10 @@ import (
 // Upgrader is the public interface to the upgrade module used by the plugin.
 type Upgrader interface {
 	Initialize(ctx context.Context, cs *api.OpenShiftManagedCluster) error
-	Deploy(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) error
```
jim-minter: the fact that Initialize() above is not changed implies to me that it should not be part of this interface.  Remove it from the interface, and convert its implementation to `initialize()` (unexported).

jim-minter:
one nit then lgtm

pkg/cluster/initialize.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
 )
 
-func (u *simpleUpgrader) Initialize(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+func initialize(u *simpleUpgrader, ctx context.Context, cs *api.OpenShiftManagedCluster) error {
```
jim-minter: no: should be `func (u *simpleUpgrader) initialize`.

kargakis:
```
pkg/cluster/ready.go:159:4: cannot use err (type error) as type *"github.com/openshift/openshift-azure/pkg/api".PluginError in return argument: need type assertion
```
/lgtm cancel

thekad:
This is odd:

```
[kad@kenpachi openshift-azure (issues/694)]$ make clean build
rm -f azure-reader.log coverage.out end-user.log e2e.test logbridge sync
go generate ./...
bindata.go
bindata.go
go build ./...
[kad@kenpachi openshift-azure (issues/694)]$ 
```

The build step obviously works on my machine, furthermore, there is no `err` variable in the location stated by the logs... What am I missing here?

PR #717, opened 2018-11-12
Use same network rules between create and update

kargakis:
@jim-minter do you remember why we did this in the first place? This change should bring our ARM template on par between create and update, thus ensure the first update is re-entrant.


kargakis:
Ah, got reminded by our unit tests :/
```
plugin_test.go:90: networkSecurityGroups should not be applied during cluster upgrade
```
So we can either exclude NSGs from hashing or wholly swallow the first update.

kargakis:
> So we can either exclude NSGs from hashing

Sorry, to be precise it's not the NSG but the dependsOn to the NSG that is causing the hash difference.

kargakis:
Based on scrum discussion, we are going to filter the dependsOn out.

PR #721, opened 2018-11-12
Gather artifacts in createorupdate

kargakis:
/cc mjudeikis charlesakalugwu

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -25,7 +25,11 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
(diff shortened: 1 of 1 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -441,6 +449,146 @@ func createResourceGroup(log *logrus.Entry) error {
...
+		return err
+	}
+
+	// gather control plane logs
+	ns := "kube-system"
+	if err := gatherLogs(kc, artifactDir, ns, "sync-master-000000"); err != nil {
+		return err
+	}
+	// TODO: Get logs from the api server and etcd dynamically by using the master count
+	if err := gatherLogs(kc, artifactDir, ns, "master-etcd-master-000000"); err != nil {
```
mjudeikis: I think at some point we need to move pod names to constants, as they starting to appear all around the code

cmd/createorupdate/createorupdate.go
```diff
@@ -441,6 +449,146 @@ func createResourceGroup(log *logrus.Entry) error {
 	return err
 }
 
+func gatherArtifacts(artifactDir, artifactKubeconfig string) error {
```
mjudeikis: Does it make sense to move these to a separate file as createorupdate.go is becoming kinda big?

cmd/createorupdate/createorupdate.go
kargakis: There is a TODO just above about constructing these names dynamically.

cmd/createorupdate/createorupdate.go
kargakis: We will have to add the new file in the `go run` invocation, and I have been resisting it so far but I definitely agree this file should be split into separate self-contained files. Certainly the server code deserves its own `server.go` place.

kargakis: Will do the split in a follow-up

PR #726, opened 2018-11-14
[wip] add e2e tests to confirm key rotation functionality

charlesakalugwu:
part of AZURE-117

@kargakis ptal


cmd/createorupdate/createorupdate.go
```diff
@@ -29,11 +29,12 @@ import (
...
 var (
-	method   = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
-	useProd  = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
-	manifest = flag.String("manifest", "_data/manifest.yaml", "Manifest to use for the initial request.")
-	update   = flag.String("update", "", "If provided, use this manifest to make a follow-up request after the initial request succeeds.")
-	cleanup  = flag.Bool("rm", false, "Delete the cluster once all other requests have completed successfully.")
+	method     = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
+	useProd    = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
+	manifest   = flag.String("manifest", "_data/manifest.yaml", "Manifest to use for the initial request.")
+	configBlob = flag.String("configBlob", "_data/containerservice.yaml", "Path on disk where the internal cluster representation is written")
```
kargakis: Is it possible to avoid exposing this here? Think of createorupdate as a tool that an end user runs. As an end user, I don't have access to the config blob.

test/e2erp/fakerp_test.go
```diff
@@ -3,15 +3,55 @@
...
+		By("Running generate on the modified config blob")
+		err = generateInternalConfig(mutated, pluginConfig)
 		Expect(err).NotTo(HaveOccurred())
+
+		By("Persisting the config blob containing the new certificates and credentials")
+		err = saveConfig(mutated, *configBlob)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Calling update on the fake rp with the updated config blob")
+		updated, err := updateCluster(ctx, external, *configBlob, logger, pluginConfig)
```
kargakis: The problem with running the upgrade here is that you have created your own way of upgrading clusters `updateCluster` that is different from what we run in development. That's why I have been working in enabling us to use createorupdate for running upgrade tests in an e2e way and it's now possible with https://github.com/openshift/openshift-azure/pull/730 + https://github.com/openshift/release/pull/2161

Not saying that you should change what you have now but long-term it's going to be more robust to have to maintain a single way to do upgrades (`createorupdate.go`) vs littering that code inside the e2e suites.

kargakis: Instead of having to write bash scripts that would need to run before and after the upgrade, we could still write ginkgo tests that get run via `ginkgo`. Or maybe we could reuse the code between createorupdate.go and here?

test/e2erp/rotation.go
```diff
@@ -0,0 +1,141 @@
+package e2erp
```
kargakis: Rename this file to helpers.go or something more general since some of these functions are not specific to key rotation.

test/e2erp/rotation.go
```diff
@@ -0,0 +1,141 @@
...
+		LogBridgeImage:  os.Getenv("LOGBRIDGE_IMAGE"),
+		AcceptLanguages: []string{"en-us"},
+		TestConfig:      tc,
+	}
+	return pluginConfig
+}
+
+// updateCluster updates an OpenshiftManagedCluster by sending both the current external manifest and internal manifest
+// which is to be used for the update
+func updateCluster(ctx context.Context, external *v20180930preview.OpenShiftManagedCluster, configBlobPath string, logger *logrus.Entry, config *api.PluginConfig) (*v20180930preview.OpenShiftManagedCluster, error) {
```
kargakis: Share this between the e2e test and createorupdate.

test/e2erp/rotation.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+//
+func generateInternalConfig(internal *api.OpenShiftManagedCluster, pluginConfig *api.PluginConfig) error {
+	configGen := config.NewSimpleGenerator(pluginConfig)
+	return configGen.Generate(internal)
+}
+
+// newPluginConfig creates a new PluginConfig from the current environment
+func newPluginConfig() *api.PluginConfig {
```
kargakis: share this between the test and fakerp

test/e2erp/fakerp_test.go
```diff
@@ -3,15 +3,55 @@
...
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Resource provider e2e tests [Fake]", func() {
 	defer GinkgoRecover()
 
-	It("dummy fake test", func() {
-		var err error
+	It("should be possible to maintain a healthy cluster after rotating all credentials", func() {
```
kargakis: Add a TODO to update this test to use the Admin API once we have it.

test/e2erp/e2erp_suite_test.go
```diff
@@ -3,20 +3,30 @@
...
-	manifest  = flag.String("manifest", "../../_data/manifest.yaml", "Path to the manifest to send to the RP")
+	c            *testClient
+	azureConf    AzureConfig
+	gitCommit    = "unknown"
+	logger       *logrus.Entry
+	ctx          context.Context
+	pluginConfig *api.PluginConfig
+	manifest     = flag.String("manifest", "../../_data/manifest.yaml", "Path to the manifest to send to the RP")
+	configBlob   = flag.String("configBlob", "../../_data/containerservice.yaml", "Path on disk where the OpenShift internal config blob should be written")
+	logLevel     = flag.String("logLevel", "Debug", "The log level to use")
```
kargakis: Nobody is going to tweak this in the e2e tests, remove it.

PR #730, opened 2018-11-14
All-in-one script to automate e2e test generation

kargakis:
This is an all-in-one script that should allow us to write e2e tests w/o the need to use multiple templates in CI just by configuring the following env variables on top of the existing variables that we pass in `hack/create.sh`:
```
EXEC - (optional) what command to run after a cluster has been created; this is ideally a make target
UPDATE_MANIFEST - (optional) the request to make to the RP to initiate a cluster upgrade
UPDATE_EXEC - (optional) what command to run after a cluster has been upgraded; this is ideally a make target
ARTIFACT_DIR - (optional) directory to place artifacts that are gathered just before a cluster is cleaned up
```
Next step should be to start rewritting [our templates](https://github.com/openshift/release/tree/master/ci-operator/templates/openshift/openshift-azure) to use this script.

@mjudeikis @charlesakalugwu @jim-minter @kwoodson thoughts?


kargakis:
What I don't like about this script is that we are duplicating `hack/create.sh`. Unfortunately until we rewrite hack/dns.sh and hack/aad.sh we are going to need to duplicate logic between the two scripts.

cmd/createorupdate/createorupdate.go
```diff
@@ -167,6 +167,58 @@ func createResourceGroup(conf *fakerp.Config) (bool, error) {
 	return true, err
 }
 
+func execute(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClustersClient, conf *fakerp.Config) error {
+	// simulate the API call to the RP
+	if err := wait.PollImmediate(time.Second, 10*time.Second, func() (bool, error) {
```
charlesakalugwu: we should increase that timeout

cmd/createorupdate/createorupdate.go
```diff
@@ -220,62 +272,20 @@ func main() {
...
+			log.Warn(err)
 		}
 	}
 
-	if *artifactDir != "" {
-		if err := fakerp.GatherArtifacts(*artifactDir, *artifactKubeconfig); err != nil {
-			log.Warnf("could not gather artifacts: %v", err)
-		}
+	if err != nil {
+		os.Exit(1)
```
mjudeikis: log.Fatal(err)?

kargakis: We already printed the error above.

cmd/createorupdate/createorupdate.go
kargakis: fixed

PR #683, opened 2018-11-01
Re-entrant node upgrades

kargakis:
Builds on top of https://github.com/openshift/openshift-azure/pull/670

Part of AZURE-199.

PR #734, opened 2018-11-15
Use the TestConfig.RunningUnderTest instead of the env

asalkeld:


pkg/cluster/update.go
```diff
@@ -48,7 +47,7 @@ func (u *simpleUpgrader) Update(ctx context.Context, cs *api.OpenShiftManagedClu
 	// we enable configuration changes and/or upgrades, uncomment this code.  At
 	// the same time, current thinking is that we will add a hash-based
 	// mechanism to avoid unnecessary VM rotations as well.
-	if os.Getenv("RUNNING_UNDER_TEST") != "" {
+	if u.pluginConfig.TestConfig.RunningUnderTest {
```
kargakis: This is removed altogether in https://github.com/openshift/openshift-azure/pull/683

/close

PR #738, opened 2018-11-15
Remove e2e image

kargakis:
It never worked, and it seems we are in agreement that MSFT should run tests directly from the repo to run the exact same commit that the RP has vendored. 

/cc mjudeikis charlesakalugwu


kargakis:
release PR that is required to merge first is https://github.com/openshift/release/pull/2145

PR #749, opened 2018-11-19
Remove quotes from bash envs to fix Go flag parsing

kargakis:
This broke the update test since EXEC is undefined and the go flag parsing would not interpret the last half of the flags.

/cc mjudeikis charlesakalugwu


mjudeikis:
/lgtm 
strange one :)

PR #754, opened 2018-11-19
replace hack/aad.sh with a go binary

asalkeld:
This replaces the shell script with a go binary.


mjudeikis:
@asalkeld we need to update `create.sh` because it is calling `aad.sh` to do an `app-update`

asalkeld:
@mjudeikis  I don't see where: https://github.com/openshift/openshift-azure/blob/master/hack/create.sh

mjudeikis:
> @mjudeikis I don't see where: https://github.com/openshift/openshift-azure/blob/master/hack/create.sh

Sorry, my bad. need rebase local copy :D

mjudeikis:
lgtm,
will wait for the second review

mjudeikis:
on the other hand. Not sure if we should do this. shell scripts are much easier to maintain. And our RP already does what's needed with the previous PR.

@jim-minter @kargakis  WDYT?

kargakis:
I am happy with the state of the fake rp, don't really care about removing
the scripts.

PR #762, opened 2018-11-20
Refactor e2e tests into various packages

charlesakalugwu:
This refactors the current layout of our e2e tests into more granular suites, scenarios and clients. This is a foundation for update tests which may need to reference test scenarios from other suites

@kargakis @kwoodson ptal


test/util/client/azure/compute.go
```diff
@@ -0,0 +1,356 @@
...
+func (az *Client) ScaleSet(resourceGroup, scaleSetName string) (*compute.VirtualMachineScaleSet, error) {
+	ss, err := az.ssc.Get(az.ctx, resourceGroup, scaleSetName)
+	if err != nil {
+		return nil, err
+	}
+	return &ss, nil
+}
+
+// ScaleOutScaleSet scales out the capacity of a given scale set to count VMs
+func (az *Client) ScaleOutScaleSet(scaleSetName string, count int) error {
```
kargakis: Collapse this with ScaleInScaleSet

test/util/client/azure/compute.go
```diff
@@ -0,0 +1,356 @@
...
+		err = vmPages.Next()
+		if err != nil {
+			return nil, err
+		}
+	}
+	return vms, nil
+}
+
+// ScaleSet returns a VirtualMachineScaleSet with a given name within a resource group
+func (az *Client) ScaleSet(resourceGroup, scaleSetName string) (*compute.VirtualMachineScaleSet, error) {
```
kargakis: Functions wrapping a single function call are probably unnecessary. Inline this Get wherever ScaleSet is used.

test/util/client/azure/azure.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	sdk "github.com/openshift/openshift-azure/pkg/util/azureclient/osa-go-sdk/services/containerservice/mgmt/2018-09-30-preview/containerservice"
+)
+
+var (
+	fakeRe = regexp.MustCompile("Fake")
+	realRe = regexp.MustCompile("Real")
+)
+
+type AzureConfig struct {
```
kargakis: TestConfig

test/util/client/azure/compute.go
```diff
@@ -0,0 +1,356 @@
...
+		return err
+	}
+	if err := future.WaitForCompletionRef(az.ctx, az.ssc.Client()); err != nil {
+		return err
+	}
+	return nil
+}
+
+// UpdateScaleSetsCapacity returns a slice of the errors it encounters as it attempts to increment the capacity
+// (by one) for all scale sets within a given resource group
```
kargakis: Not sure what's the utility of this function. If you are writting the scale up/down tests you should just be scaling up/down compute.

kargakis: Nevermind, this is part of the permissions tests.

test/util/client/azure/compute.go
```diff
@@ -0,0 +1,356 @@
...
+// UpdateCluster updates an OpenshiftManagedCluster by sending both the current external manifest and internal manifest
+// which is to be used for the update
+func (az *Client) UpdateCluster(external *v20180930preview.OpenShiftManagedCluster, configBlobPath string, config *api.PluginConfig) (*v20180930preview.OpenShiftManagedCluster, error) {
+	// Remove the provisioning state before updating
+	external.Properties.ProvisioningState = ""
+
+	var oc *v20180930preview.OpenShiftManagedCluster
+	var err error
+	// simulate the API call to the RP
+	if err := wait.PollImmediate(5*time.Second, 1*time.Hour, func() (bool, error) {
```
kargakis: We should be sharing this with createorupdate.

test/util/client/cluster/cluster.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// GenerateInternalConfig calls generate on a OpenshiftManagedCluster
+func GenerateInternalConfig(internal *api.OpenShiftManagedCluster) error {
+	configGen := config.NewSimpleGenerator(NewPluginConfig())
+	return configGen.Generate(internal)
+}
+
+// NewPluginConfig creates a new PluginConfig from the current environment
+func NewPluginConfig() *api.PluginConfig {
```
kargakis: W should be sharing this with createorupdate.

kargakis:
Shouldn't you be removing any code ie. the existing suites?

charlesakalugwu:
> Shouldn't you be removing any code ie. the existing suites?

I'll remove the current suite as the final commit in this PR

pkg/fakerp/fakerp.go
```diff
@@ -41,13 +41,23 @@ func CreateOrUpdate(ctx context.Context, oc *v20180930preview.OpenShiftManagedCl
 		return nil, err
 	}
 
+	// determine the paths where various configs should be written on disk
+	configBlob := os.Getenv("CONFIG_BLOB_PATH")
+	helpers := os.Getenv("HELPERS_PATH")
```
kargakis: Why do you need these changes?

env.example
```diff
@@ -66,3 +66,9 @@ export DEPLOY_VERSION=v3.11
 
 # Path to the request to send to the OSA RP.
 #export MANIFEST=test/manifests/normal/create.yaml
+
+# Path to the internal config blob written by the fake rp
+export CONFIG_BLOB_PATH=${PWD}/_data/containerservice.yaml
+
+# Path to a directory where openshift development files should be written
+export HELPERS_PATH=${PWD}/_data/_out/
```
kargakis: It's unclear these envs add any value or are going to be tweaked ever.

test/util/client/azure/azure.go
```diff
@@ -0,0 +1,109 @@
...
+	subID := conf.SubscriptionID
+	gc := resources.NewGroupsClient(subID)
+	gc.Authorizer = authorizer
+
+	var rpc sdk.OpenShiftManagedClustersClient
+	focus := []byte(config.GinkgoConfig.FocusString)
+	switch {
+	case fakeRe.Match(focus):
+		fmt.Println("Creating a cluster using the fake resource provider")
+		rpc = sdk.NewOpenShiftManagedClustersClientWithBaseURI("http://localhost:8080", subID)
```
kargakis: There is a const in pkg/fakerp about the location the fake RP is going to serve. Expose that if it's not exposed already and use it here.

test/util/client/azure/compute.go
```diff
@@ -0,0 +1,347 @@
...
+// managed application instance
+func (az *Client) ManagedResourceGroup(applicationResourceGroup string) (string, error) {
+	apps, err := az.appsc.ListByResourceGroup(az.ctx, applicationResourceGroup)
+	if err != nil {
+		return "", err
+	}
+
+	for _, app := range apps.Values() {
+		if appid := *app.ManagedResourceGroupID; appid != "" {
+			r, err := az.ResourceGroupFromManagedResourceID(appid)
```
kargakis:
```go
return az.ResourceGroupFromManagedResourceID(appid)
```

test/util/client/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	return res.Status.Allowed, nil
+}
+
+func (t *Client) CreatePodDisruptionBudget(p *policy.PodDisruptionBudget) error {
+	_, err := t.kc.Policy().PodDisruptionBudgets(t.namespace).Create(p)
+	return err
+}
+
+func (t *Client) GetRoute(template string, options *metav1.GetOptions) (*v1.Route, error) {
```
kargakis: Please inline back all wrapper function that wrap single-line api calls. Trust me, you don't want to maintain these going forward. We have been through the pain before.

test/util/client/azure/azure.go
kargakis: Disregard, this is not a thing anymore.

kargakis:
Please split the key rotation e2e test into its own PR, on top of the PR where you do the refactoring.

test/util/client/azure/azure.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	sdk "github.com/openshift/openshift-azure/pkg/util/azureclient/osa-go-sdk/services/containerservice/mgmt/2018-09-30-preview/containerservice"
+)
+
+var (
+	fakeRe = regexp.MustCompile("Fake")
+	realRe = regexp.MustCompile("Real")
+)
+
+type Config struct {
```
kargakis: Can you just reuse fakerp.Config?

test/util/client/cluster/cluster.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var cs *v20180930preview.OpenShiftManagedCluster
+	if err := yaml.Unmarshal(b, &cs); err != nil {
+		return nil, err
+	}
+	return cs, nil
+}
+
+// ParseInternalConfig parses an internal manifest located at path and returns an internal OpenshiftManagedCluster struct
+func ParseInternalConfig(path string) (*api.OpenShiftManagedCluster, error) {
```
kargakis: Use `pkg/util/managedcluster.ReadConfig`

env.example
jim-minter: +1 - suggest to remove

test/util/client/cluster/cluster.go
charlesakalugwu: done

pkg/fakerp/fakerp.go
charlesakalugwu: I needed these changes because
1) Allow a developer to change where the config blob path and helpers are written to disk. Without that, I found that e2e helpers and the config blob were being written to the various e2e suite root directories. For key rotation tests, I wanted each rotation to cause the previous helpers and config blob files to be overwritten.
2) These values were hardcoded in the code. Those values still remain the default. These env vars will help to tweak those locations on an ad-hoc basis.

jim-minter:
Hey @charlesakalugwu, thank-you for your effort here.  There are a lot
of good ideas and approaches in this PR which I would like to take
forwards, but I think there are some significant barriers to straight
merging this PR as it stands.  I see at least the following points
which I'd to discuss on a team call in order to take the ideas in this
PR forwards:

* The PR is reworking and "moving" tests without actually deleting the
  old ones.  I'm reluctant to have an interim step where the same test
  exists twice in the same repo.  I think the deletions need to happen
  in this PR too.

* I'm not convinced about the architecture of adding suites for things
  like "keyrotation" and "scaleupdown".  I think we're going to have
  lots of tests of the form "validate cluster / mutate state /
  revalidate cluster" and I don't think it makes sense for these to be
  separate suites with individually curated validation test lists.

  I think we should consider some model of describing different
  mutations and reusing similar lists of validation tests on top, but
  I admit I don't know exactly what that looks like.  Fundamentally I
  think we have to bound and reduce the number of test suites we need
  to think about.

* I'm not convinced about the model of separating neat test specs
  (test/suites/*/spec.go) and inevitably messy implementations
  (test/util/scenarios) across separate packages.  I fear that tests
  can never be super-neat by their nature, and I don't think that
  aiming for this level of neat test specs justifies the maintenance
  costs involved.  I'm not comfortable with the very high number of
  arbitrarily defined exported functions to try to maintain over time.
  I think it's essential that package scope and APIs are tightly
  defined.

  I wonder whether we should get to a model where many of our current
  It()s end up being a free-standing Describe()s, including
  implementation, each in a separate .go file in a given
  package/suite.  Sub-packages per functional area.  This more closely
  matches origin's e2e tests and I think may have advantages around
  maintainability and future parallelism.

* I see advantages in maintaining the ability to compile our tests
  into a single free-standing executable / single entrypoint, although
  I don't know how that fits with the "validate cluster / mutate state
  / revalidate cluster" approach for all of the tests we're going to
  need in the future.  I'm concerned about moving to have multiple
  test entrypoints now without discussing the longer term vision and
  the knock-on effects of this approach over others.

* I think we should aim to reuse library code where possible between
  production and test.  Using non-production error handling idioms
  (e.g. Expect(err).NotTo(HaveOccurred()) / panic(err)) in libraries
  prevents that.  Are we able to use/extend pkg/azureclient instead of
  adding test/util/client/azure, for example?  I think we need to
  agree a direction and guidelines here as a team.

* In general I'm not comfortable with test/util/client/azure becoming
  a random grab-bag of Azure methods.  We ended up there in origin e2e
  and it's unmaintainable.  Lots of the functions there should
  probably be private to specific tests (example:
  `UpdateScaleSetSSHKey`).  Some should probably be collected up and
  go in a separate, well defined, library (e.g. `func (az *Client)
  ApplicationResourceGroup` and friends).

  Also, building and maintaining a new API which coalesces multiple
  underlying API calls into a new paradigm is a pretty slippery slope
  - there always turn out to be good reasons why the underlying API
  works as it does.  So in general I'm worried about things like `func
  (az *Client) ScaleSets` as well.  Even more so all the helpers in
  test/util/client/kubernetes.

* I don't want to needlessly add more environment variables and flags,
  and I think CONFIG_BLOB_PATH and HELPERS_PATH fit in that category.
  Less is more.  Can you ensure you start the fake RP in the right
  directory instead of adding these?

@charlesakalugwu, I welcome and am grateful for your proactiveness and
effort on this PR.  I feel like it uncovers multiple significant
underlying questions that we need to work through as a team and with
which we probably need to start.  I think our next steps should be to
talk through the above (and any additional points others may raise),
and then come to a consensus about how we want to proceed in that
light.

env.example
asalkeld: they probably need to be consts in the code somewhere

kargakis:
Agreed with @jim-minter on all points except the keyrotation suite. I think each upgrade test has to be its own _suite_, we cannot really batch multiple upgrade tests in a single run for obvious reasons.

kargakis:
Also, I have created https://jira.coreos.com/browse/AZURE-222 which I think is our top priority in testing and we should be able to resolve some of the issues this PR has come across as part of that Jira ticket.

charlesakalugwu:
@jim-minter Thanks for your thorough analysis of my work in this PR. I'm excited to discuss all of the issues with you and the team to get this to a place where we are all happy.

jim-minter:
closing for #780

PR #764, opened 2018-11-20
harmonise logging, ensure that we use a *logrus.Entry wherever possible

jim-minter:


jim-minter:
@amanohar among other things this PR should hopefully deal with the missing logs we discussed this morning.

pkg/cluster/types.go
```diff
@@ -41,15 +40,16 @@ type simpleUpgrader struct {
 	vmc            azureclient.VirtualMachineScaleSetVMsClient
 	ssc            azureclient.VirtualMachineScaleSetsClient
 	kubeclient     kubernetes.Interface
+	log            *logrus.Entry
```
kargakis: +100000 on moving into the struct

PR #765, opened 2018-11-21
Improve the check for isCreate: use the existence of the generated manifest

### LINKED ISSUE #759: GET openshift resource instead of resource group
When creating new cluster we rely on the creation of RG to determine if this is `create` or `update`
https://github.com/openshift/openshift-azure/blob/298535a71a1fce290eafb7300e9783f8e1788d88/cmd/createorupdate/createorupdate.go#L267-L272

This does not represent real RP as if we want to rerun script it will treat existing group as `update` and will not even initiate arm template. 

we should be ~~checking for something like ARM deployment, which would correspond to managedApp in realRP~~ reading the openshift object instead of solely checking the existence of the resource group. 

/kind bug

asalkeld:
closes #759

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -262,12 +262,12 @@ func main() {
 	}
 	log = logrus.NewEntry(logger).WithFields(logrus.Fields{"resourceGroup": conf.ResourceGroup})
 
-	var isCreate bool
+	var rgCreated bool
 	if strings.ToUpper(*method) != http.MethodDelete {
 		log.Infof("creating resource group %s", conf.ResourceGroup)
-		if isCreate, err = createResourceGroup(conf); err != nil {
+		if rgCreated, err = createResourceGroup(conf); err != nil {
 			log.Fatal(err)
-		} else if !isCreate {
+		} else if !rgCreated {
 			log.Infof("reusing existing resource group %s", conf.ResourceGroup)
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -301,7 +301,8 @@ func main() {
 		return
 	}
 
-	if isCreate {
+	if _, err := os.Stat(path.Join(outputDirectory, "manifest.yaml")); !os.IsNotExist(err) {
```
asalkeld: oops, needs to be without the !

cmd/createorupdate/createorupdate.go
```diff
@@ -301,7 +301,8 @@ func main() {
 		return
 	}
 
-	if isCreate {
+	if _, err := os.Stat(path.Join(outputDirectory, "manifest.yaml")); os.IsNotExist(err) {
```
kargakis: This is simpler than a GET but if we want to simulate the real RP as much as possible we should be using the osa sdk to check that the openshift object actually exists. It may be problematic to do though since we don't persist objects on the fake rp side so running a hack/create.sh followed by a hack/upgrade.sh is not going to work as is.

asalkeld: I guess that just felt like overkill @jim-minter what do you think?

asalkeld: i think it should be the same way as: https://github.com/openshift/openshift-azure/blob/master/pkg/fakerp/fakerp.go#L47-L63
otherwise we could end up with one thinking it's an update and the other a create.

asalkeld:
closing as i prefer this: https://github.com/openshift/openshift-azure/pull/778

PR #766, opened 2018-11-21
Add another spec to e2erp

thekad:
This spec will ensure a user cannot read the config blob from the
storage account (or read the storage account at all)

### DIFF
--- a/test/e2erp/client_test.go
+++ b/test/e2erp/client_test.go
@@ -27,7 +27,6 @@ type AzureConfig struct {
 	ClientID        string   `envconfig:"AZURE_CLIENT_ID" required:"true"`
 	ClientSecret    string   `envconfig:"AZURE_CLIENT_SECRET" required:"true"`
 	ResourceGroup   string   `envconfig:"RESOURCEGROUP" required:"true"`
-	ClusterName     string   `envconfig:"CLUSTERNAME"`
 	AcceptLanguages []string `envconfig:"ACCEPT_LANGUAGES" default:"en-us"`
 }
 
@@ -43,7 +42,6 @@ type testClient struct {
 
 	resourceGroup    string
 	location         string
-	clusterName      string
 	appResourceGroup string
 }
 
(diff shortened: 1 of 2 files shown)


test/e2erp/client_test.go
```diff
@@ -25,19 +27,24 @@ type AzureConfig struct {
 	ClientID        string   `envconfig:"AZURE_CLIENT_ID" required:"true"`
 	ClientSecret    string   `envconfig:"AZURE_CLIENT_SECRET" required:"true"`
 	ResourceGroup   string   `envconfig:"RESOURCEGROUP" required:"true"`
+	ClusterName     string   `envconfig:"CLUSTERNAME"`
```
kargakis: Let's add this when we are going to use it. Remove it for now.

test/e2erp/realrp_test.go
```diff
@@ -3,45 +3,61 @@
 package e2erp
 
 import (
-	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/sirupsen/logrus"
-
```
kargakis: make verify breaks here

test/e2erp/realrp_test.go
```diff
@@ -3,45 +3,61 @@
...
-	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/log"
+	"github.com/sirupsen/logrus"
```
mjudeikis: https://github.com/openshift/openshift-azure/pull/764

test/e2erp/client_test.go
thekad: @kargakis actually I use it right now, the `az openshift` tool allows you to create several cluster per "parent" resourcegroup and of course the cluster and resource group may not match, so I feel like we should account for that already

kargakis: I meant are there any plans to use this in CI?

thekad: Ah, you're right. Gonna remove.

test/e2erp/realrp_test.go
thekad: latest commit fixes this

test/e2erp/realrp_test.go
thekad: had to rework to get rid of logger stuff, using logrus.* now

PR #769, opened 2018-11-21
Validate pkg/util is empty

kargakis:
@jim-minter thoughts?

/cc jim-minter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ verify:
 	./hack/validate-generated.sh
 	go vet ./...
 	./hack/verify-code-format.sh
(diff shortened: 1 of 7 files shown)


mjudeikis:
I think this is ok. 
And we should agree (I think it is a case already) that anything going to utils is used only for fakeRP, tests, or our dev work. Anything related to vendored code should land into root packages

kargakis:
That's a different discussion and I don't think I agree. Everything
non-prod, related to the fake rp should go into pkg/fakerp, and anything
test-related should go under /test.

Issue #698, opened 2018-11-06
fake rp needs to reuse status code from group deletion error


kargakis:
```
+ go run cmd/createorupdate/createorupdate.go -request=DELETE -use-prod=false
INFO[2018-11-06T14:26:41+01:00] starting the fake resource provider           resourceGroup=mkargakiosa
INFO[2018-11-06T14:26:41+01:00] deleting cluster                              resourceGroup=mkargakiosa
INFO[2018-11-06T14:26:41+01:00] starting server on localhost:8080             resourceGroup=mkargakiosa
INFO[2018-11-06T14:26:42+01:00] deleting resource group mkargakiosa           resourceGroup=mkargakiosa
DEBU[2018-11-06T14:26:42+01:00] 500 Internal Error: Failed to delete resource group: resources.GroupsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'mkargakiosa' could not be found."  resourceGroup=mkargakiosa
INFO[2018-11-06T14:27:12+01:00] deleting resource group mkargakiosa           resourceGroup=mkargakiosa
DEBU[2018-11-06T14:27:12+01:00] 500 Internal Error: Failed to delete resource group: resources.GroupsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'mkargakiosa' could not be found."  resourceGroup=mkargakiosa
INFO[2018-11-06T14:28:12+01:00] deleting resource group mkargakiosa           resourceGroup=mkargakiosa
DEBU[2018-11-06T14:28:13+01:00] 500 Internal Error: Failed to delete resource group: resources.GroupsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'mkargakiosa' could not be found."  resourceGroup=mkargakiosa
^Csignal: interrupt
```
Otherwise, we return a 500 and the client retries the deletion.
/kind bug

kargakis:
got a fix for this 

/assign

PR #771, opened 2018-11-21
Add NO_WAIT functionality for DELETEs

kargakis:
Also:
* fix resource group deletion to exit successfully if the resource
group is already deleted (fixes https://github.com/openshift/openshift-azure/issues/698)
* use the delete future appropriately

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -76,23 +76,28 @@ func validate() error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


kargakis:
This is not ready yet, need to avoid waiting for the group deletion in the server.

kargakis:
@jim-minter @mjudeikis ptal, I have updated our fake RP to do DELETEs appropriately and the NO_WAIT env is going to give us back 4 minutes in the e2e test.

PR #772, opened 2018-11-22
3.11.43 release 

mjudeikis:
bumpt to new release while we still can :)


mjudeikis:
/hold 
until image publishing goes through

mjudeikis:
@kargakis  @jim-minter should we try and test "upgrade" flow with this minor releases on our code base? See how it would play out in the real world?

PR #773, opened 2018-11-22
geneva metrics integration

mjudeikis:
Geneva metrics integration 

- [x] Check if we can move to new namespace for metrics and not to run in prometheus namespace
- [x] add post-install check deployments


pkg/addons/data/Deployment.apps/openshift-azure-monitoring/mdm.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - name: azure-registry
+      restartPolicy: Always
+      terminationGracePeriodSeconds: 30
+      containers:
+        - name: statsd
+          image: "*** GENERATED ***"
+          imagePullPolicy: Always
+          env:
+            - name: MDMENDPOINT
+              value: "https://az-int.metrics.nsatc.net/"
```
mjudeikis: @sozercan I assume these will be configured from RP?

pkg/addons/data/Deployment.apps/openshift-azure-monitoring/prom-mdm-converter.yaml
```diff
@@ -0,0 +1,49 @@
...
+          imagePullPolicy: "Always"
+          ports:
+            - containerPort: 8080
+          env:
+            - name: SERVER_PORT
+              value: "8080"
+            - name: REGION
+              value: "*** GENERATED ***"
+            - name: STATSD_ENDPOINT
+              value: "mdm.openshift-monitoring"
```
mjudeikis: TODO: make this work from separate namespaces

pkg/addons/data/Deployment.apps/openshift-monitoring/mdm.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: mdm-shard
+  name: mdm
+  namespace: openshift-monitoring
```
mjudeikis: We running in Prometheus namespace. Not nice :/
We need this because statsD calls prometheus endpoints directly. And we do not allow inter-project communication if we run this in separate namespace

pkg/addons/data/Namespace/openshift-azure-monitoring.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
+  name: openshift-azure-monitoring
```
mjudeikis: We dont use this. Work out how to use it.

pkg/addons/data/ConfigMap/openshift-azure-monitoring/prom-mdm-converter-rules.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: ConfigMap
+metadata:
+  name: prom-mdm-converter-rules
+  namespace: openshift-azure-monitoring
+data:
+  rules.yaml: |
+    defaultMetricAllow: false
+    defaultDimensionAllow: true
+    metrics:
+      kube_node_info:
```
mjudeikis: @sozercan can we make this "Forward all" in any way? So we would control those on our Prometheus level so we can limit how much work it needs to do.

mjudeikis:
/cc @jim-minter @kargakis 
Most simple implementation of the metric integration

pkg/api/config.go
```diff
@@ -56,6 +56,8 @@ type Config struct {
 
 	// geneva logging sector
 	GenevaLoggingSector string `json:"genevaLoggingSector,omitempty"`
+	GenevaMDMAccount    string `json:"genevaMDMAccount,omitempty"`
```
kargakis: We could use docs about these new fields.

mjudeikis:
I was under the impression I did this already. But it might be just in my head :D

mjudeikis:
/hold
Need to validate if metrics are going through first

pkg/addons/data/ConfigMap/openshift-azure-monitoring/prometheus-forwarder-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    scrape_configs:
+    - job_name: cluster-prometheus
+      scrape_interval: 30s
+      bearer_token_file: "/run/secrets/kubernetes.io/serviceaccount/token"
+      scheme: https
+      tls_config:
+        ca_file: "/config/ca.crt"
+      params:
+        'match[]':
+          - '{__name__=~".+"}'
```
mjudeikis: We should remove rules from config above and just control metrics we pass here as this ipacts our prometheus performance directly

pkg/api/config.go
mjudeikis: docs.go?

mjudeikis:
@openshift/sig-azure ready for review

pkg/api/config.go
kargakis: godoc, comments on top of the fields

pkg/addons/data/Secret/openshift-azure-monitoring/mdm-cert.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mdm-cert
+  namespace: openshift-azure-monitoring
+type: Opaque
+stringData:
+  cert.pem: "*** GENERATED ***"
+  key.pem: "*** GENERATEAD ***"
```
jim-minter: GENERATED

pkg/cluster/ready.go
```diff
@@ -47,6 +47,10 @@ var deploymentWhitelist = []struct {
 		Name:      "cluster-monitoring-operator",
 		Namespace: "openshift-monitoring",
 	},
+	{
+		Name:      "metrics-bridge",
```
jim-minter: in the bad old days we didn't wait for prometheus to come up before declaring a cluster ready - but we checked that it did in all the e2es.  Can't we do the same for metrics-bridge and cluster-monitoring-operator?  A bit of a cheat, but it might help our startup times?

jim-minter:
Just a couple of niggles to fix here otherwise this is good to merge.  I believe I have seen metrics from this in Geneva so this can be unheld :)

pkg/api/config.go
jim-minter: @mjudeikis bump

pkg/cluster/ready.go
mjudeikis: https://github.com/openshift/openshift-azure/issues/842

PR #777, opened 2018-11-22
fix typo and missing debuglevel setting

jim-minter:


jim-minter:
lgtm'ing as this will help with flakes

PR #778, opened 2018-11-23
Fix update with no manifest defined

asalkeld:


asalkeld:
cc @jim-minter 

Haha, we both did the debug level but in different ways :-O

env.example
```diff
@@ -66,3 +66,6 @@ export DEPLOY_VERSION=v3.11
 
 # Path to the request to send to the OSA RP.
 #export MANIFEST=test/manifests/normal/create.yaml
+
+# Set the loglevel to pass to createorupdate (defaults to Debug)
+#export LOGLEVEL=Info
```
asalkeld: should show the default, will change

cmd/createorupdate/createorupdate.go
```diff
@@ -46,10 +44,8 @@ var (
 
 	artifactDir        = flag.String("artifact-dir", "", "Directory to place artifacts before a cluster is deleted.")
 	artifactKubeconfig = flag.String("artifact-kubeconfig", "", "Path to kubeconfig to use for gathering artifacts.")
-)
 
-const (
-	outputDirectory = "_data"
+	logLevel = flag.String("loglevel", "Debug", "Valid values are Debug, Info, Warning, Error")
```
kargakis: No need to have this, we are always going to use Debug in createorupdate.

env.example
```diff
@@ -66,3 +66,6 @@ export DEPLOY_VERSION=v3.11
 
 # Path to the request to send to the OSA RP.
 #export MANIFEST=test/manifests/normal/create.yaml
+
+# Set the loglevel to pass to createorupdate
+#export LOGLEVEL=Debug
```
kargakis: remove

hack/create.sh
```diff
@@ -12,11 +12,15 @@ mkdir -p _data/_out
 
 set -x
 
+if [[ -n "$LOGLEVEL" ]]; then
+LOGLEVEL_ARG="-loglevel=$LOGLEVEL"
```
kargakis: remove

pkg/fakerp/fakerp.go
```diff
@@ -23,6 +23,14 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
+// IsUpdate return whether or not this is an update or create.
+func IsUpdate() bool {
+	if _, err := os.Stat("_data/containerservice.yaml"); err == nil {
+		return true
```
kargakis: nit: return err == nil

cmd/createorupdate/createorupdate.go
```diff
@@ -201,8 +189,13 @@ func execute(ctx context.Context, log *logrus.Entry, rpc sdk.OpenShiftManagedClu
 	if *update != "" {
 		updateCtx, updateCancel := context.WithTimeout(context.Background(), *updateTimeout)
 		defer updateCancel()
-		updateManifestFile := path.Join(outputDirectory, "update.yaml")
-		if err := createOrUpdate(updateCtx, log, rpc, conf.ResourceGroup, *update, updateManifestFile); err != nil {
+
+		oc, err := fakerp.LoadClusterConfigFromManifest(log, *update, conf)
```
kargakis: Move this in `createOrUpdate`?

PR #779, opened 2018-11-23
Etcd cluster recovery 

asalkeld:
/cc @jim-minter 

this is working nicely now


hack/etcdrecovery.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+if [[ $# -lt 2 ]]; then
+    echo error: usage "$0 resourceGroup backupblobname"
+    exit 1
+fi
+export RESOURCEGROUP=$1
+export RECOVER_ETCD_FROM_BACKUP=$2
+
+oc delete configmap angus-test || true
```
kargakis: Hrmmm, I wouldn't expect to have access to the api server when recovering etcd.

hack/etcdrecovery.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
kargakis: Can't you reuse upgrade.sh?

pkg/cluster/update.go
```diff
@@ -138,11 +156,15 @@ func getCount(cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) in
 
 func (u *simpleUpgrader) listVMs(ctx context.Context, cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) ([]compute.VirtualMachineScaleSetVM, error) {
 	vmPages, err := u.vmc.List(ctx, cs.Properties.AzProfile.ResourceGroup, "ss-"+string(role), "", "", "")
+	var vms []compute.VirtualMachineScaleSetVM
```
kargakis: nit: define this before List

pkg/cluster/update.go
```diff
@@ -138,11 +156,15 @@ func getCount(cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) in
...
 	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
```
kargakis: This is a typed error; cast it to the type returned by List and read the error message correctly.

pkg/arm/arm.go
```diff
@@ -54,8 +54,10 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
 		"Startup": func(role api.AgentPoolProfileRole) ([]byte, error) {
 			if role == api.AgentPoolProfileRoleMaster {
 				return util.Template(string(masterStartup), nil, cs, map[string]interface{}{
-					"Role":       role,
-					"TestConfig": g.pluginConfig.TestConfig,
+					"IsRecovery":     len(g.pluginConfig.RecoverEtcdFromBackup) > 0,
+					"BackupBlobName": g.pluginConfig.RecoverEtcdFromBackup,
```
jim-minter: One route here is to pass this in as an argument to `Generate()` - usually empty, only populated to do a restore

pkg/plugin/plugin.go
```diff
@@ -100,6 +100,24 @@ func (p *plugin) GenerateARM(ctx context.Context, cs *api.OpenShiftManagedCluste
 	return p.armGenerator.Generate(ctx, cs, isUpdate)
 }
 
+func (p *plugin) RecoverEtcdCluster(ctx context.Context, cs *api.OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployFn api.DeployFn) *api.PluginError {
+	p.log.Info("deleting master scaleset")
+	if err := p.clusterUpgrader.DeleteScaleSet(ctx, cs, "ss-master"); err != nil {
```
jim-minter: I've got a few worries.
1) what happens if the restored etcd has a different view of what nodes should be around than what is actually there?
2) how can we make very sure we don't accidentally run the restore a second time around?

I'm wondering whether this workflow should look more like the following:
1) delete *all* the scalesets and the hash blob.
2) run a deploy with the flag which includes the restore functionality
3) run an update with the flag unset (should just rotate the masters removing the script addition)

Still need to manually check (and then have an automated test) to cover the differing nodes problem.

@kargakis @mjudeikis thoughts here?

pkg/arm/data/master-startup.sh
```diff
@@ -651,6 +651,36 @@ sed -i -re "s#( *server: ).*#\1https://$(hostname)#" /etc/origin/node/node.kubec
 # daemonset will not fail and set the master node labels correctly.
 cp /etc/origin/node/node.kubeconfig /etc/origin/node/bootstrap.kubeconfig
 
+{{- if $.Extra.IsRecovery }}
+logger -t master-startup.sh "starting recovery on $(hostname)"
+# step 1 get the backup
+tempBackDir=$(mktemp -d -t etcdbackup-XXXXXX)
```
jim-minter: I'm worried that disk space is going to be an issue here in any non-toy example.  If you save the snapshot to the same disk, presumably it should work for dbs up to approx 1/2 of the disk size?  Presumably any bigger will actually require a temporary restore disk or something?  Suggest you implement the former and open a story for the latter.

env.example
```diff
@@ -66,3 +66,6 @@ export DEPLOY_VERSION=v3.11
 
 # Path to the request to send to the OSA RP.
 #export MANIFEST=test/manifests/normal/create.yaml
+
+# install etcd backup data from this storage blob (don't include the etcd/ container)
+#export RECOVER_ETCD_FROM_BACKUP=backup-foo
```
jim-minter: don't believe this makes sense - as mentioned below, suggest creating a small dumb freestanding harness under cmd/ to call RecoverEtcdCluster. Once the admin API exists and the fake RP supports it, we can roll it in.

hack/etcdrecovery.sh
jim-minter: I'm fine with the separate script, but it should be minimal and calling a separate helper prior to rolling it in after the admin API work

pkg/cluster/types.go
```diff
@@ -54,6 +55,9 @@ func NewSimpleUpgrader(log *logrus.Entry, pluginConfig *api.PluginConfig) Upgrad
 }
 
 func (u *simpleUpgrader) createClients(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+	if u.ssc != nil {
```
jim-minter: drop

pkg/cluster/update.go
```diff
@@ -12,6 +12,22 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
 
+func (u *simpleUpgrader) DeleteScaleSet(ctx context.Context, cs *api.OpenShiftManagedCluster, scaleSetName string) *api.PluginError {
+	err := u.createClients(ctx, cs)
```
jim-minter: do this in the caller

pkg/cluster/update.go
jim-minter: I'm worried why this change is here - does the suggestion below of calling Deploy, not Upgrade, to do the restore mean this can be taken out?

asalkeld: it's because we need to delete the whole scaleset prior to running the update, so the "list" op can't find the scaleset

pkg/api/plugin.go
```diff
@@ -109,4 +111,7 @@ type Plugin interface {
 	// CreateOrUpdate either deploys or runs the update depending on the isUpdate argument
 	// this will call the deployer.
 	CreateOrUpdate(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, isUpdate bool, deployer DeployFn) *PluginError
+
+	// RecoverEtcdCluster recovers the cluster's etcd using the backup specified in the pluginConfig
+	RecoverEtcdCluster(ctx context.Context, cs *OpenShiftManagedCluster, azuretemplate map[string]interface{}, deployer DeployFn) *PluginError
```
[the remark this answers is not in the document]
asalkeld: the generate needs the blob, not this. this is just a special version of update.

pkg/arm/data/master-startup.sh
asalkeld: https://github.com/openshift/openshift-azure/issues/785

pkg/plugin/plugin.go
kargakis: I am not sure deleting all the scalesets is the right thing to do. Wouldn't that have an impact on customer workloads? It should be possible to restore clusters seemlessly. I would expect the cloud controller manager should be able to reconcile the current VMs with the node objects and delete stray objects, and kubelets with no node objects are able to recreate theirs.

env.example
mjudeikis: Agree. We need to think how we will do this in AdminAPI world.

pkg/api/plugin.go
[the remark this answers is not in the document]
jim-minter: We should implement one function in the interface which the geneva action will call, not ask it to call multiple.

pkg/arm/data/master-startup.sh
```diff
@@ -651,6 +651,36 @@ sed -i -re "s#( *server: ).*#\1https://$(hostname)#" /etc/origin/node/node.kubec
...
+docker pull {{ .Config.Images.EtcdBackup }}
+docker run --dns 8.8.8.8 -v /etc/origin/master:/etc/origin/master -v /etc/origin/cloudprovider/:/_data/_out -v $tempBackDir:/out:z {{ .Config.Images.EtcdBackup }} -blobname={{ .Extra.BackupBlobName }} -destination=/out/backup.db "download"
```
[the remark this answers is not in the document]
asalkeld: do you know what that ip is?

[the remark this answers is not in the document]
mjudeikis: and apperently they dont have one public :/

cmd/recoveretcdcluster/recoveretcdcluster.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, api.ContextKeyClientID, conf.ClientID)
+	ctx = context.WithValue(ctx, api.ContextKeyClientSecret, conf.ClientSecret)
+	ctx = context.WithValue(ctx, api.ContextKeyTenantID, conf.TenantID)
+
+	config, err := fakerp.GetPluginConfig()
```
kargakis: fakerp was never meant to be used in production code and recoveretcdcluster is production code. We should probably move the fakerp functions used here in a different package and ensure they are properly unit-tested. Please open follow-up issue(s) accordingly

@jim-minter

pkg/cluster/deploy.go
```diff
@@ -9,15 +9,39 @@ import (
...
+		keys, err := u.accountsClient.ListKeys(ctx, cs.Properties.AzProfile.ResourceGroup, cs.Config.ConfigStorageAccount)
+		if err != nil {
+			return &api.PluginError{Err: err, Step: api.PluginStepClientCreation}
+		}
+		u.storageClient, err = storage.NewClient(cs.Config.ConfigStorageAccount, *(*keys.Keys)[0].Value, storage.DefaultBaseURL, storage.DefaultAPIVersion, true)
+		if err != nil {
+			return &api.PluginError{Err: err, Step: api.PluginStepClientCreation}
+		}
+	}
+	err = u.deleteUpdateBlob()
```
kargakis: This will result in unnecessarily rotating the compute and infra nodes.

kargakis: Only the master entries should be removed from the blob.

cmd/recoveretcdcluster/recoveretcdcluster.go
kargakis: https://github.com/openshift/openshift-azure/pull/798 needs reviews

PR #783, opened 2018-11-26
add validate-imports.go

jim-minter:


kargakis:
@jim-minter is the added benefit on top of our current use of goimports the dot imports handling?

kargakis:
tcp timeout

/retest

jim-minter:
dot imports, stricter group order (std library, dot imports, !local, local), can't have multiple separated groups of the same type.

mjudeikis:
I'm still surprised how you just pull out code like this out of your magic hat...
/lgtm

PR #789, opened 2018-11-27
move to networkpolicy plugin

mjudeikis:
This will align us with upstream default option and OCP Dedicated.

In addition we will be able to create cross-namespace pod traffic Network rules for our own use.
In example `openshift-monitoring` and `openshift-azure-monitoring` namespaces 

Follow-up stories if we chose to merge this:
Create a set of Default network-policies which will be managed by Azure Operator and enable users to create their own network policy objects.

/cc @jim-minter @pweil- @kargakis WDYT?

https://docs.openshift.com/container-platform/3.11/admin_guide/managing_networking.html#admin-guide-networking-networkpolicy


kargakis:
I am in favor of switching to the network policy. Are there any cons? Things we are going to miss by doing the switch?

PR #744, opened 2018-11-16
Ensure unlink command always exits with status code 0

charlesakalugwu:
Without this change, the compute scaleset becomes unhealthy during an update due to the missing link.

closes #731

Issue #731, opened 2018-11-14
Compute scale set CSE script is being re-executed on scale up


charlesakalugwu:
We ran into an issue where it appeared that the node startup scripts were being reapplied to the vms upon update. 

We've been under the understanding that these scripts are run only once the first time a VM is created and never again.

We encountered this issue when developing key rotation e2e tests.

```
VM has reported a failure when processing extension 'cse'. Error message: "Enable failed: failed to execute command: command terminated with exit status=1 [stdout] [stderr] + SERVICE_TYPE=origin + '[' -f /etc/sysconfig/atomic-openshift-node ']' + SERVICE_TYPE=atomic-openshift + unlink /etc/docker/certs.d/registry.access.redhat.com/redhat-ca.crt unlink: cannot unlink '/etc/docker/certs.d/registry.access.redhat.com/redhat-ca.crt': No such file or directory ".
```

@jim-minter @kargakis

kargakis:
This issue came up during @charlesakalugwu's [key rotation upgrade test](https://github.com/openshift/openshift-azure/pull/726). During the upgrade, when the compute scaleset is rotated for some reason this issue happens. All VMs seem to come up correctly (master, infra, even the new compute VM is joining the cluster).

kargakis:
I don't think this question is an issue in openshift-azure and this issue tracker is probably the wrong venue. Maybe take this question in the openshiftazure slack?

mjudeikis:
/open
This is repeatable and reproducible. let's reopen so we can track progress, and if we end up with non-azure issue, we can move. But now we don't know where the issue is. 

I was thinking this might be related to our update code but now think about this less and less.
1. Tried to change the order of upgrades (master, compute, infra) as was thinking something leaks to update blob - No luck.
2. Compute is being created with count=1. Changes to Count=2 to start with as was thinking about how we do "PlussOne" Updates - No luck. 
3. Checked logs on compute0 just to check if we can get confirmation about this - no luck. Logs does not show anything. I was expecting confirmation on a script running twice. - No luck

At this point, because compute and infra is theoretically the same I don't know what might be the issue.

@sozercan reported that logging stack somehow cant connect from compute0 to Geneva. All other nodes connect properly. This raises the question what is different about how we configure Compute nodes compared to infra nodes, and to be specific - first compute node we bootstrap. As if we scale and add compute1 - it works just fine. 

cc @charlesakalugwu @kargakis

kargakis:
[...]

PR #794, opened 2018-11-28
Key rotation e2e tests

charlesakalugwu:
closes AZURE-117

/cc @kargakis @mjudeikis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,4 +90,7 @@ e2e-prod:
 e2e-etcdbackuprecovery:
 	FOCUS="\[EtcdRecovery\]\[Fake\]" TIMEOUT=70m ./hack/e2e.sh
 
+e2e-keyrotation:
+	FOCUS="\[KeyRotation\]\[Fake\]" TIMEOUT=70m ./hack/e2e.sh
+
 .PHONY: clean sync-image sync-push verify unit e2e
--- a/test/clients/azure/osa.go
+++ b/test/clients/azure/osa.go
@@ -4,6 +4,13 @@ import (
 	"context"
 	"fmt"
 	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/util/log"
 )
 
 // OSAResourceGroup returns the name of the resource group holding the OSA
@@ -23,4 +30,27 @@ func (cli *Client) OSAResourceGroup(ctx context.Context, resourcegroup, name, lo
 	// /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}
 	rgid := *app.ApplicationProperties.ManagedResourceGroupID
 	return rgid[strings.LastIndexByte(rgid, '/')+1:], nil
+}
+
+// UpdateOSACluster updates an OpenshiftManagedCluster by providing both the
+// currently applied external manifest and an updated internal manifest
+// NOTE: This method will go away as soon as the admin API is ready. Until then,
+// only the KeyRotation or other e2e tests needing admin functionality should
+// call this method.
+func (cli *Client) UpdateOSACluster(ctx context.Context, external *v20180930preview.OpenShiftManagedCluster, config *api.PluginConfig) (*v20180930preview.OpenShiftManagedCluster, error) {
+	// Remove the provisioning state before updating, if set
+	external.Properties.ProvisioningState = nil
+
+	// create logger
+	logrus.SetLevel(log.SanitizeLogLevel("Debug"))
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	logger := logrus.NewEntry(logrus.StandardLogger())
+
+	var oc *v20180930preview.OpenShiftManagedCluster
+	var err error
+	// simulate the API call to the RP
+	if oc, err = fakerp.CreateOrUpdate(ctx, external, logger, config); err != nil {
+		return nil, err
+	}
+	return oc, nil
 }
--- a/test/e2e/specs/fakerp/keyrotation.go
+++ b/test/e2e/specs/fakerp/keyrotation.go
@@ -1,0 +1,151 @@
+package fakerp
+
+import (
+	"context"
+	"flag"
+	"io/ioutil"
+	"os"
+	"reflect"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/ghodss/yaml"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/config"
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+)
+
+var _ = Describe("Key Rotation E2E tests [KeyRotation][Fake][LongRunning]", func() {
+	var (
+		cli        *azure.Client
+		manifest   = flag.String("manifest", "../../_data/manifest.yaml", "Path to the manifest to send to the RP")
+		configBlob = flag.String("configBlob", "../../_data/containerservice.yaml", "Path to the OpenShift internal config blob")
+	)
+
+	BeforeEach(func() {
+		var err error
+		cli, err = azure.NewClientFromEnvironment()
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should be possible to maintain a healthy cluster after rotating all credentials", func() {
+		ctx := context.Background()
+		ctx = context.WithValue(ctx, api.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
+		ctx = context.WithValue(ctx, api.ContextKeyClientSecret, os.Getenv("AZURE_CLIENT_SECRET"))
+		ctx = context.WithValue(ctx, api.ContextKeyTenantID, os.Getenv("AZURE_TENANT_ID"))
+
+		By("Parsing the external manifest")
+		external, err := parseExternalConfig(*manifest)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(external).NotTo(BeNil())
+
+		By("Parsing the internal manifest containing config blob")
+		internal, err := managedcluster.ReadConfig(*configBlob)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(internal).NotTo(BeNil())
+
+		By("Deleting all non-ca cluster certificates and credentials from the config blob...")
+		mutated := deleteSecrets(internal)
+		Expect(mutated).NotTo(BeNil())
+
+		By("Running generate on the modified config blob")
+		pluginConfig, err := fakerp.GetPluginConfig()
+		Expect(err).NotTo(HaveOccurred())
+		configGen := config.NewSimpleGenerator(pluginConfig)
+		Expect(configGen).NotTo(BeNil())
+		err = configGen.Generate(mutated)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Persisting the config blob containing the new certificates and credentials")
+		err = saveConfig(mutated, *configBlob)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Calling update on the fake rp with the updated config blob")
+		updated, err := cli.UpdateOSACluster(ctx, external, pluginConfig)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(updated).NotTo(BeNil())
+
+		By("Parsing the config blob after the update")
+		internalAfterUpdate, err := managedcluster.ReadConfig(*configBlob)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(internalAfterUpdate).NotTo(BeNil())
+
+		By("Verifying that the initial config blob does not match the one created after the update")
+		configMatch := reflect.DeepEqual(internal.Config.Certificates, internalAfterUpdate.Config.Certificates)
+		Expect(configMatch).To(BeFalse())
+
+		By("Verifying that the mutated config blob matches the one created after the update")
+		configMatch = reflect.DeepEqual(mutated.Config.Certificates, internalAfterUpdate.Config.Certificates)
+		Expect(configMatch).To(BeTrue())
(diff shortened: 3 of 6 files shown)


kargakis:
```
S [SKIPPING] in Spec Setup (BeforeEach) [0.001 seconds]
Key Rotation E2E tests [KeyRotation] [Update]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/updates/keyrotation/keyrotation.go:23
  should be possible to maintain a healthy cluster after rotating all credentials [BeforeEach]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/updates/keyrotation/keyrotation.go:40

  failed to get oauth token from client credentials: parameter 'clientID' cannot be empty

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/updates/keyrotation/keyrotation.go:36
```

kargakis:
We should avoid skipping in the e2e framework cc @jim-minter

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,149 @@
...
+		cli        *azure.Client
+		manifest   = flag.String("manifest", "_data/manifest.yaml", "Path to the manifest to send to the RP")
+		configBlob = flag.String("configBlob", "_data/containerservice.yaml", "Path to the OpenShift internal config blob")
+	)
+
+	BeforeEach(func() {
+		var err error
+		cli, err = azure.NewClientFromEnvironment()
+		if err != nil {
+			Skip(err.Error())
```
kargakis: Expect(err).ToNot(HaveOccurred())

kargakis:
```
# github.com/openshift/openshift-azure/test/e2e
test/e2e/e2e_test.go:10:2: cannot find package "github.com/onsi/ginkgo" in any of:
	/go/src/github.com/openshift/openshift-azure/vendor/github.com/onsi/ginkgo (vendor tree)
	/usr/local/go/src/github.com/onsi/ginkgo (from $GOROOT)
	/go/src/github.com/onsi/ginkgo (from $GOPATH)
make: *** [e2e-keyrotation] Error 1
```
Makes no sense

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,149 @@
...
+		cli        *azure.Client
+		manifest   = flag.String("manifest", "_data/manifest.yaml", "Path to the manifest to send to the RP")
+		configBlob = flag.String("configBlob", "_data/containerservice.yaml", "Path to the OpenShift internal config blob")
+	)
+
+	BeforeEach(func() {
+		var err error
+		cli, err = azure.NewClientFromEnvironment()
+		if err != nil {
+			panic(err)
```
kargakis: Expect(err).NotTo(HaveOccurred())

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,149 @@
...
+		Expect(external).NotTo(BeNil())
+
+		By("Parsing the internal manifest containing config blob")
+		internal, err := managedcluster.ReadConfig(*configBlob)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(internal).NotTo(BeNil())
+
+		By("Deleting all non-ca cluster certificates and credentials from the config blob...")
+		mutated := deleteSecrets(internal)
+		Expect(err).NotTo(HaveOccurred())
```
mjudeikis: Where this `err` comes from?

charlesakalugwu: removed :)

Makefile
```diff
@@ -72,4 +72,7 @@ e2e:
 e2e-prod:
 	FOCUS="\[Real\]" ./hack/e2e.sh
 
+e2e-keyrotation:
```
jim-minter: This should be a target which will run all long running tests.  We'd run this via a prow periodic.

test/clients/azure/osa.go
```diff
@@ -24,3 +38,39 @@ func (cli *Client) OSAResourceGroup(ctx context.Context, resourcegroup, name, lo
 	rgid := *app.ApplicationProperties.ManagedResourceGroupID
 	return rgid[strings.LastIndexByte(rgid, '/')+1:], nil
 }
+
+// UpdateOSACluster updates an OpenshiftManagedCluster by providing both the currently applied external manifest and
+// an updated internal manifest
+func (cli *Client) UpdateOSACluster(ctx context.Context, external *v20180930preview.OpenShiftManagedCluster, config *api.PluginConfig) (*v20180930preview.OpenShiftManagedCluster, error) {
```
jim-minter: Please open a follow-up to remove this once the admin API lands.

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,147 @@
...
+		Expect(configMatch).To(BeFalse())
+
+		By("Verifying that the mutated config blob matches the one created after the update")
+		configMatch = reflect.DeepEqual(mutated.Config.Certificates, internalAfterUpdate.Config.Certificates)
+		Expect(configMatch).To(BeTrue())
+	})
+})
+
+// deleteSecrets removes all non-ca certificates, private keys and secrets from OpenShiftManagedCluster.Config
+func deleteSecrets(config *api.OpenShiftManagedCluster) *api.OpenShiftManagedCluster {
```
jim-minter: This code should move into a newly created function inside the plugin, which would be listed in the external plugin interface, which would be responsible for resetting these values and calling Update on the cluster.  You'd call into that from here via the admin API, which would call that function.  This work should also be part of the follow-up mentioned above.

test/clients/azure/osa.go
```diff
@@ -24,3 +38,41 @@ func (cli *Client) OSAResourceGroup(ctx context.Context, resourcegroup, name, lo
...
+	var err error
+	// simulate the API call to the RP
+	if err := wait.PollImmediate(5*time.Second, 1*time.Hour, func() (bool, error) {
+		if oc, err = fakerp.CreateOrUpdate(ctx, external, logger, config); err != nil {
+			if autoRestErr, ok := err.(autorest.DetailedError); ok {
+				if urlErr, ok := autoRestErr.Original.(*url.Error); ok {
+					if netErr, ok := urlErr.Err.(*net.OpError); ok {
+						if sysErr, ok := netErr.Err.(*os.SyscallError); ok {
+							if sysErr.Err == syscall.ECONNREFUSED {
+								return false, nil
```
jim-minter: Do you actually see this happening in the test context?  ie is this code actually needed or is it superstition? ;-)

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/openshift/openshift-azure/pkg/api"
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/config"
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+)
+
+var _ = Describe("Key Rotation E2E tests [KeyRotation] [Update]", func() {
+	defer GinkgoRecover()
```
jim-minter: I don't believe this should be necessary, can it be removed?

test/e2e/specs/updates/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,147 @@
+package keyrotation
```
jim-minter: I think this can go in specs/fakerp/keyrotation.go for now?

Makefile
mjudeikis: are we saying we want to test this with each e2e run?

kargakis: We need to be able to run these long running tests individually. At that point, I am not sure what a single job that runs all long-running tests gives us since each test is going to create/update its own cluster.

kargakis: > are we saying we want to test this with each e2e run?

No, these tests should run periodically and we should also be able to run them individually and optionally on PRs.

test/clients/azure/osa.go
kargakis: We were seeing this in the fake RP create case. Probably not needed for updates since we know the RP is already running.

Makefile
charlesakalugwu: I think the idea here is to condense all of the make e2e targets into a single `make e2e` and then we tag the various e2e specs as follows:

`[EveryPR]` - These tests should be run on every PR (duration-based tag)
`[LongRunning]` - These tests are long running and should be run periodically (duration-based tag)
`[Real]` - These tests should be run against the real RP (endpoint-based tag)
`[Fake]` - These tests should be run against the fake RP until the admin API is ready (endpoint-based tag)
`[RP]` - These tests could be run against either the fake or real RP after the admin API is ready (endpoint-based tag)

The developer of a spec is free to add more tags however those 5 above should be the supported ones. For example:

* Key rotation test (before admin api): `[KeyRotation] [Fake] [LongRunning]`
* Key rotation test (after admin api): `[KeyRotation] [RP] [LongRunning]`
* Scale up/down test (before admin api): `[ScaleUpDown] [Fake] [Real] [LongRunning] [EveryPR]`
* Scale up/down test (after admin api): `[ScaleUpDown] [RP] [LongRunning] [EveryPR]`

Then we create prow jobs using using focus combinations to filter through eligible tests:

* Run all long running tests in one job: `FOCUS=[LongRunning] make e2e`
  * Run the long running key rotation test in one job: `FOCUS=[KeyRotation] make e2e`
  * Run the long running scale up/down test in one job: `FOCUS=[ScaleUpDown] make e2e`
* Run all tests against the real RP: `FOCUS=Real make e2e`
* Run the set of tests before every PR in one job: `FOCUS=EveryPR make e2e`
* Run all tests against the fake RP (until admin API is ready): `FOCUS=Fake make e2e`
* Run all tests against any RP (after admin API is ready): `FOCUS=RP make e2e`

test/clients/azure/osa.go
charlesakalugwu: I lifted this bit of code from cmd/createorupdate.go but it is indeed not needed. I'll take it out

test/e2e/specs/updates/keyrotation/keyrotation.go
charlesakalugwu: alright

test/e2e/specs/updates/keyrotation/keyrotation.go
charlesakalugwu: alright

test/clients/azure/osa.go
charlesakalugwu: here https://github.com/openshift/openshift-azure/issues/807

test/e2e/specs/updates/keyrotation/keyrotation.go
charlesakalugwu: here #807

Makefile
mjudeikis: Agreed. And all this sound like small readme in docs/tests.md with convention or smth like that. as I think this will come up again :)

mjudeikis:
> @mjudeikis should the dns call run during an update?

No. Not in current flow. It will potentially in the future, but not now.

mjudeikis:
Something new :) good!

charlesakalugwu:
> Something new :) good!

actually this isn't new. this was the first issue I discovered locally. https://github.com/openshift/openshift-azure/pull/744/files fixes it. In fact, throughout my development of this test, I always had to have the changes in 744 present locally.

charlesakalugwu:
I have a test cluster in `charlesakalugwu-test` resource group which I allowed to undergo a failed key rotation similar to the scenario we're seeing in the test. It will come in handy for investigating this issue.

Makefile
```diff
@@ -87,4 +87,7 @@ e2e:
...
+	FOCUS="\[KeyRotation\] \[Fake\]" TIMEOUT=70m ./hack/e2e.sh
```
kargakis: Remove the space to conform with the way we structure tags in the e2e make target (this is actually what we do everywhere else in openshift/kube).

kargakis: Remove the space between the tags to conform with the way we structure tags in the e2e make target (this is actually what we do everywhere else in openshift/kube).

test/e2e/specs/fakerp/keyrotation.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/config"
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+)
+
+var _ = Describe("Key Rotation E2E tests [KeyRotation] [Fake] [LongRunning]", func() {
```
kargakis: remove the space between the tags

test/clients/azure/osa.go
```diff
@@ -24,3 +31,26 @@ func (cli *Client) OSAResourceGroup(ctx context.Context, resourcegroup, name, lo
...
+
+	// create logger
+	logrus.SetLevel(log.SanitizeLogLevel("Debug"))
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	logger := logrus.NewEntry(logrus.StandardLogger())
+
+	var oc *v20180930preview.OpenShiftManagedCluster
+	var err error
+	// simulate the API call to the RP
+	if oc, err = fakerp.CreateOrUpdate(ctx, external, logger, config); err != nil {
```
kargakis: Here you should be using the OSA client we have in https://github.com/openshift/openshift-azure/blob/master/pkg/api/2018-09-30-preview/api/client.go#L94, similar to how it's used in createorupdate instead of using the fake RP backend function.

charlesakalugwu: we'll take care in a follow up #807

kargakis: Actually I still think we should be fixing this here. After the admin API lands, you still need a client call from this test so `UpdateOSACluster` is not going to be removed but repurposed, so I think it's worth structuring it correctly from start, ie. you should be querying the fake RP instead of running the update yourself.

kargakis: Should be fixed https://github.com/openshift/openshift-azure/pull/795 where the OSA client is re-introduced into the e2e framework.

kargakis:
Only blocking issue seems to be https://github.com/openshift/openshift-azure/issues/731 - since that's not caused by this PR, lgtm.

charlesakalugwu:
The latest ci e2e tests fail because of https://github.com/openshift/openshift-azure/issues/865

PR #793, opened 2018-11-28
Remove OSA sdk

kargakis:
As discussed w/ @jim-minter, we are getting rid of the OSA sdk to enable using the admin api in the fake rp since the admin api is not going to be part of the sdk.

Fixes https://github.com/openshift/openshift-azure

PR #798, opened 2018-11-28
Add support for the Admin API

kargakis:
This PR only includes the types and the conversions. I'd prefer to get this merged asap and work on validation and making the fake RP use it in separate PRs.

Part of AZURE-119

@jim-minter


kargakis:
Of course, this is also blocked on https://github.com/openshift/openshift-azure/pull/750 since I am reusing some of the functions added in that PR.

pkg/api/admin/api/config.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+type Config struct {
+	// configuration of VMs in ARM template
+	ImageOffer     *string `json:"imageOffer,omitempty"`
+	ImagePublisher *string `json:"imagePublisher,omitempty"`
+	ImageSKU       *string `json:"imageSku,omitempty"`
+	ImageVersion   *string `json:"imageVersion,omitempty"`
+
+	SSHKey *rsa.PrivateKey `json:"sshKey,omitempty"`
```
jim-minter: There should be no keys or secrets in the admin API.  We neither want to risk accidentally returning these to an SRE, nor make it theoretically expressible to set these to a value known by an SRE.  Please remove all keys, secrets, passwords, kubeconfigs across the board.  In the future there may be a separate action which returns a kubeconfig to an SRE, under separate audit rules.

pkg/api/admin/api/config.go
```diff
@@ -0,0 +1,142 @@
...
+	// misc certificates
+	AzureClusterReader *CertKeyPair `json:"azureClusterReader,omitempty"`
+
+	// geneva integration certificates
+	GenevaLogging *CertKeyPair `json:"genevaLogging,omitempty"`
+}
+
+// CertKeyPair is an rsa private key and x509 certificate pair.
+type CertKeyPair struct {
+	Key  *rsa.PrivateKey   `json:"key,omitempty"`
```
jim-minter: Think I'd take the approach of keeping the type, removing the `Key` field, and possibly renaming the type (can't think of a great name though).  That will keep the actual JSON structure as similar as possible.

pkg/api/admin/api/types.go
```diff
@@ -0,0 +1,233 @@
+package api
+
+const (
+	// APIVersion is the version of this API
+	APIVersion = "admin"
+)
+
+// OpenShiftManagedCluster complies with the ARM model of resource definition in
+// a JSON template.
+type OpenShiftManagedCluster struct {
```
asalkeld: wow there is now going to be 4 versions of this struct in our code base.

kargakis: Soon to be brought down to 3 - https://github.com/openshift/openshift-azure/pull/793

But of course long-term we are going to have more, every version of the external apis is going to be its own type.

pkg/api/admin/api/config.go
kargakis: How will we be able to rotate them if we don't expose them in this API? How will Charle's key rotation test work with this API then?

mjudeikis: (brain dump) I think we need something different for this. Like "Rotate keys" and we could have something like validationAPI which would return if backend objects are valid. If we rotated sshKey, it would return ssh key type and validity. overall as SRE I would like to check integrity of my config after action was done. So some callback is needed without sensitive data.

mjudeikis:
So much boilerplace... my eyes hurts :/

pkg/api/admin/api/config.go
kargakis: Yeah, makes sense. We discussed this already with @jim-minter but I hadn't fully grasped how it was going to look like.

kargakis: @jim-minter should I also remove ServicePrincipalProfile and AzProfile?

kargakis:
The reflective test is kinda broken because it does not work recursively. Needs to be reworked.

pkg/api/admin/api/types.go
```diff
@@ -0,0 +1,215 @@
...
+	NetworkProfile *NetworkProfile `json:"networkProfile,omitempty"`
+
+	// RouterProfiles (in,optional/out): Configuration for OpenShift router(s).
+	RouterProfiles []RouterProfile `json:"routerProfiles,omitempty"`
+
+	// AgentPoolProfiles (in): configuration of OpenShift cluster VMs.
+	AgentPoolProfiles []AgentPoolProfile `json:"agentPoolProfiles,omitempty"`
+
+	// AuthProfile (in): configures OpenShift authentication
+	AuthProfile *AuthProfile `json:"authProfile,omitempty"`
```
kargakis: @jim-minter while on removing Secret from AADIdentityProvider, I am still trying to figure out whether exposing most of these options (parts of Properties) makes any sense for an SRE to manipulate.

pkg/api/admin/api/types.go
```diff
@@ -0,0 +1,214 @@
...
+	// Failed means the OSA resource is in failed state.
+	Failed ProvisioningState = "Failed"
+	// Succeeded means the last create/update succeeded.
+	Succeeded ProvisioningState = "Succeeded"
+	// Deleting means the OSA resource is being deleted.
+	Deleting ProvisioningState = "Deleting"
+	// Migrating means the OSA resource is being migrated from one subscription
+	// or resource group to another.
+	Migrating ProvisioningState = "Migrating"
+	// Upgrading means the existing OAS resource is being upgraded.
```
charlesakalugwu: OSA

pkg/api/converttoadmin.go
```diff
@@ -0,0 +1,169 @@
...
+		})
+	}
+
+	oc.Properties.AuthProfile = &admin.AuthProfile{}
+	oc.Properties.AuthProfile.IdentityProviders = make([]admin.IdentityProvider, len(cs.Properties.AuthProfile.IdentityProviders))
+	for i := range cs.Properties.AuthProfile.IdentityProviders {
+		ip := cs.Properties.AuthProfile.IdentityProviders[i]
+		oc.Properties.AuthProfile.IdentityProviders[i].Name = &ip.Name
+		switch provider := ip.Provider.(type) {
+		case (*AADIdentityProvider):
```
charlesakalugwu: The parenthesis are not needed

kargakis:
> Looking good. Please add a new `ValidateAdmin(ctx context.Context, new, old *OpenShiftManagedCluster) []error` function to the plugin interface and implement it. Don't think you'll need `externalOnly bool` for this. The admin API will not be able to create a model (so if old == nil, it's a validation error). For now, ValidateAdmin should reject any requested change to the model.

Working on validation in a different branch to ease reviews. I'd prefer to get this merged and out of the way if there aren't any more comments on conversions.

kargakis:
The reflective test is now ready, pulling this out of wip

kargakis:
sorry, still needs some work

kargakis:
Should be OK now

kargakis:
Validation added in https://github.com/openshift/openshift-azure/pull/821

pkg/api/admin/api/types.go
mjudeikis: I think in the future customer will be able to change these via customerAPI. Question is do we want to have this now here or not.

mjudeikis:
lgtm, will wait for second pair of eyes

PR #801, opened 2018-11-29
Simplify manifest generation

kargakis:
Simplify some of the manifest generation code to reduce code duplication. Also got rid of the `MANIFEST` env since it's not actually used by CI and brought back `-manifest` to be on par with `-update` (although the latter is probably going away soon).


kargakis:
/hold

probably need to revert the MANIFEST removal

kargakis:
@jim-minter @asalkeld @mjudeikis  my initial thought on moving the create manifest into its own directory out of hack/create.sh was to allow testing different CREATE configurations (namely, I was anticipating custom vnet testing). It seems we won't need that after all and I am not sure we will need to test any other configuration on CREATE. I am keen on getting rid of the golang template generation stuff but I am also keen on moving the CREATE manifest inside createorupdate.go vs keeping it in bash or in a text template. It is going to keep hack/create.sh small as it already is. Thoughts?

mjudeikis:
I kinda liked shell script (yes, thorow stone to me) because it was easy to modify and update any manifest values in one place together with execution (shell is most univeral linux glue as much we hate it).
But at the same time Im not locked about it.

kargakis:
> I kinda liked shell script (yes, thorow stone to me) because it was easy to modify and update any manifest values in one place together with execution (shell is most univeral linux glue as much we hate it).
> But at the same time Im not locked about it.

I am not against moving it back to shell, maybe adding back all the -z env checks is not that bad. We would still need to preprocess the manifest in go for adding the generated aad secret since we moved aad processing in go.

kargakis:
The vnet peering test will need to send its own manifest

PR #806, opened 2018-11-29
move secret to absolute path

mjudeikis:
@kargakis @charlesakalugwu  WDYT?

### DIFF
--- a/pkg/fakerp/config.go
+++ b/pkg/fakerp/config.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"path"
 	"time"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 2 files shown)


pkg/fakerp/config.go
```diff
@@ -87,16 +88,17 @@ func getPluginConfig() (*api.PluginConfig, error) {
 	}
 
 	// populate geneva artifacts
-	artifactDir := "secrets/"
-	logCert, err := readCert(artifactDir + "logging-int.cert")
+	gp := os.Getenv("GOPATH")
```
charlesakalugwu: i think this could work but more care needs to be taken in the code for the case where the system has multiple `GOPATH` directories configured

mjudeikis: I think good practice is to have to have one GOPATH, or just switch it between projects if you really need?

kargakis:
Not a big fan of this - I remember being opposed to `SECRETSDIR` but I mainly didn't like the name and the fact that it was made available in `env.example`. I still prefer making this directory configurable via TestConfig.

kargakis:
No, it's not artifacts but it's secrets specific to Geneva so I would probably set something along those lines. `GENEVA_SECRETS_DIR` or `GENEVA_CREDENTIALS_DIR`. @jim-minter thoughts?

mjudeikis:
Its just now. I expect more secrets to come here. Like `registry.io` token, subscription details, etc. And I would not want to start maintaining more complex directory structure `secrets/geneva secrets/rh secrets/stuff`. Same as our CI secrets directory - as flat as my home country :lithuania:

jim-minter:
I'd like fewer environment variables, not more.  Why do we need to do this?

mjudeikis:
> I'd like fewer environment variables, not more. Why do we need to do this?

because it is breaking tests who run not from project root. @charlesakalugwu tests in particual. 
check first commit for alternative.

mjudeikis:
file walker it is
/close

PR #809, opened 2018-11-29
Handle possible errors from os.Setenv

charlesakalugwu:
We should handle possible errors from os.Setenv in `cmd/createorupdate.go`

/cc @mjudeikis @kargakis

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -244,8 +244,14 @@ func updateAadApplication(ctx context.Context, log *logrus.Entry, conf *fakerp.C
 		conf.AADClientSecret = conf.ClientSecret
 	}
 	// set env variable so enrich() still works
-	os.Setenv("AZURE_AAD_CLIENT_ID", conf.AADClientID)
-	os.Setenv("AZURE_AAD_CLIENT_SECRET", conf.AADClientSecret)
+	err := os.Setenv("AZURE_AAD_CLIENT_ID", conf.AADClientID)
+	if err != nil {
+		return err
+	}
+	err = os.Setenv("AZURE_AAD_CLIENT_SECRET", conf.AADClientSecret)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


thekad:
/lgtm

although I am curious what kinda errors `os.Setenv` could throw?

charlesakalugwu:
> /lgtm
> 
> although I am curious what kinda errors `os.Setenv` could throw?

@thekad `os.Setenv` calls `syscall.Setenv` which returns EINVAL if the function arguments are invalid e.g. key is the empty string or the key contains `0` or `=` or the value contains `0`

```
#define EINVAL          22      /* Invalid argument */
```

asalkeld:
> > /lgtm
> > although I am curious what kinda errors `os.Setenv` could throw?
> 
> @thekad `os.Setenv` calls `syscall.Setenv` which returns the linux system error code `EINVAL` if the function arguments are invalid e.g. key is the empty string or the key contains `0` or `=` or the value contains `0`
> 
> ```
> #define EINVAL          22      /* Invalid argument */
> ```

ok, but that seems super unlikely given we are passing in a good key and the only way we are this far is if the value is good. shrug

charlesakalugwu:
@asalkeld @kargakis are you guys now trying to say that we should leave unhandled errors lying around because you think those error situations are rare. I think callers of any function which returns an error must have that error handled.

kargakis:
My take on this part of code is that it doesn't really matter if we aren't 100% defensive since it's not production code so the simpler it is the better.

mjudeikis:
I agree that this is not needed, BUT considering how people well knows AAD flow it might prevent somebody in the future causing errors while refactoring it.
At the moment AAD flow above is built so that both values will be set in any case so these SetEnv command should never get `nil` as input. But if this would change because somebody will change the logic without understanding the consequences - this will potentially prevent "clusterfrackups". 

I would say lets merge it :)

asalkeld:
/lgtm

not a biggie

PR #810, opened 2018-11-29
Enable partial requests to update clusters

kargakis:
Managed to scale up a cluster with the following partial request. @charlesakalugwu we should probably reuse it since that's how scale up requests are going to look like when they land in the prod RP.

/cc jim-minter charlesakalugwu

### DIFF
--- a/test/manifests/normal/scaleup.yaml
+++ b/test/manifests/normal/scaleup.yaml
@@ -1,0 +1,4 @@
+properties:
+  agentPoolProfiles:
+  - name: compute
+    count: 2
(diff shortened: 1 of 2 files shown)


test/manifests/normal/scaleup.yaml
```diff
@@ -0,0 +1,4 @@
+properties:
+  agentPoolProfiles:
+  - name: compute
+    count: 2
```
charlesakalugwu: I see you have one file for the scale up partial request scenario. That means we would need another file for the scale down part in the case of a scale up/down e2e test.

For such a test it seems to me that its simpler to just manipulate the full cluster struct before sending it to the RP. Do you have something different in mind as to how this should operate?

kargakis: > For such a test it seems to me that its simpler to just manipulate the full cluster struct before sending it to the RP. Do you have something different in mind as to how this should operate?

@jim-minter can keep me honest but requests landing on the production RP can be partial hence my latest work on merging (eg. on a scale up/down only the agent pool name and count are needed). We want to simulate that behavior and ensure we support partial requests.

kargakis:
@jim-minter @charlesakalugwu ptal

test/manifests/normal/scaleup.yaml
charlesakalugwu: @kargakis I will integrate the partial updates techniques in my PR. It works for me. My earlier comment was just stating the fact that it would require a little more work from my side.

Issue #814, opened 2018-11-30
should deploy a template with persistent storage timeout


kargakis:
```
• Failure [296.533 seconds]
Openshift on Azure end user e2e tests [EndUser]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:29
  should deploy a template with persistent storage and test failure modes [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:153

  Expected error:
      <*errors.errorString | 0xc0000838b0>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
  not to have occurred

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:238
```
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/798/pull-ci-openshift-openshift-azure-master-e2e-azure/1087/build-log.txt
Pods in the namespace:
<details>

```
metadata:
  annotations:
    openshift.io/build.name: django-psql-persistent-1
    openshift.io/scc: privileged
  creationTimestamp: 2018-11-30T09:55:09Z
  labels:
    openshift.io/build.name: django-psql-persistent-1
  name: django-psql-persistent-1-build
  namespace: e2e-test-4rt82
  ownerReferences:
  - apiVersion: build.openshift.io/v1
    controller: true
    kind: Build
    name: django-psql-persistent-1
    uid: 0590cdfd-f486-11e8-8b10-000d3a149b33
  resourceVersion: "5883"
  selfLink: /api/v1/namespaces/e2e-test-4rt82/pods/django-psql-persistent-1-build
  uid: 05a6a9b5-f486-11e8-8b10-000d3a149b33
spec:
  containers:
  - args:
    - --loglevel=0
    command:
    - openshift-sti-build
    env:
    - name: BUILD
      value: |
[...]

PR #821, opened 2018-12-03
Support admin requests in the fake RP

kargakis:
This PR creates a new endpoint in the fake RP where the admin API is served
and enables handling of admin requests.

@jim-minter @charlesakalugwu 

Part of AZURE-119

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -124,6 +124,9 @@ type Plugin interface {
 	// should be excluded.
 	Validate(ctx context.Context, new, old *OpenShiftManagedCluster, externalOnly bool) []error
 
+	// ValidateAdmin is used for validating admin API requests.
+	ValidateAdmin(ctx context.Context, new, old *OpenShiftManagedCluster) []error
+
 	// GenerateConfig ensures all the necessary in-cluster config is generated
 	// for an Openshift cluster.
 	GenerateConfig(ctx context.Context, cs *OpenShiftManagedCluster) error
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"reflect"
@@ -117,10 +118,15 @@ var (
 // Validator shares objects between validation methods
 type Validator struct {
 	runningUnderTest bool
+	isAdmin          bool
 }
 
 func NewValidator(runningUnderTest bool) *Validator {
 	return &Validator{runningUnderTest: runningUnderTest}
+}
+
+func NewAdminValidator(runningUnderTest bool) *Validator {
+	return &Validator{runningUnderTest: runningUnderTest, isAdmin: true}
 }
 
 func init() {
@@ -209,7 +215,18 @@ func (v *Validator) Validate(new, old *OpenShiftManagedCluster, externalOnly boo
 		return errs
 	}
 	if old != nil {
-		return v.validateUpdateContainerService(new, old, externalOnly)
+		if errs := v.validateUpdateContainerService(new, old, externalOnly); len(errs) > 0 {
+			return errs
+		}
+	}
+	if v.isAdmin {
+		if old == nil {
+			errs = append(errs, errors.New("admin requests cannot create clusters"))
+			return errs
+		}
+		if errs := v.validateUpdateConfig(new.Config, old.Config); len(errs) > 0 {
+			return errs
+		}
 	}
 	return nil
 }
@@ -498,6 +515,13 @@ func (v *Validator) validateProvisioningState(ps ProvisioningState) (errs []erro
 		Upgrading:
 	default:
 		errs = append(errs, fmt.Errorf("invalid properties.provisioningState %q", ps))
+	}
+	return
+}
+
+func (v *Validator) validateUpdateConfig(internalConfig, adminConfig Config) (errs []error) {
+	if !reflect.DeepEqual(internalConfig, adminConfig) {
+		errs = append(errs, fmt.Errorf("invalid change %s", deep.Equal(internalConfig, adminConfig)))
 	}
 	return
 }
--- a/pkg/fakerp/server.go
+++ b/pkg/fakerp/server.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
-	"github.com/openshift/openshift-azure/pkg/api"
 	internalapi "github.com/openshift/openshift-azure/pkg/api"
 	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
 	admin "github.com/openshift/openshift-azure/pkg/api/admin/api"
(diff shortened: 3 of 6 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -289,6 +293,9 @@ func main() {
 		s := fakerp.NewServer(log, conf.ResourceGroup, fakeRpAddr, conf)
 		go s.ListenAndServe()
 	}
+	if *isAdmin {
+		rpURL += "/admin"
```
kargakis: Instead of doing this we should probably create a client in pkg/admin/api and use that here instead of using `v20180930preview.NewOpenShiftManagedClustersClientWithBaseURI` below.

pkg/fakerp/server.go
```diff
@@ -52,6 +53,7 @@ func NewServer(log *logrus.Entry, resourceGroup, address string, c *Config) *Ser
 func (s *Server) ListenAndServe() {
 	// TODO: match the request path the real RP would use
 	http.Handle("/", s)
+	http.Handle("/admin", s)
```
kargakis: Unclear how this endpoint will look like in prod.

kargakis:
@jim-minter @charlesakalugwu @mjudeikis ptal, I have updated the fake RP to support accepting admin requests. Follow-up PRs to add admin validation and enable createorupdate to send admin requests.

kargakis:
> I personally dont like how request path is being implemented here, (`/` vs `/admin`)

@jim-minter and I agreed on a separate endpoint for the admin api, I am open to better suggestions on the paths :)

pkg/fakerp/server.go
```diff
@@ -249,70 +267,121 @@ func (s *Server) handlePut(w http.ResponseWriter, req *http.Request) {
...
+		http.Error(w, resp, http.StatusBadRequest)
+		return
 	}
 
-	if _, err := CreateOrUpdate(ctx, oc, s.log, config); err != nil {
-		s.writeState(v20180930preview.Failed)
-		resp := "400 Bad Request: Failed to apply request"
-		s.log.Debugf("%s: %v", resp, err)
+	// apply the request
+	cs, err = createOrUpdate(ctx, s.log, cs, oldCs, config)
```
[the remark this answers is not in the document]
mjudeikis: What is the long run plan? Is the plan to "abstract" all `Update` queries behind `createOrUpdate` function, or we will split those to smaller ones if needed? In example `MigrateDisk` ?

mjudeikis:
> > I personally dont like how request path is being implemented here, (`/` vs `/admin`)
> 
> @jim-minter and I agreed on a separate endpoint for the admin api, I am open to better suggestions on the paths :)

separate endpoint is must, but each time I see something like:
```
isAdminRequest := strings.HasPrefix(req.URL.Path, "/admin")
```
I think about "path router" for http. But as I say, 2 Paths - fine. when we start doing this for 4 - we might need to think about router

kargakis:
Oh yeah, that can be a follow-up enhancement.

kargakis:
connection timeout

/retest

pkg/fakerp/server.go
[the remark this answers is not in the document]
kargakis: I think we are going to have separate functions to describe common flows like key rotation or backup+restore.

pkg/fakerp/config.go
```diff
@@ -76,7 +76,7 @@ func NewConfig(log *logrus.Entry, needRegion bool) (*Config, error) {
 	return &c, nil
 }
 
-func GetPluginConfig() (*api.PluginConfig, error) {
+func GetPluginConfig(isAdmin bool) (*api.PluginConfig, error) {
 	tc := api.TestConfig{
 		RunningUnderTest:      os.Getenv("RUNNING_UNDER_TEST") != "",
```
charlesakalugwu: do you need to rebase? this line is `os.Getenv("RUNNING_UNDER_TEST") == "true"` in master

kargakis: Not necessary, my branch is not on top of HEAD but there is no conflict.

kargakis:

/retest

kargakis:
https://github.com/openshift/openshift-azure/issues/863

/retest

pkg/api/plugin.go
```diff
@@ -73,8 +73,8 @@ type PluginConfig struct {
 	SyncImage       string
 	AcceptLanguages []string
 	GenevaConfig    GenevaConfig
-
-	TestConfig TestConfig
+	IsAdmin         bool
```
jim-minter: why is this here?  surely it shouldn't be?

test/clients/azure/osa.go
```diff
@@ -44,13 +43,12 @@ func (cli *Client) UpdateOSACluster(ctx context.Context, external *v20180930prev
...
-	logger := logrus.NewEntry(logrus.StandardLogger())
 
 	var oc *v20180930preview.OpenShiftManagedCluster
 	var err error
 	// simulate the API call to the RP
-	if oc, err = fakerp.CreateOrUpdate(ctx, external, logger, config); err != nil {
-		return nil, err
-	}
-	return oc, nil
+	//if oc, err = fakerp.CreateOrUpdate(ctx, external, logger, config); err != nil {
```
jim-minter: why is this commented out?  is a middle ground necessary before this entire function gets deleted?

kargakis: Waiting on https://github.com/openshift/openshift-azure/pull/880 to merge, then I am going to fix this.

pkg/api/plugin.go
kargakis: Hrm, shouldn't be here - I have split validation into its own PR, see https://github.com/openshift/openshift-azure/pull/854

pkg/fakerp/server.go
```diff
@@ -19,9 +18,12 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
+	"github.com/openshift/openshift-azure/pkg/api"
 	internalapi "github.com/openshift/openshift-azure/pkg/api"
```
charlesakalugwu: remove one of the two imports

pkg/fakerp/server.go
```diff
@@ -126,44 +118,61 @@ func (s *Server) validate(w http.ResponseWriter, r *http.Request) bool {
...
+// be consistent with how the RP runs in production so we need
+// to restore the internal state of the cluster from the
+// filesystem.
+func (s *Server) restore(path string) error {
+	dataDir, err := FindDirectory(path)
+	if err != nil {
+		return err
 	}
-	data, err := ioutil.ReadFile(path)
+	data, err := ioutil.ReadFile(dataDir)
```
charlesakalugwu: you should read the containerservice.yaml and not path (which is most likely _data)

pkg/api/validate.go
```diff
@@ -501,3 +518,10 @@ func (v *Validator) validateProvisioningState(ps ProvisioningState) (errs []erro
 	}
 	return
 }
+
+func (v *Validator) validateUpdateConfig(internalConfig, adminConfig Config) (errs []error) {
+	if !reflect.DeepEqual(internalConfig, adminConfig) {
```
charlesakalugwu: wouldn't this be a problem for key rotation which needs to have certain fields in the Config invalidated and regenerated before using that in the update?

kargakis: Yes, we agreed that in the first cut of this, we should invalidate every change. Your changes should enable validating certificate updates.

pkg/fakerp/server.go
kargakis: good catch, fixed

pkg/fakerp/server.go
kargakis: fixed

PR #829, opened 2018-12-04
metrics add health and ready

mjudeikis:
We need some kind of very basic health check for this container. What do people think?

### DIFF
--- a/cmd/metricsbridge/metricsbridge.go
+++ b/cmd/metricsbridge/metricsbridge.go
@@ -46,6 +46,8 @@ type config struct {
 	rootCAs *x509.CertPool
 	http    *http.Client
 	statsd  *statsd.Client
+
+	ready bool
 }
 
 func (c *config) load(path string) error {
@@ -113,7 +115,9 @@ func (c *config) init() error {
 			},
 		},
 	}
-
+	// not ready by default
+	c.ready = false
+	go c.health()
 	return nil
 }
 
@@ -178,9 +182,10 @@ func (c *config) runOnce(req *http.Request) error {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
+		c.ready = false
 		return fmt.Errorf("prometheus returned status code %d", resp.StatusCode)
 	}
-
+	c.ready = true
 	d := expfmt.NewDecoder(resp.Body, expfmt.FmtText)
 
 	for {
@@ -214,6 +219,25 @@ func (c *config) runOnce(req *http.Request) error {
 	}
 
 	return c.statsd.Flush()
+}
+
+func (c *config) health() {
+	c.log.Debug("starting health endpoints")
+	http.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	},
+	)
+	http.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
+		if c.ready {
+			// 200
+			w.WriteHeader(http.StatusOK)
+		} else {
+			// 503
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+	},
+	)
+	http.ListenAndServe(":8080", nil)
 }
 
 func main() {


asalkeld:
seems ok to me

cmd/metricsbridge/metricsbridge.go
```diff
@@ -216,6 +221,25 @@ func (c *config) runOnce(req *http.Request) error {
 	return c.statsd.Flush()
 }
 
+func (c *config) health() {
+	c.log.Debug("starting health endpoints")
+	http.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
```
kargakis: /healthz

cmd/metricsbridge/metricsbridge.go
```diff
@@ -46,6 +46,8 @@ type config struct {
 	rootCAs *x509.CertPool
 	http    *http.Client
 	statsd  *statsd.Client
+
+	ready bool
```
kargakis: since /health can be accessed in parallel, you should put this behind a RWLock.

kargakis:
I am wondering what's the intention here; what's the problem with considering the metricsbridge ready even if the federated prometheus is not? Also, are there any plans to loadbalance the metrics bridge?

mjudeikis:
> I am wondering what's the intention here; what's the problem with considering the metricsbridge ready even if the federated prometheus is not? Also, are there any plans to loadbalance the metrics bridge?

If Prometheus is not ready metrics bridge does not have anything to query. And this might be the issue with networking, dns or genuinely Prometheus is down. This would give us one more possible place to monitor interconnected components and potentially detect possible issues is some occurs (I hope not). 

For now, loadbalancing is not possible as metrics bridge scapes and pushes, so we at best can make it Active-Passive with some leader election.

kargakis:
> If Prometheus is not ready metrics bridge does not have anything to query. And this might be the issue with networking, dns or genuinely Prometheus is down. This would give us one more possible place to monitor interconnected components and potentially detect possible issues is some occurs (I hope not).

So why would you turn the metricsbridge pod to Unready in that case? As far as it is concerned, it works fine. Wouldn't that be extra noise? (Think "I have both Prometheus and metricsbridge failing but it's not obvious that metricsbridge is failing because of Prom or because of its own issue".)

> For now, loadbalancing is not possible as metrics bridge scapes and pushes, so we at best can make it Active-Passive with some leader election.

So we don't really care when a metricsbridge pod turns to ready.

mjudeikis:
> So why would you turn the metricsbridge pod to Unready in that case? As far as it is concerned, it works fine. Wouldn't that be extra noise? (Think "I have both Prometheus and metricsbridge failing but it's not obvious that metricsbridge is failing because of Prom or because of its own issue".)

Yes, we might need to add something similar to monitor statsd container too from metricsbridge. As it can be ready only when both containers are alive. 
Use case:
1. Prometheus is failing -> metrics bridge is not ready 
2. Statsd is failing -> metrics bridge is not ready
3. Metricsbridge is not healthy - k8s restarts the pod. 

I just want to get notified when for let's say network segregation between nodes Prometheus is healthy, but metrics bridge is not able to reach it. So its not ready, and I should get alert via logging integration (or any other ways to do this down the stream). Its an extra noise we can choose to ignore if we want. But if we need it - its there.

kargakis:
> I just want to get notified when for let's say network segregation between nodes Prometheus is healthy, but metrics bridge is not able to reach it. So its not ready, and I should get alert via logging integration (or any other ways to do this down the stream). Its an extra noise we can choose to ignore if we want. But if we need it - its there.

I still don't understand why this alert has to go through monitoring a different component than the one you want to monitor. Doesn't prometheus expose metrics about the responses it sends back to clients? Also, in the particular case where you want to monitor clients, turning the client to Unready is a smelly design vs exposing the response status in a metric because you are masking other potential errors that the client might have.

mjudeikis:
> I still don't understand why this alert has to go through monitoring a different component than the one you want to monitor. Doesn't prometheus expose metrics about the responses it sends back to clients? Also, in the particular case where you want to monitor clients, turning the client to Unready is a smelly design vs exposing the response status in a metric because you are masking other potential errors that the client might have.

I little to disagree with smelliness of this. Overall I think if microservice is not ready due problems of itself or its upstream dependency - it should be marked unhealthy/unready. Any components not doing its job have to be marked appropriately, and it does not matter the reasons, reasons are business logic related.

In this case, it's tightly connected into the cluster Prometheus. It is a question of avoiding cascading failures. Its not a case in this one, but let's say if we make any other component dependent on `metrics-bridge`, we want other components to know if it healthy. 

In this hypothetical situation if Prometheus goes down, or "hangs" and our metrics bridge just keeps "ready", but does not get any metrics to forward to its downstream. From the observer in the downstream, no metrics might look like a healthy situation (nothing is running and no metrics is sent, so nothing to alert on) when the of the components is in the failure state. On the other hand, if downstream knows that `metrics-bridge` is not ready, it can mark itself appropriately or raise an alert.

kargakis:
> In this hypothetical situation if Prometheus goes down, or "hangs" and our metrics bridge just keeps "ready", but does not get any metrics to forward to its downstream. From the observer in the downstream, no metrics might look like a healthy situation (nothing is running and no metrics is sent, so nothing to alert on) when the of the components is in the failure state. On the other hand, if downstream knows that `metrics-bridge` is not ready, it can mark itself appropriately or raise an alert.

How is downstream going to know that metricsbridge is Unready? I assume the alert manager would have to use different sources of metrics and not only Prometheus (since Prometheus is down in your scenario). At that point, why not directly alert that Prometheus is down?

jim-minter:
@mjudeikis I too do not see the point to this PR.  Recommend closing.

We will report a freshness metric from Prometheus to Geneva so that we can detect problems in the monitoring stack.

mjudeikis:
:(

PR #836, opened 2018-12-04
Remove colons from addons

sozercan:
Looks like a file got renamed and a new file was added with colons, which broke windows dev envs. Removed bootstrap-signer (same exists with dashes), renamed cloud-provider and generated bindata.

cc @jim-minter @mjudeikis


jim-minter:
/hold

need a unit test to prevent this happening again before we merge

jim-minter:
added test

/hold cancel

/lgtm

jim-minter:
thanks @sozercan

PR #818, opened 2018-12-03
Fix Md5Mismatch error during update in place

asalkeld:
fixes #812

PR #838, opened 2018-12-05
Follow-up to https://github.com/openshift/openshift-azure/pull/818

jim-minter:
Avoid storing the Blob object in the simpleUpgrader, evidently this was a
mistake.  On a Blob.Get() the blob MD5 is latched into the Blob object.  If
Blob.CreateBlockBlobFromReader() is called, the latched MD5 is not cleared,
resulting in an error on the subsequent Get().  Store the Container in the
simpleUpgrader to force the user to get the Blob themselves.

### DIFF
--- a/pkg/cluster/initialize.go
+++ b/pkg/cluster/initialize.go
@@ -30,12 +30,11 @@ func (u *simpleUpgrader) initialize(ctx context.Context, cs *api.OpenShiftManage
 	}
 
 	// update tracking container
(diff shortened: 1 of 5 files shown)


jim-minter:
@asalkeld @kargakis ptal - sorry for screwing this up.

asalkeld:
looks fine, just waiting for the e2e test

mjudeikis:
lgtm

Issue #834, opened 2018-12-04
Logging namespace needs to be configurable


sozercan:
For Geneva logging, realized that we'll need to make `namespace` and account `moniker` configurable from the RP side.
https://github.com/openshift/openshift-azure/blob/master/pkg/addons/data/ConfigMap/openshift-azure-logging/mdsd-customerdata-config.yaml#L9-L11

For quick short term fix, can we change these to `OpenShift` and `openshiftdiag` on a new branch, so we can vendor without all the new changes?

PR #840, opened 2018-12-05
logging add namespace and account

mjudeikis:
Adds namespace and account to geneva config for logging 

fix: https://github.com/openshift/openshift-azure/issues/834


mjudeikis:
@sozercan https://github.com/Azure/moffet/pull/6 this need to get in first and new image built on MSFT side, so we could update image reference too.

/hold

mjudeikis:
/hold 
need split for PP2 :/

mjudeikis:
Should be fine now

PR #844, opened 2018-12-05
Set RUNNING_UNDER_TEST only when true

kargakis:
Should allow us to set it to false explicitly in [tests](https://github.com/openshift/release/pull/2309).

### DIFF
--- a/pkg/fakerp/config.go
+++ b/pkg/fakerp/config.go
@@ -78,7 +78,7 @@ func NewConfig(log *logrus.Entry, needRegion bool) (*Config, error) {
 
 func GetPluginConfig() (*api.PluginConfig, error) {
 	tc := api.TestConfig{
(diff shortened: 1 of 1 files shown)


kargakis:
@stevekuznetsov @bbguimaraes @petr-muller @droslean the e2e-no-test job was just added as [an optional job](https://github.com/openshift/release/blob/165c469c1ed9f3b583f0ea3f391649e87ec72159/ci-operator/jobs/openshift/openshift-azure/openshift-openshift-azure-master-presubmits.yaml#L189), any idea why it runs by default?

stevekuznetsov:
@kargakis I messed up on a new component that makes sure that new blocking presubmits are triggered -- it is too eager 

https://github.com/kubernetes/test-infra/pull/10348

PR #854, opened 2018-12-06
Add admin API validation

kargakis:
Add admin api validation. Needs to be plugged into the fake RP after https://github.com/openshift/openshift-azure/pull/821 merges.

Part of AZURE-119

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -124,6 +124,9 @@ type Plugin interface {
 	// should be excluded.
 	Validate(ctx context.Context, new, old *OpenShiftManagedCluster, externalOnly bool) []error
 
+	// ValidateAdmin is used for validating admin API requests.
+	ValidateAdmin(ctx context.Context, new, old *OpenShiftManagedCluster) []error
+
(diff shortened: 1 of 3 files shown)


pkg/api/plugin.go
```diff
@@ -73,8 +73,8 @@ type PluginConfig struct {
 	SyncImage       string
 	AcceptLanguages []string
 	GenevaConfig    GenevaConfig
-
-	TestConfig TestConfig
+	IsAdmin         bool
```
jim-minter: I really don't think it can be right to have `IsAdmin bool` as well as a separate ValidateAdminUpdate() entrypoint in the interface.
I'd rather remove `IsAdmin bool`, add a ValidateAdmin() in the interface instead of ValidateAdminUpdate(), have that function call `func (v *Validator) ValidateAdmin`, which can then reuse validateContainerService and have its own copy validateAdminUpdateContainerService, as you've implemented

kargakis:
@jim-minter updated, ptal

kargakis:
Cherrypicked in https://github.com/openshift/openshift-azure/pull/821

Issue #853, opened 2018-12-06
dns conflict


kargakis:
```
time="2018-12-06T08:33:37Z" level=debug msg="400 Bad Request: Failed to apply request: dns.ZonesClient#CreateOrUpdate: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> <nil>"
```
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/825/pull-ci-openshift-openshift-azure-master-e2e-azure-etcdbackuprecovery/8/build-log.txt
/kind test-flake

jim-minter:
Activity log on the dns RG in the portal showed that zones weren't being created because we hit our quota (100).  I've deleted the following zones:
```
ci-op-0lym3lpm-2e2aa-201811161500-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811162017-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811172019-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811182018-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811190940-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811192020-e2e.osadev.cloud
ci-op-0lym3lpm-2e2aa-201811202022-e2e.osadev.cloud
ci-op-0lym3lpm-37b6c-201811172218-e2e.osadev.cloud
ci-op-0lym3lpm-37b6c-201811182219-e2e.osadev.cloud
ci-op-0lym3lpm-37b6c-201811192222-e2e.osadev.cloud
ci-op-0my9brsv-5d5fb.osadev.cloud
ci-op-1tpvhtwq-8474e.osadev.cloud
ci-op-2ltttb51-5d5fb.osadev.cloud
ci-op-54qr6nrv-f1aa2-201812061128-e2e.osadev.cloud
ci-op-8kcz33w0-8474e.osadev.cloud
ci-op-9933fnq7-48b5d-201812011404-e2e.osadev.cloud
ci-op-9933fnq7-48b5d-201812021405-e2e.osadev.cloud
ci-op-9933fnq7-48b5d-201812031406-e2e.osadev.cloud
ci-op-9933fnq7-99ef1-201811291644-e2e.osadev.cloud
ci-op-9933fnq7-99ef1-201811301650-e2e.osadev.cloud
[...]

PR #858, opened 2018-12-06
add dns logs

mjudeikis:
Add logging to DNS to confirm we create right records

@jim-minter @kwoodson @kargakis 

I suspect flakes might be due random-regions in the CI and DNS names getting the wrong payload.


jim-minter:
Not convinced this is necessary.  See https://github.com/openshift/openshift-azure/issues/853 for an update on the DNS trouble.

Issue #850, opened 2018-12-06
  should start kube-state-metrics correctly


kargakis:
```
Openshift on Azure admin e2e tests [AzureClusterReader] 
  should start kube-state-metrics correctly
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/fakerp/admin.go:66
panic: test timed out after 20m0s

goroutine 637 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:1296 +0xfd
created by time.goFunc
	/usr/local/go/src/time/sleep.go:172 +0x44

goroutine 1 [chan receive, 20 minutes]:
testing.(*T).Run(0xc00017eb00, 0x12a6d92, 0x7, 0x1335028, 0x481296)
	/usr/local/go/src/testing/testing.go:879 +0x37a
testing.runTests.func1(0xc00017ea00)
	/usr/local/go/src/testing/testing.go:1119 +0x78
testing.tRunner(0xc00017ea00, 0xc0004d3e08)
	/usr/local/go/src/testing/testing.go:827 +0xbf
testing.runTests(0xc000186980, 0x1f793e0, 0x1, 0x1, 0x40d8df)
	/usr/local/go/src/testing/testing.go:1117 +0x2aa
testing.(*M).Run(0xc000160180, 0x0)
	/usr/local/go/src/testing/testing.go:1034 +0x165
main.main()
	_testmain.go:42 +0x13d

goroutine 5 [syscall, 20 minutes]:
os/signal.signal_recv(0x0)
	/usr/local/go/src/runtime/sigqueue.go:139 +0x9c
os/signal.loop()
	/usr/local/go/src/os/signal/signal_unix.go:23 +0x22
created by os/signal.init.0
	/usr/local/go/src/os/signal/signal_unix.go:29 +0x41

goroutine 6 [chan receive]:
github.com/openshift/openshift-azure/vendor/github.com/golang/glog.(*loggingT).flushDaemon(0x1fa05c0)
	/go/src/github.com/openshift/openshift-azure/vendor/github.com/golang/glog/glog.go:879 +0x8b
created by github.com/openshift/openshift-azure/vendor/github.com/golang/glog.init.0
	/go/src/github.com/openshift/openshift-azure/vendor/github.com/golang/glog/glog.go:410 +0x203

goroutine 12 [chan receive]:
[...]

PR #860, opened 2018-12-07
Use the new FindDirectory in e2e tests

asalkeld:
seems better than hardcoded path.


kargakis:
https://github.com/openshift/openshift-azure/issues/850

/retest

PR #862, opened 2018-12-07
package names added by goimports

Makdaam:
Result of 
goimports -e -w -local=github.com/openshift/openshift-azure pkg cmd test
required for make verify to go through.

### DIFF
--- a/cmd/metricsbridge/metricsbridge.go
+++ b/cmd/metricsbridge/metricsbridge.go
@@ -16,7 +16,7 @@ import (
 	"time"
 
 	"github.com/ghodss/yaml"
-	"github.com/prometheus/client_model/go"
+	io_prometheus_client "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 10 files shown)


Makdaam:
Please ignore for now, trying to find out what's wrong with my goimports settings.

mjudeikis:
/hold
Just in case :) this is not needed as per current content

Makdaam:
Resolved the issue with my goimports.
Checked out commit 04b5d21 of goimports to match @mjudeikis's version, and rebuilt it.

mjudeikis:
Do we know if this is "new feature" or "bug"?

Makdaam:
@mjudeikis There's a long discussion here https://github.com/golang/go/issues/28428 and it seems like it went in as a feature of new goimports.

kargakis:
The very next moment we rebuild our test base image, we are going to hit this issue in CI. We should probably pin goimports in the test-base image into that commit for now.

kargakis:
https://github.com/openshift/release/blob/87bee64d8fb540296a0d7b479e58065cfb88b38c/projects/azure/base-images/test-base.yaml#L64

mjudeikis:
> The very next moment we rebuild our test base image, we are going to hit this issue in CI. We should probably pin goimports in the test-base image into that commit for now.

Good call! @Makdaam can you do this?

Makdaam:
I'll take it.

PR #872, opened 2018-12-10
Move updateblob code into a seperate package

asalkeld:
This is another step in reducing the size of cluster to make it more testable.


jim-minter:
as with pkg/util/kubeclient, I think there is a strong argument that this would be better as a subpackage of pkg/cluster?

asalkeld:
this is too big, i am going to split it up

PR #873, opened 2018-12-10
Add unit test for HealthCheck

asalkeld:


pkg/cluster/healthcheck.go
```diff
@@ -10,35 +10,36 @@ import (
...
 			TLSClientConfig: &tls.Config{
 				RootCAs: pool,
 			},
 		},
 		Timeout: 10 * time.Second,
 	}
+}
 
-	req, err := http.NewRequest("HEAD", "https://"+cs.Properties.FQDN+"/console/", nil)
+func (u *simpleUpgrader) doHealthCheck(ctx context.Context, cli wait.SimpleHTTPClient, uri string, sleepDuration time.Duration) *api.PluginError {
```
kargakis: Feels like the client should be part of simpleUpgrader.

Issue #869, opened 2018-12-07
AZURE_REGION is not shared appropriately in e2e tests


kargakis:
The problem is that ever since we stopped hardcoding the region to eastus, the e2e tests that need access to AZURE_REGION, lost it (production RP tests). We write the external manifest down after the cluster is deployed so potentially if the AZURE_REGION variable is unset in the e2e framework we can read the location from the manifest.

/area tests
/kind bug
/assign

kargakis:
Taking into account https://github.com/openshift/openshift-azure/issues/868, we should do this seamlessly w/o MSFT having to use the external manifest themselves.

PR #874, opened 2018-12-10
Set region from the external manifest in e2e tests if unset

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/869

This will allow us to continue using random region selection in our tests.


kargakis:
@thekad the production job is back to green with this fix. Once it merges, we can test your PR.

Issue #876, opened 2018-12-10
Cannot merge partial requests in production


kargakis:
I am trying to scale up a production cluster in `mkargakiprod` to 2 compute nodes, sending the following manifest:
```yaml
location: westeurope
properties:
  agentPoolProfiles:
  - name: compute
    count: 2
```
and I am getting the following error:
```
FATA[2018-12-10T12:31:04+01:00] containerservice.OpenShiftManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="An error has occurred in subscription 225e02bc-43d0-43d1-a01a-17e584a4ef69, resourceGroup: mkargakiprod request: [invalid role \"\" in properties.agentPoolProfiles[\"compute\"] invalid properties.agentPoolProfiles[\"compute\"].subnetCidr \"\" invalid properties.agentPoolProfiles[\"compute\"].osType \"\" missing role \"compute\" in properties.agentPoolProfiles missing role \"infra\" in properties.agentPoolProfiles missing role \"master\" in properties.agentPoolProfiles]." 
```
which means that validation is run on the request before merging it with the existing manifest. Is this expected? @amanohar @sozercan @jim-minter

kargakis:
@charlesakalugwu fyi

kargakis:
This is actually because my changes in the merge code are yet to be vendored in the real RP. Would be nice to be able to identify via an API which commit is vendored by the real RP.

/close

PR #877, opened 2018-12-10
Set prod flag correctly in hack/delete.sh 

kargakis:


test/manifests/normal/scaleup.yaml
```diff
@@ -1,3 +1,4 @@
+location: "{{ .Env.AZURE_REGION }}"
```
kargakis: https://github.com/openshift/openshift-azure/issues/879

charlesakalugwu: just ran into this :)

kargakis: Note that scale up with a partial request is still not going to work in prod, see https://github.com/openshift/openshift-azure/issues/876.

Issue #552, opened 2018-10-03
WaitForCompletion: context has been cancelled  flake


kargakis:
Valid test flake
```
time="2018-10-03T14:24:31Z" level=info msg="write helpers" resourceGroup=ci-op-d9wxcrm6-8474e
time="2018-10-03T14:24:31Z" level=info msg="applying arm template deployment" resourceGroup=ci-op-d9wxcrm6-8474e
time="2018-10-03T14:24:33Z" level=info msg="waiting for arm template deployment to complete" resourceGroup=ci-op-d9wxcrm6-8474e
time="2018-10-03T14:39:33Z" level=fatal msg="Future#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded"
```
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/387/pull-ci-openshift-openshift-azure-master-e2e-azure/152/build-log.txt
/kind test-flake

kargakis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/624/pull-ci-openshift-openshift-azure-master-e2e-azure/463?log#log

kargakis:
We are seeing a lot of different timeouts

Latest
```
time="2018-11-20T12:57:21Z" level=info msg="waiting for arm template deployment to complete" resourceGroup=ci-op-5fy5djqy-8474e
time="2018-11-20T13:01:01Z" level=debug msg="400 Bad Request: Failed to apply request: Initialize: Put https://4paisvz0oxrjp2fpxcv6adek.blob.core.windows.net/etcd?restype=container: dial tcp 52.239.214.132:443: i/o timeout" resourceGroup=ci-op-5fy5djqy-8474e
```
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/758/pull-ci-openshift-openshift-azure-master-e2e-azure/835/build-log.txt

@amanohar

kargakis:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/758/pull-ci-openshift-openshift-azure-master-e2e-azure/833/build-log.txt
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/758/pull-ci-openshift-openshift-azure-master-e2e-azure/832/build-log.txt
[...]

PR #880, opened 2018-12-10
Enable e2e to run against either fake rp or real rp.

charlesakalugwu:
Also add support for the admin api

/cc @kargakis

### DIFF
--- a/pkg/fakerp/server.go
+++ b/pkg/fakerp/server.go
@@ -19,10 +19,12 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
-	"github.com/openshift/openshift-azure/pkg/api"
+	internalapi "github.com/openshift/openshift-azure/pkg/api"
 	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
 )
+
+var once sync.Once
 
 type Server struct {
 	// the server will not process more than a single
@@ -47,6 +49,18 @@ func NewServer(log *logrus.Entry, resourceGroup, address string, c *Config) *Ser
 		address:    address,
 		conf:       c,
 	}
+}
+
+// StartServer starts the fake rp server, ensuring that is done only once. It returns the
+// address to the running http server
+func StartServer(log *logrus.Entry, conf *Config, address string) string {
+	// once ensures that the server start is invoked only once
+	once.Do(func() {
+		log.Info("starting the fake resource provider")
+		s := NewServer(log, conf.ResourceGroup, address, conf)
+		go s.ListenAndServe()
+	})
+	return "http://" + address
 }
 
 func (s *Server) ListenAndServe() {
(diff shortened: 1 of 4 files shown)


pkg/fakerp/server.go
```diff
@@ -19,11 +19,13 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
-	"github.com/openshift/openshift-azure/pkg/api"
-	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	internalapi "github.com/openshift/openshift-azure/pkg/api"
+	externalapi "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
```
kargakis: Revert this change, I am adding support for the admin api which is also external.

charlesakalugwu:
https://github.com/openshift/openshift-azure/issues/552

test/clients/azure/client.go
```diff
@@ -19,12 +35,41 @@ func NewClientFromEnvironment() (*Client, error) {
...
+	focus := []byte(config.GinkgoConfig.FocusString)
+	switch {
+	case adminRpFocus.Match(focus), fakeRpFocus.Match(focus):
+		fmt.Println("configuring the fake resource provider")
+		rpURL = realfakerp.StartServer(log, conf, realfakerp.LocalHttpAddr)
+	case realRpFocus.Match(focus):
+		fmt.Println("configuring the real resource provider")
+		rpURL = externalapi.DefaultBaseURI
+	default:
+		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\]", config.GinkgoConfig.FocusString))
```
jim-minter: doesn't this imply the Makefile needs to be updated?

test/clients/azure/client.go
```diff
@@ -19,12 +35,41 @@ func NewClientFromEnvironment() (*Client, error) {
...
+
+	conf, err := realfakerp.NewConfig(log, true)
+	if err != nil {
+		return nil, err
+	}
+
+	var rpURL string
+	focus := []byte(config.GinkgoConfig.FocusString)
+	switch {
+	case adminRpFocus.Match(focus), fakeRpFocus.Match(focus):
```
jim-minter: the reason this is succeeding without the Makefile being updated is that this is matching the Admin in "[CustomerAdmin]" - I'd have expected you would be looking for "[someTag]" in square brackets and updating the Makefile?

jim-minter:
this can go in now for expediency, but longer term I would prefer that the fake rp ran in a separate process which was set up via the test scripts, and the tests took in an argument which allowed overriding the RP endpoint to communicate with.

test/clients/azure/client.go
kargakis: @charlesakalugwu please raise an issue about this

charlesakalugwu: #890

PR #882, opened 2018-12-10
Add documentation on e2e inputs and running with container image

### LINKED ISSUE #868: documentation on e2e framework input expectations
Please add short documentation which indicates what the e2e framework input expectations are.

I.E.: for real RP tests ([Real]), these env vars:

AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_TENANT_ID (user authentication)
AZURE_SUBSCRIPTION_ID, AZURE_REGION, RESOURCEGROUP (existing cluster location)

for everything else:

_data/containerservice.yaml should be populated with the cluster manifest

charlesakalugwu:
Please let me know if I am missing something.

closes #868 

/cc @kargakis

### DIFF
--- a/docs/e2e/README.md
+++ b/docs/e2e/README.md
@@ -1,0 +1,83 @@
+#### Running e2e tests locally with go test
+
+If you are developing tests locally, you'd probably want to use native `go test` to invoke your ginkgo test specs.
+
+Example: Run the end user e2e tests in a container
+```
+source ./env
+export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
+export TIMEOUT=20m
+./hack/e2e.sh
+```
+
+As a shorthand to the above, you could simply run `make e2e`
+
+Example: Run the scale up/down e2e tests against the fake rp in a container
+```
+source ./env
+export FOCUS="\[ScaleUpDown\]\[Fake\]"
+export TIMEOUT=30m
+export EXTRA_FLAGS="-scaleUpManifest=/path/to/scaleup.yaml -scaleDownManifest=/path/to/scaledown.yaml"
+./hack/e2e.sh
+```
+
+#### Running e2e tests with the container image
+
+To run the e2e tests in a container use the [e2e](https://quay.io/repository/openshift-on-azure/e2e-tests) container 
+image
+
+```
+docker pull quay.io/openshift-on-azure/e2e-tests:latest
+```
+
+Example: Run the end user e2e tests in a container
+```
+export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
+export TIMEOUT=20m
+source ./env
+
+docker run \
+--rm \
+-it \
+--volume $PWD/_data:/_data \
+--volume $PWD/artifacts:/artifacts \
+--volume $PWD/secrets:/secrets \
+--volume $HOME/.kube/config:/root/.kube/config \
+quay.io/openshift-on-azure/e2e-tests \
+-test.v \
+-test.timeout=${TIMEOUT} \
+-ginkgo.noColor \
+-ginkgo.v \
+-ginkgo.focus=${FOCUS:-} \
+-artifact-dir=/artifacts
+```
+
+Example: Run the scale up/down e2e tests against the fake rp in a container
+```
+export FOCUS="\[ScaleUpDown\]\[Fake\]"
+export TIMEOUT=30m
+source ./env
+
+docker run \
+--rm \
+-it \
+--env AZURE_CLIENT_ID=${AZURE_CLIENT_ID} \
+--env AZURE_CLIENT_SECRET=${AZURE_CLIENT_SECRET} \
+--env AZURE_SUBSCRIPTION_ID=${AZURE_SUBSCRIPTION_ID} \
+--env AZURE_TENANT_ID=${AZURE_TENANT_ID} \
+--env RESOURCEGROUP=${RESOURCEGROUP} \
+--volume $PWD/_data:/_data \
+--volume $PWD/artifacts:/artifacts \
+--volume $PWD/secrets:/secrets \
+--volume $PWD/test/manifests:/manifests \
+--volume $HOME/.kube/config:/root/.kube/config \
+quay.io/openshift-on-azure/e2e-tests \
+-test.v \
+-test.timeout=${TIMEOUT} \
+-ginkgo.noColor \
+-ginkgo.v \
+-ginkgo.focus=${FOCUS:-} \
+-artifact-dir=/artifacts \
+-scaleUpManifest=/manifests/normal/scaleup.yaml \
+-scaleDownManifest=/manifests/normal/scaledown.yaml
+```
--- a/docs/e2e/README.md
+++ b/docs/e2e/README.md
@@ -1,23 +1,22 @@
 #### Running e2e tests locally with go test
 
-If you are developing tests locally, you'd probably want to use native `go test` to invoke your ginkgo test specs.
+If you are developing tests locally, you can use the native `go test` to invoke your ginkgo test specs.
 
-Example: Run the end user e2e tests in a container
+Example: Run the end user e2e tests
 ```
 source ./env
 export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
-export TIMEOUT=20m
 ./hack/e2e.sh
 ```
 
 As a shorthand to the above, you could simply run `make e2e`
 
-Example: Run the scale up/down e2e tests against the fake rp in a container
+Example: Run the scale up/down e2e tests against the fake rp
 ```
 source ./env
 export FOCUS="\[ScaleUpDown\]\[Fake\]"
 export TIMEOUT=30m
-export EXTRA_FLAGS="-scaleUpManifest=/path/to/scaleup.yaml -scaleDownManifest=/path/to/scaledown.yaml"
+export EXTRA_FLAGS="-scaleUpManifest=test/manifests/normal/scaleup.yaml -scaleDownManifest=test/manifests/normal/scaledown.yaml"
 ./hack/e2e.sh
 ```
 
--- a/docs/e2e/requirements.md
+++ b/docs/e2e/requirements.md
@@ -1,0 +1,88 @@
+#### Supported test focuses
+
+The following are ginkgo test focuses supported by openshift on azure e2e tests
+
+* `[EveryPR]` - These tests should be run on every PR
+* `[LongRunning]` - These tests are long running and should be run periodically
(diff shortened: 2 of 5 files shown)


docs/e2e/requirements.md
```diff
@@ -0,0 +1,38 @@
...
+| `AZURE_TENANT_ID` | authentication |
+| `AZURE_SUBSCRIPTION_ID` | existing cluster |
+| `AZURE_REGION` | existing cluster location |
+| `RESOURCEGROUP` | existing cluster group |
+
+The following are the files required for running the e2e tests against the fake RP
+
+| Name | Notes |
+|--- | --- |
+| `_data/containerservice.yaml` | allows the test to determine if it is an update or not |
```
kargakis: Hrm, if this is needed only for determining whether we are on an update vs create, I would try to think how to get rid of it.

charlesakalugwu: at a minimum it is required to discriminate against an update vs create. However, it is also needed for certain admin-oriented, update-based e2e tests such as key rotation.

docs/e2e/requirements.md
```diff
@@ -0,0 +1,38 @@
...
+| Name | Notes |
+|--- | --- |
+| `AZURE_CLIENT_ID` | authentication |
+| `AZURE_CLIENT_SECRET` | authentication |
+| `AZURE_TENANT_ID` | authentication |
+| `AZURE_SUBSCRIPTION_ID` | existing cluster |
+| `AZURE_REGION` | existing cluster location |
+| `RESOURCEGROUP` | existing cluster group |
+
+#### Fake RP environment requirements
```
jim-minter: These are all things needed by the fake RP itself, not by the test.  That documentation belongs somewhere else.
In the fake RP case, the test needs the _data/containerservice.yaml for the existing cluster to pull out the credentials with which to connect.

jim-minter: If #880 merges as is, the e2e tests will also need a [Real] / [Fake] focus input - this needs to be documented here.

charlesakalugwu:
@jim-minter can you have another look?

docs/e2e/requirements.md
```diff
@@ -0,0 +1,53 @@
...
+| `Environment Variable` | `RESOURCEGROUP` | The resource group for an existing cluster |
+
+#### Fake RP environment requirements
+
+The following are required for running the e2e tests against the fake RP
+
+| Artifact Kind | Name | Notes |
+| --- | --- | --- |
+| `Ginkgo Focus` | `[Fake]` | Allows the test to configure its `OpenshiftManagedCluster` client against the fake RP |
+| `File` | `_data/containerservice.yaml` | Allows the test to retrieve credentials as well as to decide if it is a `create` vs `update` |
```
kargakis: Actually _data/manifest.yaml is also required (used by the reentrant update test now and by the key rotation test although it should be removed from the latter once key rotation is a supported action in the fake RP. It can potentially be removed from the former too maybe in favor of reading the manifest from containerservice.yaml), and test/manifests/normal/scaleup.yaml and test/manifests/normal/scaledown.yaml are used by the scaleupdown test. Can you provide examples on how to make the scale up/down manifests configurable since we have flags for them?

charlesakalugwu: @kargakis mostly took care of your suggestions. thanks. Regarding examples on how to make the scale up/down configurable through the use of flags, it should be possible to run the binary with the flags as shown below

```
./e2e -scaleUpManifest=/some/path/to/a/scale/down/manifest  (other flags)
```

I'll update the documentation in this PR with examples once #914 merges.

docs/e2e/running.md
```diff
@@ -0,0 +1,59 @@
+#### Running e2e tests
```
kargakis: Rename this file to README.md

docs/e2e/running.md
```diff
@@ -0,0 +1,59 @@
...
+
+To run the e2e tests locally it is recommended to use the [e2e](https://quay.io/repository/openshift-on-azure/e2e-tests)
```
kargakis: This approach cannot possibly work when working locally with e2e tests, eg. you make a change in the test that you want to test out. I would suggest explaining how to run tests locally using ginkgo and having the docker image as an alternative option that does not require the user to use ginkgo or work in the context of openshift-azure in order to run the tests.

docs/e2e/requirements.md
```diff
@@ -0,0 +1,88 @@
...
+| Artifact Kind | Name | Notes |
+| --- | --- | --- |
+| `Ginkgo Focus` | `[Fake]` | Allows the test to configure its `OpenshiftManagedCluster` client against the fake RP |
+| `Environment Variable` | `AZURE_CLIENT_ID` | Required for authentication against the RP |
+| `Environment Variable` | `AZURE_CLIENT_SECRET` | Required for authentication against the RP |
+| `Environment Variable` | `AZURE_TENANT_ID` | Required for authentication against the RP |
+| `Environment Variable` | `AZURE_SUBSCRIPTION_ID` | The subscription id for an existing cluster |
+| `Environment Variable` | `RESOURCEGROUP` | The resource group for an existing cluster |
+| `File` | `_data/containerservice.yaml` | Allows the test to retrieve credentials as well as to decide if it is a `create` vs `update` |
+| `File` | `_data/manifest.yaml` | The external cluster manifest (required for key rotation and reentrant updates tests) |
```
kargakis: Needs a follow-up issue to remove this (either in favor of the manifest stored in containerservice.yaml or by removing the restriction altogether from the tests).

docs/e2e/running.md
```diff
@@ -0,0 +1,59 @@
...
+--volume $PWD/test/manifests:/manifests \
+--volume $HOME/.kube/config:/root/.kube/config \
+quay.io/openshift-on-azure/e2e-tests \
+-test.v \
+-test.timeout=${TIMEOUT} \
+-ginkgo.noColor \
+-ginkgo.v \
+-ginkgo.focus=${FOCUS:-} \
+-artifact-dir=/artifacts \
+-scaleUpManifest= /manifests/normal/scaleup.yaml \
```
kargakis: Is the extra space between the `=` sign not breaking the command? For completeness, I would remove the space.

docs/e2e/running.md
charlesakalugwu: fair enough

docs/e2e/README.md
```diff
@@ -0,0 +1,81 @@
+#### Running e2e tests locally with go test
+
+If you are developing tests locally, you could use the native `go test` to invoke your ginkgo test specs.
```
kargakis: nit: s/could/can/

docs/e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Example: Run the end user e2e tests
+```
+export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
+export TIMEOUT=20m
```
kargakis: We default to 20m so this can be omitted.

docs/e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+Example: Run the end user e2e tests in a container
+```
+export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
+export TIMEOUT=20m
+
+docker run \
+--rm \
+-it \
+--volume $PWD/_data:/_data \
+--volume $PWD/artifacts:/artifacts \
```
kargakis: Haven't used this `artifacts` directory at the root of `openshift-azure` before. May be worth adding in `.gitignore`

docs/e2e/requirements.md
```diff
@@ -0,0 +1,88 @@
...
+| `Environment Variable` | `AZURE_CLIENT_ID` | Required for authentication against the real RP |
+| `Environment Variable` | `AZURE_CLIENT_SECRET` | Required for authentication against the real RP |
+| `Environment Variable` | `AZURE_TENANT_ID` | Required for authentication against the real RP |
+| `Environment Variable` | `AZURE_SUBSCRIPTION_ID` | The subscription id for an existing cluster |
+| `Environment Variable` | `AZURE_REGION` | The region for an existing cluster |
+| `Environment Variable` | `RESOURCEGROUP` | The resource group for an existing cluster |
+
+#### Fake RP environment requirements
+
+The following are required for running all e2e tests against the fake RP
```
kargakis: The fake RP is deployed separately from the test suite now so you should give directions on how to deploy the fake rp (maybe revert https://github.com/openshift/openshift-azure/pull/948)

docs/e2e/requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+* `manifest`: Path to a customer manifest specifying a cluster for which keys should be rotated (default: `_data/manifest.yaml`)
+* `configBlob`: Path to an internal config blob for the cluster in question (default: `_data/containerservice.yaml`)
+
+These flags are required for any test focus which includes the key rotation test.
+
+##### Reentrant Updates
+The reentrant updates e2e test accepts one flag for customizing its functionality
+
+* `manifest`: Path to a customer manifest specifying a cluster for which reentrant updates should be tested (default: `_data/manifest.yaml`)
```
kargakis: This should not really be customizable and the same request that was used during cluster creation should be reused. Also the actual flag is `request` and not `manifest` since `manifest` is already used by the key rotation test.

docs/e2e/README.md
```diff
@@ -0,0 +1,80 @@
...
+image
+
+```
+docker pull quay.io/openshift-on-azure/e2e-tests:latest
+```
+
+Example: Run the end user e2e tests in a container
+```
+export FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]"
+export TIMEOUT=20m
```
kargakis: You also need to `source ./env` here?

hack/e2e.sh
```diff
@@ -14,6 +14,21 @@ if [[ -z "$TIMEOUT" ]]; then
     TIMEOUT=20m
 fi
 
-go test ./test/e2e -timeout "$TIMEOUT" -test.v -ginkgo.v "${FOCUS:-}" -ginkgo.noColor -tags e2e "${ARTIFACT_FLAG:-}"
+# start the fake rp server if needed
```
kargakis: Doing this here is wrong since this script already runs in CI alongside a running fake RP?

kargakis: I guess the script will not fail even though 8080 is already bound to a different process because you run it aysnchronously here?

kargakis: I think this is fine, we probably want to remove https://github.com/openshift/release/blob/dcf126e623dae74f034189b6177901a518d23877/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L232

charlesakalugwu: :+1:

PR #813, opened 2018-11-30
More unit tests for cluster and move some code out to make testing easier

asalkeld:
So testing Update() would be very hard as it pulls in a lot of code and you would have to mock a *lot* of calls.
I have tried to reduce the volume of calls by moving the kubenetes functions into a new package . This make the mocking you have to do in the cluster unit tests simpler and then we can unit test the kubeclient package easily too.

So my basic aim is to create another interface that we can mock, if there is a better way of splitting cluster up so we have more interfaces, I am all ears.

PR #885, opened 2018-12-10
fix stale generated code

jim-minter:
@kargakis is master currently broken?  we don't retest as we merge code?


kargakis:
You need to update your local gomock repo, I've sent an email in aos-azure

jim-minter:
yes, but I think our repo is actually broken since #813 merged

kargakis:
ouch, I thought those tests were fresh. my bad.

jim-minter:
@kargakis is there a relatively painless way of having the CI prevent this from happening again?

PR #886, opened 2018-12-10
Add support for vnet testing

kwoodson:
Adding ability to precreate the vnet for vnet testing purposes.

### DIFF
--- a/pkg/util/azureclient/resources.go
+++ b/pkg/util/azureclient/resources.go
@@ -59,3 +59,49 @@ func NewResourcesClient(subscriptionID string, authorizer autorest.Authorizer, l
 		Client: client,
 	}
 }
+
+// GroupsClient is a minimal interface for azure Resources Client
+type GroupsClient interface {
+	CreateOrUpdate(ctx context.Context, resourceGroupName string, parameters resources.Group) (result resources.Group, err error)
+	CheckExistence(ctx context.Context, resourceGroupName string) (exists bool, err error)
+	List(ctx context.Context, filter string, top *int32) (result resources.GroupListResultPage, err error)
+	Delete(ctx context.Context, resourceGroupName string) (result resources.GroupsDeleteFuture, err error)
+	Client
+}
+
+type groupsClient struct {
+	resources.GroupsClient
+}
+
+var _ GroupsClient = &groupsClient{}
+
+// NewGroupsClient creates a new ResourcesClient
+func NewGroupsClient(subscriptionID string, authorizer autorest.Authorizer, languages []string) GroupsClient {
+	client := resources.NewGroupsClient(subscriptionID)
+	client.Authorizer = authorizer
+	client.RequestInspector = addAcceptLanguages(languages)
+
+	return &groupsClient{
+		GroupsClient: client,
+	}
+}
+
+func (c *groupsClient) Client() autorest.Client {
+	return c.GroupsClient.Client
+}
+
+func (c *groupsClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, parameters resources.Group) (resources.Group, error) {
+	return c.GroupsClient.CreateOrUpdate(ctx, resourceGroupName, parameters)
+}
+
+func (c *groupsClient) CheckExistence(ctx context.Context, resourceGroupName string) (bool, error) {
+	resp, err := c.GroupsClient.CheckExistence(ctx, resourceGroupName)
+	switch resp.StatusCode {
+	case 404:
+		return false, nil
+	case 200:
+		return true, nil
+	default:
+		return false, err
+	}
+}
--- a/test/clients/azure/client.go
+++ b/test/clients/azure/client.go
@@ -26,7 +26,6 @@ var (
 	adminRpFocus = rpFocus(regexp.QuoteMeta("[Admin]"))
 	fakeRpFocus  = rpFocus(regexp.QuoteMeta("[Fake]"))
 	realRpFocus  = rpFocus(regexp.QuoteMeta("[Real]"))
-	vnetRpFocus  = rpFocus(regexp.QuoteMeta("[Vnet]"))
 )
 
 func (tf rpFocus) match(focusString string) bool {
@@ -83,11 +82,11 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
 	case adminRpFocus.match(focus), fakeRpFocus.match(focus):
 		fmt.Println("configuring the fake resource provider")
 		rpURL = fmt.Sprintf("http://%s", shared.LocalHttpAddr)
-	case realRpFocus.match(focus), vnetRpFocus.match(focus):
+	case realRpFocus.match(focus):
 		fmt.Println("configuring the real resource provider")
 		rpURL = externalapi.DefaultBaseURI
 	default:
-		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\] or -ginkgo.focus=\\[Vnet\\]", config.GinkgoConfig.FocusString))
+		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\]", config.GinkgoConfig.FocusString))
 	}
 
 	rpc := externalapi.NewOpenShiftManagedClustersClientWithBaseURI(rpURL, subscriptionID)
--- a/test/clients/azure/client.go
+++ b/test/clients/azure/client.go
@@ -26,6 +26,7 @@ var (
 	adminRpFocus = rpFocus(regexp.QuoteMeta("[Admin]"))
 	fakeRpFocus  = rpFocus(regexp.QuoteMeta("[Fake]"))
 	realRpFocus  = rpFocus(regexp.QuoteMeta("[Real]"))
+	vnetRpFocus  = rpFocus(regexp.QuoteMeta("[Vnet]"))
 )
 
 func (tf rpFocus) match(focusString string) bool {
@@ -43,6 +44,9 @@ type Client struct {
 	VirtualMachineScaleSetExtensions azureclient.VirtualMachineScaleSetExtensionsClient
 	VirtualMachineScaleSetVMs        azureclient.VirtualMachineScaleSetVMsClient
 	Resources                        azureclient.ResourcesClient
+	VirtualNetworks                  azureclient.VirtualNetworksClient
+	VirtualNetworksPeerings          azureclient.VirtualNetworksPeeringsClient
+	Groups                           azureclient.GroupsClient
 }
 
 // NewClientFromEnvironment creates a new azure client from environment variables.
@@ -79,11 +83,11 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
 	case adminRpFocus.match(focus), fakeRpFocus.match(focus):
 		fmt.Println("configuring the fake resource provider")
 		rpURL = fmt.Sprintf("http://%s", shared.LocalHttpAddr)
-	case realRpFocus.match(focus):
+	case realRpFocus.match(focus), vnetRpFocus.match(focus):
 		fmt.Println("configuring the real resource provider")
 		rpURL = externalapi.DefaultBaseURI
 	default:
-		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\]", config.GinkgoConfig.FocusString))
+		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\] or -ginkgo.focus=\\[Vnet\\]", config.GinkgoConfig.FocusString))
 	}
 
 	rpc := externalapi.NewOpenShiftManagedClustersClientWithBaseURI(rpURL, subscriptionID)
@@ -102,5 +106,8 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
 		VirtualMachineScaleSetExtensions: azureclient.NewVirtualMachineScaleSetExtensionsClient(subscriptionID, authorizer, nil),
 		VirtualMachineScaleSetVMs:        azureclient.NewVirtualMachineScaleSetVMsClient(subscriptionID, authorizer, nil),
 		Resources:                        azureclient.NewResourcesClient(subscriptionID, authorizer, nil),
+		VirtualNetworks:                  azureclient.NewVirtualNetworkClient(subscriptionID, authorizer, nil),
+		VirtualNetworksPeerings:          azureclient.NewVirtualNetworksPeeringsClient(subscriptionID, authorizer, nil),
+		Groups:                           azureclient.NewGroupsClient(subscriptionID, authorizer, nil),
 	}, nil
 }
--- a/test/e2e/specs/realrp/vnetpeering.go
+++ b/test/e2e/specs/realrp/vnetpeering.go
@@ -2,7 +2,6 @@ package realrp
 
 import (
 	"context"
-	"errors"
 	"os"
 	"time"
 
@@ -11,14 +10,14 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-06-01/network"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
-	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
 	"github.com/openshift/openshift-azure/pkg/fakerp/client"
 	"github.com/openshift/openshift-azure/test/clients/azure"
+	tlog "github.com/openshift/openshift-azure/test/util/log"
 )
 
-var _ = Describe("Peer Vnet tests [Vnet]", func() {
+var _ = Describe("Peer Vnet tests [Vnet][Real][LongRunning]", func() {
 	var (
 		ctx          = context.Background()
 		cli          *azure.Client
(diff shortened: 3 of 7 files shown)


cmd/createorupdate/createorupdate.go
```diff
@@ -126,11 +131,78 @@ func createResourceGroup(conf *fakerp.Config) (bool, error) {
 	return true, err
 }
 
-func execute(ctx context.Context, log *logrus.Entry, rpc v20180930preview.OpenShiftManagedClustersClient, conf *fakerp.Config) error {
+func deployVnet(ctx context.Context, log *logrus.Entry, conf *fakerp.Config, rpc v20180930preview.OpenShiftManagedClustersClient) (string, error) {
+	vnet, err := ioutil.ReadFile("../../test/templates/" + *createPeerVnet + ".json")
```
kwoodson: The preceding `../..` are for my own debug purposes.  Will need to reset this pathing before merge.

pkg/fakerp/manifest.go
```diff
@@ -39,7 +39,7 @@ func LoadClusterConfigFromManifest(log *logrus.Entry, manifestTemplate string, c
 		manifestTemplate = conf.Manifest
 	}
 	if manifestTemplate == "" {
-		manifestTemplate = "test/manifests/normal/create.yaml"
+		manifestTemplate = "../../test/manifests/normal/create.yaml"
```
kwoodson: The preceeding `../..` are for my own debug purposes.  Will need to reset this pathing before merge.

test/templates/peervnet.json
```diff
@@ -0,0 +1,152 @@
...
+    }
+  ],
+  "outputs": {
+    "hostname": {
+      "type": "string",
+      "value": "[reference('publiciptest').dnsSettings.fqdn]"
+    },
+    "vnetid": {
+      "type": "string",
+      "value": "[reference('vnet').resourceGuid]"
```
kwoodson: This is not the value required for `peerVnetID`.  Will need to either return it or derive it above.

test/templates/peervnet.json
```diff
@@ -0,0 +1,152 @@
...
+                "id": "[resourceId('Microsoft.Network/virtualNetworks/subnets', 'vnet', 'subnettest')]"
+              }
+            }
+          }
+        ]
+      }
+    },
+    {
+      "apiVersion": "2018-10-01",
+      "type": "Microsoft.Compute/virtualMachines",
```
asalkeld: our of curiosity why do we need  a vm to create vnet?

kwoodson: We need something on the other side to answer the call (ping).  @jim-minter  and I had a chat about this story and we've decided we don't want to pay the cost of having a VM on each test.  I'd like to get this working and then remove the VM as to verify that vnet peering is possible and provide a way to get it set up.

cmd/createorupdate/createorupdate.go
```diff
@@ -249,8 +321,16 @@ func main() {
 			log.Fatal(err)
 		}
 	}
+	var vnet string
+	if !isDelete && *useProd && *createPeerVnet != "" {
```
kargakis: If we are only planning to test this in production then you don't need to deploy the VNET stuff first, then get the vnet string and set it in the OSA deployment. You can follow @jim-minter's advice and add the OSA resource as part of your ARM template, then you should be able to share the VNET string in the ARM template?

kargakis:
For clarity, I was discussing this with @jim-minter yesterday, we agreed it should be a self-contained test in the e2e framework (likely its own suite under eg. `test/e2e/specs/realrp/vnetpeering.go`) and no changes should go in `createorupdate.go`. Also we should ditch the ARM template idea in favor of using the golang sdk to create the vnet and then once the cluster is deployed, ensure the vnet shows up as peered.

On a sidenote, such a test wouldn't be able to use our existing e2e template as is because we create a cluster and then run the tests, whereas this test needs to create the cluster itself. We agreed we should rework how our tests run (instead of running hack/create.sh in the setup container to create clusters we should instead deploy the fakerp to run, then the cluster creation can be requested by the e2e test framework). For the purposes of this PR, we could simply avoid running `hack/create.sh` in the setup container via an env variable for now ...

cmd/createorupdate/createorupdate.go
```diff
@@ -124,6 +124,7 @@ func execute(ctx context.Context, log *logrus.Entry, rpc v20180930preview.OpenSh
 	if err != nil {
 		return err
 	}
+
```
kargakis: remove

kwoodson: removed.

kwoodson:
@kargakis, thanks for the comments.  I met with @jim-minter after he had spoken with you.  He repeated your comments so I'm going to work on the e2e test and remove my changes from createorupdate and add them to e2e.  I'll remove the arm template as well.

kwoodson:
@kargakis, @asalkeld, I think this is ready for review.

Few questions I had for @kargakis,
1. How does the delete of the resources happen?  I wasn't sure so I added a delete on the `aftertest` section.  Is that correct?
2. Should we add this to the prow config to run this test? If so, how often should we run it?

Thanks!

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+		if os.Getenv("AZURE_REGION") == "" {
+			Expect(errors.New("AZURE_REGION is not set")).ToNot(HaveOccurred())
+		}
+		if os.Getenv("RESOURCEGROUP") == "" {
+			Expect(errors.New("RESOURCEGROUP is not set")).ToNot(HaveOccurred())
+		}
+	})
+
+	AfterEach(func() {
+		result, _ := cli.Groups.Delete(ctx, rg)
```
kargakis: You shouldn't need to delete. Today, we are cleaning everything up in the `teardown` container of the e2e template. We may want to move resource deletion in the e2e framework in the future but we should be doing that holistically.

kargakis: Hrm, actually you may need to keep this for now. I will have another pass tomorrow and think about this more.

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-06-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp/client"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+)
+
+var _ = Describe("Peer Vnet tests [Vnet]", func() {
```
kargakis: Also [Real] to denote that it is meant to run using the production RP.

kargakis:
> * How does the delete of the resources happen?  I wasn't sure so I added a delete on the `aftertest` section.  Is that correct?

All of our tests are using the `teardown` container in the e2e template in release. I think you should be able to reuse it, but I can verify it tomorrow.

> * Should we add this to the prow config to run this test? If so, how often should we run it?

Yes, we should. We can start with daily like all the rest of the periodics.

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+		region       = os.Getenv("AZURE_REGION")
+		clientID     = os.Getenv("AZURE_CLIENT_ID")
+		clientSecret = os.Getenv("AZURE_CLIENT_SECRET")
+	)
+
+	BeforeEach(func() {
+		var err error
+		cli, err = azure.NewClientFromEnvironment(false)
+		Expect(err).ToNot(HaveOccurred())
+		if os.Getenv("AZURE_REGION") == "" {
```
charlesakalugwu: I guess we could simplify this to `Expect(region).NotTo(BeEmpty())`

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+			Expect(errors.New("AZURE_REGION is not set")).ToNot(HaveOccurred())
+		}
+		if os.Getenv("RESOURCEGROUP") == "" {
```
charlesakalugwu: I guess we could simplify this to `Expect(rg).NotTo(BeEmpty())`

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+						Name: &subnetName,
+						SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+							AddressPrefix: &subnetAddressPrefix,
+						},
+					},
+				},
+			},
+			Location: &region,
+		})
+		Expect(err).ToNot(HaveOccurred())
```
charlesakalugwu: also check the `future` with something like `Expect(future).NotTo(BeNil())`

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+		err = future.WaitForCompletionRef(ctx, cli.VirtualNetworks.Client())
+		Expect(err).ToNot(HaveOccurred())
+
+		vnet, err := cli.VirtualNetworks.Get(ctx, rg, vnetPeerName, "")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(*vnet.VirtualNetworkPeerings)).To(BeEquivalentTo(0))
```
charlesakalugwu: or `To(Equal(0))`

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+				},
+			},
+			Location: &region,
+		})
+		Expect(err).ToNot(HaveOccurred())
+		err = future.WaitForCompletionRef(ctx, cli.VirtualNetworks.Client())
+		Expect(err).ToNot(HaveOccurred())
+
+		vnet, err := cli.VirtualNetworks.Get(ctx, rg, vnetPeerName, "")
+		Expect(err).ToNot(HaveOccurred())
```
charlesakalugwu: also check the vnet with something like `Expect(vnet).NotTo(BeNil())`

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+		Expect(len(*vnet.VirtualNetworkPeerings)).To(BeEquivalentTo(0))
+
+		// load cluster config
+		config, err := client.LoadClusterConfigFromManifest(log, "../../test/manifests/normal/create.yaml")
```
charlesakalugwu: `Expect(config).NotTo(BeNil())`

charlesakalugwu: you might also want to consider making the path to the cluster config configurable using flags..

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+		}
+		// Set pre-created peer vnetid in cluster config
+		config.Properties.NetworkProfile.PeerVnetID = vnet.ID
+
+		// create a cluster with the peerVnet
+		_, err = cli.OpenShiftManagedClusters.CreateOrUpdateAndWait(ctx, rg, rg, *config)
+		Expect(err).NotTo(HaveOccurred())
+
+		clusterVnet, err := cli.VirtualNetworks.Get(ctx, rg, vnetPeerName, "")
```
charlesakalugwu: check `clusterVnet` with something like `Expect(clusterVnet).NotTo(BeNil())`

test/e2e/specs/realrp/vnetpeering.go
kargakis: Yeah, you can probably keep this, it would make the test more self-contained. Eventually, we should do this for all other e2e tests.

test/clients/azure/client.go
```diff
@@ -26,6 +26,7 @@ var (
 	adminRpFocus = rpFocus(regexp.QuoteMeta("[Admin]"))
 	fakeRpFocus  = rpFocus(regexp.QuoteMeta("[Fake]"))
 	realRpFocus  = rpFocus(regexp.QuoteMeta("[Real]"))
+	vnetRpFocus  = rpFocus(regexp.QuoteMeta("[Vnet]"))
```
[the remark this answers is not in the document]
kwoodson: Will update.

test/clients/azure/client.go
```diff
@@ -79,11 +83,11 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
 	case adminRpFocus.match(focus), fakeRpFocus.match(focus):
 		fmt.Println("configuring the fake resource provider")
 		rpURL = fmt.Sprintf("http://%s", shared.LocalHttpAddr)
-	case realRpFocus.match(focus):
+	case realRpFocus.match(focus), vnetRpFocus.match(focus):
```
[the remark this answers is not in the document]
kwoodson: Will update.

test/clients/azure/client.go
```diff
@@ -79,11 +83,11 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
...
 		fmt.Println("configuring the real resource provider")
 		rpURL = externalapi.DefaultBaseURI
 	default:
-		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\]", config.GinkgoConfig.FocusString))
+		panic(fmt.Sprintf("invalid focus %q - need to -ginkgo.focus=\\[Admin\\], -ginkgo.focus=\\[Fake\\] or -ginkgo.focus=\\[Real\\] or -ginkgo.focus=\\[Vnet\\]", config.GinkgoConfig.FocusString))
```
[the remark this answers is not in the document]
kwoodson: Will update

test/e2e/specs/realrp/vnetpeering.go
kwoodson: Done.

test/e2e/specs/realrp/vnetpeering.go
kwoodson: I took this from realrp.go.  Should I update that file as well?

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,129 @@
...
+	AfterEach(func() {
+		result, _ := cli.Groups.Delete(ctx, rg)
+		_ = result.WaitForCompletionRef(ctx, cli.Groups.Client())
+	})
+
+	It("should create the vnet and cluster and verify peering", func() {
+		logrus.SetLevel(logrus.DebugLevel)
+		logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+		logrus.SetOutput(GinkgoWriter)
+		log := logrus.NewEntry(logrus.StandardLogger())
```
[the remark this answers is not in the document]
kwoodson: Will update.

test/e2e/specs/realrp/vnetpeering.go
kwoodson: Will add.

test/e2e/specs/realrp/vnetpeering.go
kwoodson: Will add.

test/e2e/specs/realrp/vnetpeering.go
kwoodson: Added.

test/e2e/specs/realrp/vnetpeering.go
kwoodson: Added.

pkg/util/azureclient/resources.go
```diff
@@ -59,3 +59,49 @@ func NewResourcesClient(subscriptionID string, authorizer autorest.Authorizer, l
...
+
+func (c *groupsClient) CheckExistence(ctx context.Context, resourceGroupName string) (bool, error) {
+	resp, err := c.GroupsClient.CheckExistence(ctx, resourceGroupName)
+	switch resp.StatusCode {
+	case 404:
+		return false, nil
+	case 200:
+		return true, nil
+	default:
+		return false, err
```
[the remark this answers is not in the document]
kwoodson: I believe the status code is included in the err message.

kwoodson:
Unless there are other comments I think this one is ready to go.

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,125 @@
...
+	BeforeEach(func() {
+		var err error
+		cli, err = azure.NewClientFromEnvironment(false)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(region).NotTo(BeEmpty())
+		Expect(rg).NotTo(BeEmpty())
+	})
+
+	AfterEach(func() {
+		result, _ := cli.Groups.Delete(ctx, rg)
```
kargakis: Instead of just deleting the resource group, I think it's more appropriate to first Delete the OSA resource, then the resource group, similarly to how you first create the resource group, then the OSA resource. There may be other actions triggered in the OSA RP backend apart from resouces in the current resource group?

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,125 @@
...
+		_ = result.WaitForCompletionRef(ctx, cli.Groups.Client())
+	})
+
+	It("should create the vnet and cluster and verify peering", func() {
+		log := tlog.GetTestLogger()
+
+		// create a resource group and a vnet
+		exists, err := cli.Groups.CheckExistence(ctx, rg)
```
kargakis: The resource group is meant to be unique per test run so you can probably remove this check. Reusing groups has been problematic in the past (eg. what if the group is in Deleting state?).

kargakis: Also I think the resource group creation code can be moved in BeforeEach.

test/e2e/specs/realrp/vnetpeering.go
kargakis: The error message from this is bad:
```
  Expected
      <string>: 
  not to be empty
```
We should either create a wrapper for this like:
```go
func ensureEnvExists(env string) {
  if os.Getenv(env) == "" {
	Expect(fmt.Errorf("%s is not set", env)).ToNot(HaveOccurred())
  } 
}
```
or use `envconfig` but that would need as to centralize config into a single struct or set of structs.

test/e2e/specs/realrp/vnetpeering.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-06-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+
+	"github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp/client"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+	tlog "github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = Describe("Peer Vnet tests [Vnet][Real][LongRunning]", func() {
```
kargakis: Probably unexpected but the [Real] tag is causing this test to run as part of `make e2e-prod`. I think we need to keep them separate.

kwoodson: I was following instructions from other reviewers.  We can add them back

PR #831, opened 2018-12-04
Add re-entrancy check e2e test

kargakis:
We will also need a test where an actual upgrade is invoked and at the end we ensure the update blob is different than what it was before the update.

PR #890, opened 2018-12-11
Fix rp focus selections in e2e tests needing azure client

charlesakalugwu:
follow-up to #880 

/cc @kargakis @jim-minter

### DIFF
--- a/test/clients/azure/client.go
+++ b/test/clients/azure/client.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/onsi/ginkgo/config"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


thekad:
Does this require a change in `./hack/e2e.sh` ?

charlesakalugwu:
@thekad no changes in `./hack/e2e.sh` required with this change

kargakis:
Just reviewed this, it still seems to be an issue. I would expect this PR to fail because `make e2e`'s FOCUS does not include Fake, right?

test/clients/azure/client.go
```diff
@@ -50,12 +59,12 @@ func NewClientFromEnvironment() (*Client, error) {
 	}
 
 	var rpURL string
-	focus := []byte(config.GinkgoConfig.FocusString)
+	focus := config.GinkgoConfig.FocusString
 	switch {
-	case adminRpFocus.Match(focus), fakeRpFocus.Match(focus):
+	case adminRpFocus.match(focus), fakeRpFocus.match(focus):
 		fmt.Println("configuring the fake resource provider")
 		rpURL = realfakerp.StartServer(log, conf, realfakerp.LocalHttpAddr)
```
kargakis: Btw, this is wrong, StartServer should ensure there is no fakerp running before trying to start a new one.

kargakis: Probably doesn't matter since we don't check the error returned from ListenAndServe.

kargakis:
> Just reviewed this, it still seems to be an issue. I would expect this PR to fail because `make e2e`'s FOCUS does not include Fake, right?

Ok, this does not fail because we don't use the azure client in any of these tests.

kargakis:
> Ok, this does not fail because we don't use the azure client in any of these tests.

I am setting the tags correctly in the second commit in https://github.com/openshift/openshift-azure/pull/831 (had to otherwise my PR fails to pass e2e).

PR #891, opened 2018-12-11
Relax dns variable restrictions from e2e tests

### LINKED ISSUE #888: Remove DNS variables restrictions from e2e tests
DNS_RESOURCEGROUP and DNS_DOMAIN are required by our e2e tests because they are set as `required: true` in the `envconfig` tag of `fakerp.Config`. Since tests may not run against the fake RP or may not even create or delete clusters via the fake RP, these variables should not be required. Once relaxed in openshift-azure, we should also remove them from the test container in [release](https://github.com/openshift/release/blob/a961f8577624f2215ea17572cc438d18eaf3846f/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml)

/kind cleanup
/assign charlesakalugwu

charlesakalugwu:
closes #888 

/cc @kargakis

### DIFF
--- a/pkg/fakerp/config.go
+++ b/pkg/fakerp/config.go
@@ -34,8 +34,8 @@ type Config struct {
 	AADClientSecret string `envconfig:"AZURE_AAD_CLIENT_SECRET"`
 
 	Region           string `envconfig:"AZURE_REGION"`
-	DnsDomain        string `envconfig:"DNS_DOMAIN" required:"true"`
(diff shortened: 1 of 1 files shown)


asalkeld:
@charlesakalugwu shouldn't we be manually checking for these in the case we do actually need them?

kargakis:
> @charlesakalugwu shouldn't we be manually checking for these in the case we do actually need them?

Yes, we should. We only need them to be set on CREATEs and DELETEs.

PR #900, opened 2018-12-12
createorupdate supports sending admin requests

0xmichalis:
@jim-minter I have been playing around with trying to send admin requests to the fake RP, it seems we cannot regenerate empty fields because merging runs before generation. Thoughts?

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


kargakis:
> @jim-minter I have been playing around with trying to send admin requests to the fake RP, it seems we cannot regenerate empty fields because merging runs before generation. Thoughts?

Eh, just realized field deletes are going to be sent to a different endpoint. Nevermind.

PR #901, opened 2018-12-12
decrease polling interval

jim-minter:
by default it's 60 seconds, changed to 10
(also adds code which could be useful to debug azure HTTPS API calls - disabled by default)

### DIFF
--- a/pkg/util/azureclient/azureclient.go
+++ b/pkg/util/azureclient/azureclient.go
@@ -12,7 +12,9 @@ package azureclient
 import (
 	"context"
 	"net/http"
+	"net/http/httputil"
 	"os"
+	"time"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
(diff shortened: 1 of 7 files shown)


pkg/util/azureclient/azureclient.go
```diff
@@ -32,16 +34,36 @@ func addAcceptLanguages(acceptLanguages []string) autorest.PrepareDecorator {
...
+		os.Stdout.Write(b)
+	}
+	return resp, err
+}
+
+func setupClient(client *autorest.Client, authorizer autorest.Authorizer, languages []string) {
+	client.Authorizer = authorizer
+	client.RequestInspector = addAcceptLanguages(languages)
+	client.PollingDelay = 10 * time.Second
+	// client.Sender = &loggingSender{client.Sender}
```
asalkeld: delete or uncomment?

pkg/util/azureclient/azureclient.go
```diff
@@ -32,16 +34,36 @@ func addAcceptLanguages(acceptLanguages []string) autorest.PrepareDecorator {
...
 	}
 }
 
+type loggingSender struct {
+	autorest.Sender
+}
+
+func (ls *loggingSender) Do(req *http.Request) (*http.Response, error) {
+	b, _ := httputil.DumpRequestOut(req, true)
+	os.Stdout.Write(b)
```
asalkeld: yikes this is going to get noisy, can't we log it in debug mode?

asalkeld: ok, i see it's dead code and you have to uncomment the line below - not my favourite form of configuration

pkg/util/azureclient/azureclient.go
kargakis: I have also been meaning to plump this via TestConfig. Seems like a useful thing to be able to enable in tests.

mjudeikis: Not sure we want all this noise in tests too. I would e fine to keep it non-configurable for now as in most cases it will be local runs, which needs this one. We already have tone of configurables :/

kargakis: > Not sure we want all this noise in tests too. I would e fine to keep it non-configurable for now as in most cases it will be local runs, which needs this one. We already have tone of configurables :/

We should be able to redirect this output into its own file. OK for now

/lgtm

PR #910, opened 2018-12-14
Set ttl on dns zones

kargakis:
Part of https://github.com/openshift/azure-misc/issues/122

### DIFF
--- a/pkg/fakerp/dns.go
+++ b/pkg/fakerp/dns.go
@@ -3,6 +3,7 @@ package fakerp
 import (
 	"context"
 	"fmt"
+	"time"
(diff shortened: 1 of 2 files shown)


kargakis:
There are a bunch of leftover dns records in the dns resourcegroup (see [here](http://pastebin.test.redhat.com/684187)). I am not sure how to get those garbage-collected via azure-purge since it doesn't seem dns.RecordSet has Tags.

kargakis:
/hold cancel 

Opened https://github.com/openshift/azure-misc/pull/124 for the azure-purge part.

Issue #912, opened 2018-12-14
Split fake RP server into its own process


kargakis:
We need to split the fake RP into its own process to ease e2e testing and avoid statically linking our plugins into the e2e tests.

cc @jim-minter @charlesakalugwu 

@kwoodson this is going to make custom vnet testing easier

/kind cleanup
/area tests
/assign

kargakis:
Since the server is going to be a long running process versus a goroutine in the client, we need to make it exit somehow in order for the test pod to complete successfully. I am thinking of adding an `/exit` endpoint that the template can query after it finishes deleting the cluster.

mjudeikis:
> Since the server is going to be a long running process versus a goroutine in the client, we need to make it exit somehow in order for the test pod to complete successfully. I am thinking of adding an /exit endpoint that the template can query after it finishes deleting the cluster.

Are we planning to run Server as a process in our CI cluster, as a separate pod or something like that? What is the architecture in your head for this?

kargakis:
> Are we planning to run Server as a process in our CI cluster, as a separate pod or something like that? What is the architecture in your head for this?

No, because this is not enabled by our CI and the fake RP pod would need to be redeployed for every PR anyway. At best this this can be its own container, for now its simpler to run it as we are already used to (as part of every hack/{action}.sh script).

kargakis:
> because this is not enabled by our CI

Actually we should be able to use two separate pods as part of the e2e template but I am not sure we want to go there.

mjudeikis:
I was wondering, is it worth doing something like "hosting our own fakeRP" which would be redeployed when fakeRP code changes? Like running our own Azure frontend and backend?

kargakis:
> I was wondering, is it worth doing something like "hosting our own fakeRP" which would be redeployed when fakeRP code changes? Like running our own Azure frontend and backend?

Each test for a PR, merged commit, or periodic run would have to deploy its own fake RP. Not sure it's worth it but that's a separate discussion, enabled but orthogonal to this PR.

PR #913, opened 2018-12-14
Split fake RP into its own process

kargakis:
Fixes https://github.com/openshift/openshift-azure/issues/912

### DIFF
--- a/cmd/fakerp/main.go
+++ b/cmd/fakerp/main.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
+)
+
+func main() {
+	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	log := logrus.NewEntry(logrus.StandardLogger())
+
+	// TODO: Validate all required env variables are in place
+
+	log.Info("starting the fake resource provider")
+	s := fakerp.NewServer(log, os.Getenv("RESOURCEGROUP"), shared.LocalHttpAddr)
+	s.ListenAndServe()
+}
--- a/cmd/recoveretcdcluster/recoveretcdcluster.go
+++ b/cmd/recoveretcdcluster/recoveretcdcluster.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
 	"github.com/openshift/openshift-azure/pkg/plugin"
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 )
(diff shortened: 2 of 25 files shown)


cmd/fakerp/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+func main() {
+	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	log := logrus.NewEntry(logrus.StandardLogger())
+
+	// TODO: Validate all required env variables are in place
+
+	log.Info("starting the fake resource provider")
+	s := fakerp.NewServer(log, os.Getenv("RESOURCEGROUP"), "localhost:8080")
```
thekad: Does this have to be hardcoded on port 8080? Can't the port be randomized in one of the upper ephemeral ports for the duration of the run?

cmd/recoveretcdcluster/recoveretcdcluster.go
```diff
@@ -22,15 +23,11 @@ func main() {
...
-	conf, err := fakerp.NewConfig(log, true)
-	if err != nil {
-		log.Fatal(err)
-	}
 
 	ctx := context.Background()
-	ctx = context.WithValue(ctx, api.ContextKeyClientID, conf.ClientID)
-	ctx = context.WithValue(ctx, api.ContextKeyClientSecret, conf.ClientSecret)
-	ctx = context.WithValue(ctx, api.ContextKeyTenantID, conf.TenantID)
+	ctx = context.WithValue(ctx, api.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
```
thekad: I sort of feel this is going in the wrong direction, maybe it should mirror how client.go loads its configuration via envconfig? Is there any hard reason why these are environment variables now?

hack/delete.sh
```diff
@@ -14,6 +14,8 @@ fi
 USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
+else
```
thekad: if you're adding `go generate` in create.sh, you should add it here as well

thekad: Actually, this should point to `hack/fakerp.sh` I think

hack/create.sh
```diff
@@ -12,10 +12,13 @@ mkdir -p _data/_out
 
 set -x
 
+go generate ./...
+
 USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
+else
+	go run cmd/fakerp/main.go &
```
thekad: should probably point to `hack/fakerp.sh`

cmd/fakerp/main.go
kargakis: Why do we care about randomizing it?

cmd/recoveretcdcluster/recoveretcdcluster.go
kargakis: Using something like envconfig can be a follow-up, for start I wanted to remove the unnecessary dependency.

hack/delete.sh
kargakis: done

hack/create.sh
kargakis: done

hack/create.sh
```diff
@@ -15,7 +15,8 @@ set -x
 USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
+else
+	hack/fakerp.sh $RESOURCEGROUP &
 fi
 
-go generate ./...
```
mjudeikis: Any particular reason this got removed?

kargakis: Runs as part of `fakerp.sh` now.

mjudeikis: I would like this better in main 'root' script at the top somewhere?

kargakis: What do you mean by root script? hack/create.sh? Generation is related to the backend code run by the fake RP so it's natural place is fakerp.sh.

hack/create.sh
```diff
@@ -15,7 +15,8 @@ set -x
 USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
+else
+	hack/fakerp.sh $RESOURCEGROUP &
```
kargakis: Unfortunately this is leaking and needs to be cleaned up.

kargakis: https://github.com/openshift/openshift-azure/issues/924

PR #916, opened 2018-12-17
Make the generation of scaleset and instance names more consistent

### LINKED ISSUE #894: shouldn't use profile role and name interchangeably when constructing scaleset name
@openshift/sig-azure 

I think we have a typing issue in at least pkg/cluster and pkg/arm where it comes to how we name scalesets.  Sometimes we do `fmt.Sprintf("ss-%s_%d", profile.Name, ...)`; sometimes we do `"ss-"+string(role)`.

Alarmingly I think *both* of the options above could potentially be wrong: the first because I suspect we don't validate the profile name strongly enough to convert it direct to a scaleset name (e.g. we allow upper and lower cases - is that a problem?  Is Azure's case sensitivity in this space the same as ours?)

The second is trivially broken whenever the compute agent pool profile isn't called "compute", and I don't think we actually enforce this anywhere any more? (or do we?)

@asalkeld have you got time to take a look at this?  I think it's going to need a thoughtful approach to avoid getting this wrong again.

asalkeld:
closes #894

### DIFF
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -399,6 +399,9 @@ func (v *Validator) validateAgentPoolProfile(app AgentPoolProfile, vnet *net.IPN
 		}
 
 	case AgentPoolProfileRoleMaster:
+		if app.Name != string(app.Role) {
+			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].name %q", app.Name, app.Name))
+		}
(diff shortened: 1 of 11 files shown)


mjudeikis:
lgtm, will wait for a second pair of eyes as its not mine to review :)

kargakis:
Sorry, just noticed that it needs a rebase.

PR #918, opened 2018-12-17
Changing pointers to empty strings to nil pointers

### LINKED ISSUE #908: fake rp writes back the internal representation of the manifest
For example, the external request includes
```yaml
plan: {}
```
but after an update succeeds, the manifest is updated with the internal representation
```yaml
plan:
  name: ""
  product: ""
  promotionCode: ""
  publisher: ""
```
This was caused by https://github.com/openshift/openshift-azure/pull/821 where we updated the fake RP to use the internal representation of OpenShiftManagedCluster in order to handle multiple versions.

The actual problem is in `pkg/api/convertto20180930preview.go`. We should avoid returning pointers to empty strings. The same fix should be applied in `pkg/api/converttoadmin.go`.
/assign Makdaam 
/kind bug

Makdaam:
Fixes #908.

Added a small func to replace pointers to "" strings with nil. Wrapped all string pointers in converttoadmin.go and converttov20180930preview.go with it.

### DIFF
--- a/pkg/api/convertertov20180930preview.go
+++ b/pkg/api/convertertov20180930preview.go
@@ -4,48 +4,76 @@ import (
 	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
 )
 
+func nilIfv20180930ProvisioningStateEmpty(s *v20180930preview.ProvisioningState) *v20180930preview.ProvisioningState {
+	if s == nil || len(*s) == 0 {
+		return nil
+	}
+	return s
+}
+
+func nilIfv20180930AgentPoolProfileRoleEmpty(s *v20180930preview.AgentPoolProfileRole) *v20180930preview.AgentPoolProfileRole {
+	if s == nil || len(*s) == 0 {
+		return nil
+	}
+	return s
+}
+
+func nilIfv20180930VMSizeEmpty(s *v20180930preview.VMSize) *v20180930preview.VMSize {
+	if s == nil || len(*s) == 0 {
+		return nil
+	}
+	return s
+}
+
+func nilIfv20180930OSTypeEmpty(s *v20180930preview.OSType) *v20180930preview.OSType {
+	if s == nil || len(*s) == 0 {
+		return nil
+	}
+	return s
+}
+
 // ConvertToV20180930preview converts from an OpenShiftManagedCluster to a
 // v20180930preview.OpenShiftManagedCluster.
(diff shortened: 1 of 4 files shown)


Makdaam:
misclick

pkg/api/util.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+// nilIfStringEmpty returns a nil if the string is 0 length and the string pointer otherwise
+func nilIfStringEmpty(s *string) *string {
+	if len(*s) == 0 {
```
kargakis: To be on the safe side:
```go
if s == nil || len(*s) == 0 {
    return nil
}
```

kargakis:
Can you also add a unit test that proves #908 is fixed?

pkg/api/util.go
Makdaam: Thanks for catching this.

Makdaam:
@kargakis added unit tests for empty string to nil conversion.

kargakis:
lgtm, please squash to a single commit and I can tag it

Makdaam:
@kargakis squashed

pkg/api/convertertov20180930preview.go
```diff
@@ -8,44 +8,44 @@ import (
...
-		Publisher:     &cs.Plan.Publisher,
+		Name:          nilIfStringEmpty(&cs.Plan.Name),
+		Product:       nilIfStringEmpty(&cs.Plan.Product),
+		PromotionCode: nilIfStringEmpty(&cs.Plan.PromotionCode),
+		Publisher:     nilIfStringEmpty(&cs.Plan.Publisher),
 	}
 
 	provisioningState := v20180930preview.ProvisioningState(cs.Properties.ProvisioningState)
 	oc.Properties = &v20180930preview.Properties{
 		ProvisioningState: &provisioningState,
```
kargakis: Hrm, I think provisioningState, count, vmSize, osType, and role also need to be handled the same way?

Makdaam: Amended the commit with provisioningState, vmSize, osType and role *""→nil conversion. Also changed the test case a bit to include internal fields of AgentPoolProfile.

Makdaam: nvm, forgot about converttoadmin.go and _test. BRB with more amends.

Makdaam: amended with converttoadmin.go changes

Makdaam:
After discussion a different approach to fixing this issue has been chosen: 
- store the plan internally as pointers and pass it on as is
- don't modify other fields as the difference between an empty value and NULL value would be lost in conversion

Issue #835, opened 2018-12-04
NewPlugin validation check


sozercan:
NewPlugin includes validation check (`validateConfig`) with #740 without a similar flag to `Validate`'s `externalOnly`. 

We are running into an issue in the RP because frontend requires `GenerateConfig` and without any values, it will fail because of validation check. 

We would like to avoid `GenerateConfig` in the front end because it is a time consuming operation and calling generateConfig in both front and backend might cause issues.

If it works, can we move `validateConfig` to `Validate` and make it possible to avoid that path with `externalOnly` flag?

PR #922, opened 2018-12-17
pluginConfigAPI refactor

mjudeikis:
Work in progress. This allows to create cluster now, but everyhting else is still broken. Working gradually. 
- [x] Refactor converterters for AdminAPI, PluginAPI
- [x] Align naming of the imports
- [x] Unit tests
- [x] Add env variable overwrite to fakeRP and remove all from the main code
- [x] Add reflective unit testing
- [x] Fix existing unit tests
- [ ] Move sync image to context (separate PR)
- [ ] Split validators to have its own interfaces (https://github.com/openshift/openshift-azure/pull/1004)
- [x] Create template bootstrapper application for local development 

/hold
/cc @jim-minter @charlesakalugwu need your input to help formalize what we want to do.


pkg/api/plugin.go
```diff
@@ -122,14 +113,14 @@ type Plugin interface {
 	// (for update, upgrade, etc.)
 	// externalOnly indicates that fields set by the RP (FQDN and routerProfile.FQDN)
 	// should be excluded.
-	Validate(ctx context.Context, new, old *OpenShiftManagedCluster, externalOnly bool) []error
+	Validate(ctx context.Context, new, old *OpenShiftManagedCluster, pc PluginConfig, externalOnly bool) []error
```
mjudeikis: use config from validator, and not pass

mjudeikis: ValidateTemplate() as separate function

mjudeikis: Include yaml file and read the file for pluginConfig and use it as tempalte/base

pkg/api/plugin.go
```diff
@@ -70,48 +68,41 @@ type DeployFn func(context.Context, map[string]interface{}) error
...
-	SyncImage       string
 	AcceptLanguages []string
-	GenevaConfig    GenevaConfig
+	Config          OverrideConfig
 
 	TestConfig TestConfig
 }
 
+// OverrideConfig is passed to GenerateConfig interface to overwrite default values
+// it is cut-down version of Config structure
```
kargakis: Is this going to end up being the admin api?

pkg/api/plugin.go
kargakis: Why does PluginConfig have to be passed as an argument in the plugin functions vs being an internal implementation detail of the interface implementation?

kargakis: I see this is related to https://github.com/openshift/openshift-azure/issues/835 - nevermind

pkg/api/plugin.go
mjudeikis: More like Plugin config bootstrap template config

mjudeikis:
@jim-minter @kargakis @charlesakalugwu before I go even deeper to this rabbit hole and start covering with unit tests and validation, can you check this, please

pkg/api/plugin.go
```diff
@@ -69,48 +69,17 @@ type DeployFn func(context.Context, map[string]interface{}) error
 
 // PluginConfig is passed into NewPlugin
 type PluginConfig struct {
-	SyncImage       string
 	AcceptLanguages []string
```
jim-minter: eventually I think this should go in the context

pkg/api/plugin/api/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+// CertKeyPair is an rsa private key and x509 certificate pair.
+type CertKeyPair struct {
+	Key  *rsa.PrivateKey   `json:"key,omitempty"`
+	Cert *x509.Certificate `json:"cert,omitempty"`
+}
+
+// Certificate is an x509 certificate.
+type Certificate struct {
+	Cert *x509.Certificate `json:"cert,omitempty"`
```
jim-minter: used?

pkg/api/validate.go
```diff
@@ -231,6 +236,155 @@ func (v *Validator) Validate(new, old *OpenShiftManagedCluster, externalOnly boo
 	return nil
 }
 
+// ValidatePluginTemplate validates an Plugin API external template/config struct
+func (v *Validator) ValidatePluginTemplate(t *pluginapi.Config) (errs []error) {
```
jim-minter: worth putting these in a separate validate_plugin.go?

pkg/api/plugin.go
mjudeikis: metrics and etcbackup too?

pkg/config/type.go
```diff
@@ -0,0 +1,24 @@
+package config
```
jim-minter: check filename?

pkg/api/plugin/api/config.go
mjudeikis: I think as I just moved from pointers, no.

pkg/fakerp/config.go
```diff
@@ -55,44 +72,36 @@ func GetPluginConfig() (*api.PluginConfig, error) {
 	if err != nil {
 		return nil, err
 	}
+	template.Certificates.GenevaLogging.Cert = logCert
```
jim-minter: I'd be tempted to ditch the separate files and put it all in the template.yaml

jim-minter: Think about ease of onboarding new people

pkg/config/type.go
mjudeikis: what do you mean?

pkg/fakerp/fakerp.go
```diff
@@ -91,8 +91,23 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
...
+	if err != nil {
+		return nil, err
+	}
+
+	// if running under test we can overwrite certain images from local env variables
+	if config.TestConfig.RunningUnderTest {
+		overridePluginTemplate(template)
+	}
+
+	errs = p.ValidatePluginTemplate(ctx, template)
```
jim-minter: can you move this up to next to plugin.NewPlugin above?
can it be done once when the fakerp is initialised, not per createUpdate?

pkg/fakerp/config.go
mjudeikis: This means we will need to do the conversion to `base64` and all file becomes sensitive information and we can commit the only example or redacted one.

pkg/plugin/plugin.go
```diff
@@ -7,49 +7,37 @@ import (
...
-	apiValidator    *api.Validator
-	adminValidator  *api.Validator
+	log                     *logrus.Entry
+	config                  api.PluginConfig
+	clusterUpgrader         cluster.Upgrader
+	armGenerator            arm.Generator
+	configGenerator         config.Generator
+	apiValidator            *api.Validator
+	adminValidator          *api.Validator
+	pluginTemplateValidator *api.Validator
```
jim-minter: 2 of these is bad, 3 is worse.  Do these different validations really want to be off the same api.Validator base type?

mjudeikis: our current Validator is super simple. If it becomes different - we can split. 
What the alternative? Create a separate base type?

pkg/fakerp/config.go
jim-minter: oh, yuck, I see, the issue is being able to check in the basics for MSFT, but keep the secrets separate.  Can you think how to try and make this as neat as poss for us and for MSFT?

pkg/config/type.go
jim-minter: types.go is standard?

pkg/fakerp/config.go
mjudeikis: We could have a small helper to populate file locally from example file. 
So we run it to "bootstrap" config once. This would basically do same as current code is doing.

pkg/plugin/plugin.go
jim-minter: yes - what does moving in the direction of separate file and base type for each separate api package type validated look like?

pkg/config/type.go
mjudeikis: ou :) my bad

pkg/plugin/plugin.go
mjudeikis: I might do it in the second PR as it will make this even more complicated after test. Will raise Issue and assign

pkg/fakerp/fakerp.go
mjudeikis: Will move this up. At the moment `createOrUpdate` represents RP process. we migh want to change name to make it more representative

mjudeikis:
To populate template added new helper `hack/populate-template/populate-template.go` which will take in `template.yaml.example.xxxx` and populate it with sensitive data from `secrets` folder. This should be used in tests and development. Can be exected with `make populate-template`. fakeRP will expect file `template.yaml` to be present from this command.

pkg/api/plugin.go
mjudeikis: https://github.com/openshift/openshift-azure/issues/966

pkg/plugin/plugin.go
mjudeikis: follow-up: https://github.com/openshift/openshift-azure/issues/967

test/templates/template.yaml.example.311.43
```diff
@@ -0,0 +1,50 @@
...
+  prometheusConfigReloaderBase: registry.access.redhat.com/openshift3/ose-prometheus-config-reloader
+  prometheusOperatorBase: registry.access.redhat.com/openshift3/ose-prometheus-operator
+  registry: registry.access.redhat.com/openshift3/ose-docker-registry:v3.11.43
+  registryConsole: registry.access.redhat.com/openshift3/registry-console:v3.11.43
+  router: registry.access.redhat.com/openshift3/ose-haproxy-router:v3.11.43
+  serviceCatalog: registry.access.redhat.com/openshift3/ose-service-catalog:v3.11.43
+  sync: quay.io/openshift-on-azure/sync:latest
+  templateServiceBroker: registry.access.redhat.com/openshift3/ose-template-service-broker:v3.11.43
+  webConsole: registry.access.redhat.com/openshift3/ose-web-console:v3.11.43
+runningUnderTest: true
```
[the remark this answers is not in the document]
kwoodson: With @asalkeld's PR for removing the locally created users and moving to a aad group sync I believe this can be removed.  In order to move this along you might want to do in a follow up.

mjudeikis:
FYI @sozercan @amanohar @shrutir25

mjudeikis:
2 failures in a row of https://github.com/openshift/openshift-azure/issues/973

pkg/plugin/plugin.go
```diff
@@ -9,49 +9,37 @@ import (
...
 }
 
 var _ api.Plugin = &plugin{}
 
 // NewPlugin creates a new plugin instance
 func NewPlugin(log *logrus.Entry, pluginConfig *api.PluginConfig, skipValidate ...bool) (api.Plugin, []error) {
```
kargakis: skipValidate is now defunct - is this expected?

mjudeikis: yes, we need to remove it. It was added as a temporary workaround

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
+package api
+
+import (
+	plugin "github.com/openshift/openshift-azure/pkg/api/plugin/api"
+)
+
+func ConvertFromPlugin(in plugin.Config) Config {
+	out := Config{}
```
jim-minter:
```golang
out := Config{
  ImageOffer: in.ImageOffer,
  ...
}
```

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
+package api
```
jim-minter: add simple unit test probably using populate.Walk()

pkg/api/plugin/api/config.go
```diff
@@ -0,0 +1,84 @@
...
+// CertificateConfig contains all certificate configuration for the cluster.
+type CertificateConfig struct {
+	// geneva integration certificates
+	GenevaLogging CertKeyPair `json:"genevaLogging,omitempty"`
+	GenevaMetrics CertKeyPair `json:"genevaMetrics,omitempty"`
+}
+
+// CertKeyPair is an rsa private key and x509 certificate pair.
+type CertKeyPair struct {
+	Key  *rsa.PrivateKey   `json:"key,omitempty"`
```
jim-minter: I think these needn't be pointers (need to cross-check what the internal API is doing).  Open a follow-up if not straightforward (probably so).

pkg/api/types_test.go
```diff
@@ -485,3 +487,79 @@ restartCheck:
...
+		regexp.MustCompile(`^\.ConfigStorageAccount$`),
+		regexp.MustCompile(`^\.RegistryStorageAccount$`),
+		regexp.MustCompile(`^\.Service`),
+		regexp.MustCompile(`^\.Session`),
+		regexp.MustCompile(`^\.Registry`),
+		regexp.MustCompile(`^\.Prometheus`),
+		regexp.MustCompile(`^\.Alert`),
+		regexp.MustCompile(`^\.Console`),
+		regexp.MustCompile(`^\.RunningUnderTest`),
+		regexp.MustCompile(`^\.Certificates.Ca\.`),
```
jim-minter:
```
regexp.MustCompile(`^\.Certificates\.Ca\.`),
```
(missing backslash)

pkg/api/validate_plugin.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+	"fmt"
+	"regexp"
+
+	pluginapi "github.com/openshift/openshift-azure/pkg/api/plugin/api"
+)
+
+var (
+	// Image version format check x{3}.y{4}.z{8}
+	imageVersion = regexp.MustCompile(`^[0-9]{3}.[0-9]{1,4}.[0-9]{8}$`)
```
jim-minter: correct or remove comment above

pkg/api/validate_plugin_test.go
```diff
@@ -0,0 +1,345 @@
...
+			},
+			expectedErrs: []error{errors.New(`invalid ImageSKU "osa_310"`)},
+		},
+		"invalid ImageVersion": {
+			f: func(t *pluginapi.Config) {
+				t.ImageVersion = "1.23.456"
+			},
+			expectedErrs: []error{errors.New(`invalid ImageVersion "1.23.456"`)},
+		},
+		"empty GenevaLoggingSector": {
```
jim-minter: try separating this to a different test / part of a different test - validate an empty file and check the list of errors returned once?

pkg/config/images.go
```diff
@@ -39,12 +21,10 @@ func openShiftVersion(imageVersion string) (string, error) {
 func (g *simpleGenerator) selectNodeImage(cs *api.OpenShiftManagedCluster) {
 	c := &cs.Config
 	c.ImagePublisher = "redhat"
```
jim-minter: should go away? double check all of these

pkg/fakerp/fakerp.go
```diff
@@ -58,6 +58,21 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
...
 	}
 
+	template, err := GetPluginTemplate()
+	if err != nil {
+		return nil, err
+	}
+
+	// if running under test we can overwrite certain images from local env variables
+	if config.TestConfig.RunningUnderTest {
+		overridePluginTemplate(template)
```
jim-minter: think this shouldn't be wrapped by the above if statement

pkg/api/validate_plugin.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	// validate certificates
+	errs = append(errs, v.validatePluginTemplateCertificates(t.Certificates)...)
+	errs = append(errs, v.validatePluginTemplateImages(t.Images)...)
+	return errs
+}
+
+func (v *Validator) validatePluginTemplateCertificates(c pluginapi.CertificateConfig) (errs []error) {
+	if c.GenevaLogging.Key == nil {
+		errs = append(errs, fmt.Errorf("invalid GenevaLogging key"))
```
kargakis: maybe align these error messages with those in `validatePluginTemplateImages` so something like "GenevaLogging key cannot be empty".

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	plugin "github.com/openshift/openshift-azure/pkg/api/plugin/api"
+)
+
+func ConvertFromPlugin(in plugin.Config) Config {
```
kargakis: This is more of a merging/overriding function than a conversion and since there is not a conversion back, I would rename this to reflect more appropriately how we use it. Also some godoc would be nice.

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
...
+	if in.GenevaLoggingControlPlaneAccount != "" {
+		out.GenevaLoggingControlPlaneAccount = in.GenevaLoggingControlPlaneAccount
+	}
+	return out
+}
+
+func mergePluginCertificateConfig(in plugin.CertificateConfig, out *CertificateConfig) {
+	mergePluginCertKeyPair(in.GenevaLogging, &out.GenevaLogging)
+	mergePluginCertKeyPair(in.GenevaMetrics, &out.GenevaMetrics)
+	return
```
kargakis: no need for return

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func mergePluginCertKeyPair(in plugin.CertKeyPair, out *CertKeyPair) {
+	if in.Cert != nil {
+		out.Cert = in.Cert
+		out.Key = in.Key
+	}
+	return
```
kargakis: no need for return

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,156 @@
...
+func mergePluginCertificateConfig(in plugin.CertificateConfig, out *CertificateConfig) {
+	mergePluginCertKeyPair(in.GenevaLogging, &out.GenevaLogging)
+	mergePluginCertKeyPair(in.GenevaMetrics, &out.GenevaMetrics)
+	return
+}
+
+func mergePluginCertKeyPair(in plugin.CertKeyPair, out *CertKeyPair) {
+	if in.Cert != nil {
+		out.Cert = in.Cert
+		out.Key = in.Key
```
kargakis: I would expect you would merge key appropriately here.

pkg/config/images.go
```diff
@@ -39,12 +21,10 @@ func openShiftVersion(imageVersion string) (string, error) {
 func (g *simpleGenerator) selectNodeImage(cs *api.OpenShiftManagedCluster) {
```
kargakis: I think it's worth renaming this to denote that we are only doing defaulting now. `setDefaults` or something across those lines.

pkg/config/images.go
kargakis: ImagePublisher is overriden by the config template so you probably want to set this here in case it's empty (otherwise the template value is never applied correctly).

pkg/config/images.go
mjudeikis: we selecting nodeImage version here explicitly, so I think name corresponds ok :)

pkg/config/images.go
mjudeikis: deleted in recent changes.

pkg/api/plugin/api/config.go
mjudeikis: https://github.com/openshift/openshift-azure/issues/986

pkg/plugin/plugin.go
kargakis: are you going to clean this up in this PR or in a follow up? If follow up, mind opening an issue to track?

pkg/config/images.go
```diff
@@ -39,12 +21,10 @@ func openShiftVersion(imageVersion string) (string, error) {
...
 	c.ImagePublisher = "redhat"
-	c.ImageOffer = g.pluginConfig.TestConfig.ImageOffer
 	if c.ImageOffer == "" {
 		c.ImageOffer = "osa"
 	}
 
-	c.ImageVersion = g.pluginConfig.TestConfig.ImageVersion
 	switch g.pluginConfig.TestConfig.DeployOS {
 	case "", "rhel7":
 		c.ImageSKU = "osa_" + strings.Replace(cs.Properties.OpenShiftVersion[1:], ".", "", -1)
```
kargakis: You should set this too only if it's empty, right?

kargakis: Ditto for centos7

pkg/config/images.go
kargakis: "select" is pretty vague, what you actually do is set defaults in case necessary values are not set by the merged plugin template.

pkg/config/images.go
mjudeikis: This would change current flow of how we define our images. But yes, this does not look right now. Let me think about this :man_shrugging:

pkg/api/validate_plugin.go
mjudeikis: done

pkg/config/images.go
mjudeikis: I think we might just remove this all selectNodeImages and make it default for PluginConfig

pkg/config/images.go
mjudeikis: I think we dont need this anymore. We pass those from plugin config now

pkg/plugin/plugin.go
mjudeikis: in this one.

pkg/fakerp/config.go
mjudeikis: moved back to RP

pkg/api/converterfromplugin.go
mjudeikis: added

pkg/api/converterfromplugin.go
mjudeikis: I agree this is not conversion in generic view. But it does convert one type to other. And it aligns to other files... @jim-minter  WDYT?

kargakis: Ah, misread this, yeah it's a conversion, nvmd

pluginconfig/pluginconfig-311.yaml
```diff
@@ -0,0 +1,59 @@
+##
```
mjudeikis: @kargakis @jim-minter PTAL to this example/template and doc please :)

pluginconfig/pluginconfig-311.yaml
```diff
@@ -0,0 +1,59 @@
...
+## Plugin config template. 
+## This template file is manually curated for the particural OSA version
```
kargakis: particular

pluginconfig/pluginconfig-311.yaml
```diff
@@ -0,0 +1,59 @@
...
+##
+## certificates, used to autenticate to external systems
```
kargakis: authenticate

pluginconfig/pluginconfig-311.yaml
```diff
@@ -0,0 +1,59 @@
...
+## Geneva intergration certificates. Example:
```
kargakis: integration

pkg/config/generate.go
```diff
@@ -32,15 +33,10 @@ func createUserHtPassEntry(name string, passwd *string, htPasswd []byte) ([]byte
 	return htPassEntry, nil
 }
 
-func (g *simpleGenerator) Generate(cs *api.OpenShiftManagedCluster) (err error) {
+func (g *simpleGenerator) Generate(cs *api.OpenShiftManagedCluster, template *pluginapi.Config) (err error) {
+	cs.Config = api.ConvertFromPlugin(*template)
```
kargakis: This line is problematic for upgrades, that's why the re-entrancy test fails to find the storage blob. You should be able to reuse the existing config and merge on top of it the template.

kargakis: At this point, cs.Config is already populated with the old config so really this function should not be a conversion but a merger.

mjudeikis: ack! thanks for saving me a hassle :)

mjudeikis:
@kargakis added preservation of old config for updates. Tried to operate with primitives as much as possible, but was not able to avoid references and pointers fully :/

pkg/config/generate.go
```diff
@@ -32,15 +33,10 @@ func createUserHtPassEntry(name string, passwd *string, htPasswd []byte) ([]byte
 	return htPassEntry, nil
 }
 
-func (g *simpleGenerator) Generate(cs *api.OpenShiftManagedCluster) (err error) {
+func (g *simpleGenerator) Generate(cs *api.OpenShiftManagedCluster, template *pluginapi.Config) (err error) {
+	cs.Config = api.ConvertFromPlugin(template, &cs.Config)
```
kargakis: &cs.Config is never going to be nil. It's always a pointer to an empty struct in case of a create.

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	plugin "github.com/openshift/openshift-azure/pkg/api/plugin/api"
+)
+
+// ConvertFromPlugin converts external plugin API config type into
+// internal API Config type
+func ConvertFromPlugin(in *plugin.Config, old *Config) Config {
+	cs := &Config{}
+	// preserve old config if this is update
+	if old != nil {
```
kargakis: old is never going to be nil the way you use it now (see my earlier comment) but this is probably fine since `cs` is either old or a pointer to an empty struct?

mjudeikis: hmm. Should we just skip `old !=nil` and `deepCopy` always?

pkg/fakerp/config.go
```diff
@@ -55,44 +75,49 @@ func GetPluginConfig() (*api.PluginConfig, error) {
...
-	return &api.PluginConfig{
-		SyncImage:       syncImage,
-		AcceptLanguages: []string{"en-us"},
-		TestConfig:      tc,
-		GenevaConfig:    genevaConfig,
-	}, nil
+	return []byte{}, fmt.Errorf("file %s does not exist", path)
+}
+
+func fileExist(path string) bool {
```
kargakis: can you revert this diff? No point in reorganizing functions when your PR is already that big. Same for `readFile`.

pkg/fakerp/config.go
```diff
@@ -8,30 +8,50 @@ import (
...
+	}, nil
+}
+
+func GetPluginTemplate() (*pluginapi.Config, error) {
+	// read template file without secrets
+	artifactDir, err := shared.FindDirectory(PluginConfigDirectory)
+	if err != nil {
+		return nil, err
+	}
+	data, err := readFile(filepath.Join(artifactDir, "pluginconfig-311.yaml"))
```
kargakis: This doesn't look right to me. Shouldn't we map plugin configs to versions and pick the correct config based on the version we are deploying?

mjudeikis: we had this discussion with @jim-minter. If I got this right, the plan is to maintain a minimal number of files `311, 4.0` and managed them via history/tags. More architecture topic discussion, not relevant for now as we dont know yet full implementation details

pkg/api/converterfromplugin.go
kargakis: I like the nil check protection here.

pkg/api/converterfromplugin.go
```diff
@@ -0,0 +1,64 @@
...
+	cs.Images.AnsibleServiceBroker = in.Images.AnsibleServiceBroker
+	cs.Images.WebConsole = in.Images.WebConsole
+	cs.Images.Console = in.Images.Console
+	cs.Images.EtcdBackup = in.Images.EtcdBackup
+	cs.Images.GenevaLogging = in.Images.GenevaLogging
+	cs.Images.GenevaTDAgent = in.Images.GenevaTDAgent
+	cs.Images.GenevaStatsd = in.Images.GenevaStatsd
+	cs.Images.MetricsBridge = in.Images.MetricsBridge
+
+	return *cs
```
kargakis: I think you should be returning a pointer instead of a struct and callers can decide to deref. This is also consistent with the rest of our conversion funcs.

mjudeikis: This would make this looks like this:
```
c := api.ConvertFromPlugin(template, &cs.Config)
cs.Config = *c
c := &cs.Config
```

If this fine, will add the comment and that we need to assign a pointer value to the existing structure, and later we use reference in below assignments.

mjudeikis:
flake: https://github.com/openshift/ci-operator/issues/239

mjudeikis:
Ping @jim-minter

mjudeikis:
/test e2e
Flake: https://github.com/openshift/openshift-azure/issues/1007

kargakis:
tests are green, tide is a bit slow so I am going to merge manually

kargakis:
heh, tide was faster than me

PR #928, opened 2018-12-17
 metrics forwarding infrastructure should use unix domain socket for reliability and security

jim-minter:


mjudeikis:
/lgtm
 will wait for succ test

PR #929, opened 2018-12-17
add lint-addons.go, this broadly checks that pkg/addons/data files are in a canonical format

jim-minter:
in addition, this PR fixes a number of bugs which were causing the sync pod to erroneously detect differences and unnecessarily update objects every cycle


asalkeld:
looking good, just waiting for the tests to pass
```

# github.com/openshift/openshift-azure/pkg/addons [github.com/openshift/openshift-azure/pkg/addons.test]
pkg/addons/addons_test.go:15:12: undefined: unmarshal
pkg/addons/clean_test.go:24:13: undefined: unmarshal
pkg/addons/clean_test.go:33:13: undefined: unmarshal
```

pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
```diff
@@ -9,71 +9,68 @@ spec:
   selector:
     matchLabels:
       app: mdsd-shard
-  updateStrategy:
-    type: RollingUpdate
```
mjudeikis: We might want to add so it would not clean UpdateStrategy.

pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
```diff
@@ -9,71 +9,68 @@ spec:
...
   template:
     metadata:
       labels:
         app: mdsd-shard
     spec:
-      tolerations:
```
mjudeikis: Same as above

pkg/addons/default.go
```diff
@@ -43,33 +43,37 @@ func Default(o unstructured.Unstructured) {
...
 
-		defaultUpdateStrategy := map[string]interface{}{
-			"rollingUpdate": map[string]interface{}{
-				"maxUnavailable": int64(1),
-			},
-			"type": "RollingUpdate",
-		}
-		jsonpath.MustCompile("$.spec.updateStrategy").DeleteIfMatch(o.Object, defaultUpdateStrategy)
+		jsonpath.MustCompile("$.spec.updateStrategy.rollingUpdate.maxUnavailable").DeleteIfMatch(o.Object, int64(1))
+		jsonpath.MustCompile("$.spec.updateStrategy.rollingUpdate").DeleteIfMatch(o.Object, map[string]interface{}{})
```
mjudeikis: Why are we deleting strategy? It might be (and in some cases are) component specific?

pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
kargakis: FWIW, if updateStrategy is missing in apps/v1, [it gets defaulted to RollingUpdate](https://github.com/kubernetes/kubernetes/blob/3dbf0abddecdeef5956466b50d407399f0e447f2/pkg/apis/apps/v1/defaults.go#L75).

pkg/addons/data/Deployment.apps/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -7,79 +7,72 @@ metadata:
   namespace: openshift-azure-monitoring
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: metrics-bridge
   strategy:
-    type: RollingUpdate
     rollingUpdate:
```
kargakis: If you are relying on defaults then you can also remove these rolling parameters.

pkg/addons/data/Deployment.apps/openshift-monitoring/cluster-monitoring-operator.yaml
```diff
@@ -14,15 +14,14 @@ spec:
     rollingUpdate:
       maxSurge: 1
```
kargakis: I am wondering what does it means to be running two different versions of the cluster monitoring operator at the same time.

pkg/addons/default.go
kargakis: It gets defaulted to RollingUpdate for all workload controllers in apps/v1 (deployments, daemonsets, statefulsets).

kargakis:
Closing in favor of https://github.com/openshift/openshift-azure/pull/979

Issue #919, opened 2018-12-17
should deploy a template and ensure a given text is in the contents 


kargakis:
```
Openshift on Azure end user e2e tests [EndUser][Fake] 
  should deploy a template and ensure a given text is in the contents
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:78
STEP: instantiating the template and getting the route (2018-12-17 11:23:18.945608688 +0000 UTC m=+689.489019541)
STEP: hitting the route and checking the contents (2018-12-17 11:24:21.532497235 +0000 UTC m=+752.075908132)

• Failure [115.529 seconds]
Openshift on Azure end user e2e tests [EndUser][Fake]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:29
  should deploy a template and ensure a given text is in the contents [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:78

  Expected error:
      <*errors.errorString | 0xc0000858b0>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
  not to have occurred

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:96
```
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/917/pull-ci-openshift-openshift-azure-master-e2e-azure/1429

/kind test-flake

kargakis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/917/pull-ci-openshift-openshift-azure-master-e2e-azure/1435

kargakis:
To reduce the flakiness of this test I would suggest deploying a standalone Deployment (no Builds, no Templates) that pulls nginx from our quay registry.

kargakis:
Another one: https://github.com/openshift/openshift-azure/pull/940#issuecomment-449049384

build failure

we should really get rid of running builds as part of the OSA merge queue.

kargakis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/992/pull-ci-openshift-openshift-azure-master-e2e-azure/1748

charlesakalugwu:
[...]

PR #930, opened 2018-12-18
random flake timeouts

mjudeikis:
Increase timeouts for tests to reduce flakes

### DIFF
--- a/test/e2e/specs/end_user.go
+++ b/test/e2e/specs/end_user.go
@@ -90,7 +90,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser][Fake]", func()
 
 		// Curl the endpoint and search for a string
 		By(fmt.Sprintf("hitting the route and checking the contents (%v)", time.Now()))
-		timeout, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		timeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 		defer cancel()
 		resp, err := waitutil.ForHTTPStatusOk(timeout, log, nil, url)
 		Expect(err).NotTo(HaveOccurred())
@@ -150,7 +150,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser][Fake]", func()
 		prevCounter := 0
 
 		loopHTTPGet := func(url string, regex *regexp.Regexp, times int) error {
-			timeout, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+			timeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 			defer cancel()
 
 			for i := 0; i < times; i++ {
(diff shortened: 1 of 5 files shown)


test/e2e/specs/end_user.go
```diff
@@ -90,7 +90,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser][Fake]", func()
 
 		// Curl the endpoint and search for a string
 		By(fmt.Sprintf("hitting the route and checking the contents (%v)", time.Now()))
-		timeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		timeout, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
mjudeikis: This timeouts quite often and is causing flakes. This might be due DNS not being registered, propagated on time, Route not being admited to router.

mjudeikis:
/cc @jim-minter 
wdyt as I think its causing this one: https://github.com/openshift/openshift-azure/issues/919

test/e2e/specs/fakerp/admin.go
```diff
@@ -92,6 +92,11 @@ var _ = Describe("Openshift on Azure admin e2e tests [AzureClusterReader][Fake]"
 		Expect(err).ToNot(HaveOccurred())
 	})
 
+	It("should start node-exporter correctly", func() {
```
kargakis: Is the node exporter depending on any other component in the stack in order to get deployed?

test/e2e/specs/fakerp/admin.go
```diff
@@ -92,6 +92,11 @@ var _ = Describe("Openshift on Azure admin e2e tests [AzureClusterReader][Fake]"
...
+		err := wait.Poll(2*time.Second, 20*time.Minute, ready.DaemonSetIsReady(cli.AppsV1.DaemonSets("openshift-monitoring"), "node-exporter"))
```
kwoodson: Unrelated to PR but I didn't realize we are waiting 20 minutes for a ds deployment.  That seems high.

test/e2e/specs/end_user.go
jim-minter: @mjudeikis I would rather see temporary diagnostic logging added to try and work out which of the above cases it is, then if applicable ensure that the plugin health check doesn't release a cluster until everything is in order.
Note that in 7930c2b6199562d1a4c013c510666fb329e1bb97 @thekad made forHTTPStatusOK return the HTTP body when the request succeeds.  We supposed a) that there is no way of telling when a route is admitted to a router, and b) there are 2 routers and just because a route is admitted to one doesn't mean it's available on the other.  That commit was supposed to deal with that problem.  So I'm curious at this point to see what the root cause is.

test/e2e/specs/fakerp/admin.go
mjudeikis: It deploys just after alert-manager in sequence, before node-exporter.

Issue #924, opened 2018-12-17
fake rp now runs asynchronously in hack scripts and needs cleanup


kargakis:
After https://github.com/openshift/openshift-azure/pull/913, we started running the fake RP into its own process in the hack scripts, resulting in the extra process leaking after the scripts are finished.

/kind bug

kargakis:
One solution in the `hack/fakerp.sh` script before running cmd/fakerp.go is
```
fuser -k 8080/tcp
```

kargakis:
Another solution is to create an `/exit` endpoint and use that in all hack scripts that start the fake rp asynchronously - see https://github.com/openshift/openshift-azure/pull/926

PR #927, opened 2018-12-17
Kill any process holding 8080

kargakis:
Another possible solution for https://github.com/openshift/openshift-azure/issues/924

Alternative to https://github.com/openshift/openshift-azure/pull/926

PR #926, opened 2018-12-17
Shutdown the fake RP via an /exit endpoint

kargakis:
One possible solution for https://github.com/openshift/openshift-azure/issues/924

PR #931, opened 2018-12-18
Trap killing process on 8080

kargakis:
@jim-minter @mjudeikis @charlesakalugwu this is the best approach I can come up with that does not involve:
* adding extra logic in the fake RP (https://github.com/openshift/openshift-azure/pull/926)
* using less cross-platform tools like `fuser` (https://github.com/openshift/openshift-azure/pull/927)

I am also trying to workaround the fact that we are using bash by replacing the callout to `hack/fakerp.sh` with using `go run` directly because running bash inside of bash is super slow and the $(lsof) subshell in the trap may not resolve correctly since the server has yet to start. The current approach is still a little racy but it avoids using weird wait logic for the server to come up in order to get the pid. Also, for some reason the process id returned from the async process execution `go run cmd/fakerp/main.go &` is different from the process id listening on the port, hence the use of `lsof`.

Fixes https://github.com/openshift/openshift-azure/issues/924

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -16,7 +16,10 @@ USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
+
+trap 'set +ex; return_id=$?; kill $(lsof -t -i :8080); wait; exit $return_id' EXIT
 
 go run cmd/createorupdate/createorupdate.go $USE_PROD_FLAG
--- a/hack/delete.sh
+++ b/hack/delete.sh
@@ -15,7 +15,10 @@ USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
+
+trap 'set +ex; return_id=$?; kill $(lsof -t -i :8080); wait; exit $return_id' EXIT
 
 go run cmd/createorupdate/createorupdate.go -request=DELETE $USE_PROD_FLAG
--- a/hack/upgrade.sh
+++ b/hack/upgrade.sh
@@ -14,11 +14,13 @@ fi
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     TEST_IN_PRODUCTION="-use-prod=true"
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
 if [[ -n "$ADMIN_MANIFEST" ]]; then
     ADMIN_MANIFEST="-admin-manifest=$ADMIN_MANIFEST"
 fi
 
+trap 'set +ex; return_id=$?; kill $(lsof -t -i :8080); wait; exit $return_id' EXIT
 
 go run cmd/createorupdate/createorupdate.go -timeout 1h ${TEST_IN_PRODUCTION:-} ${ADMIN_MANIFEST:-}


hack/create.sh
```diff
@@ -16,7 +16,10 @@ USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
 
+trap 'return_id=$?; kill $(lsof -t -i :8080); exit $return_id' EXIT
```
charlesakalugwu: do `return_id=$?;  kill $(lsof -t -i :8080); wait $(lsof -t -i :8080) 2>/dev/null; exit $return_id` instead to suppress the `signal terminated` message

hack/delete.sh
```diff
@@ -15,7 +15,10 @@ USE_PROD_FLAG="-use-prod=false"
 if [[ -n "$TEST_IN_PRODUCTION" ]]; then
     USE_PROD_FLAG="-use-prod=true"
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
 
+trap 'return_id=$?; kill $(lsof -t -i :8080); exit $return_id' EXIT
```
charlesakalugwu: same here

hack/upgrade.sh
```diff
@@ -14,11 +14,13 @@ fi
...
 else
-	hack/fakerp.sh $RESOURCEGROUP &
+    go generate ./...
+    go run cmd/fakerp/main.go &
 fi
 if [[ -n "$ADMIN_MANIFEST" ]]; then
     ADMIN_MANIFEST="-admin-manifest=$ADMIN_MANIFEST"
 fi
 
+trap 'return_id=$?; kill $(lsof -t -i :8080); exit $return_id' EXIT
```
charlesakalugwu: same here

kargakis:

/retest

hack/create.sh
kargakis: What's the point of the `wait` command?

charlesakalugwu: without wait, `signal terminated` is returned in the bash terminal 

```
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ go run cmd/fakerp/main.go &
++ generate_command_executed_sequence
++ printf '\e\7'
[1] 34951
+ go run cmd/fakerp/main.go
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ INFO[2018-12-18T13:29:45+01:00] starting the fake resource provider          
INFO[2018-12-18T13:29:45+01:00] starting server on localhost:8080            
^C
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ kill $(lsof -t -i :8080)
++ generate_command_executed_sequence
++ printf '\e\7'
++ lsof -t -i :8080
+ kill 35193
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ signal: terminated
^C
[1]+  Exit 1                  go run cmd/fakerp/main.go
```

with wait

```
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ go run cmd/fakerp/main.go &
++ generate_command_executed_sequence
++ printf '\e\7'
[1] 35360
+ go run cmd/fakerp/main.go
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ INFO[2018-12-18T13:32:39+01:00] starting the fake resource provider          
INFO[2018-12-18T13:32:39+01:00] starting server on localhost:8080            
^C
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ kill $(lsof -t -i :8080); wait $(lsof -t -i :8080)
++ generate_command_executed_sequence
++ printf '\e\7'
++ lsof -t -i :8080
+ kill 35593
++ generate_command_executed_sequence
++ printf '\e\7'
++ lsof -t -i :8080
signal: terminated
+ wait
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ 
```

charlesakalugwu: fwiw, i used a debian system to test the changes. on fedora your changes work fine.

thekad: more specifically, the wait command will wait for the given subprocess id to finish or, if no PID specified, will wait for all subprocesses spawned from the current shell to finish, sample:

```bash
#!/bin/bash

for x in {1..10};
do
  bash -c "sleep $x && echo \"Done waiting $x seconds\"" &
done

echo "Waiting..."
wait
echo "Done."
```

kargakis: Added `wait`. FWIW, the message is still not supressed so I removed the stderr redirection to /dev/null.

PR #932, opened 2018-12-18
adding VnetID field to APIs and code to populate it after deploy

### LINKED ISSUE #544: Returning vnet name
It might be good to return the full virtual network name, so CLI or other clients can initiate peering using this value.

cc @amanohar

Makdaam:
Fixes #544 
Adding the VnetID field to all 3 APIs, converters, validator, unit tests. Also the code to populate the field + subset of the azure Network client.

### DIFF
--- a/pkg/api/2018-09-30-preview/api/types.go
+++ b/pkg/api/2018-09-30-preview/api/types.go
@@ -89,9 +89,14 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+	// VnetCIDR (in): the CIDR with which the OSA cluster's Vnet is configured
 	VnetCIDR *string `json:"vnetCidr,omitempty"`
 
-	// PeerVnetID is expected to be empty or match
+	// VnetID (out): the ID of the Vnet created for the OSA cluster
+	VnetID *string `json:"vnetId,omitempty"`
+
+	// PeerVnetID (in, optional): ID of a Vnet with which the OSA cluster Vnet should be peered.
+	// If non-empty, this should match
 	// `^/subscriptions/[^/]+
 	//   /resourceGroups/[^/]+
 	//   /providers/Microsoft.Network
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -86,9 +86,14 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+	// VnetCIDR (in): the CIDR with which the OSA cluster's Vnet is configured
 	VnetCIDR string `json:"vnetCidr,omitempty"`
 
-	// PeerVnetID is expected to be empty or match
+	// VnetID (out): the ID of the Vnet created for the OSA cluster
+	VnetID string `json:"vnetId,omitempty"`
+
+	// PeerVnetID (in, optional): ID of a Vnet with which the OSA cluster Vnet should be peered.
+	// If non-empty, this should match
 	// `^/subscriptions/[^/]+
 	//   /resourceGroups/[^/]+
 	//   /providers/Microsoft.Network
--- a/pkg/api/validate.go
+++ b/pkg/api/validate.go
@@ -454,6 +454,9 @@ func (v *Validator) validateNetworkProfile(np *NetworkProfile) (errs []error) {
 		errs = append(errs, fmt.Errorf("networkProfile cannot be nil"))
 		return
 	}
+	if np.VnetID != "" && !rxVNetID.MatchString(np.VnetID) {
+		errs = append(errs, fmt.Errorf("invalid properties.networkProfile.vnetId %q", np.VnetID))
+	}
 	if !isValidIPV4CIDR(np.VnetCIDR) {
 		errs = append(errs, fmt.Errorf("invalid properties.networkProfile.vnetCidr %q", np.VnetCIDR))
 	}
--- a/pkg/api/validate_test.go
+++ b/pkg/api/validate_test.go
@@ -165,6 +165,17 @@ func TestValidate(t *testing.T) {
 			},
 			expectedErrs: []error{errors.New(`invalid properties.publicHostname "www.example.com"`)},
 		},
+		"network profile valid VnetId": {
+			f: func(oc *OpenShiftManagedCluster) {
+				oc.Properties.NetworkProfile.VnetID = "/subscriptions/b07e8fae-2f3f-4769-8fa8-8570b426ba13/resourceGroups/test/providers/Microsoft.Network/virtualNetworks/vnet"
+			},
+		},
+		"network profile invalid VnetId": {
+			f: func(oc *OpenShiftManagedCluster) {
+				oc.Properties.NetworkProfile.VnetID = "foo"
+			},
+			expectedErrs: []error{errors.New(`invalid properties.networkProfile.vnetId "foo"`)},
+		},
 		"network profile bad vnetCidr": {
 			f: func(oc *OpenShiftManagedCluster) {
 				oc.Properties.NetworkProfile.VnetCIDR = "foo"
--- a/pkg/arm/resources.go
+++ b/pkg/arm/resources.go
@@ -2,7 +2,6 @@ package arm
 
 import (
 	"encoding/base64"
-	"fmt"
 	"sort"
 	"strings"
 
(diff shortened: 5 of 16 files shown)


Makdaam:
/hold
for review of API changes by more people

Makdaam:
rebased,
switched from populating the VnetID via an azureclient request to populating while the ARM template is generated (less requests to the API).

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -89,6 +89,14 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+	// VnetID is expected to be empty or match
```
kwoodson: Could we line up VnetID and PeerVnetID so they can share the comment?

pkg/api/types.go
```diff
@@ -86,6 +86,14 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+	// VnetID is expected to be empty or match
```
kwoodson: Same as above.

pkg/api/validate_test.go
```diff
@@ -165,6 +165,17 @@ func TestValidate(t *testing.T) {
 			},
 			expectedErrs: []error{errors.New(`invalid properties.publicHostname "www.example.com"`)},
 		},
+		"network profile valid VnetId": {
+			f: func(oc *OpenShiftManagedCluster) {
+				oc.Properties.NetworkProfile.VnetID = "/subscriptions/b07e8fae-2f3f-4769-8fa8-8570b426ba13/resourceGroups/test/providers/Microsoft.Network/virtualNetworks/vnet"
```
kwoodson: Not sure if this subscription id is real or not but we should make sure that it isn't.

pkg/api/types.go
Makdaam: Moved VnetID field below VnetCIDR, edited the comment to describe contents of both fields.

pkg/api/2018-09-30-preview/api/types.go
Makdaam: Moved VnetID field below VnetCIDR, edited the comment to describe contents of both fields.

pkg/api/types.go
```diff
@@ -86,13 +86,16 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+
```
kargakis: nit: kill this newline

pkg/api/validate.go
```diff
@@ -454,6 +454,9 @@ func (v *Validator) validateNetworkProfile(np *NetworkProfile) (errs []error) {
 		errs = append(errs, fmt.Errorf("networkProfile cannot be nil"))
 		return
 	}
+	if np.VnetID != "" && !rxVNetID.MatchString(np.VnetID) {
+		errs = append(errs, fmt.Errorf("invalid properties.networkProfile.VnetId %q", np.VnetID))
```
kargakis: the rest of the error messages start the field lowercased so this should be "invalid properties.networkProfile.vnetId %q"

pkg/arm/resources.go
```diff
@@ -128,11 +128,14 @@ func fixupDepends(azProfile *api.AzProfile, template map[string]interface{}) {
...
 }
 
+// resourceID constructs an Azure resource ID offline from the provided components
 func resourceID(subscriptionID, resourceGroup, resourceProvider, resourceName string) string {
 	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s/%s", subscriptionID, resourceGroup, resourceProvider, resourceName)
 }
 
 func vnet(cs *api.OpenShiftManagedCluster) *network.VirtualNetwork {
+	// populating VnetID for customer's convenience
+	cs.Properties.NetworkProfile.VnetID = resourceID(cs.Properties.AzProfile.SubscriptionID, cs.Properties.AzProfile.ResourceGroup, "Microsoft.Network/virtualNetworks", vnetName)
```
kargakis: It doesn't look like this is the right place to be populating `cs`. Probably something that is better suited during conversion from external?

pkg/plugin/plugin.go
```diff
@@ -135,6 +135,9 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
 			return err
 		}
 	}
+	if cs != nil {
```
kargakis: Remove?

pkg/arm/resources.go
Makdaam: Ok, modifying cs in the arm package is against intuition.
What would be the better solution:
exporting resourceID in the arm package?
moving ResourceID to a separate util since it's not ARM specific?

Makdaam: Discussed on call with @kargakis. Will copy the function to the convert package for now. Created #978 to track that.

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -89,13 +89,15 @@ const (
 
 // NetworkProfile contains configuration for OpenShift networking.
 type NetworkProfile struct {
+	// VnetCIDR is the subnet used by the Openshift cluster in CIDR format
```
jim-minter: // VnetCIDR (in): the CIDR with which the OSA cluster's Vnet is configured

pkg/api/2018-09-30-preview/api/types.go
```diff
@@ -89,13 +89,15 @@ const (
...
 	VnetCIDR *string `json:"vnetCidr,omitempty"`
 
-	// PeerVnetID is expected to be empty or match
+	// VnetID and PeerVnetID are expected to be empty or match
 	// `^/subscriptions/[^/]+
 	//   /resourceGroups/[^/]+
 	//   /providers/Microsoft.Network
 	//   /virtualNetworks/[^/]+$`
+	VnetID     *string `json:"vnetId,omitempty"`
```
jim-minter:
```golang
// VnetID (out): the ID of the Vnet created for the OSA cluster
VnetID     *string `json:"vnetId,omitempty"`

// PeerVnetID (in, optional): ID of a Vnet with which the OSA cluster Vnet should be peered.
// If non-empty, this should match
// `^/subscriptions/[^/]+
//   /resourceGroups/[^/]+
//   /providers/Microsoft.Network
//   /virtualNetworks/[^/]+$`
PeerVnetID *string `json:"peerVnetId,omitempty"`
```

jim-minter: [also please apply these changes to the other relevant files]

pkg/arm/resources.go
jim-minter: let's catch up on this.  I think it can/should all be done in this PR (#978 not necessary).  Moving resourceID() to somewhere under util is fine.  I think the point where we modify cs actually needs to be a new step right at the bottom of `func (p *plugin) CreateOrUpdate`.

pkg/arm/resources.go
```diff
@@ -149,6 +152,7 @@ func vnet(cs *api.OpenShiftManagedCluster) *network.VirtualNetwork {
 				},
 			},
 		},
+		ID:       to.StringPtr(resourceID(cs.Properties.AzProfile.SubscriptionID, cs.Properties.AzProfile.ResourceGroup, "Microsoft.Network/virtualNetworks", vnetName)),
```
jim-minter: this shouldn't/needn't be here.  Azure is assigning the ID; we're reporting it back to the user.  We're not changing how we request it.

pkg/plugin/plugin.go
jim-minter: +1.  The setting of cs.Properties.NetworkProfile.VnetID will happen at the bottom of this function.

jim-minter:
> I think we should also invalidate populating this field on create.

Possibly.  I don't believe we have the easy facility to do this right now, I'm happy to skip this at this stage.

pkg/api/types.go
Makdaam: done

pkg/api/validate.go
Makdaam: fixed, changed to camelCase

pkg/arm/resources.go
Makdaam: moved to `func (p *plugin) CreateOrUpdate` added `pkg/util/resourceid` with only the `ResourceID` func.

pkg/plugin/plugin.go
Makdaam: removed Debug output

pkg/api/2018-09-30-preview/api/types.go
Makdaam: changed godoc comments in v20180930preview API, admin API and internal representation.

pkg/arm/resources.go
Makdaam: Reverted change to the ARM file.

pkg/plugin/plugin.go
```diff
@@ -134,6 +135,11 @@ func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedClu
 		return err
 	}
 
+	if cs != nil {
```
jim-minter: if cs is nil by this point, we're really in trouble... ;-)

jim-minter:
great job!

/lgtm
/retest

PR #935, opened 2018-12-19
Preparation for upgrade rework

jim-minter:
Move to using standard load balancer
Various code refactors in preparation for the upgrade rework


jim-minter:
@asalkeld please can I ask you review this today - I'm going to want to build on this tomorrow to sort out the upgrade trouble.

jim-minter:
(review it commit by commit and it should be much less daunting)

pkg/api/validate.go
```diff
@@ -24,7 +24,7 @@ var (
 		`/virtualNetworks/[^/]+` +
 		`$`)
 
-	rxAgentPoolProfileName = regexp.MustCompile(`(?i)^[a-z0-9]{1,12}$`)
```
asalkeld: The docs seem to say they are case insensitive, why do this?

pkg/cluster/ready.go
```diff
@@ -11,19 +11,41 @@ func (u *simpleUpgrader) WaitForInfraServices(ctx context.Context, cs *api.OpenS
 	return u.kubeclient.WaitForInfraServices(ctx)
 }
 
+func (u *simpleUpgrader) waitForNode(ctx context.Context, cs *api.OpenShiftManagedCluster, role api.AgentPoolProfileRole) error {
```
asalkeld: the function name seems odd, we are waiting for all the vms in a pool
waitForPool() ?

asalkeld:
overall seems fine, some questions

1. why remove PrimaryScaleSetName?
2. why drop support for upper case app names when they are case insensitive? - is that purely for the sorting? if so you could just lower() the names at the time of sorting - seems strange to potentially product errors for things that azure wouldn't
3. the "use standard load balancer " commit changes networkSecuritygroups not the loadbalancer (as far as I can see) - what gives there?

pkg/cluster/update.go
```diff
@@ -248,34 +242,36 @@ func ssNameForVM(vm *compute.VirtualMachineScaleSetVM) scalesetName {
...
 	blob, err := u.readUpdateBlob()
 	if err != nil {
 		return &api.PluginError{Err: err, Step: api.PluginStepUpdateInPlaceReadBlob}
 	}
 
-	sorted = u.filterOldVMs(sorted, blob, ssHashes)
-	for _, vm := range sorted {
+	// range our vms in order, so that if we previously crashed half-way through
+	// updating one and it is broken, we pick up where we left off.
+	sort.Slice(vms, func(i, j int) bool {
```
kargakis: Isn't this sort mute since `updateInPlace` runs in `sortedAgentPoolProfilesForRole`?

kargakis: Sorry, I misread `sortedAgentPoolProfilesForRole`, nevermind.

kargakis:
lgtm overall

pkg/arm/data/azuredeploy.json
```diff
@@ -185,7 +190,11 @@
                 "[resourceId('Microsoft.Network/loadBalancers', 'lb-apiserver')]"
 {{- end }}
 {{- if not IsUpgrade }},
-                "[resourceId('Microsoft.Network/networkSecurityGroups', 'nsg-{{ $app.Name | js }}')]"
+{{- if eq $app.Role "master" }}
+                "[resourceId('Microsoft.Network/networkSecurityGroups', 'nsg-master')]"
+{{- else }}
+                "[resourceId('Microsoft.Network/networkSecurityGroups', 'nsg-nonmaster')]"
```
mjudeikis: `s/nonmaster/worker/` ? this `nonmaster` sounds strange

pkg/api/validate.go
jim-minter: Our converter merges on agent pool names and does not do so in a case insensitive way so we're actually broken here currently.  I am going to insert the agent pool name into the scale set name and don't want to litter the code everywhere with strings.EqualFold yet.  Follow-up in #939.

pkg/arm/data/azuredeploy.json
jim-minter: ok, will fix in follow up

jim-minter:
Am marking lgtm because no-one else did!

pkg/cluster/ready.go
jim-minter: agreed, thanks, will fix in follow-up

jim-minter:
@asalkeld 

> why remove PrimaryScaleSetName?

It is a hack which MSFT required when Standard LBs were not available in k8s.  Basic LBs can only front a single scale set.  Standard LBs can front multiple.

> why drop support for upper case app names when they are case insensitive? - is that purely for the sorting? if so you could just lower() the names at the time of sorting - seems strange to potentially product errors for things that azure wouldn't

See comment above.

> the "use standard load balancer " commit changes networkSecuritygroups not the loadbalancer (as far as I can see) - what gives there?

There are 2 LBs - one in the ARM template which fronts the masters, which MJ had already changed to Standard.  A second one fronts k8s services and is created by the cloud controller.  That's the one being changed here.  Because it now fronts all non-master scale sets, the NSGs had to be rationalised.

kargakis:
@stevekuznetsov seems the status contexts are flapping between pending and success. Maybe an issue with the status controller?

PR #936, opened 2018-12-19
replacing strings in internal representation ResourcePurchasePlan with pointers

### LINKED ISSUE #908: fake rp writes back the internal representation of the manifest

Makdaam:
Fixes #908

### DIFF
--- a/pkg/api/converterfromadmin.go
+++ b/pkg/api/converterfromadmin.go
@@ -33,7 +33,7 @@ func ConvertFromAdmin(oc *admin.OpenShiftManagedCluster, old *OpenShiftManagedCl
 	}
 
 	if oc.Plan != nil {
-		mergeResourcePurchasePlanAdmin(oc, cs)
+		mergeFromResourcePurchasePlanAdmin(oc, cs)
 	}
 
 	if oc.Properties != nil {
@@ -49,18 +49,23 @@ func ConvertFromAdmin(oc *admin.OpenShiftManagedCluster, old *OpenShiftManagedCl
 	return cs, nil
 }
 
-func mergeResourcePurchasePlanAdmin(oc *admin.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
+// mergeFromResourcePurchasePlanAdmin merges filled out fields from the admin API to the internal representation, doesn't change fields which are nil in the input.
+// This reflects the behaviour of the external API.
+func mergeFromResourcePurchasePlanAdmin(oc *admin.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
+	if cs.Plan == nil {
+		cs.Plan = &ResourcePurchasePlan{}
+	}
 	if oc.Plan.Name != nil {
-		cs.Plan.Name = *oc.Plan.Name
+		cs.Plan.Name = oc.Plan.Name
 	}
 	if oc.Plan.Product != nil {
-		cs.Plan.Product = *oc.Plan.Product
+		cs.Plan.Product = oc.Plan.Product
 	}
 	if oc.Plan.PromotionCode != nil {
-		cs.Plan.PromotionCode = *oc.Plan.PromotionCode
+		cs.Plan.PromotionCode = oc.Plan.PromotionCode
 	}
 	if oc.Plan.Publisher != nil {
-		cs.Plan.Publisher = *oc.Plan.Publisher
+		cs.Plan.Publisher = oc.Plan.Publisher
 	}
 }
 
--- a/pkg/api/converterfromv20180930preview.go
+++ b/pkg/api/converterfromv20180930preview.go
@@ -36,7 +36,9 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster, o
 		}
 	}
 
-	mergeResourcePurchasePlan(oc, cs)
+	if oc.Plan != nil {
+		mergeFromResourcePurchasePlanv20180930preview(oc, cs)
+	}
 
 	if err := mergeProperties(oc, cs); err != nil {
 		return nil, err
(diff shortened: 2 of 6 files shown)


pkg/api/convertertov20180930preview.go
```diff
@@ -20,10 +20,10 @@ func ConvertToV20180930preview(cs *OpenShiftManagedCluster) *v20180930preview.Op
 	}
 
 	oc.Plan = &v20180930preview.ResourcePurchasePlan{
```
kargakis: The internal ResourcePlan field probably has to change to a pointer too and avoid setting it here if cs.Plan == nil

pkg/api/converterfromv20180930preview_test.go
```diff
@@ -34,15 +34,19 @@ func internalManagedCluster() *OpenShiftManagedCluster {
...
 	return &OpenShiftManagedCluster{
 		ID:       "ID",
 		Location: "Location",
 		Name:     "Name",
 		Plan: ResourcePurchasePlan{
-			Name:          "Plan.Name",
-			Product:       "Plan.Product",
-			PromotionCode: "Plan.PromotionCode",
-			Publisher:     "Plan.Publisher",
+			Name:          &planName,
```
jim-minter: use `to.StringPtr("Plan.Name")` here, etc.

Makdaam: switched to the "to" package

pkg/api/convertertov20180930preview.go
Makdaam: Changed ResourcePurchasePlan to a pointer, added checks in converters for nil. So that it's nil if admin/v20180903preview has a nil ResourcePurchasePlan.

pkg/api/converterfromv20180930preview.go
```diff
@@ -47,19 +47,14 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster, o
 
 func mergeResourcePurchasePlan(oc *v20180930preview.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
 	if oc.Plan == nil {
-		return
```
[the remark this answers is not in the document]
kargakis:
```go
if oc.Plan == nil {
   return
}
cs.Plan = &ResourcePurchasePlan{
			Name:          oc.Plan.Name,
			Product:       oc.Plan.Product,
			PromotionCode: oc.Plan.PromotionCode,
			Publisher:     oc.Plan.Publisher,
		}
```
?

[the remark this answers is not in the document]
kargakis: Actually even though we don't touch Plan in our code, this function should probably continue behaving like a merging function.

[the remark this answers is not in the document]
kargakis: So just remove the dereferences of the oc.Plan fields.

pkg/api/converterfromadmin.go
```diff
@@ -50,17 +50,13 @@ func ConvertFromAdmin(oc *admin.OpenShiftManagedCluster, old *OpenShiftManagedCl
...
-	}
-	if oc.Plan.Product != nil {
-		cs.Plan.Product = *oc.Plan.Product
-	}
-	if oc.Plan.PromotionCode != nil {
-		cs.Plan.PromotionCode = *oc.Plan.PromotionCode
-	}
-	if oc.Plan.Publisher != nil {
-		cs.Plan.Publisher = *oc.Plan.Publisher
+	if oc.Plan != nil {
```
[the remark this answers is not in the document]
Makdaam: Thanks. Removed internal if.

pkg/api/converterfromv20180930preview.go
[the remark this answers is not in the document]
Makdaam: Thanks for explaining the purpose, missed that the plan could be provided via the API in parts. Changed it back + adjusted for pointer types.

Makdaam:
Rolled back unit tests to verify nil merge as making no changes instead of setting all fields to nil.

pkg/api/convertertoadmin.go
```diff
@@ -16,14 +16,14 @@ func ConvertToAdmin(cs *OpenShiftManagedCluster) *admin.OpenShiftManagedCluster
...
-		Publisher:     &cs.Plan.Publisher,
+	if cs.Plan != nil {
+		oc.Plan = &admin.ResourcePurchasePlan{
+			Name:          cs.Plan.Name,
+			Product:       cs.Plan.Product,
+			PromotionCode: cs.Plan.PromotionCode,
+			Publisher:     cs.Plan.Publisher,
+		}
 	}
-
```
kargakis: nit: remove this change

Makdaam: readded \n

pkg/api/converterfromadmin.go
```diff
@@ -50,17 +50,11 @@ func ConvertFromAdmin(oc *admin.OpenShiftManagedCluster, old *OpenShiftManagedCl
 }
 
 func mergeResourcePurchasePlanAdmin(oc *admin.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
```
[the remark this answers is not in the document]
Makdaam: renamed to a more representative name.

pkg/api/converterfromv20180930preview.go
```diff
@@ -45,21 +45,25 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster, o
 	return cs, nil
 }
 
+// mergeResourcePurchasePlan merges filled out fields from the external API to the internal representation, doesn't change nil fields.
 func mergeResourcePurchasePlan(oc *v20180930preview.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
```
kargakis: Can you change it here to how we handle the resource plan conversion in pkg/api/converterfromadmin.go? Add an `if oc.Plan != nil` statement in L38, rename to `convertFromResourcePurchasePlanv20180930preview`, and passthrough cs.Plan as is if != nil. No reason to have such diff laying around.

Makdaam: Moved nil check to ConvertFromV20180930preview func. 
Changed name to match convention, but left behaviour the same (doesn't change fields that have nil in external representation).

pkg/api/converterfromv20180930preview.go
```diff
@@ -45,21 +47,22 @@ func ConvertFromV20180930preview(oc *v20180930preview.OpenShiftManagedCluster, o
 	return cs, nil
 }
 
-func mergeResourcePurchasePlan(oc *v20180930preview.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
-	if oc.Plan == nil {
-		return
+// convertFromResourcePurchasePlanv20180930preview merges filled out fields from the external API to the internal representation, doesn't change nil fields.
+func convertFromResourcePurchasePlanv20180930preview(oc *v20180930preview.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
```
kargakis: This is still merging and not just conversion. Any reason why we would merge here but passthrough via the admin api?

kargakis: /cc @jim-minter

PR #937, opened 2018-12-19
Use create futures appropriately in the fake rp

kargakis:
I was playing around with futures and I was thinking that the way we have implemented PUTs currently in the fake RP is wrong since it's a synchronous process whereas it probably should be asynchronous. I am not sure how this is implemented in the prod RP though and the fact that the clients are not treating http.StatusAccepted as a valid status returned during a PUT makes me think that this may work differently in prod or it's probably an oversight in the OSA sdk? 

@julienstroheker @amanohar @jim-minter

### DIFF
--- a/pkg/fakerp/middleware.go
+++ b/pkg/fakerp/middleware.go
@@ -13,7 +13,7 @@ func (s *Server) logger(handler http.Handler) http.Handler {
 
 func (s *Server) validator(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Method == http.MethodPut {
+		if r.Method != http.MethodGet {
 			select {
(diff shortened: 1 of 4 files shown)


kargakis:

/retest

jim-minter:
@kargakis please can you bring up this PR on scrum for us to talk about?

kargakis:
> @kargakis please can you bring up this PR on scrum for us to talk about?

Sure thing.

jim-minter:
@kargakis (@mjudeikis) I recognise that the fake rp differs in behaviour from the real rp regarding sync/async; I have forgotten why it is that we need to fix that - please can you remind me?

kargakis:
We don't really need to fix it unless we want to adhere to the azure api
patterns.

jim-minter:
My inclination for now is to avoid the extra complexity if we aren't causing the client library problems.

PR #940, opened 2018-12-19
Update production RP tests to be self-contained

kargakis:


test/e2e/specs/realrp/realrp.go
```diff
@@ -2,59 +2,73 @@ package realrp
...
 	fakepubkey = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7laRyN4B3YZmVrDEZLZoIuUA72pQ0DpGuZBZWykCofIfCPrFZAJgFvonKGgKJl6FGKIunkZL9Us/mV4ZPkZhBlE7uX83AAf5i9Q8FmKpotzmaxN10/1mcnEE7pFvLoSkwqrQSkrrgSm8zaJ3g91giXSbtqvSIj/vk2f05stYmLfhAwNo3Oh27ugCakCoVeuCrZkvHMaJgcYrIGCuFo6q0Pfk9rsZyriIqEa9AtiUOtViInVYdby7y71wcbl0AbbCZsTSqnSoVxm2tRkOsXV6+8X4SnwcmZbao3H+zfO1GBhQOLxJ4NQbzAa8IJh810rYARNLptgmsd4cYXVOSosTX azureuser"
 )
 
+var (
+	ctx = context.Background()
+	cli *azure.Client
+	cfg *client.Config
+)
+
 var _ = BeforeSuite(func() {
```
kargakis: This runs globally for all tests so I cannot really deploy a cluster in here. Maybe we are not segregating correctly our test suites.

kargakis:
This should be ready for reviews now, just need to ensure the job works

/cc @mjudeikis @charlesakalugwu

kargakis:
The prod job works fine now, just need to extend the timeouts since everything now runs inside the job (creation+tests+deletion).

pkg/fakerp/client/config.go
```diff
@@ -13,9 +13,13 @@ import (
 )
 
 var supportedRegions = []string{
-	"australiasoutheast",
+	//"australiaeast",
```
kargakis: @jim-minter we need more quota in these regions in order to enable them.

test/e2e/specs/realrp/realrp.go
```diff
@@ -2,59 +2,55 @@ package realrp
...
+var _ = Describe("Resource provider e2e tests [Default][Real]", func() {
 	var (
 		ctx = context.Background()
 		cli *azure.Client
+		cfg *client.Config
 	)
 
-	BeforeEach(func() {
+	// NOTE: Ensure this is always the first test in the [Default][Real] spec!
+	It("should deploy a cluster using the production RP", func() {
```
kargakis: The alternative solution to this that I though is to create a new `test/e2e/specs/realrp/default.go` that would simply deploy a prod cluster such as the vnet peering test is doing and then somehow reuse all tests in this file in both `default.go` and `vnetpeering.go`. That was problematic because the vnet peering test is cleaning up itself in a `AfterEach` function. Maybe we want to fix this in a follow up, didn't seem pressing to do here.

kargakis:

/retest

test/e2e/specs/realrp/realrp.go
kwoodson: I'm fine removing the clean up of itself.  I only did that because I wanted to play nice.  If we want to reap the resourcegroup/cluster after an entire run then that's a :+1:  from me.

kargakis:
> This /lgtm

fyi, bot commands need their own line in order to work correctly

kargakis:
```
pulling image error : error pulling image configuration: unknown blob
          error: build error: unable to get docker-registry.default.svc:5000/openshift/nginx@sha256:0f22060514f1fb710529437b2e5c9be61fec7ebc7e4a21379d0646f267950de2
```
failed pulling the image to build
https://github.com/openshift/openshift-azure/issues/919
/retest

PR #941, opened 2018-12-19
added a parity test between 20180930preview and internal API

### LINKED ISSUE #933: add unit tests for internal and external representation parity with appropriate exceptions
There should be an explicit parity unit test similar to TestAdminAPIParity in pkg/api, but for the External representation.
This will allow for a more concise error messages in unit tests and separates it from converter unit tests.

Makdaam:
Fixes #933 
Not sure if the AuthProfile.IdentityProviders Secret field should be missing from the internal representation. The test ignores that difference for now.

### DIFF
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/to"
 
+	v20180930preview "github.com/openshift/openshift-azure/pkg/api/2018-09-30-preview/api"
 	admin "github.com/openshift/openshift-azure/pkg/api/admin/api"
 	"github.com/openshift/openshift-azure/pkg/util/structtags"
 	"github.com/openshift/openshift-azure/test/util/populate"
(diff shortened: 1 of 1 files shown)


pkg/api/types_test.go
```diff
@@ -416,3 +417,82 @@ loop:
 		t.Fatalf("mismatch between internal and admin API fields: afield=%q, ifield=%q", afield, ifield)
 	}
 }
+
+func Test20180930previewAPIParity(t *testing.T) {
+	// the algorithm is: list the fields of both types.  If the head of either
```
kwoodson: Good comment!

Makdaam: All credit to @jim-minter who wrote the Admin API parity test which I copied and adapted.

pkg/api/types_test.go
```diff
@@ -416,3 +417,82 @@ loop:
...
+				reflect.TypeOf(AADIdentityProvider{}),
+			},
+		},
+	)
+
+	// TODO: I don't believe this should be in the admin type at all
+	notInInternal := []*regexp.Regexp{
+		regexp.MustCompile(`^\.Response`),
+		regexp.MustCompile(`^\.Properties\.MasterPoolProfile`),
+		regexp.MustCompile(`^\.Properties\.AuthProfile\.IdentityProviders\.Provider\.Secret`),
```
kargakis: This field should exist in both the internal and the customer api

pkg/api/types_test.go
```diff
@@ -416,3 +417,82 @@ loop:
...
+	}
+
+	// TODO: why don't we just include all of these in the admin type?
+	notInV20180930preview := []*regexp.Regexp{
+		regexp.MustCompile(`^\.Properties\.ServicePrincipalProfile`),
```
kargakis: This field should exist in both the internal and the customer api

pkg/api/types_test.go
```diff
@@ -416,3 +417,82 @@ loop:
...
+		regexp.MustCompile(`^\.Properties\.AzProfile`),
```
kargakis: ditto

kargakis: Sorry, servicePrincipalProfile and azProfile are not included in the customer profile, disregard

pkg/api/types_test.go
Makdaam: Thanks for catching this, there was a second list of exceptions, I removed that and everything matches my expectations now.

kargakis:
https://github.com/openshift/openshift-azure/issues/949

/retest

PR #943, opened 2018-12-20
Preparation for upgrade rework (2)

jim-minter:
@asalkeld @mjudeikis @kargakis a second round of preparations.  Should all be no-op.  I think I'm now ready to implement the upgrade changes needed.  Please lgtm if happy so we can push this through.

Same as yesterday, please can you review commit by commit?

Main direction is:
1) get rid of the azuredeploy.json and ARM template templating (hooray!) - this is needed because we're going to be creating VMSSs outside of ARM and I want to reuse the same code to do it
2) clearly split out which update flows are for masters, and which for workers.

The idea in my head for the next step is:

There will be many worker AgentPoolProfiles (one infra, one compute, in the future more compute)
In a healthy state, an AgentPoolProfile will correspond to one VMSS.  The names will not exactly match any more (e.g. you'll see something like ss-mycompute-<timestamp>)
In crash recovery, a worker AgentPoolProfile could have many VMSSs, each corresponding to a previous failed update, which we'll have to be able to clean up.

Thanks.


asalkeld:
I guess lgtm - The main one (azuredeploy -> golang) is really hard to see what has changed. Also the azuredeploy.json is starting to look really understandable in comparison to the golang :-O

mjudeikis:
I guess too. *closes eyes* and lgtm, @kargakis do we have 3 lgtm and this PR can go to the Finals? :D

pkg/arm/resources.go
```diff
@@ -0,0 +1,530 @@
...
+		switch typ {
+		case "Microsoft.Compute/virtualMachineScaleSets":
+			apiVersion = "2018-06-01"
+		case "Microsoft.Network/loadBalancers",
+			"Microsoft.Network/networkSecurityGroups",
+			"Microsoft.Network/publicIPAddresses",
+			"Microsoft.Network/virtualNetworks":
+			apiVersion = "2017-10-01"
+		case "Microsoft.Storage/storageAccounts":
+			apiVersion = "2015-06-15"
```
kargakis: Question for a different PR but is there any reason not to move to a newer version?

pkg/arm/resources.go
```diff
@@ -0,0 +1,530 @@
...
+		default:
+			panic("unimplemented: " + typ)
+		}
+		jsonpath.MustCompile("$.apiVersion").Set(resource, apiVersion)
+	}
+}
+
+// fixupDepends inserts a dependsOn field into the ARM template for each
+// resource that needs it (the field is missing from the internal Azure type).
+func fixupDepends(azProfile *api.AzProfile, template map[string]interface{}) {
```
kargakis: Follow-up issue to unit test these.

pkg/arm/resources.go
```diff
@@ -0,0 +1,530 @@
...
+			{
+				ID: to.StringPtr(resourceID(
+					cs.Properties.AzProfile.SubscriptionID,
+					cs.Properties.AzProfile.ResourceGroup,
+					"Microsoft.Network/loadBalancers",
+					lbAPIServerName,
+				) + "/backendAddressPools/" + lbAPIServerBackendPoolName),
+			},
+		}
+		(*vmss.VirtualMachineProfile.NetworkProfile.NetworkInterfaceConfigurations)[0].VirtualMachineScaleSetNetworkConfigurationProperties.NetworkSecurityGroup = &compute.SubResource{
```
kargakis: nit: there is potential to dry this

pkg/cluster/deploy_test.go
```diff
@@ -21,81 +22,33 @@ const masterHash = "RmID82LhPjuQbCEdiVa5cGCVEkdLGD6iU6ozX3vxkD0="
...
 	tests := []struct {
 		name string
-		t    map[string]interface{}
-		exp  map[scalesetName]hash
+		vmss *compute.VirtualMachineScaleSet
+		exp  hash
 	}{
 		{
 			name: "expect a scale set",
-			t: map[string]interface{}{
```
kargakis: Didn't look that bad to refactor this test, I am glad.

pkg/cluster/kubeclient/ready.go
```diff
@@ -158,11 +145,11 @@ func (u *kubeclient) MasterIsReady(computerName ComputerName) (bool, error) {
 	return ready.PodIsReady(u.client.CoreV1().Pods("kube-system"), "controllers-"+computerName.toKubernetes())()
 }
 
-func (u *kubeclient) nodeWaitForReady(ctx context.Context, computerName ComputerName) error {
+func (u *kubeclient) WaitForReadyWorker(ctx context.Context, computerName ComputerName) error {
```
kargakis: After going back to this for a while, it's not obvious to me why a wait for readiness method sets the node to unschedulable. Since this is exported now, probably a good chance to add godoc.

kargakis: It seems the node is set to schedulable (unschedulable=false). Still unclear why this is needed for worker nodes.

kargakis:
The monitoring stack timeouts are all over the place, hopefully most will go away with https://github.com/openshift/openshift-azure/pull/946

pkg/cluster/kubeclient/ready.go
jim-minter: I don't believe it is, I intend to remove it.

jim-minter: (in a subsequent PR)

jim-minter:
@asalkeld 

> The main one (azuredeploy -> golang) is really hard to see what has changed. Also the azuredeploy.json is starting to look really understandable in comparison to the golang :-O

The functionality should be 1:1 - I used https://github.com/shurcooL/go-goon to help migrate it.  I think we'll get used to it, but note there are significant advantages now: we don't have to worry about escaping / breakouts, unit testing is easier, and it's easier to share constants with the rest of the codebase.

PR #950, opened 2018-12-20
Wait specifically for lsof to exit

kargakis:
/cc @charlesakalugwu 

Fixes https://github.com/openshift/openshift-azure/issues/953


kargakis:
This fixes the case where createorupdate fails (eg. the request it tries to send does not exist in the filesystem) before fakerp has even started

kargakis:
Figured out that I am stupid; `return_id=$?` needs to run before `set` to actually get the correct last exit code propagated :)

Issue #934, opened 2018-12-18
perftests cluster has lots of restarted td-agent pods


jim-minter:
```
# oc get pods -n openshift-azure-logging |grep ^td-agent |head
td-agent-27kvj   1/1       Running            14         15d
td-agent-2mqhg   1/1       Running            1          15d
td-agent-2sdjr   1/1       Running            430        15d
td-agent-45vhj   1/1       Running            1          15d
td-agent-4lh2d   1/1       Running            3          15d
td-agent-4pdlh   1/1       Running            1          15d
td-agent-4wbjn   1/1       Running            3          15d
td-agent-4xmc4   1/1       Running            253        15d
td-agent-598lk   1/1       Running            671        15d
td-agent-5pn65   1/1       Running            1          15d

# oc logs -n openshift-azure-logging td-agent-598lk
...no output...

# oc get pods -n openshift-azure-logging td-agent-598lk -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    openshift.io/scc: privileged
    scheduler.alpha.kubernetes.io/critical-pod: ""
  creationTimestamp: 2018-12-03T17:49:05Z
  generateName: td-agent-
  labels:
    app: td-agent-ds
    controller-revision-hash: "2399226709"
    pod-template-generation: "1"
  name: td-agent-598lk
  namespace: openshift-azure-logging
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: td-agent
    uid: 7becfbef-f723-11e8-964e-000d3a2bb4d5
  resourceVersion: "18027950"
  selfLink: /api/v1/namespaces/openshift-azure-logging/pods/td-agent-598lk
  uid: ba4c9f91-f723-11e8-9656-000d3a2fbdfc
spec:
  containers:
  - env:
    - name: FLUENTD_CONF
      value: config/fluent.conf
    image: osarpint.azurecr.io/acs/td-agent:latest
    imagePullPolicy: Always
    livenessProbe:
      exec:
        command:
        - /bin/bash
        - /td-agent/liveness-probe.sh
      failureThreshold: 5
      initialDelaySeconds: 60
      periodSeconds: 60
      successThreshold: 1
      timeoutSeconds: 1
    name: td-agent
    ports:
    - containerPort: 9102
      protocol: TCP
    resources:
      limits:
        cpu: 400m
        memory: 800Mi
      requests:
        cpu: 200m
[...]

PR #955, opened 2018-12-20
make logging pods guaranteed

mjudeikis:
make logging pods guaranteed so QoS would be applied. 

This is "finger in the air" and we will need to profile them when we have access to Geneva


mjudeikis:
/hold 
https://github.com/openshift/openshift-azure/pull/929 need to go first
potentially fixes https://github.com/openshift/openshift-azure/issues/934

PR #959, opened 2018-12-20
These are prone to error w/ shuffling files around

thekad:
Using fakerp's shared dir discovery mechanism, we can make these a bit
more robust and not worry about updating locations every single time


test/e2e/specs/fakerp/admin.go
```diff
@@ -27,12 +29,16 @@ import (
...
+	if err != nil {
+		dataDir = "../../_data"
+	}
 	var (
 		cli *openshift.Client
 		// TODO: Unfortunately cannot use "manifest" because the flag name is already
 		// used by the key rotation test; Figure out whether we want to collapse these
 		// into a single flag and do it.
-		manifest = flag.String("request", "../../_data/manifest.yaml", "Path to the manifest to send to the RP")
+		manifest = flag.String("request", path.Join(dataDir, "manifest.yaml"), "Path to the manifest to send to the RP")
```
kargakis: use filepath instead of path, it's compatible with windows

test/e2e/specs/fakerp/admin.go
```diff
@@ -27,12 +29,16 @@ import (
 )
 
 var _ = Describe("Openshift on Azure admin e2e tests [AzureClusterReader][Fake]", func() {
+	dataDir, err := shared.FindDirectory(shared.DataDirectory)
```
kargakis: Can we ensure this is used only once during test setup and every test that needs to grab a manifest from the filesystem uses that instead of having to reimplement this logic in every test?

PR #962, opened 2018-12-21
Simplify manifest helpers

kargakis:
Since defaulting is only used in createorupdate and
every test is using its own manifest, it makes sense
to simplify the helper and move the defaulting logic
in createorupdate.


kargakis:
/hold cancel

This will help parsing templated manifests for admin requests so we can use ./hack/create.sh to test admin validation.

mjudeikis:
all fakeRP logic is becoming project of itself now :/

mjudeikis:
it was a generic comment :)

PR #964, opened 2019-01-02
typo Openshift -> OpenShift

mjudeikis:
we use `OpenShiftMaster -> openShiftMaster` everywhere, so it should be `OpenShiftConsole -> openShiftConsole` too.


mjudeikis:
This bugged me for a while :/

PR #965, opened 2019-01-02
add make delete create

mjudeikis:
Add make targets so we could simplify templates
This will help to add things like "populate-template" for https://github.com/openshift/openshift-azure/pull/922

will fix release repo post merge

/cc @kargakis @charlesakalugwu

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,12 @@ test: unit e2e
 generate:
 	go generate ./...
 
+create:
+	timeout 1h ./hack/create.sh ${RESOURCEGROUP}
+
+delete:
+	./hack/delete.sh ${RESOURCEGROUP}
+
 TAG ?= $(shell git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


pkg/fakerp/server.go
```diff
@@ -208,28 +208,31 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
...
-		// If the resource group deletion is successful, cleanup the object
-		// from the memory so the next GET from the client waiting for this
-		// long-running operation can exit successfully.
-		if resp.StatusCode == http.StatusOK {
-			s.log.Infof("deleted resource group %s", resourceGroup)
-			s.write(nil)
-		}
-	}()
+	// no-wait for CI
+	if os.Getenv("NO_WAIT") == "" {
```
kargakis: I don't think we need to do this server-side. Clients choose whether they want to wait or not.

mjudeikis: removed. missed client side check.

PR #893, opened 2018-12-12
Move updateblob and updatehash into separate sub-packages of cluster

asalkeld:
This is a further attempt at separating out logical chucks of code to make testing easier.

PR #970, opened 2019-01-02
 reimplement worker update to use a new scaleset and migrate VMs to it

jim-minter:
@mjudeikis @kargakis Happy new year and ptal as a priority:

With this PR, I believe our update process should work without using UpdateInstances and colliding with the k8s cloud provider code.  I'm sorry it's not going to be a pleasant PR to review.  Suggestions on how to improve the approach are welcome.  I'm aware that this is short on unit test coverage.  The general sketch is:

* on create, ARM will deploy master and worker scalesets.  On update, ARM now only redeploys the master scaleset (we handle the worker part in our own code).
* master update process is actually unchanged.
* worker update process has code comments documenting the approach - see pkg/cluster/update_worker.go.  In brief, if we want to change the worker scaleset configuration (excluding straight scales), we create a new worker scaleset, then loop creating new VMs in it and deleting old ones elsewhere.
* worker scalesets are now be named `ss-$name-$timestamp`. Master scaleset is still named `ss-master`.
* in the update blob, we'll now store **VM instance** hashes for the master VMs, and **Scale set** hashes for worker scale sets.  **Scale set** hashes are expected to be immutable.  Example:

```json
{
    "scalesetHashes": [
        {
            "scalesetName": "ss-compute-1546444018",
            "hash": "qjKcTaw4WK55ruzamYAeRsk30v4zH2uqctgxk9PVQpI="
        },
        {
            "scalesetName": "ss-infra-1546444018",
            "hash": "EVCpZXjRgd6ccfipmhgBNXnPwqrVBv6BnY6m5c9nUTg="
        }
    ],
    "instanceHashes": [
        {
            "instanceName": "ss-master_0",
            "hash": "4w5Vw/2h4gS5DGXmSjdxmk7wSkzo8xzRnJBshvmq5t0="
        },
        {
            "instanceName": "ss-master_1",
            "hash": "4w5Vw/2h4gS5DGXmSjdxmk7wSkzo8xzRnJBshvmq5t0="
        },
        {
            "instanceName": "ss-master_2",
            "hash": "4w5Vw/2h4gS5DGXmSjdxmk7wSkzo8xzRnJBshvmq5t0="
        }
    ]
}
```
* worker update scaling mechanism has been abstracted to pkg/cluster/update_worker_scaler.go.


jim-minter:
I needed to push more code to fix an e2e test :(

Now in addition, this adds some simplifications which will help improve our unit testing, and begins to roll in some of @asalkeld 's work from #893

pkg/cluster/hash.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil, err
+	}
+
+	hf := sha256.New()
+	hf.Write(data)
+
+	return hf.Sum(nil), nil
+}
+
+// hashScaleSets returns the set of desired state scale set hashes
```
kargakis: // HashScaleSet ...

kargakis:
Unfortunate that we have to work around the cloud controller. The new upgrade code lgtm.

/test keyrotation
/lgtm

pkg/cluster/update_master.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// range our vms in order, so that if we previously crashed half-way through
+	// updating one and it is broken, we pick up where we left off.
+	sort.Slice(vms, func(i, j int) bool {
+		return *vms[i].VirtualMachineScaleSetVMProperties.OsProfile.ComputerName <
+			*vms[j].VirtualMachineScaleSetVMProperties.OsProfile.ComputerName
+	})
+
+	vms = u.filterOldVMs(vms, blob, ssHash)
```
mjudeikis: This part is a little bit confusing to me. If we hash each individual instances, we just pass one hash here and compare. Is the expectation that all instances hashes will match?

mjudeikis: ignore, tests explained

PR #975, opened 2019-01-04
VMSS loss of internet update fix

jim-minter:
Create, but then don't subsequently touch, the kubernetes ELB.  Add newly
created worker VMSSes to the kubernetes ELB backend ourselves, rather than
waiting for the cloud provider to do it.  This protects us from racing with the
cloud provider code with the potential result of a newly created VMSS instance
not having external connectivity.


pkg/arm/resources.go
```diff
@@ -387,6 +493,9 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
 		},
 		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
 			UpgradePolicy: &compute.UpgradePolicy{
+				AutoOSUpgradePolicy: &compute.AutoOSUpgradePolicy{
+					DisableAutoRollback: to.BoolPtr(false),
```
kargakis: In which case are we going to need a scale set to automatically rollback?

jim-minter: this is mirroring a change that the cloud provider appears to make to the VMSS by default - I put it in here to hopefully save the cloud provider finding a diff it needed to make and applying it.

Issue #973, opened 2019-01-03
pod Antiaffinity missing


mjudeikis:
Pod not even distribution causes resources exhaustion on one of the nodes so DS and other resources depended pods fails to get scheduled. 

```
NAMESPACE                           NAME                                           READY     STATUS      RESTARTS   AGE       IP            NODE
default                             docker-registry-656486f769-hdnzp               1/1       Running     0          15m       10.131.0.5    infra-1546522962-000000
default                             docker-registry-656486f769-kd84f               1/1       Running     0          15m       10.131.0.6    infra-1546522962-000000
default                             registry-console-bd9c6cf8f-rrdgw               1/1       Running     0          15m       10.131.0.12   infra-1546522962-000000
default                             router-5f48c68ccf-mzsrm                        1/1       Running     0          15m       10.131.0.15   infra-1546522962-000000
default                             router-5f48c68ccf-rdxp2                        1/1       Running     0          15m       10.131.0.11   infra-1546522962-000000
kube-service-catalog                apiserver-5jdxn                                1/1       Running     0          15m       10.129.0.4    master-000000
kube-service-catalog                apiserver-krqjp                                1/1       Running     0          15m       10.128.0.3    master-000002
kube-service-catalog                apiserver-nc7wz                                1/1       Running     0          15m       10.130.0.6    master-000001
kube-service-catalog                controller-manager-6dgrm                       1/1       Running     0          15m       10.128.0.5    master-000002
kube-service-catalog                controller-manager-7qcfp                       1/1       Running     0          15m       10.130.0.5    master-000001
kube-service-catalog                controller-manager-ct9nh                       1/1       Running     2          15m       10.129.0.2    master-000000
kube-system                         controllers-master-000000                      1/1       Running     0          15m       10.0.0.7      master-000000
kube-system                         controllers-master-000001                      1/1       Running     0          15m       10.0.0.8      master-000001
kube-system                         controllers-master-000002                      1/1       Running     1          15m       10.0.0.9      master-000002
kube-system                         master-api-master-000000                       1/1       Running     0          15m       10.0.0.7      master-000000
kube-system                         master-api-master-000001                       1/1       Running     0          15m       10.0.0.8      master-000001
kube-system                         master-api-master-000002                       1/1       Running     0          15m       10.0.0.9      master-000002
kube-system                         master-etcd-master-000000                      1/1       Running     0          15m       10.0.0.7      master-000000
[...]

PR #976, opened 2019-01-04
Fixes to etcd restore

jim-minter:
Fixing errors introduced into etcd restore by my previous commits.  Includes #975.


mjudeikis:
/test e2e
might be related to https://github.com/openshift/openshift-azure/issues/973

mjudeikis:
lgtm, will wait for other review.
I think biggest change here is VMSS model changes to include kuberenetes loadbalancer from the start.

pkg/arm/resources.go
```diff
@@ -387,6 +494,9 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
 		},
 		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
 			UpgradePolicy: &compute.UpgradePolicy{
+				AutoOSUpgradePolicy: &compute.AutoOSUpgradePolicy{
+					DisableAutoRollback: to.BoolPtr(false),
```
kargakis: In which case are we going to need this?

kargakis:
one question about the autorollback feature but this can go in

/lgtm

pkg/arm/resources.go
jim-minter: this is mirroring a change that the cloud provider appears to make to the VMSS by default - I put it in here to hopefully save the cloud provider finding a diff it needed to make and applying it.

jim-minter: (it should be a no-op from our perspective)

PR #985, opened 2019-01-07
Enable fake RP GETs to work ootb

kargakis:
This change should help in writting more robust e2e tests where we can use GET OSA before applying a change in order to enable comparisons of the state of an OSA resource before and after an update from a client's POV.

/cc @charlesakalugwu @mjudeikis


kargakis:
updated to handle not found errors appropriately

PR #990, opened 2019-01-07
Setup codecov integration

charlesakalugwu:
This prepares the openshift-azure project for codecov integration. 

Before merging, we need to do the following:

- [x] We need to signup for a codecov account using an openshift account not attached to any given contributor https://docs.google.com/document/d/1UxVr2py_tAup-8WQl8SJ8mZ0dBdGfSOnzUd5MLRSXcs/edit
- [x] After registering openshift-azure for codecov integration, we need to get the `report upload token` and make sure that is set in a secret mounted by our CI test containers. The `CODECOV_UPLOAD_TOKEN` environment variable should be available to the container from which `make codecov` is executed. If the instrumented repo is a private one, then `CODECOV_GRAPH_TOKEN` should also be made available.
- [x] Create a `codecov` prow job which is run before every pull request and on every branch (pre and post submits) which internally calls `make codecov` (https://github.com/openshift/release/pull/2512)

See a working example of codecov PR comments, linking, the codecov badge amongst other things at: https://github.com/charlesakalugwu/codecov/pulls?q=is%3Apr+is%3Aclosed

/cc @kargakis @mjudeikis @jim-minter 
/hold

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "50...100"
+
+  status:
+    project: no
+    patch: no
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "header, diff"
+  behavior: default
+  require_changes: no
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,9 @@ endif
 cover: unit
 	go tool cover -html=coverage.out
 
+codecov: unit
+	./hack/codecov-report.sh
+
 e2e:
 	FOCUS="\[AzureClusterReader\]|\[CustomerAdmin\]|\[EndUser\]\[Fake\]" TIMEOUT=60m ./hack/e2e.sh
 
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ verify:
 	go run ./hack/lint-addons/lint-addons.go -n
 
 unit: generate
-	go test ./... -coverprofile=coverage.out
+	go test ./... -coverprofile=coverage.out -covermode=atomic
 ifneq ($(ARTIFACT_DIR),)
 	mkdir -p $(ARTIFACT_DIR)
 	cp coverage.out $(ARTIFACT_DIR)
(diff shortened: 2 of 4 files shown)


jim-minter:
/lgtm

@charlesakalugwu if this is ready to go, do a hold cancel

kargakis:
I think this lgtm but in CI we probably will run it in its own pod instead of using ci-operator.

codecov.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+coverage:
+  precision: 2
+  round: down
+  range: "50...100"
+
+  status:
+    project:
+      default:
+        # allow the project's overall coverage to drop by X%
```
kargakis: This may be problematic. Do we want to be blocking PRs that drop unit test coverage by 1%? I think not.

charlesakalugwu: @kargakis so what do you suggest we put here? I could also block the project check by setting `project: no`. In that case only the `patch` check (for PRs) will be performed. And right now that check is configured to pass if and only if the changes introduced in the PR have a coverage ratio of at least 50%

charlesakalugwu:
> I think this lgtm but in CI we probably will run it in its own pod instead of using ci-operator.

@kargakis yes it looks like. I'm thinking about using a template but I know you try to avoid templates. Do you have any other ideas how we can get codecov to run in its own pod?

codecov.yaml
kargakis: I see codecov as more of a reporting tool rather than something that would block PRs in CI so I am not really sure this setting makes sense at all but that's just my opinion. May be worth to bring it up during scrum, I hadn't thought before we were going to be blocking PRs with this and it's probably unexpected (devs writting unit tests in order to make their PRs merge is smelly).

kargakis:
> @kargakis yes it looks like. I'm thinking about using a template but I know you try to avoid templates. Do you have any other ideas how we can get codecov to run in its own pod?

Prowjob pod spec uses our test base image and runs `make codecov`. I can look into setting it up.

codecov.yaml
charlesakalugwu: Thats also fine. We could disable the codecov checks altogether and depend only the PR report comments by setting:

```
status:
    project: no
    patch: no
    changes: no
```

Lets discuss that during scrum

kargakis:
```
Error: Package: systemd-219-62.el7_6.2.x86_64 (updates)
           Requires: libcryptsetup.so.12()(64bit)
Error: Package: systemd-219-62.el7_6.2.x86_64 (updates)
           Requires: libcryptsetup.so.12(CRYPTSETUP_2.0)(64bit)
```
/shrug

PR #992, opened 2019-01-07
Use the RP instead of local file to get cluster state

kargakis:
Enables the re-entrancy test to work with the production RP in the future.


kargakis:
https://github.com/openshift/openshift-azure/issues/919

/retest

PR #995, opened 2019-01-08
Update scale test to be self-contained

kargakis:
Avoid reading from the filesystem in order to make
scale up and down requests. We stop using partial
requests but at the same time enable the test to
work with prod environments and not needing external
files.

/cc @charlesakalugwu @jim-minter


kargakis:
```
azure#createPollingTracker: unsupported HTTP method GET
```
trying to figure out where this is coming from

kargakis:
> trying to figure out where this is coming from

Seems to be caused because I am overwritting the whole `OpensShiftManagedCluster` object in `scaleUp` which probably overwrites `autorest.Response`? Unclear why that should cause this issue but at least I can work around it.

/shrug

kargakis:
https://github.com/openshift/ci-operator/issues/237

/retest

PR #999, opened 2019-01-08
Restore endpoint in the fake RP

kargakis:
/cc @jim-minter @asalkeld 

Need to figure out how to read the blob name in the RP (from the request or from the internal config?)

@charlesakalugwu since you are updating the http routing in the fake RP - I just want to enable restore for now. Can rebase on top of your changes if they get in faster.


kargakis:
> Need to figure out how to read the blob name in the RP (from the request or from the internal config?)

I chose to include the blob name in the request. This is ready for a review.

kargakis:
@bbguimaraes @petr-muller prow is not responding to my requests for running tests - github has a bad day or something else?

kargakis:
Going to merge and remove the containerservice.yaml dependency in a follow-up.

Issue #776, opened 2018-11-22
Check if mdsd-customerdata-config.yaml is populated by naive substitution and avoid if possible


jim-minter:
We have stanzas like `<IdentityComponent name="SubscriptionId">_SUBSCRIPTIONID_</IdentityComponent>` in mdsd-customerdata-config.yaml - investigate how these are populated: if done by naive substitution, it would be preferable to template them in directly using an XML parser in the sync plugin.

@mjudeikis

mjudeikis:
yes, its just XML with replacing done in bootstrap script on startup
https://github.com/Azure/moffet/blob/master/logging/images/mdsd/bootstrap.sh#L27

We could move all this logic into our code base if we want so

mjudeikis:
These 2 looks like candidated for code reuse:
```
"gopkg.in/xmlpath.v2"
"encoding/xml"
```

Its getting more complicated as its "XML inside of JSON/YAML":
```
"ConfigMap/openshift-azure-logging/mdsd-customerdata-config": {
		{
			JsonPath:      jsonpath.MustCompile("$.data.'mdsd.xml'"),
			XMLNesterPath: xmlpath.MustCompile("/MonitoringManagement/Management/Identity/IdentityComponent[@name='RPSector']"),
			Template:      "{{ String (CertAsBytes .Config.Certificates.Ca.Cert) }}",
		},
```

TranslateXML:
```
var nestedObject interface{}
	err = xml.Unmarshal(nestedBytes, &nestedObject)
	if err != nil {
		panic(err)
	}

	fmt.Println(string(nestedBytes))
	spew.Dump(nestedObject)
	return nil
```

And this part is where I got stuck. We need to `path.Set` implementation for `xmlplath` which challenge I lack the background to take on :/

PR #1002, opened 2019-01-09
Fixes to metrics

jim-minter:
* workaround an mdm issue that was preventing our metrics from appearing
* fix double-counting of metrics (set 60s interval)
* add cluster identification to metrics


pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
```diff
@@ -20,13 +20,13 @@ spec:
         - ./bootstrap.sh
         env:
         - name: SECTOR
-          value: '*** GENERATED ***'
+          value: dummy
```
jim-minter: fixing #776 will fix this ugliness

cmd/metricsbridge/example-config.yaml
```diff
@@ -1,4 +1,4 @@
-intervalNanoseconds: 30000000000 # every 30 seconds
```
mjudeikis: not sure how this would work because of prometheus scrape intervals. Prometheus scrapes each 30s, and we do each 10s. Need more reading on how it works :/

sozercan:
@jim-minter @amanohar we can make the interval configurable through RP as well if that makes sense

PR #1004, opened 2019-01-09
Validators split

mjudeikis:
This PR is proposal basis for now:

- Splits Validators to sub-package inside api package
- Splits 3 validators to separate interfaces
- has reusable validator functions for adminAPI and API
- isolates use of `RunningUnderTest` variable from validators to keep it to minimal. 

PR is built on top of https://github.com/openshift/openshift-azure/pull/922 as a follow-up 
fixes: https://github.com/openshift/openshift-azure/pull/922

/cc @jim-minter @kargakis @charlesakalugwu WDYT?

### DIFF
--- a/pkg/api/validate/admin.go
+++ b/pkg/api/validate/admin.go
@@ -1,0 +1,31 @@
+package validate
+
+import (
+	"errors"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+// Validate validates a OpenShiftManagedCluster struct
+func (v *AdminAPIValidator) Validate(cs, oldCs *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
+	// TODO are these error messages confusing since they may not correspond with the external model?
+	if oldCs == nil {
+		errs = append(errs, errors.New("admin requests cannot create clusters"))
+		return errs
+	}
+	if errs := validateContainerService(cs, externalOnly); len(errs) > 0 {
+		return errs
+	}
+	if errs := validateUpdateContainerService(cs, oldCs); len(errs) > 0 {
+		return errs
+	}
+	if errs := validateUpdateConfig(cs.Config, oldCs.Config); len(errs) > 0 {
+		return errs
+	}
+
+	for _, app := range cs.Properties.AgentPoolProfiles {
+		errs = append(errs, validateVMSize(app, v.runningUnderTest)...)
+	}
+
+	return nil
+}
--- a/pkg/api/validate/api.go
+++ b/pkg/api/validate/api.go
@@ -1,0 +1,18 @@
+package validate
+
+import (
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+// Validate validates a OpenShiftManagedCluster struct
+func (v *APIValidator) Validate(cs, oldCs *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
+	errs = append(errs, validateContainerService(cs, externalOnly)...)
+	errs = append(errs, validateUpdateContainerService(cs, oldCs)...)
+	// this limits use of RunningUnderTest variable inside our validators
+	// TODO: When removed this should be part of common validators
+	for _, app := range cs.Properties.AgentPoolProfiles {
+		errs = append(errs, validateVMSize(app, v.runningUnderTest)...)
+	}
+
+	return errs
+}
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -1,0 +1,143 @@
+package validate
+
+import (
+	"net"
+	"regexp"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+var (
+	rxRfc1123 = regexp.MustCompile(`(?i)^` +
+		`([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])` +
+		`(\.([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]))*` +
+		`$`)
+
+	rxVNetID = regexp.MustCompile(`(?i)^` +
+		`/subscriptions/[^/]+` +
+		`/resourceGroups/[^/]+` +
+		`/providers/Microsoft\.Network` +
+		`/virtualNetworks/[^/]+` +
+		`$`)
+
+	rxAgentPoolProfileName = regexp.MustCompile(`^[a-z0-9]{1,12}$`)
+
+	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
+	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
+
+	// This regexp is to check image version format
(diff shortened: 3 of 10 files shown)


pkg/api/validate/admin.go
```diff
@@ -0,0 +1,21 @@
+package validate
+
+import (
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+// Validate validates a OpenShiftManagedCluster struct
+func (v *AdminAPIValidator) Validate(new, old *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
+	if errs := validateContainerService(new, externalOnly); len(errs) > 0 {
```
kargakis: The first thing you should validate here is that oldCs is non-nil, otherwise you will return an error about the aad secret not being provided which is not available to admins at all. This is already implemented in the existing admin validation.

mjudeikis: One more reason to split as logic as becoming complicated...

pkg/api/validate/types.go
```diff
@@ -0,0 +1,145 @@
...
+
+var (
+	clusterNetworkCIDR *net.IPNet
+	serviceNetworkCIDR *net.IPNet
+)
+
+// APIValidator validator for external API
+type APIValidator struct {
+	runningUnderTest bool
+	isAdmin          bool
```
jim-minter: `isAdmin` can go away now, right?

pkg/api/validate/validators.go
```diff
@@ -278,13 +119,140 @@ func (v *Validator) validateUpdateContainerService(cs, oldCs *OpenShiftManagedCl
 	return
 }
 
-func (v *Validator) validateProperties(p *Properties, location string, externalOnly bool) (errs []error) {
+func validateFQDN(p *api.Properties, location string) (errs []error) {
```
jim-minter: note: the reordering of functions as part of this PR makes reviewing harder - separating the reordering to an otherwise no-op commit would have been handy, so that commit could be excluded at review time

pkg/plugin/plugin.go
```diff
@@ -22,9 +23,9 @@ type plugin struct {
 	clusterUpgrader         cluster.Upgrader
 	armGenerator            arm.Generator
 	configGenerator         config.Generator
-	apiValidator            *api.Validator
-	adminValidator          *api.Validator
-	pluginTemplateValidator *api.Validator
+	apiValidator            *validate.APIValidator
```
jim-minter: I don't believe there is any value to having these fields in the plugin.  The validators are not interfaces nor do I think they immediately need to become so.  Refactor the validate.NewFooValidator() calls to the point where each validator is actually used.

jim-minter:
+1 on the direction this is going

pkg/api/validate/types.go
mjudeikis: It is needed, but now its part of the check, not a struct.

pkg/api/validate/admin.go
```diff
@@ -0,0 +1,26 @@
...
+		errs = append(errs, errors.New("admin requests cannot create clusters"))
+		return errs
+	}
+	if errs := validateContainerService(new, externalOnly); len(errs) > 0 {
+		return errs
+	}
+	if errs := validateUpdateContainerService(new, old, externalOnly); len(errs) > 0 {
+		return errs
+	}
+	if errs := validateUpdateConfig(new.Config, old.Config); len(errs) > 0 {
```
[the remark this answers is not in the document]
mjudeikis: Tried some structure, but it is complex so check it our and let me know WDYT

pkg/api/validate/validators.go
mjudeikis: rollbacked where I could

pkg/plugin/plugin.go
mjudeikis: added as part of Validate main interfaces.

pkg/api/validate/api.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Validate validates a OpenShiftManagedCluster struct
+func (v *APIValidator) Validate(new, old *api.OpenShiftManagedCluster, externalOnly bool) (errs []error) {
+	errs = append(errs, validateContainerService(new, externalOnly)...)
+	errs = append(errs, validateUpdateContainerService(new, old, false)...)
+	// this limits use of RunningUnderTest variable inside our validators
+	// TODO: When removed this should be part of common validators
+	for _, app := range new.Properties.AgentPoolProfiles {
+		errs = append(errs, v.validateVMSize(app)...)
```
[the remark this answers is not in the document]
mjudeikis: I really dont like how it will propagate runningUnderTest variable... 
```
	for _, app := range cs.Properties.AgentPoolProfiles {
		errs = append(errs, validateVMSize(app, v.runningUnderTest)...)
	}
```
We force VM size in tests:
```
func isValidMasterAndInfraVMSize(size api.VMSize, runningUnderTest bool) bool {
	if runningUnderTest && size == api.StandardD2sV3 {
		return true
	}

	_, found := validMasterAndInfraVMSizes[size]
	return found
}
```

Should we just ditch it and make sure we create the right VM sizes in the first place?

pkg/api/validate/validators.go
```diff
@@ -245,46 +83,44 @@ func (v *Validator) validateContainerService(c *OpenShiftManagedCluster, externa
...
-func (v *Validator) validateUpdateContainerService(cs, oldCs *OpenShiftManagedCluster, externalOnly bool) (errs []error) {
-	if cs == nil || oldCs == nil {
-		errs = append(errs, fmt.Errorf("openShiftManagedCluster cannot be nil"))
+// validateUpdateContainerService should be called to validate ContainerServices object
+// validation conditions:
+// 1. cs not nil, oldCs nil, isAdmin - false - Create - no update validation
+// 2. cs not nil. oldCs not nil, isAdmin - true/false - Common Updates
+// 3. cs not nil. oldCs not nil, isAdmin - true - Admin Update
+// 4. cs not nil, oldCs not nil, isAdmin - false - Customer Update - NOT APPLICABLE YET
+// Any other use case is not valid call of the function
```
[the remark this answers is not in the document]
mjudeikis: done

pkg/api/validate/api.go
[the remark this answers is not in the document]
mjudeikis: This is still pending what we want to do. I would say remove this and control it via manifest file, where it should be.

pkg/api/validate/admin.go
```diff
@@ -0,0 +1,31 @@
...
+		return errs
+	}
+	if errs := validateUpdateContainerService(cs, oldCs); len(errs) > 0 {
+		return errs
+	}
+	if errs := validateUpdateConfig(cs.Config, oldCs.Config); len(errs) > 0 {
+		return errs
+	}
+
+	for _, app := range cs.Properties.AgentPoolProfiles {
```
kargakis: Why isn't this loop part of `validateContainerService`?

mjudeikis:
/retest

pkg/api/validate/admin.go
mjudeikis: Could be moved.

PR #1005, opened 2019-01-09
Add http routing fakerp

### LINKED ISSUE #994: Extend fakerp with simple http routing
We need to extend the fake rp with a simple http routing mechanism to allow us to handle various backend actions such as `rotate certificates`

/kind feature
/assign

charlesakalugwu:
closes #994


pkg/fakerp/server.go
```diff
@@ -57,57 +61,22 @@ func NewServer(log *logrus.Entry, resourceGroup, address string) *Server {
...
-		s.handleDelete(w, req)
-	case http.MethodGet:
-		s.handleGet(w, req)
-	case http.MethodPut:
-		s.handlePut(w, req)
-	}
-}
+	s.router.Delete("/", s.handleDelete)
+	s.router.Get("/", s.handleGet)
+	s.router.Put("/", s.handlePut)
```
mjudeikis: These now can be more gradually split to handler functions. we should not pass the request from admin and customer API to same handler function now.

pkg/fakerp/server.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/ghodss/yaml"
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
```
mjudeikis:
```
gnericapiserver "k8s.io/apiserver/pkg/server"
genericmux "k8s.io/apiserver/pkg/server/mux"
```

Did you checked these ones? They are already in our glide.

mjudeikis: OR in some sub-packages in k8s they use `"github.com/gorilla/mux` which very similar to what you use here,

pkg/fakerp/server.go
charlesakalugwu: yes those can be done in a series of follow-ups.

kargakis: You *may* want to use different entrypoints but you definitely want to share as much logic as possible.

pkg/fakerp/server.go
kargakis: I was wondering whether we could get away with avoiding the use of an external library. Not sure it's worth it for testing code. @jim-minter thoughts?

pkg/fakerp/middleware.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (s *Server) logger(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		s.log.Debugf("%s %s %s", r.RemoteAddr, r.Method, r.URL)
+		handler.ServeHTTP(w, r)
+	})
+}
+
+func (s *Server) validator(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
kargakis: Shouldn't you invalidate anything other than GET, PUT, and DELETE here too?

pkg/fakerp/server.go
```diff
@@ -57,70 +63,25 @@ func NewServer(log *logrus.Entry, resourceGroup, address string) *Server {
...
-	// TODO: match the request path the real RP would use
-	http.HandleFunc("/", s.handleFunc)
+func (s *Server) Run() {
+	s.router.Use(middleware.DefaultCompress)
+	s.router.Use(middleware.RedirectSlashes)
+	s.router.Use(middleware.Recoverer)
+	s.router.Use(s.logger)
+	s.router.Use(s.validator)
+
+	s.router.Get("/", s.handleGet)
```
kargakis: Open a follow-up issue to update the Location headers we return from deletes so we can stop serving at /.

pkg/fakerp/middleware.go
charlesakalugwu: The router takes care of that. A mapping for `s.router.Delete("/foo", s.handleFoo)` allows the router to only handle DELETES for /foo and returns a `405 Method not allowed` for all other methods

*Start fake rp*
```
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ go run cmd/fakerp/main.go 
INFO[2019-01-11T10:41:36+01:00] starting the fake resource provider          
INFO[2019-01-11T10:41:36+01:00] starting server on localhost:8080            
```
*Test*
```
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ curl -I -XGET http://localhost:8080/foo
HTTP/1.1 405 Method Not Allowed
Date: Fri, 11 Jan 2019 09:41:56 GMT
Content-Length: 0

charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ curl -I -XDELETE http://localhost:8080/foo
HTTP/1.1 200 OK
Date: Fri, 11 Jan 2019 09:42:07 GMT
Content-Length: 0

```

*Logs*
```      
DEBU[2019-01-11T10:41:56+01:00] 127.0.0.1:47362 GET /foo (curl/7.58.0)       
DEBU[2019-01-11T10:42:07+01:00] 127.0.0.1:47364 DELETE /foo (curl/7.58.0) 
```

pkg/fakerp/server.go
charlesakalugwu: https://github.com/openshift/openshift-azure/issues/1028

kargakis:

/retest

PR #1008, opened 2019-01-09
move AcceptLanguage to context

mjudeikis:
This further cleans PluginConfig and moves accept languages field into context object

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -44,7 +44,6 @@ const (
 var (
 	method  = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
 	useProd = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
-	timeout = flag.Duration("timeout", 30*time.Minute, "Timeout of the request to the OpenShift resource provider.")
 
 	adminManifest   = flag.String("admin-manifest", "", "If set, use the admin API to send this request.")
 	restoreFromBlob = flag.String("restore-from-blob", "", "If set, request a restore of the cluster from the provided blob name.")
@@ -173,7 +172,7 @@ func updateAadApplication(ctx context.Context, log *logrus.Entry, conf *fakerp.C
 		if err != nil {
 			return fmt.Errorf("cannot get authorizer from username+password: %v", err)
 		}
-		aadClient := azureclient.NewRBACApplicationsClient(conf.TenantID, authorizer, []string{"en-us"})
+		aadClient := azureclient.NewRBACApplicationsClient(ctx, conf.TenantID, authorizer)
 		callbackURL := fmt.Sprintf("https://%s.%s.cloudapp.azure.com/oauth2callback/Azure%%20AD", conf.ResourceGroup, conf.Region)
 		conf.AADClientSecret, err = fakerp.UpdateAADAppSecret(ctx, aadClient, conf.AADClientID, callbackURL)
 		if err != nil {
@@ -252,8 +251,7 @@ func main() {
 	adminClient.Authorizer = authorizer
 	v20180930previewClient.Authorizer = authorizer
 
-	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
-	defer cancel()
+	ctx := context.Background()
 	if isDelete {
 		if err := wait.PollImmediate(time.Second, 1*time.Hour, func() (bool, error) {
 			if err := delete(ctx, log, v20180930previewClient, conf.ResourceGroup, conf.NoWait); err != nil {
--- a/pkg/fakerp/config.go
+++ b/pkg/fakerp/config.go
@@ -1,6 +1,7 @@
 package fakerp
 
 import (
+	"context"
 	"crypto/rsa"
 	"crypto/x509"
 	"fmt"
@@ -11,6 +12,7 @@ import (
 	"github.com/ghodss/yaml"
 
 	"github.com/openshift/openshift-azure/pkg/api"
+	internalapi "github.com/openshift/openshift-azure/pkg/api"
 	pluginapi "github.com/openshift/openshift-azure/pkg/api/plugin/api"
 	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
 	"github.com/openshift/openshift-azure/pkg/tls"
@@ -30,8 +32,7 @@ func GetPluginConfig() (*api.PluginConfig, error) {
 	}
 
 	return &api.PluginConfig{
-		AcceptLanguages: []string{"en-us"},
-		TestConfig:      tc,
+		TestConfig: tc,
 	}, nil
 }
 
@@ -82,6 +83,16 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	template.Images.GenevaImagePullSecret = pullSecret
 
 	return template, nil
+}
+
+// enrichContext returns enriched context for plugin initiation
+func enrichContext(ctx context.Context) context.Context {
+	// simulate Context with property bag
+	// TODO: Get the azure credentials from the request headers
+	ctx = context.WithValue(ctx, internalapi.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
+	ctx = context.WithValue(ctx, internalapi.ContextKeyClientSecret, os.Getenv("AZURE_CLIENT_SECRET"))
+	ctx = context.WithValue(ctx, internalapi.ContextKeyTenantID, os.Getenv("AZURE_TENANT_ID"))
+	return ctx
 }
 
 func overridePluginTemplate(template *pluginapi.Config) {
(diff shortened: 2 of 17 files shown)


pkg/util/azureclient/azureclient.go
```diff
@@ -57,7 +57,13 @@ func (ls *loggingSender) Do(req *http.Request) (*http.Response, error) {
 	return resp, err
 }
 
-func setupClient(client *autorest.Client, authorizer autorest.Authorizer, languages []string) {
+func setupClient(ctx context.Context, client *autorest.Client, authorizer autorest.Authorizer) {
+	// if context does not provide languages (sync pod, tests) - use default
+	languages := []string{"en-us"}
```
jim-minter: nil is a fine default (`var languages []string`)

pkg/fakerp/config.go
```diff
@@ -84,6 +86,19 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
...
+// GetContext returns enriched context for plugin initiation
+func GetContext(duration time.Duration) (context.Context, context.CancelFunc) {
+	// simulate Context with property bag
+	ctx, cancel := context.WithTimeout(context.Background(), duration)
+	// TODO: Get the azure credentials from the request headers
+	ctx = context.WithValue(ctx, internalapi.ContextKeyClientID, os.Getenv("AZURE_CLIENT_ID"))
+	ctx = context.WithValue(ctx, internalapi.ContextKeyClientSecret, os.Getenv("AZURE_CLIENT_SECRET"))
+	ctx = context.WithValue(ctx, internalapi.ContextKeyTenantID, os.Getenv("AZURE_TENANT_ID"))
+	// AcceptLanguages via context
+	ctx = context.WithValue(ctx, internalapi.ContextAcceptLanguages, []string{"en-us"})
```
jim-minter: wouldn't bother setting it

pkg/fakerp/config.go
```diff
@@ -84,6 +86,19 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	return template, nil
 }
 
+// GetContext returns enriched context for plugin initiation
+func GetContext(duration time.Duration) (context.Context, context.CancelFunc) {
```
jim-minter: why do we hard code a timeout in any of this PR? can't we have the caller decide what they want? can this function be `func(ctx context.Context) context.Context` ?

cmd/createorupdate/createorupdate.go
```diff
@@ -252,7 +253,7 @@ func main() {
 	adminClient.Authorizer = authorizer
 	v20180930previewClient.Authorizer = authorizer
 
-	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
+	ctx, cancel := fakerp.GetContext(1 * time.Hour)
```
jim-minter: I'd like to see this being `ctx := fakerp.EnrichContext(context.Background())` and the timeout flag to die in a fire :)

cmd/etcdbackup/etcdbackup.go
```diff
@@ -69,7 +69,8 @@ func main() {
 		flag.Usage()
 	}
 
-	ctx := context.Background()
+	ctx, cancel := fakerp.GetContext(1 * time.Hour)
```
jim-minter: no hard coded timeout

jim-minter:
+1 on direction

cmd/createorupdate/createorupdate.go
kargakis: The problem is that you have different types of requests coming from the client (create takes ~15-20m, update is variable based on what is updated, restore is ~30m, key rotation is the longest since everything needs to get updated (~40-50m).

kargakis: deletes are ~10-15m

cmd/etcdbackup/etcdbackup.go
mjudeikis: Added as a flag

cmd/createorupdate/createorupdate.go
mjudeikis: we need timeout, as if somebody messes the PR, it might block CI for quite a while :/ 
Added it configurable via flag

cmd/etcdbackup/etcdbackup.go
```diff
@@ -24,6 +25,7 @@ var (
 	blobName    = flag.String("blobname", "", "Name of the blob (without the container)")
 	destination = flag.String("destination", "", "Where to place the blob on the filesystem")
 	maxBackups  = flag.Int("maxbackups", 6, "Maximum number of backups to keep")
+	timeout     = flag.Duration("timeout", 1*time.Hour, "Timeout of the request to the OpenShift resource provider.")
```
kargakis: This can probably be hardcoded since it's always going to be a single request that we know how much it takes more or less.

pkg/fakerp/server.go
```diff
@@ -136,12 +138,9 @@ func (s *Server) readAdminRequest(body io.ReadCloser) (*admin.OpenShiftManagedCl
 
 func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 	// simulate Context with property bag
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), s.timeout)
```
kargakis: There is no point in making this configurable too.

pkg/fakerp/server.go
```diff
@@ -265,12 +264,10 @@ func (s *Server) handlePut(w http.ResponseWriter, req *http.Request) {
 	}
 
 	// simulate Context with property bag
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
+	// TODO: Populate context from request header
+	ctx, cancel := context.WithTimeout(context.Background(), s.timeout)
```
kargakis: Not much gain from making this configurable too

cmd/etcdbackup/etcdbackup.go
mjudeikis: I would keep it consistent with other cmd files. Makes easier to understand. each cmd, which calls fakeRP would have this and it is set via flags.

pkg/fakerp/server.go
mjudeikis: just consistency, I think it just makes the behavior more understandable. but agreed, not much value

cmd/createorupdate/createorupdate.go
jim-minter: why? we already have code that uses /usr/bin/timeout, why do we need to maintain our own implementation of that inside createorupdate too?

pkg/fakerp/server.go
jim-minter: this has gone in the wrong direction - why can't we get rid of the timeout code across the board?

mjudeikis:
@jim-minter @kargakis timeouts gone from everywhere except https://github.com/openshift/openshift-azure/blob/master/pkg/fakerp/server.go#L137-L138

cmd/createorupdate/createorupdate.go
```diff
@@ -252,8 +252,7 @@ func main() {
 	adminClient.Authorizer = authorizer
 	v20180930previewClient.Authorizer = authorizer
 
-	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
-	defer cancel()
+	ctx := fakerp.EnrichContext(context.Background())
```
jim-minter: Remove the EnrichContext calls where they are not specifically needed (i.e. inside the fakepr process when it's going to pass that context into the plugin).  Remove this one and most if not all of the rest of the diff in this file will go back too.

cmd/etcdbackup/etcdbackup.go
```diff
@@ -69,7 +70,7 @@ func main() {
 		flag.Usage()
 	}
 
-	ctx := context.Background()
+	ctx := fakerp.EnrichContext(context.Background())
```
jim-minter: This is another to remove

test/clients/azure/client.go
```diff
@@ -57,6 +58,7 @@ func NewClientFromEnvironment(setStorageClient bool) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
+	ctx := fakerp.EnrichContext(context.Background())
```
jim-minter: also shouldn't be here?

pkg/fakerp/config.go
```diff
@@ -84,6 +85,16 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	return template, nil
 }
 
+// GetContext returns enriched context for plugin initiation
```
jim-minter: s/GetContext/enrichContext/

pkg/fakerp/config.go
```diff
@@ -84,6 +85,16 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
...
+func EnrichContext(ctx context.Context) context.Context {
```
jim-minter: unexport this

cmd/createorupdate/createorupdate.go
```diff
@@ -44,7 +44,6 @@ const (
 var (
 	method  = flag.String("request", http.MethodPut, "Specify request to send to the OpenShift resource provider. Supported methods are PUT and DELETE.")
 	useProd = flag.Bool("use-prod", false, "If true, send the request to the production OpenShift resource provider.")
-	timeout = flag.Duration("timeout", 30*time.Minute, "Timeout of the request to the OpenShift resource provider.")
```
kargakis: I think some of our hack scripts are using this flag, mind updating them?

kargakis:
Apart from updating `hack/`, the rest of the changes lgtm. Do we need MSFT to sign off?

PR #1012, opened 2019-01-10
Move deploy and update into the plugin for ease of unittesting

asalkeld:
This enables us to test the deploy and update flows with simple mocking instead of within cluster.


pkg/plugin/plugin_test.go
```diff
@@ -20,211 +20,52 @@ func TestCreateOrUpdate(t *testing.T) {
...
-			name:     "deploy: deploy error",
-			isUpdate: false,
-			wantErr:  true,
-			errStep:  api.PluginStepDeploy,
-		},
-		{
-			name:     "deploy: initialize error",
-			isUpdate: false,
-			wantErr:  true,
-			errStep:  api.PluginStepInitialize,
```
asalkeld: These tests were not very useful, so removing. the mocking below confirms the steps that we want to take for each flow.

pkg/cluster/healthcheck.go
```diff
@@ -68,3 +68,8 @@ func (u *simpleUpgrader) HealthCheck(ctx context.Context, cs *api.OpenShiftManag
 	u.log.Info("checking admin console health")
 	return u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs), "https://console."+cs.Properties.RouterProfiles[0].PublicSubdomain, 10*time.Second)
 }
+
+func (u *simpleUpgrader) WaitForHTTPStatusOk(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
```
mjudeikis: This name is confusing now. It is checking master healthz endpoint, but name is not define what we are checking :man_shrugging:

mjudeikis:
lgtm, will wait for next reviewer too

pkg/cluster/healthcheck.go
kwoodson: Probably makes sense to update this or create a follow up (issue/PR) that does.

pkg/plugin/plugin_test.go
```diff
@@ -20,211 +21,52 @@ func TestCreateOrUpdate(t *testing.T) {
...
+			AgentPoolProfiles: []api.AgentPoolProfile{
+				{Role: api.AgentPoolProfileRoleMaster, Name: "master"},
+				{Role: api.AgentPoolProfileRoleCompute, Name: "compute"},
+				{Role: api.AgentPoolProfileRoleInfra, Name: "infra"},
+			},
 		},
 	}
+
 	for _, tt := range tests {
 		mockGen := mock_arm.NewMockGenerator(mockCtrl)
```
jim-minter: it's a complete nit but I wonder if it's worth trying to standardise in the direction of calling the mocks by the fieldname they go into, ie s/mockGen/armGenerator/, s/mockUp/clusterUpgrader/. See update_master_test.go as an example.

pkg/plugin/plugin_test.go
```diff
@@ -20,211 +21,52 @@ func TestCreateOrUpdate(t *testing.T) {
...
-				} else {
-					mockUp.EXPECT().Deploy(nil, nil, nil, nil, gomock.Any()).Return(nil)
-				}
-				mockUp.EXPECT().WaitForInfraServices(nil, nil).Return(err)
-			}
+		mockGen.EXPECT().Generate(nil, cs, "", tt.isUpdate, gomock.Any()).Return(nil, nil)
+		mockUp.EXPECT().CreateClients(nil, cs).Return(nil)
+		mockUp.EXPECT().Initialize(nil, cs).Return(nil)
+		if tt.isUpdate {
+			mockUp.EXPECT().UpdateMasterAgentPool(nil, cs, &cs.Properties.AgentPoolProfiles[0]).Return(nil)
```
jim-minter: I'm interested to know whether it's plausible to strengthen this as in:
```golang
c := foo.EXPECT().Bar().Return(baz)
c = bar.EXPECT().Baz().Return(bash).After(c)
c = baz.EXPECT().Bash().Return(qux).After(c)
```
(again as in update_master_test.go)
or whether we think it'll make us go mad.  Thoughts?

pkg/cluster/healthcheck.go
```diff
@@ -68,3 +68,8 @@ func (u *simpleUpgrader) HealthCheck(ctx context.Context, cs *api.OpenShiftManag
 	u.log.Info("checking admin console health")
 	return u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs), "https://console."+cs.Properties.RouterProfiles[0].PublicSubdomain, 10*time.Second)
 }
+
+func (u *simpleUpgrader) WaitForHealthzStatusOk(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
+	_, err := wait.ForHTTPStatusOk(ctx, u.log, u.rt, "https://"+cs.Properties.FQDN+"/healthz")
```
jim-minter: can you think about whether it's worth removing `u.rt` at this point if the layering is better?  not sure either way - just have a think about it

pkg/plugin/plugin_test.go
asalkeld: done

pkg/plugin/plugin_test.go
asalkeld: done, easy enough

pkg/cluster/healthcheck.go
asalkeld: we *could*, but after trying it, it makes the plugin_test.go messy

PR #1014, opened 2019-01-10
[PROPOSAL] proposal for tagging releases with plugin config

mjudeikis:
Proposal for how we could track releases using git tags. 
This would allow us to test code upgrades between clusters in the production retrospectively, and our master code base.

Adding as a doc PR so we could convert it into doc as we discuss. 

/hold
/sig @openshift/sig-azure


docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
+# Initiating Plugin - Versioning model
+
+When initiating plugin code and calling `GenerateConfig` method we need to
+pass in `Plugin Template/PluginConfig` as a parameter.
```
jim-minter: add link

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
+# Initiating Plugin - Versioning model
+
+When initiating plugin code and calling `GenerateConfig` method we need to
+pass in `Plugin Template/PluginConfig` as a parameter. 
+
+Plugin template defines main characteristics for the cluster. In example 
+cluster, images versions, VM image, external integration details.
```
jim-minter: expand

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
...
+
+Config file examples can be found in `pluginconfig/pluginconfig-{version}.yaml`, but ideally config object should be constructed using `pkg/api/plugin/config.go Config` struct.
```
jim-minter: a config file is used to simplify prod configuration change without new binary rollout, however in general the struct contains secrets and it may not be wanted to store these in a yaml file

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
...
+
+`openshift-azure` repository code always contains stable newest version of file references.
+
+Before updating example file and code base for new release aka. code cut for MSFT RP, make sure you tag code with corresponding release tag version
```
jim-minter: I'm struggling to understand what the topic of the doc is from this point forwards.  What do you mean by tags here?

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+master--311.43.20181121--*--R1----*--R2------R3---
```
jim-minter: I don't understand what this diagram represents

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
...
+R1 tag -> `311.43.2018112`
+R2 tag -> `311.51.20190103`
+R3 tag should be `311.85.20180205` if no new image version will be released before code cut. 
+
+
+Where:
+* R1 will be reference to released 3.11.43 code base 
+* R2 will be reference to released 3.11.51 code base. 
+
+Note: Branch updates with newer version of the code base might (and in most cases will not) match with release tag. This is because in the end we are interested only in code base running in realRP.
```
jim-minter: I don't understand this either :-(

docs/pluginconfig.md
```diff
@@ -0,0 +1,39 @@
...
+
+This will allow us to create same version of the cluster using fakeRP by
+check-out previous version of the code. And check how cluster, created
+using older version of the code, will interact with newer version of the 
+code during upgrades.
+
+If needed, previous release might be branched into separate branch to enable additional code fixes. But this should not be practice.
```
jim-minter: Or this

docs/release.md
```diff
@@ -0,0 +1,68 @@
...
+5. Make sure VM image is released and is available to be used by plugin code.
+
+## Minor releases
+
+If there is a need to update already released code - minor release is done.
+This might happen if we make a mistake, want to update configuration, or not code related fix.
+
+Minor release creation flow:
+1. Add commit to release branch with required changes.
+2. Git tag minor version (`v3.1`)
```
thekad: if needed: commit to update `PluginConfig` referencing new images

docs/release.md
```diff
@@ -0,0 +1,68 @@
...
+Plugin code release is a new branch and considered a major release. Any VM image or config change is a minor release.
+
+# Versioning
+
+## Major releases
+
+Major release creation flow:
+1. Create a branch with a plugin code you wish to release.
+2. Commit to update `PluginConfig` referencing release image versions.
+3. Git tag major release version (`v3.0`)
```
thekad: Update docs/release notes/etc

docs/release.md
```diff
@@ -0,0 +1,68 @@
...
+3. Build release images (sync, metricsbridge, etc)
+
+Branching model:
+```
+Master-branch--*----------*-----------------------
+                \          \
+                 \-Release-v3------T(v3.0)----
+                             \-Release-v4----T(v4.0)-----T(v4.1)-
+```
+Note: Branch creation doesn't correspond to release, tags does.
```
thekad: There is really no change between e.g. "tagging v3.0" and "tagging v3.1", as both may require:
* code changes
* config changes
* doc changes

The _only_ difference lies in cutting a new branch "release-v3" which is a pre-requisite for "tag-v3.0"

docs/release.md
```diff
@@ -0,0 +1,68 @@
...
+1. Add commit to release branch with required changes.
+2. Git tag minor version (`v3.1`)
+3. Build release images (sync, metricsbridge, etc)
+
+Branching model:
+```
+-------master--*----------*-----------------------
+                \          \
+                 \-release-v3------T(v3.0)----
+                             \-release-v4----T(v4.0)-----T(v4.1)-
```
asalkeld: this doesn't make a lot of sense to me, maybe it's just the graphic.
imho major releases should be made from master not the previous release.

```
-------master--*----------*-----------------------
               \          \
                \           \-release-v4----T(v4.0)-----T(v4.1)-
                 \-release-v3------T(v3.0)----

```

mjudeikis: yes, its graphic,  it is as you pained it :) will use your ASCII :)

docs/release.md
```diff
@@ -0,0 +1,68 @@
+# OSA Release process
+
+OSA Release process consist of multiple dimensions:
+* Plugin code release
+* Config release
+* VM image release
+
+Update might be one or any combination of all dimensions.
```
asalkeld: I assume that this is not totally true, as in we *may* not support upgrading between openshift 3.x and 4.x (so how do we restrict this). we should probably assume that the vm images and config are at the same major version as the plugin code - right/no?
also are we going to use the same versioning scheme between the 3 entities or are the versions unrelated? (just really asking you to be specifc) - is the version based off of the openshift version or independent? - given that the addons are probably bound to a specifc openshift version it would make sense these are the same and related to the openshift version. arguably we could place the stuff that runs within the cluster in a seperate repo (sync pod/controller/backup and addons)

docs/release.md
```diff
@@ -0,0 +1,72 @@
...
+
+Config file examples can be found in `pluginconfig/pluginconfig-{version}.yaml`. A config file is used to simplify production configuration change without new binary rollout. Struct contains secrets and it may not be wanted to store these in a yaml file.
+
+# Testing architecture consideration
+
+1. Hourly image sync job should be smart enough to detect tags and build "sync list" based on those.
+2. CI infrastructure should be able to rebuild all container images retrospectively in case of CI cluster rebuild.
+   Q: How we can build container images using git tags instead of branches using CI-Operator?
+   Looks like this is not possible, and we might need to develope way to do this using OpenShift build
+
```
mjudeikis: This might be a big challenge. Currently, if you commit to private-preview-1.*, or master branch, it will build container images and if tests are successful - promotes them as part of the post-submit job. But if we want this to be based on tags - we can't do this straight forward. We can test HEAD of the branch, but not specific tags. 
The flow would be:
Git tag release-1.0 branch
CI operator will build images into IS: `release-1:sync,metricsbridge`
Now we are responsible for `oc tag release-1:sync release-1.1:sync,metricsbridge`

And if new tag is added to `release-1` branch - it will overwrite the initial image, so we need step above happen before we move on in the branch to the next tag. Because step is not automated, we have half of the process automatable, and half not.

docs/release.md
```diff
@@ -0,0 +1,72 @@
...
+3. We should publish fakeRP images with same semantic versioning as part of release. These images will be used to 
+   test all possible update/upgrade scenarios.
+4. Test should stop using emptyDir for passing configuration around and download it from azure. This will enable 
+   update tests
+5. Prow configuration for testing the releases should be generated automatically using reference file with test
+   matrix.
+
```
mjudeikis: 6. How we validate if adminAPI call SRE is submitting is part of the "supported and tested path" 
    We could validate it as part of validation code, but this means we would need to maintain code based validation of each release, and in some cases if we doing hotfixes this might not be possible to update

mjudeikis: Notes: 
```
1) the customer create API uses the yaml plugin config template to work out the versions used for a brand new cluster.  Perhaps we should have a specialised admin API which does the same for upgrade - i.e. upgrade my cluster to the versions in your on-disk config template, i.e. what you believe is the current state of the world.   Then we use the generic admin API only for fiddling if we really need it
2) we want to simplify our supported version graph as much as possible: limit the number of concurrent versions in the wild so the graph is small, don't allow rollback.  Excluding major OpenShift version upgrade, I think we want to try allowing upgrade of n-2 to n without going through an n-1 step.
that way the supported version graph really matters only for testing, not for actual RP behaviour

mjudeikis [8:54 PM]
> perhaps we should have a specialised admin API which does the same for upgrade - i.e. upgrade my cluster to the versions in your on-disk config template
so this still goes to point 2. We need to keep this graph for validation somewhere.
and I dont want one more config file, but at this point I dont see any other way(?)

Jim Minter [8:57 PM]
"the graph" hopefully reduces to: "the list of all cluster versions in the wild" (say v5, v6, v7)
then we test upgrade(v5, master); upgrade(v6, master); upgrade(v7, master)
and once v5 clusters are all gone we remove v5 from the list
ok, v5.0, v6.0, v6.1, v7.0
to be typesafe :wink:
@mj it's on you to get all this down in a doc :wink:

mjudeikis [9:00 PM]
but we need to validate so somebody would not submit v5.0 to v7.0 upgrade and do it. But we need to support it as guy who was built with v5.0 already do not have access to 6.0 RP to make step by step upgrade
````

File: docs/release.md (hunk 12 above)
Comment (mjudeikis) [reply docs/release.md]: > we should probably assume that the vm images and config are at the same major version as the plugin code - right/no? 

Not sure about this. I think we can version plugin code separately, so if we release fast we will not end up with `3.11.something.354` 

> also are we going to use the same versioning scheme between the 3 entities or are the versions unrelated? (just really asking you to be specifc) - is the version based off of the openshift version or independent? - given that the addons are probably bound to a specifc openshift version it would make sense these are the same and related to the openshift version. arguably we could place the stuff that runs within the cluster in a seperate repo (sync pod/controller/backup and addons)

We should lock config version with VM version, but plugin code versioning might be different and independent. 

Do you have versioning standard in mind?

File: docs/release.md (hunk 8 above)
Comment (mjudeikis) [reply docs/release.md]: Part of point 1

File: docs/release.md (hunk 10 above)
Comment (mjudeikis) [reply docs/release.md]: I think if we release new code (if this is not typo fixing or super small logic change) we will do a new release. But we need to be cautious so we would not get ourselves into "upgrade hell"

File: docs/release.md [hunk 15]
```diff
@@ -0,0 +1,77 @@
+# OSA Release process
+
+OSA Release process consist of multiple dimensions:
+* Plugin code release
+* Config release
+* VM image release
```
kargakis: Missing the container images?

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+
+Update might be one or any combination of all dimensions. But individual upgrade path is validated using adminAPI. 
+Plugin code release is a new branch and considered a major release. Any VM image or config change is a minor release.
+
+# Versioning
+
+## Major releases
+
+Major release creation flow:
+1. Create a branch with a plugin code you wish to release.
```
kargakis: Checkout a new branch out of `master`.

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+2. Commit to update `PluginConfig` referencing release image versions.
+3. Git tag major release version (`v3.0`)
+4. Build release images (sync, metricsbridge, etc)
+5. Make sure VM image is released and is available to be used by plugin code.
+6. Update docs & release notes
+
+## Minor releases
+
+If there is a need to update already released code - minor release is done.
+This might happen if we make a mistake, want to update configuration, or not code related fix.
```
kargakis: Can you give me an example of a "not code related fix"?

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+
+Config file examples can be found in `pluginconfig/pluginconfig-{version}.yaml`. A config file is used to simplify production configuration change without new binary rollout. Struct contains secrets and it may not be wanted to store these in a yaml file.
+
+#  Architecture consideration / Topics
+
+1. Hourly image sync job should be smart enough to detect tags and build "sync list" based on those.
+   
+2. CI infrastructure should be able to rebuild all container images retrospectively in case of CI cluster rebuild.
+   Q: How we can build container images using git tags instead of branches using CI-Operator?
+   Looks like this is not possible, and we might need to develope way to do this using OpenShift build
```
kargakis: ci-operator is using `git ls-remote` to resolve git references into commit hashes so tags should be supported.

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+   Potentially customer build generator to generate build based on tags.
```
kargakis: We actually need automation to generate new prowjobs for every new tag.

kargakis: Yes, I think this is what we need. ci-operator can build from tags but we should consider avoiding prow altogether since these builds are one-off.

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+   
+2. CI infrastructure should be able to rebuild all container images retrospectively in case of CI cluster rebuild.
+   Q: How we can build container images using git tags instead of branches using CI-Operator?
+   Looks like this is not possible, and we might need to develope way to do this using OpenShift build
+   Potentially customer build generator to generate build based on tags. 
+
+3. We should publish fakeRP images with same semantic versioning as part of release. These images will be used to 
+   test all possible update/upgrade scenarios.
+
+4. Test should stop using emptyDir for passing configuration around and download it from azure. This will enable
```
kargakis: and instead download the config blob from azure.

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+   Looks like this is not possible, and we might need to develope way to do this using OpenShift build
+   Potentially customer build generator to generate build based on tags. 
+
+3. We should publish fakeRP images with same semantic versioning as part of release. These images will be used to 
+   test all possible update/upgrade scenarios.
+
+4. Test should stop using emptyDir for passing configuration around and download it from azure. This will enable 
+   update tests
+
+5. Prow configuration for testing the releases should be generated automatically using reference file with test
```
kargakis: As I said in a previous comment, we should consider avoiding prow altogether or ensure these prowjobs are created directly in the `ci` namespace, since they are one-off and do not need to be rebuilt a second time (remember that prow needs to run jobs either as presubmits, postsubmits, or periodics). There has been some discussion in enabling a specific type of a job that runs on tag creation but I am not sure where we stand on that @stevekuznetsov @smarterclayton

docs/release.md
mjudeikis: I think all non code related fix can be put under 'configuration' section

docs/release.md
```diff
@@ -0,0 +1,77 @@
...
+images:
+  alertManagerBase: registry.access.redhat.com/openshift3/prometheus-alertmanager # operators base images 
+  ansibleServiceBroker: registry.access.redhat.com/openshift3/ose-ansible-service-broker:v3.11.43 # specific image version
+```
+
+Config file examples can be found in `pluginconfig/pluginconfig-{version}.yaml`. A config file is used to simplify production configuration change without new binary rollout. Struct contains secrets and it may not be wanted to store these in a yaml file.
+
+#  Architecture consideration / Topics
+
+1. Hourly image sync job should be smart enough to detect tags and build "sync list" based on those.
```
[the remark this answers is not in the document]
mjudeikis: This was about our image-sync from CI cluster into quay. will clarify

docs/release.md
mjudeikis: perfect!

jim-minter:
For now, closing for https://docs.google.com/document/d/1einHXiv6r6tI5i1ce_PkZybmCYOWAqtYz5bgzWhKr5Q, may reopen to commit the results of that doc

Issue #1011, opened 2019-01-09
use same co-ordinates (region, subscriptionId, resourceGroupName) to identify logs as well as metrics


jim-minter:
at the moment in mdsd.xml we use a different set of `IdentityComponents` to reference clusters than we do for metrics (see cmd/metricsbridge).  Confirm with MSFT and change the logging references to match metrics.  Consider adding in the cluster UUID as an identifier in both cases as well.

I'm hoping we can see RPSector / LoggingSector, etc. go away; also Service, Host, ResourceName can go; also s/RPTenant/Region/.

Will need to fixup the use of bootstrap.sh in the mdsd pod - maybe get rid of its use like we did for mdm?

I tried doing all of the but something broke - I wonder whether a backend configuration change is also necessary.  Check with MSFT.

mjudeikis:
Form metrics bridge:
```
		"region":            {},
		"subscriptionid":    {},
		"resourcegroupname": {},
```

mjudeikis:
/close
Should be fixed now

PR #1018, opened 2019-01-10
populate cluster id

mjudeikis:
This populates clusterID field same as realRP.

In preparation for https://github.com/openshift/openshift-azure/issues/1011 to have same data in realRP and fakeRP

/cc @jim-minter

### DIFF
--- a/pkg/fakerp/fakerp.go
+++ b/pkg/fakerp/fakerp.go
@@ -19,6 +19,7 @@ import (
 	"github.com/openshift/openshift-azure/pkg/plugin"
 	"github.com/openshift/openshift-azure/pkg/tls"
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	"github.com/openshift/openshift-azure/pkg/util/resourceid"
 )
 
 func GetDeployer(log *logrus.Entry, cs *api.OpenShiftManagedCluster, config *api.PluginConfig) api.DeployFn {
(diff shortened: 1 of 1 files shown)


mjudeikis:
@jim-minter  you mentioned this:
> Consider adding in the cluster UUID as an identifier in both cases as well.

We already have this in prodRP:
`"id": "/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourcegroups/mjudeikis-prd1/providers/Microsoft.ContainerService/openshiftmanagedClusters/mjudeikis-prd1",` is this the one you are talking about?

pkg/fakerp/fakerp.go
```diff
@@ -212,6 +212,8 @@ func enrich(cs *api.OpenShiftManagedCluster) error {
 		Secret:   os.Getenv("AZURE_CLIENT_SECRET"),
 	}
 
+	// /subscriptions/{subscription}/resourcegroups/{resource_group}/providers/Microsoft.ContainerService/openshiftmanagedClusters/{cluster_name}
+	cs.ID = fmt.Sprintf("/subscriptions/%s/resourcegroups/%s/providers/Microsoft.ContainerService/openshiftmanagedClusters/%s", cs.Properties.AzProfile.SubscriptionID, cs.Properties.AzProfile.ResourceGroup, cs.Properties.AzProfile.ResourceGroup)
```
kargakis: Use `cs.Name` instead of `cs.Properties.AzProfile.ResourceGroup` for the cluster name.

jim-minter: use resourceid.ResourceID() instead of fmt.Sprintf()

PR #1020, opened 2019-01-10
enable r/o cache on etcd disk

jim-minter:
I don't think this will fundamentally change our performance, but I also think it's better on than off.

### DIFF
--- a/pkg/arm/resources.go
+++ b/pkg/arm/resources.go
@@ -582,6 +582,7 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
 		vmss.VirtualMachineProfile.StorageProfile.DataDisks = &[]compute.VirtualMachineScaleSetDataDisk{
 			{
 				Lun:          to.Int32Ptr(0),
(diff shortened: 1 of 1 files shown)


kargakis:
From the go sdk godoc
```go
// Caching - Specifies the caching requirements. <br><br> Possible values are:
// <br><br> **None** <br><br> **ReadOnly** <br><br> **ReadWrite** <br><br> 
// Default: **None for Standard storage. ReadOnly for Premium storage**. Possible
// values include: 'CachingTypesNone', 'CachingTypesReadOnly', 'CachingTypesReadWrite'
```
https://godoc.org/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute#VirtualMachineScaleSetDataDisk

Is there any other place we can read more about this change?

jim-minter:
The godoc is wrong because it is defaulting to no caching despite the fact we're using premium storage.

See https://docs.microsoft.com/en-us/azure/virtual-machines/windows/premium-storage-performance#disk-caching for more information.

I think this change could help us with read-based IOPS, even if not for write-based.

PR #1021, opened 2019-01-10
make PeerVnetID a pointer in the internal model again

### LINKED ISSUE #1009: returning peerVnetId
`peerVnetId` is returned as an empty string (`""`) now for non-custom vnet scenarios. This used to be not returned at all I believe. This is causing issues with CLI/ARM as they weren't expecting any value.

@jim-minter do you have any thoughts whether it should be returned as an empty string or not returned at all?

cc @amanohar @julienstroheker

jim-minter:
fixes #1009 
@sozercan


jim-minter:
I suspect there may be a few more of these to do in the future.

PR #1022, opened 2019-01-10
Move router and docker-registry from deployments to daemonsets

y-cote:
This change puts router and docker-registry instances and run them as
daemonsets so that a copy of each runs on each infra nodes.


kargakis:
We can remove `affinity` from both manifests now.

kargakis:
#560 

/retest

kargakis:
Hrm, these errors are underlying issues in the CI cluster. @y-cote a good chance for you to log into the test namespace and see what's failing. I would suggest triaging the latest test runs and see whether the old test results are getting reused. I suspect some of the failed builds are not restarting so the e2e test fails to pull the right images in order to complete.

PR #1027, opened 2019-01-11
Add Worker scaling and update unittests

asalkeld:
This is just a start, I'll add more


pkg/cluster/scaler/scaler.go
```diff
@@ -12,6 +18,22 @@ import (
...
+}
+
+type scalerFactory struct{}
+
+func NewFactory() Factory {
+	return &scalerFactory{}
+}
+
+type Scaler interface {
+	Scale(ctx context.Context, count int64) *api.PluginError
```
kargakis: Just by looking at this interface it's not clear to me what's the point of Scale when there is ScaleUp and ScaleDown. Certainly comments would help if this is still required.

kargakis: Hrm, Scale seems to encapsulate ScaleUp and ScaleDown. I don't think this is the right interface to expose to clients :thinking:

asalkeld: @kargakis I am just exposing the function that are used here:
https://github.com/openshift/openshift-azure/pull/1027/files#diff-848159614a4ba38049bdf602a8654627R100

kargakis: would be useful to add docs since `count` is used differently in Scale than ScaleUp and ScaleDown.

asalkeld: will do

pkg/cluster/scaler/scaler.go
```diff
@@ -12,6 +18,27 @@ import (
...
+
+// NewFactory create a new Factory instance.
+func NewFactory() Factory {
+	return &scalerFactory{}
+}
+
+// Scaler is an interface that changes the number of VMs in a scaleset until they meet
+// the 'count' provided in each method. Scale function simply checks which direction
+// to scale up or down.
+type Scaler interface {
```
kargakis: It still feels strange to me that we are exposing methods encapsulated in other methods. Why would a caller want to use ScaleUp or ScaleDown if Scale achieves the same thing?

kargakis: It still feels strange to me that we are exposing methods encapsulated in other methods in the same interface. Why would a caller want to use ScaleUp or ScaleDown if Scale achieves the same thing?

pkg/cluster/update_worker_test.go
```diff
@@ -0,0 +1,114 @@
...
+			ssc := mock_azureclient.NewMockVirtualMachineScaleSetsClient(gmc)
+			kclient := mock_kubeclient.NewMockKubeclient(gmc)
+			hasher := mock_cluster.NewMockHasher(gmc)
+			ss := &compute.VirtualMachineScaleSet{
+				Name: to.StringPtr("ss-compute-1"),
+				Sku:  &compute.Sku{Capacity: to.Int64Ptr(0)}}
+
+			targetScaler := mock_scaler.NewMockScaler(gmc)
+			var count int64 = 0
+			targetScaler.EXPECT().Scale(ctx, count)
```
jim-minter: could s/count/int64(0)/ here fwiw

jim-minter: I'm a bit confused that ss is a scaleset with 0 capacity and the test is expecting a scale to 0.  I don't call that "basic coverage".

pkg/cluster/scaler/scaler.go
jim-minter: Think I don't mind either way.

pkg/cluster/scaler/scaler_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+			// update
+			ssc := mock_azureclient.NewMockVirtualMachineScaleSetsClient(gmc)
+			ssc.EXPECT().Update(ctx, testRg, testSS, compute.VirtualMachineScaleSetUpdate{
+				Sku: &compute.Sku{
+					Capacity: to.Int64Ptr(tt.count),
+				},
+			})
+			// list with new vms
+			vmc.EXPECT().List(ctx, testRg, testSS, "", "", "").Return(tt.vmsAfter, nil)
```
asalkeld: i'll make sure the order is correct

pkg/cluster/update_worker_test.go
asalkeld: agreed, i have a better test on the way

jim-minter:
@asalkeld can you squash this?

asalkeld:
@jim-minter squashed into 2 logical commits

Issue #1029, opened 2019-01-11
fluentd errors in the logs


mjudeikis:
With current fluentd config multiple errors in the logs:
```
root@mdsd-p8kv5:/td-agent# tail -f /var/log/td-agent.log 
2019-01-11T13:14:13.8196120Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196210Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196270Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196340Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196410Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196480Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8196550Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8259310Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8259560Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
2019-01-11T13:14:13.8259730Z: error DataReader.cc:224 unexpected mdsd ack status: ACK_DECODE_ERROR, tag '0'
```

/kind bug

jim-minter:
I feel like these could easily be caused by fluentd being killed, restarting and the djsonsocket being out of sync.

goashut:
Hi Mangirdas, hope you are doing great. I am facing same error in our fluentd set up. Were you guys able to figure this out  ? It would be great if you could share the info..

mjudeikis:
@goashut we use your fluentd for this as its integrated with Geneva and MDSD. Issues are with support for UTF-8 support on MDSD. it needs to be converted to ASCII-7:

[...]

PR #1030, opened 2019-01-11
Logging & Metrics dimensions

### LINKED ISSUE #776: Check if mdsd-customerdata-config.yaml is populated by naive substitution and avoid if possible

mjudeikis:
This unifies all metrics and logging under 3 values:

1. ResourceGroupName - unique in prodRP, not in fakeRP
2. ResourceName - cluster name
3. SubscriptionID 

Using these 3, we will always be able to identify cluster in metrics and logging systems.
For debugging we should be asking customers to provide us clusterID: 
`"id": "/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourcegroups/mjudeikis-prd2/providers/Microsoft.ContainerService/openshiftmanagedClusters/mjudeikis-prd2",` 

which will contain all 3 required metrics/logging dimensions

/cc @jim-minter 

Still have this: https://github.com/openshift/openshift-azure/issues/1029
But this is in MSFT configuration too. Still need to drill down where is the issue

closes #776


mjudeikis:
partially solves: https://github.com/openshift/openshift-azure/issues/1011

pkg/addons/translate.go
```diff
@@ -382,6 +383,14 @@ var Translations = map[string][]struct {
 			Path:     jsonpath.MustCompile("$.spec.template.spec.containers[1].env[?(@.name='MONITORING_GCS_ACCOUNT')].value"),
 			Template: "{{ .Config.GenevaLoggingControlPlaneAccount }}",
 		},
+		{
+			Path:     jsonpath.MustCompile("$.spec.template.spec.containers[1].env[?(@.name='MONITORING_GCS_ENVIRONMENT')].value"),
+			Template: "{{ .Config.GenevaLoggingControlPlaneEnvironment }}",
+		},
+		{
+			Path:     jsonpath.MustCompile("$.spec.template.spec.containers[1].env[?(@.name='MONITORING_GCS_REGION')].value"),
```
mjudeikis: @sozercan What is the reason behind this part of the code? https://github.com/Azure/moffet/blob/master/logging/images/mdsd/bootstrap.sh#L17-L18

Can it be just a parameter passed from RP?

sozercan: It can be passed from RP, checking imds makes less variables to be passed in

mjudeikis: with current pluginConfig we either way will be configuring most of the things so +1 will not make huge diff :)

pkg/addons/data/ConfigMap/openshift-azure-logging/mdsd-config.yaml
```diff
@@ -7,14 +7,11 @@ data:
             <Account moniker="_ACCOUNT_" isDefault="true" autoKey="false" />
         </Accounts>
         <Management eventVolume="Large" defaultRetentionInDays="90" >
-            <Identity tenantNameAlias="RPTenant" roleNameAlias="Service" roleInstanceNameAlias="Host">
-                <IdentityComponent name="RPSector">_SECTOR_</IdentityComponent>
-                <IdentityComponent name="RPTenant">_REGION_</IdentityComponent>
-                <IdentityComponent name="Service">OSA</IdentityComponent>
-                <IdentityComponent name="Host" useComputerName="true" />
+            <Identity tenantNameAlias="ResourceName" roleNameAlias="ResourceGroupName" roleInstanceNameAlias="SubscriptionId">
```
jim-minter: it is not clear to me that this is valid - do these need to be OSA service identifiers?

pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
```diff
@@ -46,26 +46,37 @@ spec:
...
         - mountPath: /var/log
           name: hostlog
-      - args:
+      - command:
+        - /bin/bash
         - -c
-        - ./bootstrap.sh
+        - |
+          #!/bin/bash
+          sed -e s/_ACCOUNT_/$ACCOUNT/g -e s/_NAMESPACE_/$NAMESPACE/g -e s/_REGION_/$REGION/g -e s/_SUBSCRIPTIONID_/$SUBSCRIPTION_ID/g -e s/_RESOURCEGROUPNAME_/$RESOURCE_GROUP_NAME/g -e s/_RESOURCENAME_/$RESOURCE_NAME/g /etc/mdsd.d/template/mdsd.xml > /etc/mdsd.d/mdsd.xml
```
jim-minter: let's get rid of this now rather than waiting more - make a function in config.Derived along the lines of https://gist.github.com/jim-minter/58c203894459498b1c9e295e433b5ad2

pkg/addons/data/DaemonSet.apps/openshift-azure-logging/mdsd.yaml
```diff
@@ -46,26 +46,37 @@ spec:
...
+          rm -f /var/log/mdsd.log
+          stdbuf -o0 mdsd -D -R 2>&1 | tee /var/log/mdsd.log
```
jim-minter: why do we need the tee or stdbuf? Can they be removed?

pkg/addons/data/ConfigMap/openshift-azure-logging/mdsd-config.yaml
mjudeikis: For this, we need to discuss. I would like this to confirm with MSFT.
This is 3 default identities for Jarvis. We can disable use of defaults and don't have these.

pkg/config/derive.go
```diff
@@ -139,3 +142,56 @@ func (derived) StatsdArgs(cs *api.OpenShiftManagedCluster) ([]interface{}, error
 		"-SourceRoleInstance", "OSA",
 	}, nil
 }
+
+func (derived) MDSDConfig(cs *api.OpenShiftManagedCluster) ([]byte, error) {
```
mjudeikis: @jim-minter I will try to incorporate this into main template engine, Or we want to keep XML out of it? Not sure if this is even possible with current state.

pkg/config/derive.go
```diff
@@ -139,3 +142,56 @@ func (derived) StatsdArgs(cs *api.OpenShiftManagedCluster) ([]interface{}, error
...
+			return b.String(), err
+		},
+	}).Parse(tmpl))
+
+	b := &bytes.Buffer{}
+
+	err := t.Execute(b, map[string]string{
+		"Namespace":         cs.Config.GenevaLoggingNamespace,
+		"Account":           cs.Config.GenevaLoggingAccount,
+		"Sector":            cs.Config.GenevaLoggingSector,
```
jim-minter: I think GenevaLoggingSector is completely unused now and can be removed?

pkg/config/derive.go
mjudeikis: this gives us circular imports... need to re-arrange stuff :/

pkg/config/derive.go
```diff
@@ -139,3 +142,56 @@ func (derived) StatsdArgs(cs *api.OpenShiftManagedCluster) ([]interface{}, error
 		"-SourceRoleInstance", "OSA",
 	}, nil
 }
+
+func (derived) MDSDConfig(cs *api.OpenShiftManagedCluster) ([]byte, error) {
+	var tmpl = `<?xml version="1.0" encoding="utf-8"?>
+    <MonitoringManagement version="1.0" namespace="{{ .Namespace | Escape }}" eventVersion="1" timestamp="2017-08-01T00:00:00.000Z">
```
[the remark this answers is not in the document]
mjudeikis: https://github.com/openshift/openshift-azure/issues/1054

pkg/config/derive.go
```diff
@@ -223,3 +226,55 @@ func (derived) MaxDataDisksPerVM(cs *api.OpenShiftManagedCluster) (string, error
...
 	return "", fmt.Errorf("unknown VMSize %q", app.VMSize)
 }
+
+func (derived) MDSDConfig(cs *api.OpenShiftManagedCluster) (string, error) {
+	var tmpl = `<?xml version="1.0" encoding="utf-8"?>
+    <MonitoringManagement version="1.0" namespace="{{ .Namespace | Escape }}" eventVersion="1" timestamp="2017-08-01T00:00:00.000Z">
+        <Accounts>
+            <Account moniker="{{ .Account | Escape }}" isDefault="true" autoKey="false" />
+        </Accounts>
+        <Management eventVolume="Large" defaultRetentionInDays="90" >
```
mjudeikis: TODO: Check if pass region (cluster region, not RP region) and make sure we log and metric same

sozercan: putting a note for passing in OS_ rg also

pkg/config/derive.go
```diff
@@ -223,3 +226,55 @@ func (derived) MaxDataDisksPerVM(cs *api.OpenShiftManagedCluster) (string, error
...
+            </Identity>
+            <AgentResourceUsage diskQuotaInMB="50000" />
+        </Management>
+        <Sources>
+            <Source name="journald" dynamic_schema="true" />
+        </Sources>
+        <Events>
+            <MdsdEvents>
+                <MdsdEventSource source="journald">
+                    <RouteEvent eventName="CustomerSyslogEvents" storeType="CentralBond" priority="Normal"/>
```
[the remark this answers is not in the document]
mjudeikis: Agreed. We already see OOM on fluentd. Need to cut this down. Potentially as separate PR to unblock this from a merge.

pkg/config/derive.go
mjudeikis: is `OS_rg` is not the same as `ResourceGroupName` ?

sozercan: `ResourceGroupName` is the one with uuid i believe, `OS_...` rg is the one that contains the managed app resource

mjudeikis: Just thinking, how this name is accessible to the plugin?

sozercan: looks like i missed this comment. not sure if plugin can access from somewhere, rp can pass this.

jim-minter:
@mjudeikis please rebase

PR #1033, opened 2019-01-11
Add a Root docker config.json

thekad:
This allows the operator to specify a `config.rootDockerConfig` entry in the configuration YAML file. This file is going to be placed in the nodes' file system in a place where docker will pick it up by default.
We need this so all nodes and all namespaces have access to private repositories.


thekad:
@jim-minter @mjudeikis ptal

thekad:
Things I am not so sure about in this PR:
* The location of the .docker/config.json file, the bug mentioned means I have to do one of a) set HOME env var for the kubelet process, b) put the file in the working directory set in the kubelet systemd unit c) put it at the very top in /.docker. I went with b) here by putting the file in /root and then symlinking but I'm still undecided
* Mock generated stuff showed changes for me, I pushed but maybe they aren't real and I need to update something in my dev env?
* Unsure how the base64 encoded string from the YAML config file gets decoded, I followed @mjudeikis steps like he did for the geneva image pull secret but not sure if that will work at the moment, since I am testing with fakerp
* End user test is extremely simple, since we just want to test access to the registry, I don't test any other pieces, if you think this should have more things, lmk

pkg/api/config.go
```diff
@@ -21,8 +21,9 @@ type Config struct {
 	ConfigStorageAccount   string `json:"configStorageAccount,omitempty"`
 	RegistryStorageAccount string `json:"registryStorageAccount,omitempty"`
 
-	Certificates CertificateConfig `json:"certificates,omitempty"`
-	Images       ImageConfig       `json:"images,omitempty"`
+	Certificates     CertificateConfig `json:"certificates,omitempty"`
+	RootDockerConfig []byte            `json:"rootDockerConfig,omitempty"`
```
mjudeikis: s/RootDockerConfig/ImagePullSecret/ ? to match existing ones?

pkg/api/converterfromplugin_test.go
```diff
@@ -46,10 +46,11 @@ func internalPluginConfig() Config {
 				Key:  tls.GetDummyPrivateKey(),
 			},
 		},
+		RootDockerConfig: []byte("RootDockerConfig"),
 		// Container images configuration
 		Images: ImageConfig{
-			GenevaImagePullSecret:        []byte("Images.GenevaImagePullSecret"),
```
mjudeikis: Is your linting working fine? These changes do not look right to me.

pkg/api/types_test.go
```diff
@@ -373,6 +374,7 @@ func TestAdminAPIParity(t *testing.T) {
 	// TODO: why don't we just include all of these in the admin type?
 	notInAdmin := []*regexp.Regexp{
 		regexp.MustCompile(`^\.Config\.RunningUnderTest$`),
+		regexp.MustCompile(`^\.Config\.RootDockerConfig$`), // TODO: I don't feel this should be in admin API, should it?
```
mjudeikis: @jim-minter @kargakis Do you think we should be able to rotate this secret using AdminAPI? My feeling is that we should be able as this is secret, which has a hard external dependency, and I would like to be able to update it as fast as I can without rolling out new config or code. 

Anything external "touching" should be easy to manage. WDYT?

pkg/arm/data/master-startup.sh
```diff
@@ -25,6 +25,14 @@ EOF
   systemctl start docker.service
 fi
 
+# write the /root/.docker/config.json file to disk, but actually need it in
+# /var/lib/origin thanks to https://github.com/kubernetes/kubernetes/issues/45487
+DPATH=/root/.docker
```
mjudeikis: Does this works if we directly write file only to  `/var/lib/origin/.docker` ? I think we might want to keep it in a single location only?

pkg/arm/resources.go
```diff
@@ -413,9 +413,9 @@ func nsgMaster(cs *api.OpenShiftManagedCluster) *network.SecurityGroup {
 						DestinationPortRange:     to.StringPtr("443-443"),
 						SourceAddressPrefix:      to.StringPtr("*"),
 						DestinationAddressPrefix: to.StringPtr("*"),
-						Access:                   network.SecurityRuleAccessAllow,
```
mjudeikis: Again, linting?

pkg/fakerp/config.go
```diff
@@ -75,11 +75,16 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
...
+	rootDockerConfig, err := readFile(filepath.Join(artifactDir, "root-docker-config.json"))
+	if err != nil {
+		return nil, err
+	}
 	template.Certificates.GenevaLogging.Cert = logCert
 	template.Certificates.GenevaLogging.Key = logKey
 	template.Certificates.GenevaMetrics.Cert = metCert
 	template.Certificates.GenevaMetrics.Key = metKey
 	template.Images.GenevaImagePullSecret = pullSecret
+	template.RootDockerConfig = rootDockerConfig
```
mjudeikis: I think it is worth moving this under `template.Images.ImagePullSecret` and accordingly `Config.Images.ImagePullSecret`

pkg/util/mocks/mock_arm/arm.go
```diff
@@ -38,7 +38,6 @@ func (m *MockGenerator) EXPECT() *MockGeneratorMockRecorder {
 
 // Generate mocks base method
 func (m *MockGenerator) Generate(ctx context.Context, cs *api.OpenShiftManagedCluster, backupBlob string, isUpdate bool, suffix string) (map[string]interface{}, error) {
-	m.ctrl.T.Helper()
```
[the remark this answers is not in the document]
thekad: ok that's taken care of now

pkg/arm/data/master-startup.sh
thekad: I'm only creating on both because that way you can interact with docker inside the node without extra config. Also I *think* if we just write it to /.docker works for all cases but it feels dirty to me.

thekad:
note: e2e test will fail until I add the new secret to prow

pkg/api/converterfromplugin_test.go
thekad: golint was being difficult, fixed now

pkg/arm/resources.go
thekad: fixed

thekad:
Opened openshift/release#2571 and #1035 as follow-up

pkg/arm/data/master-startup.sh
```diff
@@ -25,6 +25,14 @@ EOF
...
+mkdir -p $DPATH
+base64 -d <<< {{ .Config.RootDockerConfig | Base64Encode }} >${DPATH}/config.json
```
kargakis: Is this something that we are going to be requiring everyone to use (even in development)? If not, you should set it optionally here.
```sh
{{- if ne .Config.RootDockerConfig "" }}
# write the /root/.docker/config.json file to disk, but actually need it in
# /var/lib/origin thanks to https://github.com/kubernetes/kubernetes/issues/45487
DPATH=/root/.docker
mkdir -p $DPATH
base64 -d <<< {{ .Config.RootDockerConfig | Base64Encode }} >${DPATH}/config.json
chmod 0600 ${DPATH}/config.json
ln -s ${DPATH} /var/lib/origin/.docker
{{ end }}
```
Otherwise, you should require it during validation?

thekad: The plan was to share the file same as with the rest of the `secrets/` directory

pkg/api/config.go
mjudeikis: I still think its part of `Images` config struct, and name Should change as its not dockerConfig, its Image pull secret. Docker config is `/etc/docker/daemon.json` and all subfolder for it.

pkg/arm/data/master-startup.sh
kargakis: Can you make sure the release repo gets updated with the new secret requirement? You will need to update the Makefile, projects/azure/README.md, and https://github.com/openshift/release/blob/master/ci-operator/SECRETS.md#cluster-secrets-azure

pkg/api/types_test.go
mjudeikis: I think this is fine with a restart. We will need to have this as Geneva action either way

pkg/arm/data/master-startup.sh
mjudeikis: lets limit to the places we use for :)

pkg/arm/data/master-startup.sh
thekad: openshift/release#2584

thekad:
@mjudeikis moved the new key into Images as per your feedback, ptal

pkg/arm/data/master-startup.sh
thekad: moved into `/var/lib/origin` only

pkg/api/admin/api/config.go
```diff
@@ -3,7 +3,7 @@ package api
 import (
 	"crypto/x509"
 
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
```
mjudeikis: Any particular reason for this change here?

pkg/api/config.go
```diff
@@ -4,8 +4,8 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 
-	"github.com/satori/go.uuid"
-	"k8s.io/client-go/tools/clientcmd/api/v1"
+	uuid "github.com/satori/go.uuid"
+	v1 "k8s.io/client-go/tools/clientcmd/api/v1"
```
mjudeikis: same as above?

pluginconfig/pluginconfig-311.yaml
```diff
@@ -58,3 +57,5 @@ images:
   webConsole: registry.access.redhat.com/openshift3/ose-web-console:v3.11.43
 ## Gevena image pull secret is used to access private components images. Format: base64 encoded .docker/config.json file "{auths:{osarpint.azurecr.io:{username:username,password:notthepasswordyouarelookingfor,email:openshift@example.com,auth:dXNlcm5hbWU6bm90dGhlcGFzc3dvcmR5b3VhcmVsb29raW5nZm9yCg==}}}"
   #genevaImagePullSecret: <base64 string of .docker/config.json file>
+## Redhat IO image pull secret is used to pull images from private registry.redhat.io
```
mjudeikis: Redhat image pull secret is used to pull images from....

pkg/fakerp/config.go
```diff
@@ -75,7 +75,7 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	rootDockerConfig, err := readFile(filepath.Join(artifactDir, "root-docker-config.json"))
+	redhatIOImagePullSecret, err := readFile(filepath.Join(artifactDir, "redhat-io-docker-config.json"))
```
mjudeikis: If we renaming this here, lets rename it in https://github.com/openshift/release/pull/2584 too

pkg/api/admin/api/config.go
thekad: gofmt changed it, wth

pkg/fakerp/config.go
thekad: yep, renamed

mjudeikis:
/test e2e

@thekad did the secret was updated in Ci cluster too?

pkg/api/types_test.go
jim-minter: yes, we need a geneva action for this.

jim-minter:
> 1. Update node build process to start using default registry.io to align us to upstream. (release repo change)
> 2. Update our code base to use the default registry when we publish the new image and it is available. (openshift-azure change)

This should go on the Jira backlog, but note: we're not going to do this in a hurry.  The risk of everything breaking due to a credential screw up are too high.

jim-minter:
@thekad please open an additional Jira story to cover implementing the geneva action and mark that as ga-proposed.

mjudeikis:
E2E still fails because of lack of the secret in CI.

thekad:
@mjudeikis nope, that's not it, it seems like the authorization we had no longer works (I did a `docker pull` manually on my laptop and I get a 401) do you know if the authorization expired or something @jim-minter ?

kargakis:
```
time="2019-01-17T14:46:21Z" level=debug msg="400 Bad Request: Failed to apply request: file /go/src/github.com/openshift/openshift-azure/secrets/system-docker-config.json does not exist"
```
You need to add the new secret in cluster-secret-azure in the azure namespace.

kargakis:
Updating that secret sucks at the moment since you need to have all secrets present in the release repo and we should split them up to separate secrets eventually.

mjudeikis:
/test e2e
@kargakis it is there already:
```
 oc get secret cluster-secrets-azure -o yaml | grep system
 system-docker-config.json: xxxxx
```

Secret still had creation timestamp of 50d (edit is not reflected)
```
 1028  oc get secret cluster-secrets-azure --export -o yaml > /tmp/secret
 1029  oc delete secret cluster-secrets-azure
 1030  oc create -f /tmp/secret
 1031  oc get secret
```

I suspect it is not get mirrored. Testing now to see if this helps

mjudeikis:
@kargakis all good, found an issue. @thekad is fixing.

jim-minter:
@thekad please rebase asap

PR #1038, opened 2019-01-14
Remove reliance on lsof

### LINKED ISSUE #1036: remove reliance on lsof
The following bash is horrid in multiple ways:
`trap 'return_id=$?; set +ex; kill $(lsof -t -i :8080); wait $(lsof -t -i :8080); exit $return_id' EXIT`

Replace all instances of it in our codebase as follows.  At the top of every bash script which should remove background spawned processes when it exits, add:
```bash
cleanup() {
    trap '' TERM
    kill 0
    wait
}

trap cleanup EXIT
```

The effect of this should be:
* in any case where the script is about to quit (normal exit, signal) call `cleanup()`
* in cleanup, set our script's SIGTERM handler to SIG_IGN
* send a SIGTERM to all processes in our process group (we'll ignore ours due to the previous line)
* wait for known child processes to exit

Also, this upgrades an incoming SIGINT to an outgoing SIGTERM, this handles the fact that by default bash presets background processes to ignore SIGINTs.

@kargakis @charlesakalugwu are you OK with the above?

charlesakalugwu:
closes #1036 

/cc @kargakis @jim-minter

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -1,5 +1,12 @@
 #!/bin/bash -ex
 
+cleanup(){
+	trap '' TERM
+	kill 0
+	wait
+}
+trap cleanup EXIT
+
 if [[ $# -ne 1 ]]; then
     echo usage: $0 resourcegroup
     exit 1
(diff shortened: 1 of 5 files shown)


charlesakalugwu:
#814

kargakis:
The test passed but the container failed
```
Terminated\nkill:
          usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill
          -l [sigspec]\n"
```
Seems related to your changed?

charlesakalugwu:
> The test passed but the container failed
> 
> ```
> Terminated\nkill:
>           usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill
>           -l [sigspec]\n"
> ```
> Seems related to your changed?

We might need to update `cluster-launch-e2e-azure`. I am able to consistently and correctly run the e2e tests locally without any issues. This makes me suspect we need to do a bit more in the aforementioned ci-operator template.

jim-minter:
Closing for #1047, sorry @charlesakalugwu I gave you bad advice :(

PR #1042, opened 2019-01-15
removing --ignored flag, extending .PHONY list

Makdaam:
dirty flag not affected by .gitignored files, metrics bridge remade at every run
Fixes 1024


Makdaam:
/retest
flake connection timeout

PR #1043, opened 2019-01-15
set hardcoded tuned profile during provisioning

### LINKED ISSUE #925: default tuned profile on ocp masters not set to be openshift-control-plane after cluster installation
After OCP cluster setup on azure , tuned profile for masters is ``` virtual-guest ```  and first after reboot it will be switched to ```openshift-control-plane``` Necessary to ensure proper profile is active immediately after cluster setup.

Makdaam:
Fixes #925.
For master nodes it's easy, because the etc/origin/node/node-config.yaml is populated before the master-startup.sh is finished.

For infra and compute nodes the recommended profile changes after the node is correctly connected to the cluster (/etc/tuned/recommend.conf matches contents of node-config.yaml). 

Since there are no inotify-tools to trigger on node-config.yaml update (EPEL required for that) I added a postponed action to the startup script.


Makdaam:
/hold
/wip
After discussion on slack, rewriting this as a controller DaemonSet.

jim-minter:
This is risking going overengineered. Can't you just write the desired profile to /etc/tuned/profile_mode and restart the tuned service?

pkg/arm/data/node-startup.sh
```diff
@@ -58,3 +58,7 @@ systemctl start ${SERVICE_TYPE}-node.service &
 # disabling rsyslog since we manage everything through journald
 systemctl disable rsyslog.service
 systemctl stop rsyslog.service
+
+#load the recommended tuned profile after node is connected to cluster
+#currently it takes <10min from beginning of arm template deployment until the end of the cluster creation process
+echo 'systemctl restart tuned' | at now + 10min
```
jim-minter: This really is not up to the standard that our codebase needs to be at.  Don't do this.

kargakis:
> This is risking going overengineered. Can't you just write the desired profile to /etc/tuned/profile_mode and restart the tuned service?

Two downsides with this approach is that tuned is not going to be restarted if options in the node config that affect its config have changed during runtime of the cluster (since node config is delivered via a config map + sync daemonset). Also, changes to the default tuned config means we have to release a new VM image.

pkg/arm/data/node-startup.sh
Makdaam: understood.

Makdaam:
Added tuned-adm command instead of restarting tuned. The profile names have been hardcoded for master, infra (openshift-control-plane) and compute node (openshift-node).

No new moving parts introduced.

pkg/arm/data/master-startup.sh
```diff
@@ -689,6 +689,9 @@ restorecon -Rv /var/lib/etcd
 logger -t master-startup.sh "restore done"
 {{- end }}
 
+#set the recommended openshift-control-plane profile on first boot
+tuned-adm profile openshift-control-plane
```
mjudeikis: is tuned-adm available on the nodes by default?

Makdaam: Yes, it's provided as part of the tuned package. And tuned is installed in our images.

Makdaam:
Rebased and regenerated bindata.go to resolve merge conflict.

PR #1047, opened 2019-01-15
Remove reliance on lsof

### LINKED ISSUE #1036: remove reliance on lsof

jim-minter:
closes #1036

### DIFF
--- a/hack/create.sh
+++ b/hack/create.sh
@@ -1,5 +1,12 @@
 #!/bin/bash -ex
 
+cleanup() {
+    kill $(jobs -p) &>/dev/null || true
+    wait
+}
+
+trap cleanup EXIT
+
 if [[ $# -ne 1 ]]; then
     echo usage: $0 resourcegroup
     exit 1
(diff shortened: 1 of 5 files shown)


hack/create.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash -ex
 
+cleanup() {
+    kill $(jobs -p) &>/dev/null || true
+    wait
+}
```
kargakis: is this trap not affecting the return code of the script?

kargakis: seems like it's not - nvmd

kargakis:
@jim-minter mind opening a PR in release to remove it from [test-base](https://github.com/openshift/release/blob/712932cadf3418e1d0da0de858f59399dfadf02a/projects/azure/base-images/test-base.yaml#L53) once this PR merges?

Issue #945, opened 2018-12-20
PVC not attached to applications pods 


ekuric:
I want to create 50 projects with 50 mongodb pods ( each project one mongodb pod ) and each pod uses dynamically allocated PVC from ```azure``` storageclass. The issue I am seeing is that PVC fails to attach to MongoDB pod. error message is [here](https://gist.github.com/ekuric/80e6ada67477247a6af4fe1ec2934e3f) 
Interesting part is 
```
Events:
  Type     Reason                  Age               From                     Message
  ----     ------                  ----              ----                     -------
  Warning  FailedScheduling        9m (x2 over 9m)   default-scheduler        pod has unbound PersistentVolumeClaims (repeated 100 times)
  Normal   Scheduled               9m                default-scheduler        Successfully assigned storage-test-mongo-2/mongodb-1-t2cjh to compute-000025
  Normal   SuccessfulAttachVolume  9m                attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-e45950ef-0431-11e9-ae29-000d3a2bb4d5"
  Warning  FailedMount             9m (x7 over 9m)   kubelet, compute-000025  MountVolume.WaitForAttach failed for volume "pvc-e45950ef-0431-11e9-ae29-000d3a2bb4d5" : Cannot find Lun for disk kubernetes-dynamic-pvc-e45950ef-0431-11e9-ae29-000d3a2bb4d5
  Warning  FailedMount             8m                kubelet, compute-000025  MountVolume.WaitForAttach failed for volume "pvc-e45950ef-0431-11e9-ae29-000d3a2bb4d5" : azureDisk - WaitForAttach failed within timeout node (compute-000025) diskId:(kubernetes-dynamic-pvc-e45950ef-0431-11e9-ae29-000d3a2bb4d5) lun:(0)
Warning FailedMount 50s (x4 over 7m) kubelet, compute-000025 Unable to mount volumes for pod "mongodb-1-t2cjh_storage-test-mongo-2(e7b53fea-0431-11e9-ae29-000d3a2bb4d5)": timeout expired waiting for volumes to attach or mount for pod "storage-test-mongo-2"/"mongodb-1-t2cjh". list of unmounted volumes=[mongodb-data]. list of unattached volumes=[mongodb-data default-token-mt84l]
```

This is on same cluster for which I reported #944

kargakis:
Do you see this in all of the pods? How many nodes do you have?

kargakis:
[...]

PR #1048, opened 2019-01-16
add MaxDataDisksPerVM to prevent trying to mount too many disks on a single node

jim-minter:
cf #945


pkg/config/derive.go
```diff
@@ -139,3 +139,87 @@ func (derived) StatsdArgs(cs *api.OpenShiftManagedCluster) ([]interface{}, error
...
+	for i := range cs.Properties.AgentPoolProfiles {
+		if cs.Properties.AgentPoolProfiles[i].Role != api.AgentPoolProfileRoleCompute {
+			continue
+		}
+
+		if app != nil {
+			return "", fmt.Errorf("found multiple compute agentPoolProfiles")
+		}
+
+		app = &cs.Properties.AgentPoolProfiles[i]
```
asalkeld: you could break here

asalkeld:
/retest

a connection timeout failure

pkg/config/derive.go
jim-minter: no, because the if test above would never succeed in that case

mjudeikis:
lgtm

jim-minter:
marking lgtm based on above

PR #1051, opened 2019-01-17
add clusterVersion field to tracking

mjudeikis:
This adds an internal version field to tracking plugin code versions we will use for the release process. 
Was considering adding it into the main `ManagedOpenshiftCluster` object, but because its total internal type did it in the `Config` struct instead. 

https://docs.google.com/document/d/1einHXiv6r6tI5i1ce_PkZybmCYOWAqtYz5bgzWhKr5Q/edit#heading=h.6jiuaxjhr1fc

/cc @jim-minter

### DIFF
--- a/pkg/api/admin/api/config.go
+++ b/pkg/api/admin/api/config.go
@@ -7,6 +7,8 @@ import (
 )
 
 type Config struct {
+	// ClusterVersion (out) defines release version of the plugin used to build the cluster
+	ClusterVersion *string `json:"clusterVersion,omitempty"`
 	// configuration of VMs in ARM template
 	ImageOffer     *string `json:"imageOffer,omitempty"`
 	ImagePublisher *string `json:"imagePublisher,omitempty"`
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -83,6 +83,7 @@ var marshalled = []byte(`{
 		"Tags.key": "Tags.val"
 	},
 	"config": {
+		"clusterVersion": "Config.ClusterVersion",
 		"imageOffer": "Config.ImageOffer",
 		"imagePublisher": "Config.ImagePublisher",
 		"imageSku": "Config.ImageSKU",
@@ -373,6 +374,7 @@ func TestAdminAPIParity(t *testing.T) {
 
 	// TODO: why don't we just include all of these in the admin type?
 	notInAdmin := []*regexp.Regexp{
+		regexp.MustCompile(`^\.Config\.ClusterVersion$`),
 		regexp.MustCompile(`^\.Config\.RunningUnderTest$`),
 		regexp.MustCompile(`^\.Config\.Images\.ImagePullSecret$`),
 		regexp.MustCompile(`^\.Properties\.AzProfile\.`),
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -374,7 +374,6 @@ func TestAdminAPIParity(t *testing.T) {
 
 	// TODO: why don't we just include all of these in the admin type?
 	notInAdmin := []*regexp.Regexp{
-		regexp.MustCompile(`^\.Config\.ClusterVersion$`),
 		regexp.MustCompile(`^\.Config\.RunningUnderTest$`),
 		regexp.MustCompile(`^\.Config\.Images\.ImagePullSecret$`),
 		regexp.MustCompile(`^\.Properties\.AzProfile\.`),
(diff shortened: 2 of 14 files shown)


pkg/api/types_test.go
```diff
@@ -372,6 +373,7 @@ func TestAdminAPIParity(t *testing.T) {
 
 	// TODO: why don't we just include all of these in the admin type?
 	notInAdmin := []*regexp.Regexp{
+		regexp.MustCompile(`^\.Config\.ClusterVersion$`),
```
jim-minter: this should be in the admin api, right? (r/o)

pkg/api/validate_plugin.go
```diff
@@ -8,11 +8,16 @@ import (
 )
 
 var (
-	imageVersion = regexp.MustCompile(`^[0-9]{3}.[0-9]{1,4}.[0-9]{8}$`)
+	imageVersion   = regexp.MustCompile(`^[0-9]{3}.[0-9]{1,4}.[0-9]{8}$`)
+	clusterVersion = regexp.MustCompile(`^v\d+.\d+$`)
```
jim-minter: `\.`, not `.`

pkg/api/types_test.go
mjudeikis: I think no. I would not want to be able to change it as part of AdminAPI. AdminAPI would initiate the upgrade process, and as part of the new code in RP it will generate new version field. This way version always corresponds only to code we vendor/release. 

Do you see a need to get this changed?

mjudeikis: Ou, read only...

mjudeikis:
/hold
considering the sensitivity of the code (a field in, but out only in AdminAPI, and not a customer) Need to add E2E test to validate his.

kargakis:
Is this field something that is going to be used by sync to determine whether it runs in the right cluster version?

mjudeikis:
Yes. It will be a combination of all components. If you release either code, images, or config - you bump this version. https://docs.google.com/document/d/1einHXiv6r6tI5i1ce_PkZybmCYOWAqtYz5bgzWhKr5Q

mjudeikis:
@openshift/sig-azure stinking PR. reviews would be great.

pkg/api/validate/types.go
```diff
@@ -28,6 +28,9 @@ var (
 	// This regexp is to check image version format
 	imageVersion = regexp.MustCompile(`^[0-9]{3}.[0-9]{1,4}.[0-9]{8}$`)
 
+	// This regexp is to check cluster version (plugin) format
+	clusterVersion = regexp.MustCompile(`^v\d+\.\d+$`)
```
kargakis: We should start unit-testing these regexps. Can you add one here?

test/e2e/specs/fakerp/admin_api.go
```diff
@@ -0,0 +1,45 @@
...
+		By("Reading the cluster state")
+		before, err := cli.OpenShiftManagedClustersAdmin.Get(context.Background(), os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(before).NotTo(BeNil())
+		Expect(before.Config.ClusterVersion).To(BeEquivalentTo(to.StringPtr("v0.0")))
+
+		By("Updating the cluster state")
+		before.Config.ClusterVersion = to.StringPtr("v0.1")
+		update, err := cli.OpenShiftManagedClustersAdmin.CreateOrUpdate(context.Background(), os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), before)
+		Expect(err).NotTo(HaveOccurred())
```
kargakis: Shouldn't this be an invalid update as opposed to silently swallowing it?

pkg/api/admin/api/config.go
```diff
@@ -7,6 +7,8 @@ import (
 )
 
 type Config struct {
+	// ClusterVersion (out) defines release version of the plugin used to build the cluster
```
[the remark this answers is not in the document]
mjudeikis: we want to be able to read it via admin API, but not modify. So we can "call the mothership" and ask what is running in the wild.

test/e2e/specs/fakerp/admin_api.go
mjudeikis: It is a valid update, which does not do anything. The field is changed, but update flow should ignore it and do nothing. Any other easy way to check if exposed API ignores the field?

pkg/api/validate/types.go
mjudeikis: added

test/e2e/specs/fakerp/admin_api.go
```diff
@@ -0,0 +1,45 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/go-autorest/autorest/to"
+
+	"github.com/openshift/openshift-azure/test/clients/azure"
+)
+
+var _ = Describe("Validate AdimAPI field readability [AdminAPI][Fake]", func() {
```
asalkeld: nit: spello "Admin" - only if you need to make a change

Issue #725, opened 2018-11-13
Start using azure mocked client


kargakis:
See https://github.com/Azure/azure-sdk-for-go/issues/409#issuecomment-438046820

Then we can stop generating our own fake client.

/area tests
/kind cleanup

kargakis:
> It will ship in v23 of the SDK at the end of this month. Take a look at [#3282](https://github.com/Azure/azure-sdk-for-go/pull/3282) to see the changes, should merge into latest later today.

kargakis:
our sdk got bumped to v24 with https://github.com/openshift/openshift-azure/pull/1064

mjudeikis:
renaming to fit work left doing

jim-minter:
what mocked client? :-(
looks like they just provide fat interfaces for you to implement...

PR #1059, opened 2019-01-27
upgrade the azure sdk.

andyliuliming:


sozercan:
looks like this is for aligning with aks-engine v0.29.1 release that'll vendored into the RP
https://github.com/Azure/aks-engine/blob/master/Gopkg.lock#L30

kargakis:
We should be able to get rid of our mocked clients after this PR gets merged

https://github.com/openshift/openshift-azure/issues/725

jim-minter:
@andyliuliming many thanks for this PR!
I'm closing in favour of #1064, which builds on your work: adds a missing import to get this through test + passing the new argument to Reimage through our client layer into the main codebase.

PR #1062, opened 2019-01-28
Add unit test for Logger http middleware

charlesakalugwu:
This adds a unit test for the Logger http middleware/handler.

/cc @kwoodson @thekad


pkg/fakerp/middleware_test.go
```diff
@@ -0,0 +1,42 @@
...
+	req.Header.Add("User-Agent", useragent)
+	w := httptest.NewRecorder()
+
+	// the bytes buffer will be used to capture all logs written during the test
+	var buf bytes.Buffer
+	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	logrus.SetOutput(&buf)
+
+	s := NewServer(logrus.NewEntry(logrus.StandardLogger()), useragent, "localhost:3456")
```
jim-minter: I don't know why this isn't working, but it's certainly preferable that unit tests don't rely on networking.  Here is a replacement test (note use of "github.com/sirupsen/logrus/hooks/test"), but honestly, it's not really worth it - this PR could just be closed IMO.

```go
package fakerp

import (
	"fmt"
	"net/http"
	"testing"

	"github.com/sirupsen/logrus"
	"github.com/sirupsen/logrus/hooks/test"
)

func TestLogger(t *testing.T) {
	logger, hook := test.NewNullLogger()
	logger.SetLevel(logrus.DebugLevel)

	s := Server{
		log: logrus.NewEntry(logger),
	}
	handler := s.logger(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))

	req, err := http.NewRequest("GET", "/", nil)
	if err != nil {
		t.Fatal(err)
	}
	req.Header.Add("User-Agent", "testlogger")

	handler.ServeHTTP(nil, req)

	expectedLogs := fmt.Sprintf("%s %s %s (%s)", req.RemoteAddr, req.Method, req.URL, req.UserAgent())

	if hook.LastEntry() == nil {
		t.Fatalf("no entry logged")
	}

	if hook.LastEntry().Message != expectedLogs {
		t.Fatalf("got entry %q, wanted %q", hook.LastEntry().Message, expectedLogs)
	}
}
```

charlesakalugwu: @jim-minter I would like to understand the problem before we close this :) So you've used the test logger. Was this really the issue with my attempt?

PR #1064, opened 2019-01-28
update azure-sdk-for-go to v24.0.0

jim-minter:
* PR is intended to align with aks-engine v0.29.1 for vendoring
* also updating etcd dependency as we were stale (using 3.10's dependency, not 3.11's (@mjudeikis fyi))
* may be able to get rid of some mocked clients after merge: #725


jim-minter:
manually tagging to merge if this passes tests

PR #1075, opened 2019-01-30
Config blob as file

asalkeld:
AZURE-171
I'll follow up with a PR to extract the certs/keys from the config file instead of being inline.


pkg/arm/resources.go
```diff
@@ -425,6 +427,16 @@ func nsgWorker(cs *api.OpenShiftManagedCluster) *network.SecurityGroup {
 	}
 }
 
+func gzipBytes(data []byte) ([]byte, error) {
+	var bufOut bytes.Buffer
+	gw, err := gzip.NewWriterLevel(&bufOut, gzip.BestCompression)
```
jim-minter: check the err returned from NewWriterLevel, and from gw.Write.

pkg/arm/resources.go
```diff
@@ -425,6 +427,16 @@ func nsgWorker(cs *api.OpenShiftManagedCluster) *network.SecurityGroup {
...
+	gw.Write(data)
+	if err := gw.Close(); err != nil {
+		return nil, err
+	}
+	return bufOut.Bytes(), err
```
jim-minter: nit: `return bufOut.Bytes(), nil`

jim-minter:
looks great. nits don't affect the functionality, just gonna lgtm this

/lgtm

PR #1078, opened 2019-01-30
DNS refactor in preparation for vanity domains.

kwoodson:
In preparation for the upcoming work for vanity domains, we would like to align our DNS setup to mirror what the real RP does.  This includes creating a DNS zone with a random name and placing it inside of the resource group beside the cluster.

### DIFF
--- a/pkg/fakerp/dns.go
+++ b/pkg/fakerp/dns.go
@@ -3,6 +3,7 @@ package fakerp
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
@@ -11,7 +12,9 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
 )
 
-func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, dnsResourceGroup, dnsDomain string, noTags bool) error {
+// CreateOCPDNS creates the dns zone for the cluster, updates the main zone in dnsResourceGroup and returns
+// the generated publicSubdomain, routerPrefix
+func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, dnsResourceGroup, dnsDomain, zoneName, routerCName string, noTags bool) error {
 	authorizer, err := azureclient.NewAuthorizerFromContext(ctx)
 	if err != nil {
 		return err
@@ -33,8 +36,9 @@ func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, 
 		z.Tags["ttl"] = &ttl
 		z.Tags["now"] = &now
 	}
-	zoneName := fmt.Sprintf("%s.%s", resourceGroup, dnsDomain)
-	zone, err := zc.CreateOrUpdate(ctx, dnsResourceGroup, zoneName, z, "", "")
+
+	// This creates creates the new zone
+	zone, err := zc.CreateOrUpdate(ctx, resourceGroup, zoneName, z, "", "")
 	if err != nil {
 		return err
 	}
@@ -67,7 +71,7 @@ func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, 
 		},
 	}
 	defaultRecordName := "@"
-	_, err = rsc.CreateOrUpdate(ctx, dnsResourceGroup, zoneName, defaultRecordName, dns.SOA, soa, "", "")
+	_, err = rsc.CreateOrUpdate(ctx, resourceGroup, zoneName, defaultRecordName, dns.SOA, soa, "", "")
 	if err != nil {
 		return err
 	}
@@ -80,13 +84,11 @@ func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, 
 			NsRecords: &nsServerList,
 		},
 	}
-	_, err = rsc.CreateOrUpdate(ctx, dnsResourceGroup, zoneName, defaultRecordName, dns.NS, ns, "", "")
+	_, err = rsc.CreateOrUpdate(ctx, resourceGroup, zoneName, defaultRecordName, dns.NS, ns, "", "")
 	if err != nil {
 		return err
 	}
 
-	// create router wildcard DNS CName record
-	routerCName := fmt.Sprintf("%s-router.%s.%s", resourceGroup, location, "cloudapp.azure.com")
 	cn := dns.RecordSet{
 		Etag: zone.Etag,
 		RecordSetProperties: &dns.RecordSetProperties{
@@ -96,7 +98,7 @@ func CreateOCPDNS(ctx context.Context, subscriptionID, resourceGroup, location, 
 			TTL: to.Int64Ptr(3600),
 		},
 	}
-	_, err = rsc.CreateOrUpdate(ctx, dnsResourceGroup, zoneName, "*", dns.CNAME, cn, "", "")
+	_, err = rsc.CreateOrUpdate(ctx, resourceGroup, zoneName, "*", dns.CNAME, cn, "", "")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/fakerp/dns.go
```diff
@@ -123,15 +125,19 @@ func DeleteOCPDNS(ctx context.Context, subscriptionID, resourceGroup, dnsResourc
 	zc.Authorizer = authorizer
 	rsc := dns.NewRecordSetsClient(subscriptionID)
 	rsc.Authorizer = authorizer
+	count := new(int32)
```
kwoodson: @mjudeikis @jim-minter I wasn't sure how to create a pointer for this.  A few other iterations resulted in nil pointer exceptions. I picked this up from SO.

Any suggestions?

pkg/fakerp/dns.go
```diff
@@ -123,15 +125,19 @@ func DeleteOCPDNS(ctx context.Context, subscriptionID, resourceGroup, dnsResourc
...
+	*count = 100
+	zoneResults, err := zc.ListByResourceGroup(ctx, resourceGroup, count)
```
kwoodson: Before this PR, when the zoneName were well known, removing zones was easier.  Now with a randomized string we need to list the dns zones from the resource group and then remove them from the main zone record.

pkg/fakerp/fakerp.go
```diff
@@ -77,11 +80,39 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
...
+
+			rrCName, err := utilrs.RandomString("abcdefghijklmnopqrstuvwxyz0123456789", 20)
+			if err != nil {
+				return nil, err
+			}
+			// create router wildcard DNS CName record
+			routerCName = fmt.Sprintf("osa%s.%s.%s", rrCName, cs.Location, "cloudapp.azure.com")
+
+			log.Info("routeprofiles[0].publicsubdomain was empty, creating dns")
+			err = CreateOCPDNS(ctx, os.Getenv("AZURE_SUBSCRIPTION_ID"), os.Getenv("RESOURCEGROUP"), cs.Location, os.Getenv("DNS_RESOURCEGROUP"), os.Getenv("DNS_DOMAIN"), publicSubdomain, routerCName, os.Getenv("NOGROUPTAGS") == "true")
```
kwoodson: Currently our tests fail if we move the DNS creation until after the cluster is created.  The initial health check that determines a cluster is healthy polls indefinitely.  I would like to move this to after cluster creation ~line 180.  I think this will be possible after the certificate generation matches the names for the cluster.
```
INFO[2019-01-30T11:32:04-05:00]/home/kwoodson/go/src/github.com/openshift/openshift-azure/pkg/util/wait/wait.go:68 github.com/openshift/openshift-azure/pkg/util/wait.forHTTPStatusOk.func1() Get https://kwoodsontest1.eastus.cloudapp.azure.com/healthz: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "openshift-signer") 
```

Or should it work now and I've made a poor assumption?

kwoodson:
This looks related:
```
• Failure [81.924 seconds]
Openshift on Azure end user e2e tests [EndUser][Fake]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:31
  should deploy a template and ensure a given text is in the contents [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:80

  Expected error:
      <*url.Error | 0xc0004cfb90>: {
          Op: "Get",
          URL: "http://nginx-example-e2e-test-irdtn.omjooyjy41kapqeaxgrh.osadev.cloud",
          Err: {
              Op: "dial",
              Net: "tcp",
              Source: nil,
              Addr: nil,
              Err: {
                  Err: "no such host",
                  Name: "nginx-example-e2e-test-irdtn.omjooyjy41kapqeaxgrh.osadev.cloud",
                  Server: "10.142.15.245:53",
                  IsTimeout: false,
                  IsTemporary: false,
              },
          },
      }
      Get http://nginx-example-e2e-test-irdtn.omjooyjy41kapqeaxgrh.osadev.cloud: dial tcp: lookup nginx-example-e2e-test-irdtn.omjooyjy41kapqeaxgrh.osadev.cloud on 10.142.15.245:53: no such host
  not to have occurred
```

pkg/fakerp/dns.go
charlesakalugwu: you can use `to.Int32Ptr()` from the `github.com/Azure/go-autorest/autorest/to` package

kwoodson: Let me try this.

kwoodson:
Fixed delete and fixed int pointer.

pkg/fakerp/fakerp.go
```diff
@@ -53,6 +54,8 @@ func GetDeployer(log *logrus.Entry, cs *api.OpenShiftManagedCluster, config *api
 }
 
 func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenShiftManagedCluster, config *api.PluginConfig, isAdmin bool) (*api.OpenShiftManagedCluster, error) {
+	publicSubdomain := ""
```
mjudeikis: Why do you initialize them outside? can we just do:
`publicSubdomain := fmt.Sprintf("%s.%s", rDomain, os.Getenv("DNS_DOMAIN"))` ?

pkg/fakerp/fakerp.go
mjudeikis: Router DNS is created during cluster bootstrap. Same in real and fakeRP.
Why do we want to move DNS creation after a cluster is created? I might miss something. RealRP creates it before, and I think we should too? :man_shrugging:

pkg/fakerp/fakerp.go
mjudeikis: Or initialize them before use itself.

mjudeikis:
overall looks great. left few comments :)

pkg/fakerp/fakerp.go
kwoodson: @mjudeikis, @jim-minter and I had a discussion and I think we wanted to mimic the real RP and not have our vanity DNS setup before the cluster creation.  If we setup DNS post cluster create then we would get a realistic test that we are using the correct internal URLs for health checks.  This uncovered the admin console issue where we are not able to connect to it with our current URL until DNS is correctly set.

pkg/fakerp/fakerp.go
kwoodson: I was using this to move the DNS setup to the end of the file after cluster creation.  This was the if statement logic.  `if publicSubdomain != "" { // do dns create`.  Since moving the DNS to the end failed with the error from above I moved it back.  I think I can take your suggestion and let the follow up PR move the DNS.

kwoodson:
```
time="2019-02-01T17:27:55Z" level=debug msg="400 Bad Request: Failed to apply request: WaitForNodes: Get https://e2e-eeqroebs.eastus.cloudapp.azure.com/api/v1/nodes/master-000000: read tcp 172.16.29.221:34166->104.45.188.173:443: read: connection timed out" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).badRequest" file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:10"
time="2019-02-01T17:27:55Z" level=fatal msg="containerservice.OpenShiftManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: error response cannot be parsed: \"400 Bad Request: Failed to apply request: WaitForNodes: Get https://e2e-eeqroebs.eastus.cloudapp.azure.com/api/v1/nodes/master-000000: read tcp 172.16.29.221:34166->104.45.188.173:443: read: connection timed out\\n\" error: json: cannot unmarshal number into Go value of type azure.RequestError"
```

/test e2e

kwoodson:
@jim-minter @mjudeikis @asalkeld I think this is ready to go.  Let me know if you have any other feedback.

asalkeld:
it looks like the e2e tests did not run, so don't want to merge until they do

kwoodson:
They have run previously but we can kick them off again.

/test e2e

kwoodson:
Retriggering tests.

kwoodson:
@asalkeld, tests passed.

PR #1079, opened 2019-01-31
Add AdminUpdating provisioning state

### LINKED ISSUE #1058: Seperate Provisioning State for Admin Operations
We will need a seperate provisioning state for admin put operations such as `AdminUpdating` here https://github.com/openshift/openshift-azure/blob/master/pkg/api/types.go#L67.

This will allow us as engineers to look up clusters logs and distinguish between a regular update operation and an admin update operation. Also it will be used to block a regular put operation when an admin update operation is ongoing.

charlesakalugwu:
This adds the `AdminUpdating` provisioning state to the internal api types. This will "allow us as engineers to look up clusters logs and distinguish between a regular update operation and an admin update operation. Also it will be used to block a regular put operation when an admin update operation is ongoing."

closes #1058 

/cc @mjudeikis @Makdaam @jim-minter

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -71,6 +71,8 @@ const (
 	Creating ProvisioningState = "Creating"
 	// Updating means the existing OSA resource is being updated.
 	Updating ProvisioningState = "Updating"
+	// AdminUpdating means the existing OSA resource is being updated with admin privileges.
+	AdminUpdating ProvisioningState = "AdminUpdating"
 	// Failed means the OSA resource is in failed state.
 	Failed ProvisioningState = "Failed"
 	// Succeeded means the last create/update succeeded.
--- a/pkg/api/2018-09-30-preview/api/types.go
+++ b/pkg/api/2018-09-30-preview/api/types.go
@@ -74,6 +74,8 @@ const (
 	Creating ProvisioningState = "Creating"
 	// Updating means the existing OSA resource is being updated.
 	Updating ProvisioningState = "Updating"
+	// AdminUpdating means the existing OSA resource is being updated with admin privileges.
+	AdminUpdating ProvisioningState = "AdminUpdating"
 	// Failed means the OSA resource is in failed state.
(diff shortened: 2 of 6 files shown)


pkg/api/types.go
```diff
@@ -71,6 +71,8 @@ const (
 	Creating ProvisioningState = "Creating"
 	// Updating means the existing OSA resource is being updated.
 	Updating ProvisioningState = "Updating"
+	// AdminUpdating means the existing OSA resource is being updated with admin privileges.
+	AdminUpdating ProvisioningState = "AdminUpdating"
```
mjudeikis: What is the difference here now between Updating and AdminUpdating? 
What about this: https://github.com/openshift/openshift-azure/blob/master/pkg/api/admin/api/types.go#L89 

In addition we have validation tests: 
https://github.com/openshift/openshift-azure/blob/e3a61726057308e5bd471fc9d4918eccd819647c/pkg/api/validate/validators.go#L312 
We have validation for this, which need to be updated.

Overall where do these states is being used apart fakeRP? How it being consumed?

jim-minter:
this is a good start but you also need to:
* add the "enum" value in pkg/api/2018-09-30-preview/api/types.go
* add the "enum" value in pkg/api/admin/api/types.go
* fix up validateProvisioningState in pkg/api/validate/validators.go
* open a follow-up issue to implement a test that would have caught at least some of these problems

pkg/api/types.go
charlesakalugwu: `AdminUpdating` will be set when an update (PUT) to the OSA resource is initiated via the admin API. `Updating` will be set when the update is unprivileged.

asalkeld:
lgtm, wait for others to review too..

charlesakalugwu:
https://github.com/openshift/openshift-azure/issues/1097

mjudeikis:
rebase :/

mjudeikis:
/test e2e
prow! I believe in u!

PR #1082, opened 2019-02-01
Delete stray objects in sync pod

y-cote:
Delete objects marked owned by the sync pod that do not show in the blueprint (config blob).

### DIFF
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -23,6 +23,8 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
 )
 
+const ownedBySyncPodAnnotationKey = "azure.openshift.io/owned-by-sync-pod"
+
 type extra struct {
 	RegistryStorageAccountKey string
 	ConfigStorageAccountKey   string
@@ -295,5 +297,10 @@ func Main(ctx context.Context, log *logrus.Entry, cs *api.OpenShiftManagedCluste
 		return nil
 	}
 
-	return writeDB(log, client, db)
+	err = writeDB(log, client, db)
+	if err != nil {
+		return err
+	}
+
+	return client.DeleteOrphans(db)
 }
--- a/pkg/addons/addons.go
+++ b/pkg/addons/addons.go
@@ -23,7 +23,7 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
 )
 
-const ownedBySyncPodAnnotationKey = "azure.openshift.io/owned-by-sync-pod"
+const ownedBySyncPodLabelKey = "azure.openshift.io/owned-by-sync-pod"
 
 type extra struct {
 	RegistryStorageAccountKey string
--- a/pkg/addons/client.go
+++ b/pkg/addons/client.go
@@ -182,8 +182,8 @@ func (c *client) DeleteOrphans(db map[string]unstructured.Unstructured) error {
 
 			for _, i := range l.Items {
 				// check that the object is marked by the sync pod
-				a := i.GetAnnotations()
-				if av, ok := a[ownedBySyncPodAnnotationKey]; ok && av == "yes" {
+				l := i.GetLabels()
+				if l[ownedBySyncPodLabelKey] == "true" {
 					// if object is marked, but not in current DB, remove it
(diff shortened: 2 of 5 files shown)


pkg/addons/addons.go
```diff
@@ -23,6 +23,8 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
 )
 
+const syncpodNote = "belongs-to-syncpod"
```
jim-minter: `const ownedBySyncPodAnnotationKey = "azure.openshift.io/owned-by-sync-pod"`

pkg/addons/client.go
```diff
@@ -34,6 +35,7 @@ type Interface interface {
 	UpdateDynamicClient() error
 	ServiceCatalogExists() (bool, error)
 	GetStorageAccountKey(ctx context.Context, resourceGroup, storageAccount string) (string, error)
+	DeleteOrphan(db map[string]unstructured.Unstructured) error
```
jim-minter: DeleteOrphans

pkg/addons/client.go
```diff
@@ -141,6 +143,69 @@ func (c *client) ApplyResources(filter func(unstructured.Unstructured) bool, db
...
+			if !contains(resource.Verbs, "list") {
+				continue
+			}
+
+			dc, err := c.dyn.ClientForGroupVersionKind(gv.WithKind(resource.Kind))
+			if err != nil {
+				return err
+			}
+
+			o, err := dc.Resource(&resource, "").List(metav1.ListOptions{})
```
jim-minter: add a log line to indicate the gvk that we're listing

pkg/addons/client.go
```diff
@@ -141,6 +143,69 @@ func (c *client) ApplyResources(filter func(unstructured.Unstructured) bool, db
...
+				continue
+			}
+
+			for _, i := range l.Items {
+				// check that the object is marked by the sync pod
+				a := i.GetAnnotations()
+				if av, ok := a[syncpodNote]; ok && av == "yes" {
+					// if object is marked, but not in current DB, remove it
+					if _, ok := db[KeyFunc(i.GroupVersionKind().GroupKind(), i.GetNamespace(), i.GetName())]; !ok {
+						c.log.Infof("   D %s\n", KeyFunc(i.GroupVersionKind().GroupKind(), i.GetNamespace(), i.GetName()))
```
jim-minter: `c.log.Infof("Delete %s\n")`

pkg/addons/client.go
```diff
@@ -190,6 +255,7 @@ func write(log *logrus.Entry, dyn dynamic.ClientPool, grs []*discovery.APIGroupR
 				// return a synthetic Conflict to force a retry.
 				err = kerrors.NewConflict(schema.GroupResource{Group: res.Group, Resource: res.Name}, o.GetName(), errors.New("synthetic"))
 			}
+			markSyncPodOwned(o)
```
jim-minter: move just above Create() call

jim-minter:
this is lgtm, just awaiting squash than can do hold cancel

mjudeikis:
I think I found there is the problem :) If your run sync twice, and the second run is on "converged cluster" you get a lot of deleted until the cluster is bricked:
https://gist.github.com/mjudeikis/8eb93ff28589faf00aa083771aa8c620

mjudeikis:
To sum up, my and @y-cote debugging yesterday:
Some deletes happens due to "child" items, who get annotations propagated. In example Deployment creates ReplicaSets who controls the pods behaviour. this case ReplicaSet has annotations but does not have record in the database. For this use-case we need to check `ownersReference` field within metadata and check if "owner" has an annotation and record in the database.

other usecase (below) is not clear. These object has record and annotation. But still being deleted.
```
time="2019-02-07T15:37:59Z" level=info msg="Delete SecurityContextConstraints/hostmount-anyuid"
time="2019-02-07T15:37:59Z" level=info msg="Delete SecurityContextConstraints/hostnetwork"
time="2019-02-07T15:37:59Z" level=info msg="Delete SecurityContextConstraints/privileged"
```

jim-minter:
@y-cote I've pushed 3 commits onto your branch - please review them and let's see if that takes us forward.  Happy to catch up on the changes made via bluejeans/slack.

PR #1085, opened 2019-02-02
[wip] Geneva action run command

charlesakalugwu:
This work implements a geneva action that allows an admin to run a generic command on a given virtual machine within a given scale set in a cluster. 

This is part of AZURE-257

/cc @mjudeikis @Makdaam @thekad @kwoodson

### DIFF
--- a/pkg/cluster/runcommand.go
+++ b/pkg/cluster/runcommand.go
@@ -1,0 +1,42 @@
+package cluster
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+func (u *simpleUpgrader) RunCommand(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string, parameters compute.RunCommandInput) (compute.RunCommandResult, *api.PluginError) {
+	result, err := u.vmc.RunCommand(ctx, oc.Properties.AzProfile.ResourceGroup, scaleSetName, instanceId, parameters)
+	if err != nil {
+		return result, &api.PluginError{Err: err, Step: "RunGenericCommand"}
+	}
+	return result, nil
+}
+
+func (u *simpleUpgrader) RestartDocker(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) (compute.RunCommandResult, *api.PluginError) {
+	params := compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
+		Script:    to.StringSlicePtr([]string{"systemctl restart docker.service"}),
+	}
+	return u.RunCommand(ctx, oc, scaleSetName, instanceId, params)
+}
+
+func (u *simpleUpgrader) RestartKubelet(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) (compute.RunCommandResult, *api.PluginError) {
+	params := compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
+		Script:    to.StringSlicePtr([]string{"systemctl restart atomic-openshift-node.service"}),
+	}
+	return u.RunCommand(ctx, oc, scaleSetName, instanceId, params)
+}
+
+func (u *simpleUpgrader) RestartNetworkManager(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) (compute.RunCommandResult, *api.PluginError) {
+	params := compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
+		Script:    to.StringSlicePtr([]string{"systemctl restart NetworkManager.service"}),
+	}
+	return u.RunCommand(ctx, oc, scaleSetName, instanceId, params)
+}
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -4,6 +4,8 @@ package api
 import (
 	"context"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+
 	plugin "github.com/openshift/openshift-azure/pkg/api/plugin/api"
 )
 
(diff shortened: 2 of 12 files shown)


pkg/arm/resources.go
```diff
@@ -502,8 +502,8 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
 		},
 		VirtualMachineScaleSetProperties: &compute.VirtualMachineScaleSetProperties{
 			UpgradePolicy: &compute.UpgradePolicy{
-				AutoOSUpgradePolicy: &compute.AutoOSUpgradePolicy{
-					DisableAutoRollback: to.BoolPtr(false),
+				AutomaticOSUpgradePolicy: &compute.AutomaticOSUpgradePolicy{
```
kwoodson: What does this do?

kwoodson: Is this due to the latest compute API?

pkg/fakerp/runcommand.go
```diff
@@ -0,0 +1,49 @@
...
+	w.Write(result)
+	s.log.Infof("ran command on %s in %s", vm, ss)
+}
+
+func readCommandInput(req *http.Request) (*compute.RunCommandInput, error) {
+	data, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read request body: %v", err)
+	}
+	var params *compute.RunCommandInput
```
kwoodson: How do you envision the RunCommandInput stuct working?  Do we have to create scripts to lay down or is the scripts section a bash string?

pkg/arm/resources.go
charlesakalugwu: yes this is due to the update of the compute api

pkg/cluster/runcommand.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+func (u *simpleUpgrader) RunCommand(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string, parameters compute.RunCommandInput) *api.PluginError {
+	return nil
+}
+
+func (u *simpleUpgrader) RestartDocker(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) *api.PluginError {
+	params := compute.RunCommandInput{}
```
kwoodson: Is this where you will define  the actual run command?  Did I miss where those commands were defined?

kwoodson:
I'd like to chat and have you show me more of this in a chat.  Just had a few additional questions:
- How do we add new commands?  Do we have to do a release?
- What does the structure of the script/runcommand look like?  Does it take complex syntax?  Do we pass bash or a path to a script?
- Should we create generic restart handlers in bash and call our bash scripts giving us built-in retries?

Thanks @charlesakalugwu

pkg/fakerp/runcommand.go
charlesakalugwu: It will be something like this:

```
func (u *simpleUpgrader) RestartDocker(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) ([]byte, *api.PluginError) {
	params := compute.RunCommandInput{
		CommandID: to.StringPtr("RunShellScript"),
		Script: to.StringSlicePtr([]string{"systemctl restart docker.service"}),
	}
	return u.RunCommand(ctx, oc, scaleSetName, instanceId, params)
}
```

pkg/cluster/runcommand.go
charlesakalugwu: yes this is where the actual command will be specified. as this is wip at this point, I didn't add those yet but you can see a proposed example usage in the previous comment thread.

charlesakalugwu:
> I'd like to chat and have you show me more of this in a discussion. Just had a few additional questions:
> 
> * How do we add new commands?  Do we have to do a release?
> * What does the structure of the script/runcommand look like?  Does it take complex syntax?  Do we pass bash or a path to a script?
> * Should we create generic restart handlers in bash and call our bash scripts giving us built-in retries?
> 
> Thanks @charlesakalugwu

@kwoodson sure thing! We can go through the PR first thing tomorrow during our call. Looking forward to it.

pkg/cluster/runcommand.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func (u *simpleUpgrader) RunCommand(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string, parameters compute.RunCommandInput) (compute.RunCommandResult, *api.PluginError) {
+	return compute.RunCommandResult{}, nil
+}
+
+func (u *simpleUpgrader) RestartDocker(ctx context.Context, oc *api.OpenShiftManagedCluster, scaleSetName, instanceId string) (compute.RunCommandResult, *api.PluginError) {
+	params := compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
+		Script:    to.StringSlicePtr([]string{"systemctl restart docker.service"}),
```
mjudeikis: just nit: all these are systemd services. we could have a wrapper for systemd managed service. We could trigger a restart, status, reload, etc.

pkg/fakerp/runcommand.go
mjudeikis: Dont think we will have this luxury to lay down the script. I think this needs to be able to accept pasted in the script of some sort. Laying down means reimage. and if stuff is broken this mostly will not be an option.

kwoodson:
Remove WIP when ready.

charlesakalugwu:
> Remove WIP when ready.

@kwoodson done :)

thekad:
I got a few overall design comments so I didn't submit a proper review:

* I feel there's lots of repetition here, I would suggest making something like `RunGenericCommand()` more robust and write thin wrappers around it, for example `RunSystemctlCommand()` which can be called from `RestartDocker()` and `RestartKubelet()`
* Also, I *feel* like the `RunGenericCommand()` interface shouldn't be exposed through the API, and only individual/more granular commands to limit the possible attack surface area
* Needs test cases, I think

We can chat directly as well, I may have misunderstood the design direction here.

charlesakalugwu:
> I got a few overall design comments so I didn't submit a proper review:
> 
> * I feel there's lots of repetition here, I would suggest making something like `RunGenericCommand()` more robust and write thin wrappers around it, for example `RunSystemctlCommand()` which can be called from `RestartDocker()` and `RestartKubelet()`
> * Also, I _feel_ like the `RunGenericCommand()` interface shouldn't be exposed through the API, and only individual/more granular commands to limit the possible attack surface area
> * Needs test cases, I think
> 
> We can chat directly as well, I may have misunderstood the design direction here.

All of these are concerns I also had/have and thats why this needs to be discussed and signed-off jointly by both teams. I've got a scoping meeting later today and together with your (and our team's) feedback we should arrive at a design that makes the most sense. 

Tests are coming of course.

mjudeikis:
Now it requires rebase after my PR got in

jim-minter:
follow-ups:
* function to list all hostnames of a cluster
* remove generic run command for now (open follow-up jira to track)
* condense other run commands with an enum to indicate docker/kubelet/networkmanager etc
* check if a *logrus.Entry will be passed in
* change scaleset name / instance id api to be hostname based

jim-minter:
@charlesakalugwu bump on the follow-ups?

jim-minter:
closing for #1236

PR #1089, opened 2019-02-04
Create storageAccounts before Generate and pass the key into sync

asalkeld:
AZURE-279
The point of this is not use azure.conf as MSFT is planning on dropping the privileges (so we will not be able to list keys), so we get the key at create time and store it in a file that the pod can load at run time.


pkg/cluster/initialize.go
```diff
@@ -38,10 +39,51 @@ func (u *simpleUpgrader) Initialize(ctx context.Context, cs *api.OpenShiftManage
...
+
+	for found := false; !found; _, found = u.storageAccountKey[accountName] {
+		keys, err := u.accountsClient.ListKeys(ctx, cs.Properties.AzProfile.ResourceGroup, accountName)
+		if err != nil {
+			return "", err
+		}
+		if len(*keys.Keys) == 0 {
+			// The WaitForCompletionRef above should be enough, but I have seen
+			// this return an empty list the first time.
+			time.Sleep(1)
```
mjudeikis: wow :/

mjudeikis:
gave one read, need more one. TBC

pkg/cluster/initialize.go
asalkeld: yeah, I know. :-(

asalkeld:
so one thing left to do here is to alter the e2e test that checks for unnecessary vm rotations.
masters will now get rotated as the config is stored on disk. we could potentially strip out things that
we don't need.

jim-minter:
/hold
want to talk this one over with @amanohar in light of the need for 2 SPs anyway

jim-minter:
closing this as the wind changed direction.

PR #1090, opened 2019-02-04
ILB creation issue

mjudeikis:
fixes: https://github.com/openshift/openshift-azure/issues/1087

When creating internal LB this is the query to Azure: 
```
[resourceId('Microsoft.Network/virtualNetworks/subnets', parameters('vnetName'), parameters('subnetname'))]",
```

Our cloud provider does not have `vnetName` and ` subnetName` set so we ended with "" values.
And so the error:
```
 Subnet "" not found with message: "network.SubnetsClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code=\"ResourceNotFound\" Message=\"The Resource 'Microsoft.Network/virtualNetworks/subnets' under resource group '5786a402-a88d-4cac-b8f9-ae2a4226a531' was not found.\""
```

This should solve this. Will raise follow-up issue to tide-up constants for vnet names and subnet names as now we hardcode them all around the place. 

/cc @jim-minter @charlesakalugwu @asalkeld 
 @amanohar Let me know how you want to approach this fix. I can produce new release with this cherry.

### DIFF
--- a/pkg/config/derive.go
+++ b/pkg/config/derive.go
@@ -107,6 +107,8 @@ func (derived) CloudProviderConf(cs *api.OpenShiftManagedCluster) ([]byte, error
 		Location:          cs.Location,
 		SecurityGroupName: "nsg-worker",
 		VMType:            "vmss",
+		SubnetName:        "default",
+		VnetName:          "vnet",
 	})
 }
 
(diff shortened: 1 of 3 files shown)


mjudeikis:
Test passed, CI is bottle-necked. Merging manually.

PR #1092, opened 2019-02-04
Deep e2e test integration

### LINKED ISSUE #811: Investigate how to inject sanity checks in our update-based e2e tests
We've started writing update-based e2e tests which involve updating part or all of the cluster.

It would be very useful if we found a way to inject a set of sanity checks at various points within update-based test specs. It should particularly be possible to resuse those sanity checks across different test specs

For example
* Scale up a clusters compute pool
* Run the set of sanity checks
* Scale down the cluster
* Run the set of sanity checks

or 

* Rotate all credentials of a cluster
* Run the set of sanity checks

/cc @jim-minter @kargakis

charlesakalugwu:
This work enables MSFT to call functions to "sanity check" a cluster as part of their e2e tests.

*Notable Changes*

* A new deep test interface in `test/clients/interface.go`
* A new deep test client in `test/clients/openshift/client.go`
* The deep test client wraps all the previous openshift clients we used for e2e tests
* A new deep test client can be created with `NewClient(log *logrus.Entry, cs *internalapi.OpenShiftManagedCluster)`
  * If developing against the fake rp (RH), sensible defaults will be set for log and cs when you call `NewClient(nil, nil)`
  * If developing against the real rp (MSFT), you can provide your log and cs and call `NewClient(log, cs)`
* Conforming to the deep test interface, the deep client provides the following methods
  * `CreateTestApp(ctx context.Context) (interface{}, []*TestError)` - creates a test app and returns the project where it was created
  * `ValidateTestApp(ctx context.Context, cookie interface{}) []*TestError` - validates functionality of the deployed test app
  * `DeleteTestApp(ctx context.Context, cookie interface{}) []*TestError` - deletes the test app and its project
  * `ValidateCluster(ctx context.Context) []*TestError` - validates cluster functionality
* Update-based e2e tests can optionally call a combination of `Create/ValidateTestApp`, perform updates and then call `ValidateCluster`
* It is still possible to granularly reference the internal user-specific clients e.g. `Client.EndUser` or `Client.AzureClusterReader`

This is part of AZURE-222

closes #811 

/cc @mjudeikis @jim-minter @kwoodson @thekad @Makdaam

### DIFF
--- a/test/clients/openshift/client.go
+++ b/test/clients/openshift/client.go
@@ -1,9 +1,6 @@
 package openshift
 
 import (
-	"os"
-	"path/filepath"
-
 	oappsv1client "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	projectv1client "github.com/openshift/client-go/project/clientset/versioned/typed/project/v1"
 	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
@@ -20,8 +17,8 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/tools/clientcmd/api/latest"
 
-	fakerp "github.com/openshift/openshift-azure/pkg/fakerp/shared"
-	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
+	internalapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
 )
 
 type Client struct {
@@ -67,29 +64,9 @@ func newClientFromKubeConfig(kc *api.Config) (*Client, error) {
 	return newClientFromRestConfig(restconfig), nil
 }
 
-func clientFromEnv() (*Client, error) {
-	restconfig, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
-	if err != nil {
-		return nil, err
-	}
-	return newClientFromRestConfig(restconfig), nil
-}
-
-func NewAzureClusterReaderClient() (*Client, error) {
-	if os.Getenv("KUBECONFIG") != "" {
-		return clientFromEnv()
-	}
-	dataDir, err := fakerp.FindDirectory(fakerp.DataDirectory)
-	if err != nil {
-		return nil, err
-	}
-	cs, err := managedcluster.ReadConfig(filepath.Join(dataDir, "containerservice.yaml"))
-	if err != nil {
-		return nil, err
-	}
-
+func NewAzureClusterReaderClient(cs *internalapi.OpenShiftManagedCluster) (*Client, error) {
 	var kc api.Config
-	err = latest.Scheme.Convert(cs.Config.AzureClusterReaderKubeconfig, &kc, nil)
+	err := latest.Scheme.Convert(cs.Config.AzureClusterReaderKubeconfig, &kc, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -97,11 +74,8 @@ func NewAzureClusterReaderClient() (*Client, error) {
 	return newClientFromKubeConfig(&kc)
 }
 
-func NewCustomerReaderClient() (*Client, error) {
-	if os.Getenv("KUBECONFIG") != "" {
-		return clientFromEnv()
-	}
-	kc, err := login("customer-cluster-reader")
+func NewCustomerReaderClient(cs *internalapi.OpenShiftManagedCluster) (*Client, error) {
+	kc, err := login("customer-cluster-reader", cs)
 	if err != nil {
 		return nil, err
 	}
@@ -109,11 +83,8 @@ func NewCustomerReaderClient() (*Client, error) {
 	return newClientFromKubeConfig(kc)
 }
 
-func NewAdminClient() (*Client, error) {
-	if os.Getenv("KUBECONFIG") != "" {
-		return clientFromEnv()
-	}
-	kc, err := login("admin")
+func NewAdminClient(cs *internalapi.OpenShiftManagedCluster) (*Client, error) {
+	kc, err := login("admin", cs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


kwoodson:
@charlesakalugwu, is this a proof of concept?  Are we going to migrate some of our other tests to this structure?  I'm just trying to understand how this will work.

charlesakalugwu:
> @charlesakalugwu, is this a proof of concept? Are we going to migrate some of our other tests to this structure? I'm just trying to understand how this will work.

@kwoodson this is not a proof of concept. This is how we'll co-jointly run e2e test cluster validation in a standardized way across both teams.

test/clients/openshift/client.go
```diff
@@ -67,80 +72,209 @@ func newClientFromKubeConfig(kc *api.Config) (*Client, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	return newClientFromKubeConfig(&kc)
 }
 
-func NewCustomerReaderClient() (*Client, error) {
+func newCustomerReaderClient(cs *internalapi.OpenShiftManagedCluster) (*client, error) {
 	if os.Getenv("KUBECONFIG") != "" {
```
charlesakalugwu: we probably don't need these anymore as you'll be able to pass the cluster struct to the client to get all internals setup correctly. Can you confirm @sozercan

test/clients/interface.go
```diff
@@ -0,0 +1,23 @@
+package clients
+
+import (
+	"context"
+)
+
+type TestError struct {
+	Bucket string
```
charlesakalugwu: @jim-minter can you give me more details on how you envisioned the usage of `TestError.Bucket`?

test/clients/openshift/client.go
```diff
@@ -20,11 +22,14 @@ import (
...
-	fakerp "github.com/openshift/openshift-azure/pkg/fakerp/shared"
+	internalapi "github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
 	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
+	"github.com/openshift/openshift-azure/test/clients"
+	testlogging "github.com/openshift/openshift-azure/test/util/log"
 )
 
-type Client struct {
+type client struct {
```
kwoodson: The reason for the lower case was to localize it here?  Encapsulation?

kwoodson: Having two structs called client, Client in this might be confusing.  Would it make sense to call this a kubeclient?

charlesakalugwu: Yes I lower-cased the previous client (change from Client to client) to prevent them from being exported from the package (and therefore being used by developers). In its stead I introduced a new `Client` which is the heart of this PR.

test/clients/openshift/client.go
```diff
@@ -67,80 +72,209 @@ func newClientFromKubeConfig(kc *api.Config) (*Client, error) {
...
+		cs:  cs,
+	}
+	var err error
+	if c.log == nil {
+		c.log = testlogging.GetTestLogger()
+	}
+	if c.cs == nil {
+		c.cs, err = c.discoverInternalConfig()
+		if err != nil {
+			panic(err)
```
kwoodson: Should these panic or return errors?

charlesakalugwu: this is fine. we want to fail the test fast if we are unable to set the client up properly

test/clients/openshift/client.go
charlesakalugwu: > Having two structs called client, Client in this might be confusing. Would it make sense to call this a kubeclient?

Makes sense. I actually initially used `client` and `DeepTestClient`. I'm open to the idea of changing the names.

kwoodson:
Just a few small questions but nothing blocking.  lgtm

test/clients/openshift/client.go
mjudeikis: I think this is fine. Because if to follow the naming convention and package naming we should avoid things like `openshift.KubeClient` as it is same as `openshift.OpenshiftClient` . Practice is that package name defines what it is, and names are generic. In this case `openshift.Client` make sense. Otherwise, we end up with duplications in the package name and method it exposes.

test/clients/openshift/client.go
```diff
@@ -67,80 +72,209 @@ func newClientFromKubeConfig(kc *api.Config) (*Client, error) {
...
+
+func (c *Client) discoverInternalConfig() (*internalapi.OpenShiftManagedCluster, error) {
+	dataDir, err := shared.FindDirectory(shared.DataDirectory)
+	if err != nil {
+		return nil, err
+	}
+	return managedcluster.ReadConfig(filepath.Join(dataDir, "containerservice.yaml"))
+}
+
+func (c *Client) CreateTestApp(ctx context.Context) (interface{}, []*clients.TestError) {
```
mjudeikis: Move to separate file?

charlesakalugwu:
https://github.com/openshift/openshift-azure/issues/1127

test/clients/openshift/client.go
```diff
@@ -67,80 +72,209 @@ func newClientFromKubeConfig(kc *api.Config) (*Client, error) {
...
+// sensible defaults for the logger and cluster struct will be set.
+// If the E2E tests are run against the real rp then the logger and/or the cluster config
+// should be set.
+func NewClient(log *logrus.Entry, cs *internalapi.OpenShiftManagedCluster) *Client {
+	c := &Client{
+		log: log,
+		cs:  cs,
+	}
+	var err error
+	if c.log == nil {
```
jim-minter: FWIW I feel that the implicit pattern of defaulting nil always goes sour in the long run, especially in test libraries.  I personally prefer explicit even if it's verbose.  Even `func NewDefaultClient() *Client { return NewClient(testlogging.GetTestLogger(), c.discoverInternalConfig() }` or whatever is better.

jim-minter:
I confirm I've given up on the idea of Ginkgo recursively calling Ginkgo.
I'm not sure this PR is necessarily the end of refactoring this stuff, but I believe it takes us several steps towards it. Thanks @charlesakalugwu.

test/clients/openshift/client.go
charlesakalugwu: going with `test/e2e/standard`

mjudeikis:
/lgtm
If stuff will be broken fix later.

jim-minter:
/hold
looking once more

jim-minter:
/hold cancel
lgtm
some follow ups in #1145

PR #1094, opened 2019-02-04
Fix var interpolation inside of awk statement

kwoodson:
Adding an extra `$` to make variables work in Makefile.


kwoodson:
I'm not seeing `dev` in the image version.

/hold

kwoodson:
@mjudeikis I think this does it

PR #1095, opened 2019-02-04
Moving docker images from centos to rhel base.

kwoodson:
Moving our base images from centos to rhel.  @pweil- sent me a link to origin and their image building here(https://github.com/openshift/origin/blob/master/images/base/Dockerfile.rhel7).  This uses a rhel7 base image which lives on registry.access.redhat.com.

### DIFF
--- a/Dockerfile.azure-controllers
+++ b/Dockerfile.azure-controllers
@@ -1,4 +1,3 @@
-FROM centos:7
-RUN yum -y update
+FROM rhel7
 COPY azure-controllers .
 ENTRYPOINT [ "/azure-controllers" ]
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,4 +1,3 @@
-FROM centos:7
-RUN yum -y update
+FROM rhel7
 COPY e2e .
 ENTRYPOINT [ "/e2e" ]
--- a/Dockerfile.etcdbackup
+++ b/Dockerfile.etcdbackup
@@ -1,4 +1,3 @@
-FROM centos:7
-RUN yum -y update
+FROM rhel7
 COPY etcdbackup .
 ENTRYPOINT [ "/etcdbackup" ]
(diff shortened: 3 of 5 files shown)


asalkeld:
any reason you don't run the "yum update" anymore?

charlesakalugwu:
> any reason you don't run the "yum update" anymore?

i'm also curious about this. It is usually a bad idea to run stuff like "yum update" when dealing with image builds..because we would have no idea what version of what software packages are baked into the image. Its usually preferable to install known versions of everything. However why have you taken out "yum update" @kwoodson ?

kwoodson:
@asalkeld @charlesakalugwu 

Great question and this question hits at the crux of the problem.  When using a rhel image and running `yum update` we would need to supply entitlements, subscriptions, yum repos, etc.  This means that locally we would need to have these credentials/repos/entitlements in order to build and test.  Then replicate those same secrets into the build system.  This ends up being a bit of management overhead.

```
imagebuilder -f Dockerfile.sync -t quay.io/openshift-on-azure-dev/sync:dev .
--> FROM rhel7 as 0
--> RUN yum -y update
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not receiving updates. You can use subscription-manager on the host to register and assign subscriptions.
There are no enabled repos.
 Run "yum repolist all" to see the repos you have.
 To enable Red Hat Subscription Management repositories:
     subscription-manager repos --enable <repo>
 To enable custom repositories:
     yum-config-manager --enable <repo>
running 'yum -y update' failed with exit code 1
make: *** [Makefile:85: sync-image] Error 1
```

The origin repository uses this rhel image for builds and this image is the base for our other components.  If OpenShift uses this image, I believe we should follow their lead.    

Now if the team feels it necessary to keep the `yum update`, then we need to talk about using rhel entitlements to do our updates.  This becomes a process of registering and deregistering our image during container builds as we cannot make these images public with entitlements.

Thus the reason for removing `yum update`.

charlesakalugwu:
/lgtm

I prefer the route of installing specific versions of software in an image over mass updating all packages.

asalkeld:
thanks for the explaination @kwoodson 

/lgtm

PR #1091, opened 2019-02-04
Geneva action reimage vm

charlesakalugwu:
This work implements a geneva action that allows an admin to reimage a given virtual machine, identified by its hostname, in a cluster.

This is part of AZURE-257

/cc @mjudeikis @Makdaam @y-cote @kwoodson

PR #1070, opened 2019-01-29
Avoid setting VMScaleSetVMReimageInput during master VM reimaging

charlesakalugwu:
closes #1068

/cc @jim-minter @kargakis @asalkeld @mjudeikis

PR #1096, opened 2019-02-05
Update azure compute api

charlesakalugwu:
An updated azure compute sdk is required for a few of the proposed geneva action implementations (https://github.com/openshift/openshift-azure/pull/1085 and  https://github.com/openshift/openshift-azure/pull/1091).

This PR updates the compute sdk to `2018-10-01` and also takes a stab at solving the `Reimage` parameter issue which was added [here](https://github.com/openshift/openshift-azure/pull/1064) and was attempted to be solved [here](https://github.com/openshift/openshift-azure/pull/1069). The eventual workaround can be found [here](https://github.com/openshift/openshift-azure/pull/1070)

After getting merged, related open geneva-related PRs (e.g. [this](https://github.com/openshift/openshift-azure/pull/1085)) would need to rebase to pull in the compute updates.

/cc @mjudeikis @Makdaam

### DIFF
--- a/pkg/cluster/update_master.go
+++ b/pkg/cluster/update_master.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"sort"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/cluster/kubeclient"
--- a/pkg/arm/resources.go
+++ b/pkg/arm/resources.go
@@ -7,7 +7,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-07-01/network"
 	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2015-06-15/storage"
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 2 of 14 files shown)


asalkeld:
@charlesakalugwu do you need to update this too: https://github.com/openshift/openshift-azure/blob/master/pkg/arm/resources.go#L57

charlesakalugwu:
> @charlesakalugwu do you need to update this too: https://github.com/openshift/openshift-azure/blob/master/pkg/arm/resources.go#L57

Good catch. Thanks :) Opened up a follow-up issue: https://github.com/openshift/openshift-azure/issues/1103

mjudeikis:
> An updated azure compute sdk is required for a few of the proposed geneva action implementations (#1085 and #1091).

Can you give more details what is needs from new SDK?

charlesakalugwu:
> > An updated azure compute sdk is required for a few of the proposed geneva action implementations (#1085 and #1091).
> 
> Can you give more details what is needs from new SDK?

Details are in the linked PRs ... `RunCommand` needs the sdk update for example

mjudeikis:
Do we need to make sure ARM is aligned too? 
https://github.com/openshift/openshift-azure/blob/master/pkg/arm/resources.go#L51

charlesakalugwu:
@mjudeikis it already is. See the first commit in this PR

charlesakalugwu:
@Makdaam there is an issue with this PR which is why i placed it on hold. After bringing a cluster up with this, master vms fail to join the control plane. Will need to investigate further

pkg/cluster/update_master.go
```diff
@@ -75,7 +75,7 @@ func (u *simpleUpgrader) UpdateMasterAgentPool(ctx context.Context, cs *api.Open
 		}
 
 		u.log.Infof("reimaging %s", computerName)
-		err = u.vmc.Reimage(ctx, cs.Properties.AzProfile.ResourceGroup, ssName, *vm.InstanceID, nil)
+		_, err = u.vmc.ReimageAll(ctx, cs.Properties.AzProfile.ResourceGroup, ssName, *vm.InstanceID)
```
jim-minter: @charlesakalugwu you need to wait on the future returned by ReimageAll.  Write a shim function like this one: https://github.com/openshift/openshift-azure/blob/master/pkg/util/azureclient/compute_addons.go#L120-L128

jim-minter: @charlesakalugwu does this wipe the etcd disk as well?  We definitely do not want to do that.

charlesakalugwu: @jim-minter i just discovered that ReimageAll wipes all managed disks of a VM. This is definitely not what we want for the masters.

PR #1099, opened 2019-02-05
Refactor fakerp package

charlesakalugwu:
- admin_handlers.go now contains admin-related handlers
- customer_handlers.go now contains unprivileged handlers (basic createorupdate, get etc)
- routes.go now contains the http routing spec and should be easier to deal with

/cc @mjudeikis @kwoodson @thekad @Makdaam


kwoodson:
I like this direction.  lgtm

PR #1105, opened 2019-02-06
Allow RP credentials to be used in plugin

y-cote:
The Authorizer comes in from the RP and is passed to the plugin for
subsequent requests. Before, the autorest.Authorizer object was created
using credentials harvested by the plugin.

In this PR, the fakerp package is also altered in order to pass along an
RP generated autorest.Authorizer mocked by enrichContext(). Otherwise,
the internal communication of the credentials is done via the
GetAuthorizerFromContext() routine.

Signed-off-by: Yannick Cote <ycote@redhat.com>

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -11,10 +11,11 @@ import (
 type ContextKey string
 
 const (
-	ContextKeyClientID     ContextKey = "ClientID"
-	ContextKeyClientSecret ContextKey = "ClientSecret"
-	ContextKeyTenantID     ContextKey = "TenantID"
-	ContextAcceptLanguages ContextKey = "AcceptLanguages"
+	ContextKeyCloudProviderClientID     ContextKey = "ClientID"
+	ContextKeyCloudProviderClientSecret ContextKey = "ClientSecret"
+	ContextKeyCloudProviderTenantID     ContextKey = "TenantID"
+	ContextKeyClientAuthorizer          ContextKey = "ClientAuthorizer"
+	ContextAcceptLanguages              ContextKey = "AcceptLanguages"
 )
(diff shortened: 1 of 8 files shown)


kwoodson:
I don't know much about this or see anything wrong. lgtm

pkg/api/plugin.go
```diff
@@ -11,10 +11,11 @@ import (
 type ContextKey string
 
 const (
-	ContextKeyClientID     ContextKey = "ClientID"
-	ContextKeyClientSecret ContextKey = "ClientSecret"
-	ContextKeyTenantID     ContextKey = "TenantID"
-	ContextAcceptLanguages ContextKey = "AcceptLanguages"
+	ContextKeyClientID         ContextKey = "ClientID"
```
amanohar: I would name these differently now e.g. `ContextKeyCloudProviderClientID` to distinguish its purpose from `ContextKeyClientAuthorizer`. 

In fact now that plugin is not going to use cloud provider creds for deploying resource is there a need to continue keeping these in the context instead of just in `OpenShiftManagedCluster.Properties.ServicePrincipalProfile`?

kwoodson:
@y-cote please rebase

pkg/api/plugin.go
y-cote: Oops, wrong button, meant to submit a comment.  Retyping : 

Possibly, only the line : ContextKeyClientAuthorizer ContextKey = "ClientAuthorizer" has been added by this commit. The other constants were just re-indented by "go fmt". Nevertheless, I will revisit this space and cleanup accordingly -- renaming and removing/moving keys that aren't used or belong only in the fakerp pkg.

PR #1107, opened 2019-02-06
remove dev namespace

mjudeikis:
remove dev namespace for now. too many moving parts. maybe later we bring it back

/cc @charlesakalugwu @Makdaam


mjudeikis:
> I don't get this...why were all the `openshift-on-azure` changed to `openshift-on-azure-dev` in the first place?

because I was thinking to do testing with staging namespace (as if my automated process) fracks up I can frack a lot of people clusters :) but at the moment it is just one more moving part and already too many of those :/

PR #1113, opened 2019-02-06
Send audit logs to Geneva

thekad:
1. Re-enable the audit logs in the master startup script
2. Change the fluentd configuration so it tags the audit logs
appropriately
3. Change the mdsd configuration so the logs are sent to different
geneva tables based on tagging


thekad:
Still need someone with geneva access to verify this for me. The tagging/etc is already tested (by sending to local FS), but need to check geneva tables have the right kind of logs each

pkg/addons/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -7,9 +7,22 @@ data:
...
       </storage>
-      tag journald
+      tag logs
     </source>
-    <match journald>
+    <match logs>
+      @type rewrite_tag_filter
+      <rule>
+        key MESSAGE
+        pattern audit\.k8s\.io
```
mjudeikis: are we currently logging all audit logs to journald? Or do we have a separate log file for it?

thekad: audit logs are in the same output as regular k8s logs, which are logged to journald, they just have this specific string as part of the message

mjudeikis: https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html#master-node-config-advanced-audit
This works, but there is option to send them to separate file, so we dont need to do any filtering in fluentd, so just read the file. Might be more operationally efficient to have filtering outside fluentd? 

Did you noticed any OOM for this work?

thekad: I did bring up the use of filtering vs split logging when @jim-minter and I discussed the approach, we decided to go with this for now.
I didn't notice any OOMs in my testing, but didn't run a logging stress test.

Issue #1118, opened 2019-02-07
unnecessary node rotation 


mjudeikis:
in E2E unnecessary node rotation happens!

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1117/pull-ci-openshift-openshift-azure-master-e2e-azure/2260 

/cc @charlesakalugwu @Makdaam @kwoodson 

/kind bug

mjudeikis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1105/pull-ci-openshift-openshift-azure-master-e2e-azure/2259 
one more

mjudeikis:
third: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1089/pull-ci-openshift-openshift-azure-master-e2e-azure/2262 in https://github.com/openshift/openshift-azure/pull/1089

mjudeikis:
forth: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1117/pull-ci-openshift-openshift-azure-master-e2e-azure/2264

mjudeikis:
fift: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1117/pull-ci-openshift-openshift-azure-master-e2e-azure/2264

PR #1119, opened 2019-02-07
rollback use of UpdateStatus

mjudeikis:
Remove/rollback `adminUpdate` state as this sets "provisioningState" inside containerservices.yaml and if test is running in the wrong order, this is causing next tests to rotate nodes.

Example:
A subset of tests ran before and now we have modified provisioning state (local and cluster) so it will trigger node rollout.
```
→ diff _data/containerservice.yaml _data/containerservice.yaml_bckp                                                                                                                                                
178d177                                                                                                                                                                                                            
<   provisioningState: AdminUpdating     
```

Need to check which test is causing these statues to leak and fix. We should be able to reset status, or have status "Running/Healthy".

Overall wrong setting and unsetting of status is causing flakes.

### DIFF
--- a/pkg/fakerp/admin_handlers.go
+++ b/pkg/fakerp/admin_handlers.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/storage"
 
-	internalapi "github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/cluster"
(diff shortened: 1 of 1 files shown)


mjudeikis:
fixes: https://github.com/openshift/openshift-azure/issues/1118

mjudeikis:
@charlesakalugwu will raise a followup issue to bring this back and fix status setting.

PR #1124, opened 2019-02-11
Adding a dedicated drive for docker data

thekad:
We're currently using the temporary instance drive for the docker data
on each of the nodes. This is a variable drive (the bigger the instance,
the bigger the drive) and since we want to support several instance
types ranging from fairly small to big-ish, it's hard to guarantee the
same space will be available. OCD is reserving 200G space for their
docker data, I am going to 256 to take advantage of dedicated IOPs in
azure disks.


thekad:
@jim-minter Jake got back to me, the plan is to support a wide range (4 VCPUs up to 16VCPUs) of instances, but all of them fall short of the 200G goal, lmk what do you think

pkg/arm/resources.go
```diff
@@ -591,15 +591,17 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
...
 	}
 
+	// The additional data disk have different purposes on master
+	// and infra/compute nodes, but they are both the same size
+	vmss.VirtualMachineProfile.StorageProfile.DataDisks = &[]compute.VirtualMachineScaleSetDataDisk{
+		{
+			Lun:          to.Int32Ptr(0),
+			Caching:      compute.CachingTypesReadOnly,
+			CreateOption: compute.DiskCreateOptionTypesEmpty,
+			DiskSizeGB:   to.Int32Ptr(256),
```
kwoodson: We might want to variablize the size `256` so that we can configure this more easily.  Or at least make a constant.

jim-minter: Please can you push this up to be part of the `vmss := &compute.VirtualMachineScaleSet{` statement above?

pkg/arm/resources.go
```diff
@@ -591,15 +591,17 @@ func Vmss(pc *api.PluginConfig, cs *api.OpenShiftManagedCluster, app *api.AgentP
 		Location: to.StringPtr(cs.Location),
 	}
 
+	// The additional data disk have different purposes on master
+	// and infra/compute nodes, but they are both the same size
```
jim-minter: good comment - please add further rationale: we use 256GiB for etcd because of the iops; for docker we use it to match OSD (size)

jim-minter:
/hold
want to run this by @amanohar and @jimzim

pkg/arm/data/master-startup.sh
```diff
@@ -32,6 +32,7 @@ mkdir -p $DPATH
 base64 -d <<< {{ .Config.Images.ImagePullSecret | Base64Encode }} >${DPATH}/config.json
 chmod 0600 ${DPATH}/config.json
 
+# The additional data disk holds the etcd data in master nodes
```
mjudeikis: What are we doing with docker on master nodes? it needs some space too. Not that much as nodes, but still

thekad: @mjudeikis master nodes keep the current configuration where the docker volumes are stored in the temporary drive, which is dependant on the size of the instance

thekad: for example, in our current QA configuration the master nodes have 16G assigned to `/var/lib/docker`

jim-minter:
we decided to abandon this, closing.

PR #1126, opened 2019-02-12
Security options for ssh access in NSG.

y-cote:
Lock down the NSGs to a particular management IP/port range.


y-cote:
Should we fork out some envvars for the configurables now or is the hardcoded values from before still enough?

pkg/api/plugin.go
```diff
@@ -77,7 +77,9 @@ type DeployFn func(context.Context, map[string]interface{}) error
 
 // PluginConfig is passed into NewPlugin
 type PluginConfig struct {
-	TestConfig TestConfig
+	SSHSrcAddrPrefix string
+	SSHDestPortRange string
```
jim-minter: don't believe we need to make SSHDestPortRange configurable

pkg/api/plugin.go
```diff
@@ -77,7 +77,9 @@ type DeployFn func(context.Context, map[string]interface{}) error
 
 // PluginConfig is passed into NewPlugin
 type PluginConfig struct {
-	TestConfig TestConfig
+	SSHSrcAddrPrefix string
```
jim-minter: unfortunately this is the wrong place for this - we've been working on getting everything out of PluginConfig.  It needs to go somewhere in pkg/api/admin/api/config.go.  That way it can be part of the on-disk plugin configuration in the RP.  You'll also need to update pluginconfig/pluginconfig-311.yaml - our version should be configured with '*' + a comment that MSFT would want to change that in prod

pkg/arm/resources.go
```diff
@@ -385,8 +385,8 @@ func nsgMaster(cs *api.OpenShiftManagedCluster) *network.SecurityGroup {
 						Description:              to.StringPtr("Allow SSH traffic"),
 						Protocol:                 network.SecurityRuleProtocolTCP,
 						SourcePortRange:          to.StringPtr("*"),
-						DestinationPortRange:     to.StringPtr("22-22"),
-						SourceAddressPrefix:      to.StringPtr("*"),
+						DestinationPortRange:     to.StringPtr(pc.SSHDestPortRange),
+						SourceAddressPrefix:      to.StringPtr(pc.SSHSrcAddrPrefix),
```
jim-minter: Bet there'll be multiple ranges in prod.  Can you try using SourceAddressPrefixes ([]string)?

pkg/api/plugin.go
jim-minter: a follow-up PR that removes PluginConfig entirely in favour of TestConfig would probably be a good plan and make things clearer for others if you'd like to do that?

y-cote: Comments help with temporary situations. In this case I will make sure this gets removed after that PR.

pkg/api/plugin.go
y-cote: Anubhuti mentions it in an email. I will double check with her just to make sure.

jim-minter:
@y-cote bump

y-cote:
There's still one little enigma I need to solve where the interaction between setting 'SourceAddressPrefix' and/or 'SourceAddressPrefixes' causes the ARM template verification to fail. If I don't set SourceAddressPrefix, the error is that it should be set, if I set both SourceAddressPrefix and SourceAddressPrefixes the error is that only one of them should be set. I'll take a look deeper in a few.

pkg/api/admin/api/config.go
```diff
@@ -15,6 +15,9 @@ type Config struct {
 	ImageSKU       *string `json:"imageSku,omitempty"`
 	ImageVersion   *string `json:"imageVersion,omitempty"`
 
+	// SSH to system nodes allowed IP ranges
+	SSHSrcAddrPrefixes *[]string `json:"sshSrcAddrPrefixes,omitempty"`
```
jim-minter: `SSHSourceAddressPrefixes` please

y-cote: done

jim-minter:
merged so soon @y-cote ?

y-cote:
Do you think I should revert it and try again? ;)

Issue #1127, opened 2019-02-12
Scale Up/Down rotates all VMs


charlesakalugwu:
The scale up/down e2e tests have been failing in CI since 1st February 2019: https://openshift-gce-devel.appspot.com/builds/origin-ci-test/logs/periodic-ci-azure-scaleupdown-e2e/

It appears that from this day, `CreateOrUpdate`  on the non-admin `OpenshiftManagedClustersClient`, with a manifest whose only update is the count of compute instances, sends a request which results in all VMs being updated/rotated instead of scaling up/down only the compute scale set.

This results in the test taking longer than it should and it effectively gets timed-out after the configured timeout of 30 minutes.

I am not yet sure if this is a problem on the MSFT RP side or on our side.

sample artifacts: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1092/pull-ci-openshift-openshift-azure-release-v1-e2e-azure-scaleupdown/2/

/cc @mjudeikis @kwoodson @julienstroheker @sozercan 
/kind bug

kwoodson:
After taking @jim-minter's advice and dumping the hash during cluster creation and then also dumping it during the `make e2e-scaleupdown` it appears that properties inside of 
`properties.virtualMachineProfile.extensionProfile.extensions[0].properties.protectedSettings.script` are changing.

Looking further into this.

kwoodson:
It appears that gzip includes the script name and the date which differs on each run and in turn causes the file to be different each time.
```
[root@master-000002 ~]# gzip  script.1.sh script.sh 
[root@master-000002 ~]# ll
total 248
-r-x------. 1 root root 123960 Feb 12 23:14 script.1.sh.gz
-r-x------. 1 root root 123958 Feb 12 23:10 script.sh.gz
[root@master-000002 ~]# md5sum script.1.sh.gz script.sh.gz 
d6dad4513977e6654c9e613b51a8d45c  script.1.sh.gz
209253fbdd03f83b6a68886a8f7fa48f  script.sh.gz
```

With the `--no-name` option removes name and time stamp:
```
[root@master-000002 ~]# gzip --no-name script.sh script.1.sh 
[root@master-000002 ~]# ll
total 248
-r-x------. 1 root root 123948 Feb 12 23:14 script.1.sh.gz
-r-x------. 1 root root 123948 Feb 12 23:10 script.sh.gz
[...]

PR #1129, opened 2019-02-12
Don't compress when calculating the hash

asalkeld:
relates to #1127


thekad:
this LGTM but will wait for @kwoodson to weigh in

kwoodson:
Running a test locally and will report back.

kwoodson:
@asalkeld I was able to reproduce the issue with this PR.  The hash when running `make e2e-scaleupdown` causes a full rotation of masters.

jim-minter:
isn't gzip deterministic?  not convinced we should go this route?

jim-minter:
/hold
for more discussion in #1127

jim-minter:
closing this, we're rolling back to the config blob

PR #1132, opened 2019-02-13
Revert config as file (cluster config back to a blob)

### LINKED ISSUE #1127: Scale Up/Down rotates all VMs

asalkeld:
this is an alternative to #1129

closes #1127


jim-minter:
@mjudeikis fyi

jim-minter:
@asalkeld still needs a rebase, ptal and check the e2e, hope to lgtm this tomorrow.

PR #1134, opened 2019-02-13
remove etcd probe and add tcp session config 

mjudeikis:
* removed etcd probe. If we remove master-000000, cluster in unhealthy, and k8s starts bouncing all other etcd pods due failed liveness probe.
* DisableKeepAlives disables HTTP keep-alives andwill only use the connection to the server for a single
HTTP request. This allows us to move from master to master without `EOF` errors

TODO:
we need to make sure we monitor etcd in other ways (ETCD metrics shim is a proposal)
potentially do same TCP config for ARM client too as we see errors there.


mjudeikis:
@jim-minter I think ready to merge

jim-minter:
/lgtm
great work on this @mjudeikis

PR #1135, opened 2019-02-13
remove context keys that are no longer used

jim-minter:
@y-cote follow up from #1105

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -11,11 +11,8 @@ import (
 type ContextKey string
 
 const (
-	ContextKeyCloudProviderClientID     ContextKey = "ClientID"
-	ContextKeyCloudProviderClientSecret ContextKey = "ClientSecret"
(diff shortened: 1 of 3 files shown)


kwoodson:
I like removing these old keys but tbh, I'm not sure I understand where they have moved to.

Is this due to us removing environment variables and taking authentication from the real RP to our plugin?

Other than that, lgtm.

y-cote:
> I like removing these old keys but tbh, I'm not sure I understand where they have moved to.
> 
> Is this due to us removing environment variables and taking authentication from the real RP to our plugin?
> 
> Other than that, lgtm.

Yeah, Real RP is including an Authorizer in the context and we don't need to carry the other keys to create one (Authorizer) out of them throughout the code like we use to. I didn't delete them as part as 1105, because I didn't see that we could call the ENV version of NewAuthorizer to do what I did in enrichContext().

PR #1137, opened 2019-02-13
Geneva action update cluster with latest plugin config

charlesakalugwu:
This work implements a geneva action that allows an admin to update a cluster using the latest plugin config known to the rp.

This is part of AZURE-257

/cc @mjudeikis @Makdaam @y-cote  @kwoodson

### DIFF
--- a/pkg/fakerp/fakerp.go
+++ b/pkg/fakerp/fakerp.go
@@ -14,11 +14,13 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/openshift/openshift-azure/pkg/api"
+	pluginapi "github.com/openshift/openshift-azure/pkg/api/plugin/api"
 	"github.com/openshift/openshift-azure/pkg/config"
 	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
 	"github.com/openshift/openshift-azure/pkg/plugin"
 	"github.com/openshift/openshift-azure/pkg/tls"
 	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	"github.com/openshift/openshift-azure/pkg/util/managedcluster"
 	utilrs "github.com/openshift/openshift-azure/pkg/util/randomstring"
 	"github.com/openshift/openshift-azure/pkg/util/resourceid"
 )
@@ -60,9 +62,17 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
 		return nil, kerrors.NewAggregate(errs)
 	}
 
-	template, err := GetPluginTemplate()
-	if err != nil {
-		return nil, err
+	var err error
+	var template *pluginapi.Config
+	if shared.IsUpdate() && shared.IsScaleOperation(cs, oldCs) {
+		log.Info("using previous plugin template (scale)")
+		template = managedcluster.PluginTemplate(cs)
+	} else {
+		log.Info("using latest plugin template (create/upgrade)")
+		template, err = GetPluginTemplate()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// This should be executed only for fakeRP
--- a/pkg/fakerp/shared/shared_test.go
+++ b/pkg/fakerp/shared/shared_test.go
@@ -1,0 +1,93 @@
+package shared
+
+import (
+	"testing"
+
+	"github.com/openshift/openshift-azure/pkg/api"
+)
+
+func TestIsScaleOperation(t *testing.T) {
+	tests := []struct {
+		name     string
+		new      *api.OpenShiftManagedCluster
+		old      *api.OpenShiftManagedCluster
+		expected bool
+	}{
+		{
+			name:     "old and new are the same",
+			new:      &api.OpenShiftManagedCluster{},
+			old:      &api.OpenShiftManagedCluster{},
+			expected: false,
+		},
+		{
+			name: "old and new differ only in the agent pool profile counts",
+			new: &api.OpenShiftManagedCluster{
+				Properties: api.Properties{
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{Role: api.AgentPoolProfileRoleMaster, Name: "master", Count: 3},
+						{Role: api.AgentPoolProfileRoleCompute, Name: "compute", Count: 10},
+						{Role: api.AgentPoolProfileRoleInfra, Name: "infra", Count: 2},
+					},
+				},
+			},
+			old: &api.OpenShiftManagedCluster{
+				Properties: api.Properties{
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{Role: api.AgentPoolProfileRoleMaster, Name: "master", Count: 3},
+						{Role: api.AgentPoolProfileRoleCompute, Name: "compute", Count: 2},
+						{Role: api.AgentPoolProfileRoleInfra, Name: "infra", Count: 3},
+					},
+				},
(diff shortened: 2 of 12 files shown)


pkg/plugin/plugin_test.go
```diff
@@ -179,3 +179,48 @@ func TestRotateClusterSecrets(t *testing.T) {
 		t.Errorf("plugin.RotateClusterSecrets error = %v", err)
 	}
 }
+
+func TestUpdatePluginConfig(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
```
kwoodson: I would like a walk through of how this works.

pkg/plugin/plugin.go
```diff
@@ -240,3 +240,17 @@ func (p *plugin) GetControlPlanePods(ctx context.Context, oc *api.OpenShiftManag
 	}
 	return json.Marshal(pods)
 }
+
+func (p *plugin) UpdateWithPluginConfig(ctx context.Context, cs *api.OpenShiftManagedCluster, deployFn api.DeployFn, pluginTemplate *pluginapi.Config) *api.PluginError {
```
jim-minter: do not pass in `pluginTemplate *pluginapi.Config` here, access it via `p.config`

pkg/fakerp/util.go
```diff
@@ -31,3 +34,16 @@ func (s *Server) isAdminRequest(req *http.Request) bool {
 	// TODO: Align with the production RP once it supports the admin API
 	return strings.HasPrefix(req.URL.Path, "/admin")
 }
+
+func readPluginConfigTemplate(req *http.Request) (*api.Config, error) {
```
jim-minter: should not be needed

jim-minter:
still need to add code which controls *when* a new RP plugin config gets applied to an existing cluster.  Example: when a user calls update to scale, an upgrade should not take place.  When an admin calls this API to update, the upgrade should take place.  @mjudeikis @charlesakalugwu can you sync on this between you on Thurs?

mjudeikis:
Re: `UpdateWithPluginConfig` interface name and little bit more:

What if we just call it `UpdateCluster`? Upgrade implies we take "newest state plugin knows about and apply". If called it would use pluginConfig with which plugin instance was initiated and regenerate config as you are doing it already. So as stated in Jim's comment, you don't need to pass it.

Now what Im missing is our plugin code function behaviour mapping. As it is very easy to make mistakes when using this now. Why?

1. Customer API call for functions `CreateOrUpdate` should never use PluginTemplate from current running plugin code, but should instead be using the one in the config blob. So it goes to how `Generate` code was called before calling this function. And this is implementation outside our code base, but how the plugin is being consumed. But without this knowledge is easy to make a mistake.

2. Admin API call for function `UpdateCluster` (see the relationship with `CreateOrUpdate` method name?) should be using almost same flow as `CreateOrUpdate` BUT plugin should be initiated with newest plugin code known configuration. And `GenerateConfig` should be called befire same as we are calling it for `CreateOrUpdate`

So all this kinda implies we should/could split our Interfaces by how their methods need to be initiated. I think that any method, which makes *changes to the cluster* from `Plugin` interface should always use `PluginTemplate` with which cluster was initiated. Hence - customer operations.

On the other hand, anything in `GenevaActions` interface should always use the current known newest `pluginTemplate` config version.

All this would imply 2 plugin code instantiations, depending on where API call is coming from or 2 Config generation instances. Otherwise mixing methods and interface behaviour is very dangerous to make an unexpected change which we cant test because consumption of all plugin code might be misinterpreted. 

@jim-minter @asalkeld @y-cote @charlesakalugwu WDYT?

mjudeikis:
Overall I dont like that this method behaves differently than `CreateOrUpdate` one. I think they should be as close as possible.

mjudeikis:
Or alternative we need to make `diff` in the client functions to make sure they are not initiated with different config (for parts which are not allowed to be changed) but this sounds complicated too and can be reason for errors.

pkg/plugin/plugin.go
jim-minter: bump

jim-minter:
/hold
so this doesn't get merged accidentally

@mjudeikis @charlesakalugwu we should discuss in scrum tomorrow and then with Microsoft, I think Charles might have found another problematic update edge case.

TL;DR:
Customer create workflow should call Generate then CreateOrUpdate.  FakeRP does this, RealRP does this.
UpdateWithPluginConfig should call do the same.

Question: what happens for customer update workflow?  If the customer cluster version is older than the RP version, we can probably support scale up/down without upgrade (don't call Generate), but it's possible that the customer could express a config change that is not supported by their cluster version, in which case an upgrade should happen as well (should call Generate).

Currently the FakeRP customer update workflow calls Generate.  I'm not sure what the RealRP does.

thekad:
@charlesakalugwu once you clear the outstanding design questions, can you either add a comment or update the PR description with the path forward so we can all be in the same channel?

pkg/fakerp/fakerp.go
```diff
@@ -60,9 +62,17 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
 		return nil, kerrors.NewAggregate(errs)
 	}
 
-	template, err := GetPluginTemplate()
-	if err != nil {
-		return nil, err
+	var err error
+	var template *pluginapi.Config
+	if shared.IsUpdate() && shared.IsScaleOperation(cs, oldCs) {
```
mjudeikis: This might do for now. But there are no limitations that we could not do scale-up for a customer using adminAPI. It is just logic ordering. But we should track which API endpoint request hit and act accordingly.

mjudeikis: + `IsUpdate` is little bit useless if we start thinking about scaleUp only and updates. ll this is fakeRP code, but logic looks weak for all this :/ dont know how to fix. need to think ...

mjudeikis:
all this is going to the direction on what I had in mind yesterday in the scrum call... need to sit and make that draft/proposal...

charlesakalugwu:
> @charlesakalugwu once you clear the outstanding design questions, can you either add a comment or update the PR description with the path forward so we can all be in the same channel?

@thekad heres the summary of the status quo as laid out in the code

* Added `IsScaleOperation(cs, oldCs)` function to `pkg/fakerp/shared` to detect when an update (from oldCs to cs) is a pure scale up/down operation
* Added `PluginTemplate(cs)` method to `pkg/util/managedcluster` to reconstruct the plugin template used to create the cluster specified by cs
* Changed the fakerp's `createOrUpdate` function to construct its plugin template as follows:
  * If the call to createOrUpdate is an update which is a pure scale operation, a plugin template is reconstructed from the cs (ensures only the intended agent pools are scaled up/down)
  * Otherwise, load the latest plugin template known to the RP (could cause a cluster upgrade)

pkg/fakerp/shared/shared_test.go
```diff
@@ -0,0 +1,79 @@
...
+				Properties: api.Properties{
+					AgentPoolProfiles: []api.AgentPoolProfile{
+						{Role: api.AgentPoolProfileRoleMaster, Name: "master", Count: 3},
+						{Role: api.AgentPoolProfileRoleCompute, Name: "compute", Count: 10},
+						{Role: api.AgentPoolProfileRoleInfra, Name: "infra", Count: 3},
+					},
+				},
+			},
+			expected: false,
+		},
```
thekad: just add another test: both old and new have the same agent pool profiles, but the configuration is different (this shouldn't trigger a scale op)

Issue #1130, opened 2019-02-13
Quay shows vulnerabilities on our Docker images


jim-minter:
We must do a `RUN yum -y update && yum clean` as part of all our rhel based Dockerfile builds.  If that means we need to go to Dockerfile.foo.rhel7 and Dockerfile.foo.centos7 for local development, then so be it.

kwoodson:
@jim-minter This was a topic of discussion around the team here https://github.com/openshift/openshift-azure/pull/1095#issuecomment-460663887.

We need to decide on which credentials we want to use for our builds.

kwoodson:
This image is updated every sprint and is maintained by our internal infrastructure and build teams (aos-art).  The image is updated here https://access.redhat.com/containers/?tab=tags&architecture=AMD64#/registry.access.redhat.com/rhel7 and shows that the versions are up-to-date.

I will figure out why we are using an older version.  I docker pull'd the latest and the packages mentioned in the quay.io have been updated.

I think we want to continue using this rhel7 image but enforce a tag or use latest as it is the official image that we are using across Openshift.

jim-minter:
@kwoodson https://github.com/openshift/release/pull/2854 should solve this

jim-minter:
fixed

PR #1142, opened 2019-02-13
Force azure image builds onto latest rhel7 image.

kwoodson:
As a follow on to #1095 and #1121, we need to ensure our images are at a specific patch level in order to keep them up to date.

The issue here is that our images are being built on an older image without the latest packages.  We should either move to a version or use latest.

### DIFF
--- a/Dockerfile.azure-controllers
+++ b/Dockerfile.azure-controllers
@@ -1,3 +1,3 @@
-FROM registry.access.redhat.com/rhel7
+FROM registry.access.redhat.com/rhel7:latest
 COPY azure-controllers .
 ENTRYPOINT [ "/azure-controllers" ]
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,3 +1,3 @@
-FROM registry.access.redhat.com/rhel7
+FROM registry.access.redhat.com/rhel7:latest
 COPY e2e .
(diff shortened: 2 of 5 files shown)


jim-minter:
gosh, does this really make a difference? If it wasn't pulling :latest before, what was it pulling?

kwoodson:
@jim-minter, good question.  I'm not sure but the version I have on my box is later than the version that is being used by the image building process.  I verified the RPMs in the azure-controllers image on quay was old and the latest rhel7 image had newer packages.  

I will attempt to build our latest images in ci-operator and verify they have the latest.

So I'm puzzled how this happens. Does the ci-operator build job use a registry that has a cached version of rhel7?

jim-minter:
@mjudeikis wdyt?
@kwoodson going to merge this but have removed the "fixes #1130" - please check to see if this did the trick.
/lgtm

mjudeikis:
@jim-minter @kwoodson I think latest or not, it will use image provided by CI-Operator. Same as OCP build can override base image, so CI-Operator too. 
And we use this as base image:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-azure/openshift-openshift-azure-master.yaml#L1

Which is this:
https://github.com/openshift/release/blob/master/projects/azure/base-images/plugin-base.yaml#L25

These dockerfiles `FROM` section is used only for local development

PR #1147, opened 2019-02-15
Relieve pressure on mdsd

jim-minter:
Set td-agent memory to 200Mi based on observation.
Set mdsd forwarding priority to high, this makes it forward data from memory to Geneva every minute rather than every five, significantly reduces memory usage.
Set mdsd pod memory to 400Mi.
Remove liveness probe for now, I'm highly sceptical it's doing us any good.

I suspect mdsd potentially has a memory leak so I still think we're going to see OOMs.


jim-minter:
@thekad @mjudeikis fyi

PR #1151, opened 2019-02-15
Make a startup container to convert much of the startup bash script

asalkeld:
The point of this PR is to create a golang startup binary that we can run from master-startup.sh
My next steps are to 
- create a vault in fakerp and insert a self signed api cert
- modify the new startup.go to download and install the cert from the vault

Note: I have included a commit that pulls the creation of storage accounts out from arm. This is so that
we can have the config blob in place by the time the startup script runs. normally this is not the case.


pkg/cluster/kubeclient/gomock_reflect_091577636/prog.go
```diff
@@ -0,0 +1,66 @@
+
```
charlesakalugwu: this file shouldn't have got in :)

asalkeld:
This will need a change to openshift/release for the new docker image

pkg/util/configblob/container.go
```diff
@@ -11,10 +14,34 @@ import (
...
+type StorageKey struct {
+	Name string `json:"name"`
+	Key  string `json:"key"`
+}
+
 // GetService is a helper function called by code running outside of the plugin.
 // It returns the blob storage interface to the storage account containing the
 // config blob, etcd backups, etc.
-func GetService(ctx context.Context, cpc *cloudprovider.Config) (azureclientstorage.BlobStorageClient, error) {
+func GetService(cloudproviderDir string) (azureclientstorage.BlobStorageClient, error) {
```
jim-minter: can you drop this change for now and stick with using ListKeys?

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+	origin = "/etc/origin"
+	etcd   = "/etc/etcd"
+)
+
+type startup struct {
+	blob azureclientstorage.Blob
+	log  *logrus.Entry
+}
+
+func (s *startup) writeKubeconfigs(cs *api.OpenShiftManagedCluster) error {
```
jim-minter: Am wondering about the following: follow a pattern of putting `{{ YamlMarshal .Config.AdminKubeconfig | Base64Encode }}` in a file /pkg/arm/data/etc/origin/node/node.kubeconfig and get rid of writeKubeconfigs().  The equivalent is possible to get rid of writeCerts.  In writeTemplatedFiles, check if the filename has suffix ".kubeconfig"/".key" to control using the file mode 0600.

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+	azureclientstorage "github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
+	"github.com/openshift/openshift-azure/pkg/util/configblob"
+	"github.com/openshift/openshift-azure/pkg/util/log"
+	"github.com/openshift/openshift-azure/pkg/util/template"
+)
+
+var (
+	logLevel         = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
+	runningUnderTest = flag.Bool("runningundertest", false, "TestConfig.RunningUnderTest")
+	dnsHostname      = flag.String("dnshostname", "", "hostname -d")
```
jim-minter: you shouldn't need to pass this in - something like the following:
```golang
	hostname, _ := os.Hostname()
	cname, _ := net.LookupCNAME(hostname)
	domainname := strings.SplitN(strings.TrimSuffix(cname, "."), ".", 2)[1]
```

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/openshift/openshift-azure/pkg/tls"
+	azureclientstorage "github.com/openshift/openshift-azure/pkg/util/azureclient/storage"
+	"github.com/openshift/openshift-azure/pkg/util/configblob"
+	"github.com/openshift/openshift-azure/pkg/util/log"
+	"github.com/openshift/openshift-azure/pkg/util/template"
+)
+
+var (
+	logLevel         = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
+	runningUnderTest = flag.Bool("runningundertest", false, "TestConfig.RunningUnderTest")
```
jim-minter: shouldn't need this? use cs.Config.RunningUnderTest instead.

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+	dnsHostname      = flag.String("dnshostname", "", "hostname -d")
+	gitCommit        = "unknown"
+)
+
+const (
+	origin = "/etc/origin"
```
jim-minter: instead of committing say pkg/arm/data/scheduler.json, commit pkg/arm/data/etc/origin/master/scheduler.json - then these consts go

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+		s.log.Debugf("processing template %s", destination)
+		templateFileName := path.Base(destination)
+		templateFile, err := arm.Asset(templateFileName)
+		if err != nil {
+			return err
+		}
+		b, err := template.Template(string(templateFile), nil, cs, map[string]interface{}{
+			"Hostname":    hostname,
+			"DNSHostname": *dnsHostname,
+			"TestConfig":  tc,
```
jim-minter: use cs.Config.RunningUnderTest instead

cmd/startup/startup.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "could not find storage account"))
+	}
+	s.blob = bsc.GetContainerReference(cluster.ConfigContainerName).GetBlobReference(cluster.ConfigBlobName)
+	s.log = log
+
+	return nil
+}
+
+func (s *startup) getBlob() (*api.OpenShiftManagedCluster, error) {
```
jim-minter: -> library function to share with sync pod, don't duplicate

pkg/arm/data/master-startup.sh
```diff
@@ -57,602 +57,41 @@ done
...
-base64 -d <<< {{ CertAsBytes .Config.Certificates.EtcdCa.Cert | Base64Encode }} >/etc/etcd/ca.crt
-base64 -d <<< {{ CertAsBytes .Config.Certificates.EtcdServer.Cert | Base64Encode }} >/etc/etcd/server.crt
-base64 -d <<< {{ PrivateKeyAsBytes .Config.Certificates.EtcdServer.Key | Base64Encode }} >/etc/etcd/server.key
-base64 -d <<< {{ CertAsBytes .Config.Certificates.EtcdPeer.Cert | Base64Encode }} >/etc/etcd/peer.crt
-base64 -d <<< {{ PrivateKeyAsBytes .Config.Certificates.EtcdPeer.Key | Base64Encode }} >/etc/etcd/peer.key
-
-base64 -d <<< {{ YamlMarshal .Config.AdminKubeconfig | Base64Encode }} >/etc/origin/node/node.kubeconfig
-base64 -d <<< {{ CertAsBytes .Config.Certificates.Ca.Cert | Base64Encode }} >/etc/origin/node/ca.crt
-
+mkdir -p /etc/etcd
```
jim-minter: writeTemplatedFiles should do this

pkg/arm/data/master-startup.sh
```diff
@@ -57,602 +57,41 @@ done
...
-  - args:
-    - start
-    - master
-    - controllers
-    - --config=/etc/origin/master/master-config.yaml
-    - --listen=https://0.0.0.0:8444
+######################################################
+# run startup container
+logger -t master-startup.sh "running startup container"
+while ! docker pull quay.io/asalkeld/aos-init; do
```
jim-minter: some wiring is missing - this should honour STARTUP_IMAGE

pkg/arm/data/master-startup.sh
```diff
@@ -57,602 +57,41 @@ done
...
-      name: master-cloud-provider
-      readOnly: true
-  hostNetwork: true
-  volumes:
-  - hostPath:
-      path: /etc/origin/cloudprovider
-    name: master-cloud-provider
-EOF
-fi
+cp /etc/origin/node/ca.crt /etc/origin/node/client-ca.crt
```
jim-minter: can be done via writeTemplatedFiles

pkg/util/configblob/container.go
asalkeld: sure can do

cmd/startup/startup.go
asalkeld: interesting, I'll give that a go

cmd/startup/startup.go
asalkeld: done

cmd/startup/startup.go
asalkeld: done

pkg/arm/data/master-startup.sh
asalkeld: yip done

cmd/startup/startup.go
asalkeld: done

cmd/startup/startup.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	for _, templateFileName := range files {
+		s.log.Debugf("processing template %s", templateFileName)
+		templateFile, err := arm.Asset(templateFileName)
+		if err != nil {
+			return errors.Wrapf(err, "Asset(%s)", templateFileName)
+		}
+		b, err := template.Template(string(templateFile), nil, cs, map[string]interface{}{
+			"Hostname":    hostname,
```
mjudeikis: Need to add this too:
https://github.com/openshift/openshift-azure/blob/master/pkg/arm/data/master-startup.sh#L489

mjudeikis:
Would it be staticPod or just `docker run` ?

mjudeikis:
Need changes in release repository to start building these images. This will never pass tests otherwise. And with changes in release repo it will not pass if we dont merge release first. But when we do, it will start blocking all other PR's (I think so). 

If you raise a PR inside release repo, I can orchestrate and merge them accordingly. 
https://github.com/openshift/release/

asalkeld:
> Would it be staticPod or just `docker run` ?

@mj I am open to other ideas, I thought that since it just runs once to install some files i'd run it with docker run. Not sure what is better?

mjudeikis:
> I am open to other ideas, I thought that since it just runs once to install some files i'd run it with docker run. Not sure what is better?

I think for now we can leave it as part of the startup script. We might want to think in the future maybe to wrap in in systemd service so we can control its behaviour and link to other dependent services if needed. Added this to arch document discussion so we can flush it out.

cmd/startup/startup.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *startup) writeTemplatedFiles(cs *api.OpenShiftManagedCluster) error {
+	tc := api.TestConfig{RunningUnderTest: cs.Config.RunningUnderTest}
+	hostname, _ := os.Hostname()
+	cname, _ := net.LookupCNAME(hostname)
+	domainname := strings.SplitN(strings.TrimSuffix(cname, "."), ".", 2)[1]
+
+	s.log.Infof("hostname is \"%s\"", hostname)
+	s.log.Infof("domainname %s", domainname)
+
+	files := []string{
```
jim-minter: use arm.AssetNames() here, then blacklist etc/origin/node/pods/sync.yaml as appropriate (and I guess {master,node}-startup.sh)

cmd/startup/startup.go
jim-minter: should be able to remove `"TestConfig": tc,`

cmd/startup/startup.go
```diff
@@ -0,0 +1,198 @@
...
+		})
+		if err != nil {
+			return errors.Wrapf(err, "Template(%s)", templateFileName)
+		}
+		parentDir := path.Base("/" + templateFileName)
+		err = os.MkdirAll(parentDir, 0700)
+		if err != nil {
+			return errors.Wrapf(err, "MkdirAll(%s)", parentDir)
+		}
+		err = ioutil.WriteFile("/"+templateFileName, b, 0600)
```
jim-minter: 0600 if it's a key or kubeconfig; 0666 otherwise (your uname will take that down to 0644 or whatever)

env.example
```diff
@@ -60,6 +60,10 @@ export DEPLOY_VERSION=v3.11
 # If overriding the image used for the backup container, customise this section.
 #export ETCDBACKUP_IMAGE=quay.io/openshift-on-azure/etcdbackup:latest
 
+# If overriding the image used for the startup pod, uncomment and customise this
+# section.
+#export STARTUP_IMAGE=quay.io/openshift-on-azure/startup:v3.11
```
jim-minter: please use :latest

jim-minter: also update /pluginconfig/pluginconfig-311.yaml

pkg/arm/arm.go
```diff
@@ -50,8 +50,6 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
 			vnet(cs),
 			ipAPIServer(cs),
 			lbAPIServer(&g.pluginConfig, cs),
-			storageRegistry(cs),
```
jim-minter: can we leave storageRegistry in the ARM template?

pkg/arm/data/etc/origin/node/pods/apiserver.yaml
```diff
@@ -0,0 +1,69 @@
...
+  name: master-api
+  namespace: kube-system
+spec:
+  containers:
+  - args:
+    - start
+    - master
+    - api
+    - --config=/etc/origin/master/master-config.yaml
+{{- if $.Extra.TestConfig.RunningUnderTest }}
```
jim-minter: @charlesakalugwu note #1175 is touching the same files as this PR

cmd/startup/startup.go
```diff
@@ -0,0 +1,198 @@
...
+		b, err := template.Template(string(templateFile), nil, cs, map[string]interface{}{
+			"Hostname":    hostname,
+			"DNSHostname": domainname,
+			"TestConfig":  tc,
+		})
+		if err != nil {
+			return errors.Wrapf(err, "Template(%s)", templateFileName)
+		}
+		parentDir := path.Base("/" + templateFileName)
+		err = os.MkdirAll(parentDir, 0700)
```
jim-minter: 0777

pkg/arm/data/master-startup.sh
```diff
@@ -57,607 +57,32 @@ done
...
-base64 -d <<< {{ CertAsBytes .Config.Certificates.EtcdServer.Cert | Base64Encode }} >/etc/etcd/server.crt
-base64 -d <<< {{ PrivateKeyAsBytes .Config.Certificates.EtcdServer.Key | Base64Encode }} >/etc/etcd/server.key
-base64 -d <<< {{ CertAsBytes .Config.Certificates.EtcdPeer.Cert | Base64Encode }} >/etc/etcd/peer.crt
-base64 -d <<< {{ PrivateKeyAsBytes .Config.Certificates.EtcdPeer.Key | Base64Encode }} >/etc/etcd/peer.key
-
-base64 -d <<< {{ YamlMarshal .Config.AdminKubeconfig | Base64Encode }} >/etc/origin/node/node.kubeconfig
-base64 -d <<< {{ CertAsBytes .Config.Certificates.Ca.Cert | Base64Encode }} >/etc/origin/node/ca.crt
-
+mkdir -p /etc/etcd
 mkdir -p /etc/origin/master/named
```
jim-minter: can delete this line?

pkg/arm/data/master-startup.sh
```diff
@@ -57,607 +57,32 @@ done
...
-          "label": "zone"
-        }
-      },
-      "name": "Zone",
-      "weight": 2
-    }
-  ]
-}
-EOF
+mkdir -p /etc/origin/node/pods
```
jim-minter: can't these be deleted?

pkg/cluster/initialize.go
```diff
@@ -58,6 +61,31 @@ func (u *simpleUpgrader) Initialize(ctx context.Context, cs *api.OpenShiftManage
...
 
+func (u *simpleUpgrader) CreateStorageAccount(ctx context.Context, cs *api.OpenShiftManagedCluster, accountName string) error {
+	parameters := azstorage.AccountCreateParameters{
+		Sku: &azstorage.Sku{
+			Name: azstorage.StandardLRS,
+		},
+		Kind:     azstorage.Storage,
+		Location: &cs.Location,
+	}
+	if accountName == cs.Config.ConfigStorageAccount {
```
jim-minter: roll the registry acct back into the arm template, make this function CreateConfigStorageAccount and get rid of the if

pkg/cluster/initialize.go
```diff
@@ -58,6 +61,31 @@ func (u *simpleUpgrader) Initialize(ctx context.Context, cs *api.OpenShiftManage
...
+		parameters.Tags = map[string]*string{
+			"type": to.StringPtr("config"),
+		}
+	}
+	future, err := u.accountsClient.Create(ctx, cs.Properties.AzProfile.ResourceGroup, accountName, parameters)
```
jim-minter: we've moved WaitForCompletionRef's into the API layer to ease unit testing.  Can you do the same with this one?

cmd/startup/startup.go
asalkeld: TestConfig is used

```
{{- if $.Extra.TestConfig.RunningUnderTest }}
    - --loglevel=4
{{- else }}
    - --loglevel=2
{{- end }}

```

cmd/startup/startup.go
asalkeld: TestConfig is used

cmd/startup/startup.go
asalkeld: i have moved this into util

pkg/arm/data/master-startup.sh
asalkeld: done

cmd/startup/startup.go
asalkeld: done

env.example
asalkeld: done both

pkg/arm/arm.go
asalkeld: ok, sure

pkg/cluster/initialize.go
asalkeld: done

pkg/cluster/initialize.go
asalkeld: done

pkg/arm/data/master-startup.sh
asalkeld: done

pkg/arm/data/master-startup.sh
asalkeld: done

cmd/startup/startup.go
jim-minter: If you rewrite `.Extra.TestConfig.RunningUnderTest` to `.Config.RunningUnderTest`, you don't need `TestConfig` any more.

cmd/startup/startup.go
```diff
@@ -0,0 +1,144 @@
...
+	gitCommit = "unknown"
+)
+
+type startup struct {
+	log *logrus.Entry
+	cs  *api.OpenShiftManagedCluster
+}
+
+func (s *startup) certFromVault(name string) ([]byte, error) {
+	if name == "PublicHostname" {
```
jim-minter: follow-up: const

cmd/startup/startup.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (s *startup) privateKeyFromVault(name string) ([]byte, error) {
+	if name == "PublicHostname" {
+		return tls.PrivateKeyAsBytes(s.cs.Config.Certificates.OpenShiftConsole.Key)
+	}
+	return []byte{}, nil
+}
+
+func (s *startup) writeTemplatedFiles() error {
+	tc := api.TestConfig{RunningUnderTest: s.cs.Config.RunningUnderTest}
```
jim-minter: follow-up: remove

pkg/plugin/plugin_test.go
```diff
@@ -34,6 +34,7 @@ func TestCreateOrUpdate(t *testing.T) {
 		return nil
 	}
 	cs := &api.OpenShiftManagedCluster{
+		Config: api.Config{ConfigStorageAccount: "config", RegistryStorageAccount: "registry"},
```
jim-minter: follow-up: nit: can remove RegistryStorageAccount population fwiw

pkg/arm/data/etc/etcd/ca.crt
```diff
@@ -0,0 +1 @@
+{{ CertAsBytes .Config.Certificates.EtcdCa.Cert | String }}
```
jim-minter: follow-up: path pkg/arm/data/etc will need to be renamed pkg/arm/data/master/etc so that we can add pkg/arm/data/node for node startup.

jim-minter:
/lgtm
going to get https://github.com/openshift/release/pull/2907 in so that this can run through tests properly

pkg/arm/data/etc/origin/node/pods/apiserver.yaml
charlesakalugwu: Roger

jim-minter:
force pushed to rebase
/lgtm

jim-minter:
force push to fix unit test after rebase
/lgtm

Makefile
```diff
@@ -87,9 +88,19 @@ sync-image: sync imagebuilder
...
+all-push: azure-controllers-push e2e-push etcdbackup-push metrics-push sync-push startup-push
+
+startup: generate
+	go build -ldflags ${LDFLAGS} ./cmd/startup
+
+startup-image: startup
+	go get github.com/openshift/imagebuilder/cmd/imagebuilder
+	imagebuilder -f Dockerfile.startup -t $(STARTUP_IMAGE) .
+
+startup-push: startup-image
```
Makdaam: I'd add startup, startup-image and startup-push to .PHONY at the end of the Makefile. Make didn't notice when source files changed for metricsbridge and sync - ended up making a new docker image with an old binary for it.

jim-minter: metricsbridge needs to be in .PHONY because it does not depend on generate, which isn't in .PHONY but should be (in practice we get away with it because a file called generate does not exist).
We can get away with not listing startup, sync, etc. in .PHONY because they depend on generate.
I agree it's a bit brittle but it's good enough for this PR.

PR #1168, opened 2019-02-18
Move writeHelpers to config generator

### LINKED ISSUE #1167: Cannot ssh into VM after running key rotation tests
After running key rotation tests, we are no longer able to ssh into the VMs because we are failing to write an updated ssh key after the key rotation. 

/kind bug
/cc @mjudeikis @mgahagan73 @jim-minter

charlesakalugwu:
This adds a `WriteHelpers` method to the `Generator` interface in `pkg/config`. The method is called from `plugin.GenerateConfig` which therefore ensures that the plugin's `CreateOrUpdate` or other config altering calls such as `RotateClusterSecrets` result in updated ssh keys, kubeconfigs etc being laid down on the machine. This will allow an operator to ssh into any VM during a key rotation run.

/closes #1167
/cc @mjudeikis @jim-minter @mgahagan73

### DIFF
--- a/pkg/fakerp/admin_handlers.go
+++ b/pkg/fakerp/admin_handlers.go
@@ -118,5 +118,9 @@ func (s *Server) handleRotateSecrets(w http.ResponseWriter, req *http.Request) {
 		s.internalError(w, fmt.Sprintf("Failed to rotate cluster secrets: %v", err))
 		return
 	}
+	err = writeHelpers(cs)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


jim-minter:
/hold
as discussed, this shouldn't go in the config pkg

charlesakalugwu:
added `writeHelpers(cs)` at the end of `handleRotateSecrets`. If key rotation fails midway through the VM rotation, writeHelpers will not have been called and we'll be locked out until we run a successful key rotation. 

/hold cancel

PR #1173, opened 2019-02-20
Substantially reduce the number of metrics being forwarded to Geneva.

jim-minter:
* no longer use Prometheus federate endpoint, this is because you can't send it
  arbitrary queries and doing so seems to be the best way to pre-aggregate
  across unwanted labels

* omitLabels is no longer necessary as we're not using the federate endpoint

* use Prometheus client API library instead of DIY

* make list of queries much more restricted, we can always add more back in

* remove excessive labels from apiserver_request_.+

* the new queries syntax in the yaml file only requires the name: field if the
  query itself aggregates.  In this circumstance Prometheus removes the name
  from the metric; the name: field is used to add it back in as part of the
  passthrough

This PR reduces the number of metrics nearly ten-fold, down to around 6000 for a newly provisioned cluster.


jim-minter:
@openshift/sig-azure this PR looks OK to me but I can't fully test it because of the Geneva throttling.  We can try merging it, get people to recreate their clusters, then see if it's working?

@Makdaam please sign off.

mjudeikis:
@jim-minter why do not enrich metrics anymore? How this data is being appended to the data? Or I missed something?

PR #1174, opened 2019-02-20
Geneva action get RP plugin version

charlesakalugwu:
This work implements a geneva action that allows an admin to fetch the RP plugin version

This is part of AZURE-257

/cc @Makdaam @mjudeikis @kwoodson @jim-minter


pkg/fakerp/admin_handlers.go
```diff
@@ -124,3 +126,15 @@ func (s *Server) handleRotateSecrets(w http.ResponseWriter, req *http.Request) {
 	}
 	s.log.Info("rotated cluster secrets")
 }
+
+// handleGetPluginVersion handles admin requests to get the RP plugin version for OpenShiftManagedClusters
+func (s *Server) handleGetPluginVersion(w http.ResponseWriter, req *http.Request) {
+	version := &api.OpenShiftManagedClustersPluginVersion{Version: s.pluginTemplate.ClusterVersion}
```
jim-minter: this is missing the implementation in the plugin which does this.

charlesakalugwu: done

jim-minter: ok, but now this function needs to call that implementation ...

pkg/plugin/plugin.go
```diff
@@ -264,3 +264,12 @@ func (p *plugin) ForceUpdate(ctx context.Context, cs *api.OpenShiftManagedCluste
 	p.log.Info("force updates successful")
 	return nil
 }
+
+func (p *plugin) GetPluginVersion(ctx context.Context, pluginTemplate *pluginapi.Config) ([]byte, error) {
```
kwoodson: The version is in the config struct.  Where does that get stored?  How does this get the pluginapi.confg?  Are we storing that somewhere other than the config blob?

pkg/api/plugin.go
```diff
@@ -125,4 +125,7 @@ type GenevaActions interface {
 
 	// ForceUpdate forces rotates all vms in a cluster
 	ForceUpdate(ctx context.Context, cs *OpenShiftManagedCluster, deployer DeployFn) *PluginError
+
+	// GetPluginVersion fetches the RP plugin version
+	GetPluginVersion(ctx context.Context, template *plugin.Config) ([]byte, error)
```
thekad: I got a basic consistency complaint here: `GetPluginVersion` handles `/admin/pluginVersion` however the values come from the `config.ClusterVersion` (and the `clusterVersion` value in the config file). I'd much rather have all these follow the same nomenclature (e.g. `/admin/clusterVersion`) to avoid confusion in the future.

pkg/plugin/plugin.go
thekad: why return a `[]byte` at all? return a string

pkg/plugin/plugin.go
```diff
@@ -264,3 +264,12 @@ func (p *plugin) ForceUpdate(ctx context.Context, cs *api.OpenShiftManagedCluste
...
+	pluginVersion := &struct {
+		Version string `json:"version,omitempty"`
```
thekad: shouldn't the json tag be `clusterVersion` ?

pkg/api/plugin.go
jim-minter: @thekad you're right.  A future PR is going to rename config.ClusterVersion to PluginVersion.

jim-minter:
@charlesakalugwu I've force pushed, fixing the structural issue whereby the plugin template was not a member of the plugin struct.
@mjudeikis @asalkeld this is a bit of a backwards step wrt supporting multiple major versions, but we can deal with that properly after GA.  We're currently in a halfway house between "plugin template per major version" and "one plugin config containing one plugin template per major version".

@charlesakalugwu marking lgtm, unhold if happy
/hold
/lgtm

mjudeikis:
@jim-minter agreed. One more thing, we might need to enhance, is to add something like:
```
clusterVersion
pluginVersion
... component versions ...
```
This is Fridays conversation from 2 weeks ago.

jim-minter:
closing for #1261

PR #1175, opened 2019-02-20
Configurable openshift component logging

charlesakalugwu:
This work makes the log level of the openshift components configurable for clusters.

This currently only allows changing the log levels for api-server, node and controller-manager components to start with. 

This is part of AZURE-257

```release-note
plugin: geneva action to change openshift component level logging
```

/cc @jim-minter @kwoodson @thekad @asalkeld


pkg/api/config.go
```diff
@@ -11,6 +11,8 @@ import (
 type Config struct {
 	// ClusterVersion defines release version of the plugin used to build the cluster
 	ClusterVersion string `json:"clusterVersion,omitempty"`
+	// ClusterLogLevel specifies the OpenShift component log level
+	ClusterLogLevel uint8 `json:"clusterLogLevel,omitempty"`
```
kwoodson: Would it be better to make this a const?  Similar to https://github.com/openshift/openshift-azure/blob/master/pkg/api/types.go#L71

pkg/api/validate/plugin.go
```diff
@@ -24,6 +24,10 @@ func (v *PluginAPIValidator) Validate(t *pluginapi.Config) (errs []error) {
 	if !clusterVersion.MatchString(t.ClusterVersion) {
 		errs = append(errs, fmt.Errorf("invalid ClusterVersion %q", t.ClusterVersion))
 	}
+	// validate ClusterLogLevel
+	if t.ClusterLogLevel%2 != 0 || (t.ClusterLogLevel < 2 || t.ClusterLogLevel > 8) {
```
kwoodson: 0 is a valid clusterloglevel.  Any reason we are excluding it?

pkg/arm/data/node-startup.sh
```diff
@@ -9,7 +9,7 @@ if [ -f "/etc/sysconfig/atomic-openshift-node" ]; then
 fi
 
 # remove registry certificate softlink from docker
-unlink /etc/docker/certs.d/registry.access.redhat.com/redhat-ca.crt 
+unlink /etc/docker/certs.d/registry.access.redhat.com/redhat-ca.crt
```
kwoodson: not for this PR but this should move to the node image build.

pkg/api/config.go
jim-minter: This is going back on what I said to you before, but it's little extra work: LogLevel would be better as a struct containing different components and their log levels.  Example: apiserver, controllers, node.  Can use an int rather than a uint8.

pkg/fakerp/admin_handlers.go
```diff
@@ -124,3 +126,45 @@ func (s *Server) handleRotateSecrets(w http.ResponseWriter, req *http.Request) {
 	}
 	s.log.Info("rotated cluster secrets")
 }
+
+// handleGetLogLevel handles admin requests to get the configured OpenShift component log level for OpenShiftManagedClusters
+func (s *Server) handleGetLogLevel(w http.ResponseWriter, req *http.Request) {
```
jim-minter: you don't need any fakerp changes or additional geneva action - with the config struct in place the generic admin API can handle getting and putting loglevel changes.

pluginconfig/pluginconfig-311.yaml
```diff
@@ -13,6 +13,14 @@
...
 ## Geneva integration other details
 clusterVersion: v0.0
+## Openshift component logging level
+#levels:
+  # 0 - Errors and warnings only
+  # 2 - Normal information
+  # 4 - Debugging-level information
+  # 6 - API-level debugging information (request / response)
+  # 8 - Body-level API debugging information
+clusterLogLevel: 4
```
jim-minter: default to 2, not 4

test/util/populate/walk.go
```diff
@@ -115,7 +115,7 @@ func (w walker) walk(v reflect.Value, path string) {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		v.SetInt(1)
 	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
-		v.SetUint(1)
+		v.SetUint(2)
```
jim-minter: revert

pkg/api/validate/plugin_test.go
```diff
@@ -14,6 +14,7 @@ func TestPluginTemplateValidate(t *testing.T) {
 			errors.New(`imagePublisher should be redhat`),
 			errors.New(`invalid ImageSKU ""`),
 			errors.New(`invalid ClusterVersion ""`),
+			errors.New(`invalid ClusterLogLevel 0`),
```
thekad: log level 0 is valid afaik

pkg/arm/data/master-startup.sh
```diff
@@ -44,9 +44,7 @@ fi
 
 echo "BOOTSTRAP_CONFIG_NAME=node-config-master" >>/etc/sysconfig/${SERVICE_TYPE}-node
 
-{{- if $.Extra.TestConfig.RunningUnderTest }}
-sed -i -e "s#DEBUG_LOGLEVEL=2#DEBUG_LOGLEVEL=4#" /etc/sysconfig/${SERVICE_TYPE}-node
-{{- end }}
+sed -i -e "s#DEBUG_LOGLEVEL=2#DEBUG_LOGLEVEL={{ .Config.ClusterLogLevel }}#" /etc/sysconfig/${SERVICE_TYPE}-node
```
thekad: I think with these changes we're losing the ability to get debug logs when `RUNNING_UNDER_TEST`, I'd suggest work that back into the dev process, either by defaulting the logLevel to 4 when running under test, or exposing the logLevel to be configurable through environment variables.

pkg/api/config.go
asalkeld: yes, otherwise you get all or nothing.

charlesakalugwu: good idea

pkg/api/validate/plugin.go
charlesakalugwu: rationale was that setting 0 wouldn't be helpful to a cluster admin. I'll revert this

pkg/api/config.go
charlesakalugwu: good idea. I'll get this in

test/util/populate/walk.go
charlesakalugwu: done

pluginconfig/pluginconfig-311.yaml
charlesakalugwu: done

pkg/api/validate/plugin_test.go
charlesakalugwu: yep

pkg/fakerp/admin_handlers.go
charlesakalugwu: alright

pkg/arm/data/master-startup.sh
charlesakalugwu: i can readd this in a follow-up

pkg/api/convertertoadmin.go
```diff
@@ -104,6 +104,7 @@ func ConvertToAdmin(cs *OpenShiftManagedCluster) *admin.OpenShiftManagedCluster
 func convertConfigToAdmin(cs *Config) *admin.Config {
 	return &admin.Config{
 		ClusterVersion:                       &cs.ClusterVersion,
+		ClusterLogLevel:                      convertClusterLogLevelsToAdmin(cs.ClusterLogLevel),
```
jim-minter: there are no changes made to convertfromadmin.go and there should be.  This should have triggered a unit test failure.  Add a unit test that does a round trip: populate a dummy `*admin.OpenShiftManagedCluster`, convert it to a `*OpenShiftManagedCluster`, convert that back again and check that what you get equals what you started with.

pkg/api/validate/plugin.go
```diff
@@ -53,6 +55,19 @@ func (v *PluginAPIValidator) Validate(t *pluginapi.Config) (errs []error) {
 	return errs
 }
 
+func (v *PluginAPIValidator) validatePluginTemplateComponentLogLevels(c *pluginapi.ComponentsLogLevel) (errs []error) {
+	if c.ApiServer%2 != 0 || (c.ApiServer < 0 || c.ApiServer > 8) {
```
jim-minter: the %2 does not make sense.  Make the types uints, then check they are not larger than, say, 20.

pluginconfig/pluginconfig-311.yaml
```diff
@@ -13,6 +13,17 @@
   #  cert: <Geveva metrics certificate value. Format: Base64 in YAML, x509.Certificate struct>
 ## Geneva integration other details
 clusterVersion: v0.0
+## Openshift component logging levels
+#levels:
```
jim-minter: this is not the right place for this documentation - please remove.

test/e2e/specs/fakerp/loglevel.go
```diff
@@ -0,0 +1,58 @@
...
+		Expect(update).NotTo(BeNil())
+		Expect(update.StatusCode).To(Equal(http.StatusOK))
+
+		By("Reading the internal config after the log level update")
+		after, err := azurecli.OpenShiftManagedClustersAdmin.Get(context.Background(), os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(after).NotTo(BeNil())
+
+		By("Verifying that the cluster log level has been updated")
+		Expect(after.Config.ClusterLogLevel).To(Equal(before.Config.ClusterLogLevel))
```
jim-minter: I don't understand why this is succeeding.  I would expect it currently to fail because of validateUpdateConfig (which this PR should be updating).  That's a fairly serious bug - please solve it before this gets merged.

pkg/api/validate/plugin.go
charlesakalugwu: valid logging levels are 0,2,4,6 and 8. The %2 checks that the supplied log level is an even number. Am i missing something?

pluginconfig/pluginconfig-311.yaml
charlesakalugwu: alright

pkg/arm/data/master-startup.sh
charlesakalugwu: @jim-minter just to make sure, are you ok with @thekad 's suggestion? I think I prefer his second option. what do you think?

pkg/api/convertertoadmin.go
charlesakalugwu: thats right, theres even more stuff that had been left out in conversions

```
=== RUN   TestRoundTrip
--- FAIL: TestRoundTrip (0.00s)
    converterfromadmin_test.go:301: unexpected diff [Config.ClusterVersion: Config.ClusterVersion !=  Config.ClusterLogLevel.ApiServer: 1 != 0 Config.ClusterLogLevel.ControllerManager: 1 != 0 Config.ClusterLogLevel.Node: 1 != 0 Config.GenevaLoggingControlPlaneEnvironment: Config.GenevaLoggingControlPlaneEnvironment !=  Config.GenevaLoggingControlPlaneRegion: Config.GenevaLoggingControlPlaneRegion != ]
```

I'll add them in

test/e2e/specs/fakerp/loglevel.go
charlesakalugwu: Having a look. I also have a PR for the prow jobs open: https://github.com/openshift/release/pull/2914 . Can we get that merged so I can run `/test changeloglevel` in CI?

pkg/api/admin/api/config.go
```diff
@@ -6,9 +6,18 @@ import (
...
+	ApiServer         *int `json:"apiServer,omitempty"`
+	ControllerManager *int `json:"controllerManager,omitempty"`
+	Node              *int `json:"node,omitempty"`
+}
+
 type Config struct {
 	// ClusterVersion (out) defines release version of the plugin used to build the cluster
 	ClusterVersion *string `json:"clusterVersion,omitempty"`
+	// ClusterLogLevel specifies the log levels for the various openshift components
+	ClusterLogLevel *ComponentsLogLevel `json:"clusterLogLevel,omitempty"`
```
jim-minter: yikes - pick one name and stick to it
```
Foo *Foo `json:"foo,omitempty"`
```

jim-minter: also - use spaces to separate logical groups

pkg/api/converterfromadmin.go
```diff
@@ -349,9 +354,27 @@ func mergeConfig(oc *admin.OpenShiftManagedCluster, cs *OpenShiftManagedCluster)
 	if in.GenevaLoggingControlPlaneAccount != nil {
 		out.GenevaLoggingControlPlaneAccount = *in.GenevaLoggingControlPlaneAccount
 	}
+	if in.GenevaLoggingControlPlaneEnvironment != nil {
```
jim-minter: your new tests found these were missing?  If so I owe you :beer:

jim-minter:
@charlesakalugwu also rebased and pushed - ptal
please can you check the test is working - lmk then will lgtm
/hold

pkg/api/converterfromadmin.go
charlesakalugwu: @jim-minter yes the test discovered the following were missing:

```
Config.GenevaLoggingControlPlaneEnvironment
Config.GenevaLoggingControlPlaneRegion
```

jim-minter:
retest broke because https://github.com/openshift/openshift-azure/pull/1220 is in flight, will try retest again once that lands

pkg/api/config.go
```diff
@@ -68,6 +72,13 @@ type Config struct {
 	GenevaMetricsEndpoint string `json:"genevaMetricsEndpoint,omitempty"`
 }
 
+// ComponentLogLevel represents the log levels for the various components of a cluster
+type ComponentLogLevel struct {
```
mjudeikis: Router is needed too.

mjudeikis: Can be follow-up

pkg/api/converterfromadmin.go
```diff
@@ -300,7 +300,12 @@ func convertIdentityProviderAdmin(in admin.IdentityProvider, old *IdentityProvid
 
 func mergeConfig(oc *admin.OpenShiftManagedCluster, cs *OpenShiftManagedCluster) {
 	in, out := oc.Config, &cs.Config
-
+	if in.ClusterVersion != nil {
```
mjudeikis: Why this one was added? I think we dont propagate this field back to OpenShiftManageCluster?

mjudeikis: Or my brains still on holiday?

mjudeikis:
@openshift/sig-azure let's stop merging release jobs before PR is merged/or we are 100% sure tests are passing. 
We can test these locally using:https://github.com/openshift/release/tree/master/projects/azure#testing-prow-jobs

charlesakalugwu:
@jim-minter I discovered and fixed the following issues and fixed those back up to your commit.

- Json tags for componentLogLevel was incorrectly specified as: 

```
ComponentLogLevel ComponentLogLevel `json:"componentLogLevels,omitempty"`
```
- I added a TestJsonTags test to check for the conformance of all the plugin template's json tags
- Changed e2e tests to perform relative updates of the log level. Instead of setting it to 4 (which results in a no-op update on clusters created with RUNNING_UNDER_TEST=true) we subtract 2 from the current log levels. As we run e2e tests on fresh clusters this should not be a problem. 
- fakerp.overridePluginTemplate now also accepts a boolean isUpdate argument. This allows us to only set loglevel=4 to clusters being created under test. We should be able to update loglevel afterwards.
- Changed the following sed commands in the node startup scripts from 
```
sed -i -e "s#DEBUG_LOGLEVEL=2#DEBUG_LOGLEVEL={{ .Config.ComponentLogLevel.Node }}#" /etc/sysconfig/${SERVICE_TYPE}-node
```
to
```
sed -i -e "s#DEBUG_LOGLEVEL=.*#DEBUG_LOGLEVEL={{ .Config.ComponentLogLevel.Node }}#" /etc/sysconfig/${SERVICE_TYPE}-node
```
This is to ensure that the change takes effect always instead of only on clusters with loglevel=2 (clusters created with RUNNING_UNDER_TEST=false)
- updated validateUpdateContainerService to allow updates to the log level.

pkg/api/converterfromadmin.go
jim-minter: this PR added a round trip test to prove that you can convert admin -> internal -> admin and get the same thing back - this discovered several missing fields which are added here

jim-minter:
thanks @charlesakalugwu 
/hold cancel
/lgtm

jim-minter:
/lgtm cancel
e2e issue looks genuine

asalkeld:
needs rebase

jim-minter:
closing for #1271

PR #1176, opened 2019-02-20
etcd metrics shim

mjudeikis:
This is work in progress for ETCD Metrics shim container/tls-proxy

* Adds TLSProxy binary/image
* Uses TLSproxy image to get metrics by running on master nodes and mounting etcd certificates as host path
* exposes metrics via auth endpoint and different certificate, generated by OCP CA
* Creates ServiceMonitor object to monitor this endpoint

```release-note
Internal: Enable ETCD metrics via tls-proxy container 
```


mjudeikis:
@jim-minter need some feedback on implementation. I think these limits expose of the certs into the secret so etcd. At the same time metrics are not flying in open-text over the network and we still do TLS.

charlesakalugwu:
https://github.com/openshift/openshift-azure/issues/1177

mjudeikis:
```
- let's not call it etcdmetrics, I bet we'll use the same component for vm image builds / kickstarts.  Call it tlsproxy or some such
-dont default to certs
- add a configurable restriction that only GET verb can be used
- ok, set the env var but use command:/args: on the daemonset to run entrypoint -foo -bar $HOSTNAME
then don't need to customise the binary
- sync pod to create the monitor object. CRD to register
```

pkg/addons/data/SecurityContextConstraints.security.openshift.io/privileged.yaml
```diff
@@ -42,6 +42,7 @@ users:
 - system:serviceaccount:openshift-node:sync
 - system:serviceaccount:openshift-etcd:etcd-backup
 - system:serviceaccount:openshift-azure-logging:geneva
+- system:serviceaccount:openshift-azure-monitoring:etcd-metrics
```
mjudeikis: This needs to go

pkg/addons/data/ClusterRoleBinding.rbac.authorization.k8s.io/geneva-cluster-reader.yaml
```diff
@@ -13,3 +13,6 @@ subjects:
 - kind: ServiceAccount
   name: geneva
   namespace: openshift-azure-monitoring
+- kind: ServiceAccount
```
mjudeikis: Need to think about how to limit this. Prom operator needs access to read endpoints, svc, and get to the metrics

mjudeikis:
follow-up : https://github.com/openshift/openshift-azure/issues/1182

mjudeikis:
missing PR to release repo to add images.
But this works fine, Just need to clean tls-proxy code. 
@jim-minter ^

mjudeikis:
ping @jim-minter

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,289 @@
...
+
+func usage() {
+	fmt.Printf("Usage:\n")
+	fmt.Printf("\"%s -hostname\" url to rewrite \n\n", os.Args[0])
+	flag.PrintDefaults()
+	os.Exit(1)
+}
+
+func (c *config) Run() error {
+
```
asalkeld: nit: rm empty line

Dockerfile.tlsproxy
```diff
@@ -0,0 +1,3 @@
+FROM registry.access.redhat.com/rhel7:latest
```
asalkeld: this probably needs to be 
images/tlsproxy/Dockerfile now..

jim-minter:
@mjudeikis please rebase

mjudeikis:
@jim-minter rebased

pkg/addons/data/ClusterRoleBinding.rbac.authorization.k8s.io/geneva-cluster-reader.yaml
kwoodson: It appears this cluster role binding binds cluster-reader to this service account.  That has access to read everything minus secrets.

If we could list the dependencies then we could come up with a policy to limit the access.  What does get metrics require?

Start with something small and play whack-a-mole until the pieces come together.  @mjudeikis, let me know how I can help.
```
rules
- apiGroups:
  - ""
  resources:
  - endpoints
  - svc
  verbs:
  - get
  - list
```

jim-minter:
@mjudeikis PR is currently checking in a stray etcdmetrics binary, please remove
@mjudeikis please rebase

cmd/tlsproxy/example-config.yaml
```diff
@@ -0,0 +1,11 @@
+loglevel: Debug
```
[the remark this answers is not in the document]
jim-minter: Adding all the boilerplate - config file, config merging, etc. makes it harder to reuse this in #1264

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (c *config) validate() []error {
+	var errs []error
+	// check variables
+	if *hostname != "" && !hostnameRegex.MatchString(*hostname) {
+		return append(errs, errors.New("hostname is not a valid hostname"))
+	}
+	if c.Ca == "" {
+		return append(errs, errors.New("cacert must be provided"))
```
jim-minter: even when insecure is set?

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	if c.Cert == "" {
+		return append(errs, errors.New("cert must be provided"))
+	}
+	if c.Key == "" {
+		return append(errs, errors.New("key must be provided"))
+	}
+	if c.Reencrypt {
```
jim-minter: why do you need the setting?  just enable it when serving{key,cert} are provided

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,288 @@
...
+		if c.ServingKey == "" {
+			return append(errs, errors.New("servingkey must be provided for re-encrypt"))
+		}
+		if c.ServingCert == "" {
+			return append(errs, errors.New("servingcert must be provided for re-encrypt"))
+		}
+	}
+	if c.password == "" && c.username != "" ||
+		c.password != "" && c.username == "" {
+		return append(errs, errors.New("USERNAME and PASSWORD variables must be provided"))
```
jim-minter: needs to be optional for #1264.  Just enable auth checking if they're set.  Write an e2e to prove that auth is required on an OSA cluster.

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,288 @@
...
+	whitelistRegexp *regexp.Regexp
+}
+
+type envConfig struct {
+	// env var config
+	Username string `envconfig:"USERNAME"`
+	Password string `envconfig:"PASSWORD"`
+}
+
+func (c *config) validate() []error {
```
jim-minter: Honestly, I think this entire function can go.  The errors will show themselves up clearly quickly anyway.

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	return errs
+}
+
+func (c *config) Init(path string) error {
+	var err error
+
+	// Resolve env to config structure
+	ec := envConfig{}
```
jim-minter: overkill - just use os.Getenv for the password

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -0,0 +1,79 @@
...
+            secretKeyRef:
+              key: password
+              name: etcd-metrics
+        image: '*** GENERATED ***'
+        name: etcd-metrics
+        ports:
+        - containerPort: 8080
+          name: metrics
+        resources:
+          requests:
```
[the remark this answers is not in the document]
mjudeikis: Thinking behind this argument?

pkg/config/generate.go
```diff
@@ -376,6 +376,17 @@ func (g *simpleGenerator) Generate(cs *api.OpenShiftManagedCluster, template *pl
 		}
 	}
 
+	if len(c.EtcdMetricsPassword) == 0 {
+		if c.EtcdMetricsPassword, err = random.AlphanumericString(10); err != nil {
+			return
+		}
+	}
+	if len(c.EtcdMetricsUsername) == 0 {
```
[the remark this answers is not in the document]
mjudeikis: considering we are building generic tlsProxy, I would make it consistent configurable? Otherwise, we will need to duplicate values here (hardcoding and , tlsProxy config where we would set to the same value)

pluginconfig/pluginconfig-311.yaml
```diff
@@ -41,6 +41,7 @@ images:
   controlPlane: registry.access.redhat.com/openshift3/ose-control-plane:v3.11.82
   etcdBackup: quay.io/openshift-on-azure/etcdbackup:latest
   startup: quay.io/openshift-on-azure/startup:latest
+  tlsProxy: quay.io/openshift-on-azure/tlsproxy:dev
```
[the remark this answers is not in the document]
mjudeikis: yes, this was for testing so not to publish under latest. Note for future me, leave notes in the code about dev values.

pkg/addons/addons.go
```diff
@@ -178,15 +178,21 @@ var (
...
 	}
+
 	scFilter = func(o unstructured.Unstructured) bool {
 		return o.GroupVersionKind().Group == "servicecatalog.k8s.io"
 	}
 	crdFilter = func(o unstructured.Unstructured) bool {
 		return o.GroupVersionKind().GroupKind() == schema.GroupKind{Group: "apiextensions.k8s.io", Kind: "CustomResourceDefinition"}
 	}
+	// targeted filter is used to target specific CRD's, which are managed not by sync pod
+	targetedCrdFilter = func(o unstructured.Unstructured) bool {
```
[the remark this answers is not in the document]
mjudeikis: I suspect we will have more of these, with the same condition. So instead of duplicating and adding more filers, more `updateDynamicClient` and more `CRDReady` statements to follow I was thinking to keep this generic and add all under one filter?

cmd/tlsproxy/example-config.yaml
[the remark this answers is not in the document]
mjudeikis: Its just question of readability. I personally like config files over 10 flags. But fine either way.

cmd/tlsproxy/tlsproxy.go
mjudeikis: it is optional. Validation happens only when one of the fields is set, but not other. Same with configuration. No fields set - no auth.

cmd/tlsproxy/tlsproxy.go
mjudeikis: If we dont do config file, this means for simplicity purpose we will just use "flags values" all around the place directly. I would like to have it sanitized, as it will be used in different places and packages we dont control (we should trust them, but still)

pkg/addons/data/ClusterRoleBinding.rbac.authorization.k8s.io/geneva-cluster-reader.yaml
mjudeikis: removed

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,220 @@
...
+
+func usage() {
+	fmt.Printf("Usage:\n")
+	fmt.Printf("\"%s -hostname\" url to rewrite \n\n", os.Args[0])
+	flag.PrintDefaults()
+	os.Exit(1)
+}
+
+func (c *config) Run() error {
+	whitelist, err := regexp.Compile(*whitelist)
```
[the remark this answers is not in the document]
mjudeikis: yes :)

jim-minter:
/hold
want to give this one more review in the next hour or two

cmd/tlsproxy/tlsproxy.go
jim-minter: fix the error message - if one is provided, both must be provided...

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,214 @@
...
+		*servingkey != "" && *servingcert == "" {
+		return append(errs, errors.New("servingkey and servingcert must be provided for re-encrypt"))
+	}
+	if c.password == "" && c.username != "" ||
+		c.password != "" && c.username == "" {
+		return append(errs, errors.New("USERNAME and PASSWORD variables must be provided"))
+	}
+
+	// check files exist
+	if _, err := os.Stat(*cacert); os.IsNotExist(err) {
```
jim-minter: `if *cacert != ""`

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,215 @@
...
+	var errs []error
+	// check variables
+	if *hostname != "" && !validate.IsValidURI(*hostname) {
+		return append(errs, errors.New("hostname is not a valid hostname"))
+	}
+	if !*insecure {
+		if *cacert == "" {
+			return append(errs, errors.New("cacert must be provided"))
+		}
+		if *cert == "" {
```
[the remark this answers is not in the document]
jim-minter: at the moment you're saying that cert and key must be set - I guess that's the point of the program - but they could harmlessly be optional

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,214 @@
...
+				RootCAs:            pool,
+				Certificates:       []tls.Certificate{cert},
+			},
+		},
+	}
+
+	return nil
+}
+
+func usage() {
```
jim-minter: no need for this now everything's a flag

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (c *config) Run() error {
+	handlerFunc := http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		req.URL.Scheme = c.redirectURL.Scheme
+		req.URL.Host = c.redirectURL.Host
+		req.RequestURI = ""
+		req.Host = ""
+
+		if !c.whitelistRegexp.MatchString(req.URL.String()) || req.Method != http.MethodGet {
```
jim-minter: preferably MethodGet would be a flag (follow up for another day)

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,214 @@
...
+		c.log.Debug("starting in reencrypt mode")
+		return http.ListenAndServeTLS(*listen, *servingcert, *servingkey, handlerFunc)
+	}
+	c.log.Debug("starting in plain text mode")
+	return http.ListenAndServe(*listen, handlerFunc)
+}
+
+func (c *config) checkAuth(rw http.ResponseWriter, req *http.Request) bool {
+	if username, password, ok := req.BasicAuth(); ok {
+		return username == c.username && password == c.password
```
jim-minter:
```
username, password, _ := req.BasicAuth()
return username == c.username && password == c.password
```

is clearer.  Also name the function something like authIsOK so that usage is clearer at the caller

hack/lint-addons/lint-addons.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
```
jim-minter: can this be reverted from this PR? I'd like to revisit if/how we should use pkg/errors separately as a team

pkg/addons/addons.go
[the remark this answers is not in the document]
jim-minter: the filter thing is exploding, but I don't think your way is the way to solve the problem.  Keeping these separate at least makes dependencies clearer - also helps my work to make the sync pod report ready when it is ready (excluding waiting for the monitoring stack at cluster startup)

pkg/addons/data/ClusterRoleBinding.rbac.authorization.k8s.io/geneva-cluster-reader.yaml
jim-minter: this is granting prometheus-k8s cluster reader.  I do not want to do that.

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -0,0 +1,83 @@
...
+    metadata:
+      labels:
+        k8s-app: etcd-metrics
+    spec:
+      containers:
+      - args:
+        - -loglevel
+        - debug
+        - -hostname
+        - https://$(HOSTNAME):2379
```
jim-minter: this works?

pkg/addons/data/Role.rbac.authorization.k8s.io/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Role
+metadata:
+  name: etcd-metrics-reader
+  namespace: openshift-azure-monitoring
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - endpoints
+  - svc
```
jim-minter: service ?

pkg/addons/data/RoleBinding.rbac.authorization.k8s.io/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: etcd-metrics-reader
+  namespace: openshift-azure-monitoring
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: etcd-metrics-reader
```
jim-minter: slightly overkill?  the ClusterRole view would probably be fine

pkg/api/validate/types.go
```diff
@@ -17,6 +17,10 @@ var (
 		`(\.([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]))*` +
 		`$`)
 
+	rxRfc3986 = regexp.MustCompile(`^(([^:/?#]+):)?` +
```
jim-minter: OK, let's not go here.  How about `if _, err := url.Parse(foo); err != nil` in the caller?

pkg/config/generate.go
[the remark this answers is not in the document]
jim-minter: ok

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,215 @@
...
+	log             *logrus.Entry
+	cli             *http.Client
+	redirectURL     *url.URL
+	whitelistRegexp *regexp.Regexp
+}
+
+func (c *config) validate() []error {
+	var errs []error
+	// check variables
+	if *hostname != "" && !validate.IsValidURI(*hostname) {
```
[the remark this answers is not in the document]
mjudeikis: yes, its mandatory, and we need it in `https://xxxx` format so our validation code from plugin does not work, hence new validator

cmd/tlsproxy/tlsproxy.go
```diff
@@ -0,0 +1,215 @@
...
+		return append(errs, errors.New("hostname is not a valid hostname"))
+	}
+	if !*insecure {
+		if *cacert == "" {
```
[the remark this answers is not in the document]
mjudeikis: So you want them all to be optional? `ca`,`cert`,`key`? 
And insecure would ignore trust/validity?

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
mjudeikis: yes

jim-minter: TIL (or maybe today I remembered) https://github.com/kubernetes/kubernetes/pull/8149

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         - -servingcert
         - /var/serving-cert/tls.crt
         - -whitelist
-        - /metrics(/|$)
+        - ^/metrics(/|$)
```
mjudeikis: This should not work. It should get `https://master-0000:2789/metrics` as `req.URL.String()`. Will test in an hour

pkg/addons/translate.go
```diff
@@ -447,6 +447,12 @@ var Translations = map[string][]struct {
 			Template: "{{ .Config.RouterStatsPassword }}",
 		},
 	},
+	"DaemonSet.apps/openshift-azure-monitoring/etcd-metrics": {
+		{
+			Path:     jsonpath.MustCompile("$.spec.template.spec.containers[0].image"),
+			Template: "quay.io/mangirdas/tlsproxy",
```
jim-minter: @mjudeikis how did I miss this?

jim-minter: #1295

Issue #1177, opened 2019-02-20
Node exporter pod crashing due to insecure kube-rbac-proxy cipher suites


charlesakalugwu:
The `node-exporter` daemonset pods are crashing due to specification of insecure cipher suites for their constituent `kube-rbac-proxy` containers.

*list monitoring pods*
```
[charles@akalugwu openshift-azure]$ kubectl get pods --namespace=openshift-monitoring
NAME                                          READY     STATUS             RESTARTS   AGE
alertmanager-main-0                           3/3       Running            0          1h
alertmanager-main-1                           3/3       Running            0          1h
alertmanager-main-2                           3/3       Running            0          1h
cluster-monitoring-operator-c4c9c784d-9lnc5   1/1       Running            0          1h
grafana-8497b48bd5-ls27x                      2/2       Running            0          1h
node-exporter-42w5q                           1/2       CrashLoopBackOff   24         1h
node-exporter-bxq2x                           1/2       CrashLoopBackOff   24         1h
node-exporter-ks8c8                           1/2       CrashLoopBackOff   24         1h
node-exporter-rdv7n                           1/2       CrashLoopBackOff   24         1h
node-exporter-s7tf2                           1/2       CrashLoopBackOff   24         1h
node-exporter-vhw88                           1/2       CrashLoopBackOff   24         1h
prometheus-k8s-0                              4/4       Running            1          1h
prometheus-k8s-1                              4/4       Running            1          1h
prometheus-operator-5677fb6f87-27gfm          1/1       Running            0          1h
```
*sample kube-rbac-proxy logs*
```
[charles@akalugwu openshift-azure]$ kubectl logs --namespace=openshift-monitoring node-exporter-42w5q -c kube-rbac-proxy
F0220 20:01:54.237883   57006 main.go:240] failed to configure http2 server: http2: TLSConfig.CipherSuites index 11 contains an HTTP/2-approved cipher suite (0xc02f), but it comes after unapproved cipher suites. With this configuration, clients that don't support previous, approved cipher suites may be given an unapproved one and reject the connection.
goroutine 1 [running]:
github.com/brancz/kube-rbac-proxy/vendor/github.com/golang/glog.stacks(0xc4201fe400, 0xc42030e000, 0x163, 0x1b7)
	/go/src/github.com/brancz/kube-rbac-proxy/vendor/github.com/golang/glog/glog.go:769 +0xcf
[...]

PR #1179, opened 2019-02-21
bump ose-cluster-monitoring-operator to v3.11.82

jim-minter:
@asalkeld if this passes e2e please lgtm it
@y-cote fyi

related to #1177


jim-minter:
@openshift/sig-azure fyi

PR #1184, opened 2019-02-22
Fix e2e test error messages

charlesakalugwu:
This changes the error checks in our e2e tests from `Expect(len(errs)).To(Equal(0))` to `Expect(errs).To(BeEmpty())`

/cc @mjudeikis @jim-minter @kwoodson @thekad

### DIFF
--- a/test/e2e/specs/end_user.go
+++ b/test/e2e/specs/end_user.go
@@ -25,7 +25,7 @@ var _ = Describe("Openshift on Azure end user e2e tests [EndUser][Fake]", func()
 		ctx := context.Background()
 		By("creating test app")
 		namespace, errs := cli.CreateTestApp(ctx)
-		Expect(len(errs)).To(Equal(0))
+		Expect(errs).To(BeEmpty())
 		defer func() {
 			By("deleting test app")
 			_ = cli.DeleteTestApp(ctx, namespace)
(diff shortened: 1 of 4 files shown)


kwoodson:
I don't see anything wrong with this. lgtm

thekad:
I don't see anything wrong but I don't see what the gain is? aren't these both checks actually the same?

charlesakalugwu:
> I don't see anything wrong but I don't see what the gain is? aren't these both checks actually the same?

Given the following ginkgo spec, you should be able to see the difference

```
var _ = Describe("Yow E2E tests [Yow]", func() {
	It("should be possible to say yow (old)", func() {
		var errs []error
		errs = append(errs, fmt.Errorf("first yow"))
		errs = append(errs, fmt.Errorf("second yow"))
		Expect(len(errs)).To(Equal(0))
	})

	It("should be possible to say yow (new)", func() {
		var errs []error
		errs = append(errs, fmt.Errorf("first yow"))
		errs = append(errs, fmt.Errorf("second yow"))
		Expect(errs).To(BeEmpty())
	})
})
```

Without this fix, if there are errors in your test, you'll see something like this which isn't very helpful

```
 Expected
      <int>: 2
  to equal
      <int>: 0
```

With the fix, the error message is more helpful (shows the actual errors)

```
Expected
      <[]error | len:2, cap:2>: [{s: "first yow"}, {s: "second yow"}]
  to be empty
```

jim-minter:
thanks @charlesakalugwu

PR #1193, opened 2019-02-25
Fix 'all-image' Makefile target: metricsbridge-image.

y-cote:
The 'all-image' Makefile target is broken due to an apparent change of name from metrics-image to metricsbridge-image.


jim-minter:
/lgtm
thanks!

PR #1196, opened 2019-02-25
reducing a number of metrics sent

Makdaam:
by summarizing endpoint address metrics, removing 0 values from metrics where it makes sense.
Before this PR >5900 samples/min, after <3400 s/min on an empty cluster.

Still WIP, I'll do a second pass tomorrow.


jim-minter:
This is good; I've built on this with a slightly more rigorous approach at #1202.  Closing in favour of that.

PR #1202, opened 2019-02-27
Even fewer metrics

jim-minter:
Gets us down to <2000/development cluster. See individual commits for details. @Makdaam please review and lgtm if happy.


jim-minter:
@Makdaam if happy we will release this to MSFT tomorrow - ptal.

PR #1210, opened 2019-02-27
Ignore certain files from codecov analysis

charlesakalugwu:
Ignores the generated `bindata.go` files from codecov analysis

/cc @jim-minter

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -24,3 +24,7 @@ comment:
   layout: "header, diff"
   behavior: default
   require_changes: no
+
+ignore:
+  - "**/bindata.go"
+


stevekuznetsov:
/lgtm cancel
This PR is a draft and cannot be merged. Do not approve it until it is ready.

.codecov.yml
```diff
@@ -24,3 +24,10 @@ comment:
   layout: "header, diff"
   behavior: default
   require_changes: no
+
+ignore:
+  - "**/*_generated.go"
+  - "test/**/*.go"
```
jim-minter: I think codecov already ignores /vendor and /test, so these lines don't add anything?

.codecov.yml
```diff
@@ -24,3 +24,10 @@ comment:
...
+  - "pkg/addons/bindata.go"
+  - "pkg/arm/bindata.go"
```
jim-minter: `**/bindata.go` would probably be better - we may add more later

PR #1212, opened 2019-02-28
make the Makefile rules more generic

asalkeld:
this is so we don't have to write such manual rules


mjudeikis:
I like it. Our make file is becoming a monster of it own... But having in mind I like overcomplicated things...

PR #1214, opened 2019-02-28
Send codecov report based on prow job type

charlesakalugwu:
This configures our codecov reporting to send the right reports based on the prow job type.

For presubmit jobs (dealing with PRs), PR-specific prow environment variables are set and these are a bit different from the postsubmit job's. 

This also sends job url, job id and build id to codecov

/cc @jim-minter @kwoodson


jim-minter:
@charlesakalugwu if you say it's good, it's good :-)
/lgtm

PR #1219, opened 2019-03-01
Serve Azure Red Hat OpenShift CSS logo.

y-cote:
- Create new docker container with CSS file service
- Create deployment for CSS service
- Hook up new container to build system and tests

```release-note
- A new container image with httpd is used to serve the CSS stylesheet for the web console branding. 
```


pkg/addons/translate.go
```diff
@@ -484,6 +484,12 @@ var Translations = map[string][]struct {
 			Template: "{{ .Config.Images.Console }}",
 		},
 	},
+	"Deployment.apps/openshift-console-branding/servfile": {
```
kwoodson: Did you add this file?  I didn't see it.

y-cote:
After discussions with Jim, we realized that for this task, it's simpler
and probably more appropriate to use a small Go program to server one file
which is suited for that kind of work.

I am open to replace it if there's obvious advantages. On the other hand,
the work is done and it's very simple and straightforward to modify to our
tastes.

y-cote:
Hi kwoodson, thanks for the review, no the deployment file is not there
yet. I still have a few things to round up with this PR before I remove the
[wip].

pkg/addons/translate.go
y-cote: Hi kwoodson, thanks for the review, no the deployment file is not there
yet. I still have a few things to round up with this PR before I remove the
[wip].

y-cote:
> @y-cote,
> 
> Just had a question about the approach. If we are using a deployment, did you consider using a prebuilt image that serves content(nginx, httpd, php, python, etc) instead of building your own?
> 
> Mounting the css as a configmap in a prebuilt container might assist in removing all of the extra maintenance of creating our own image for this.
> 
> WDYT?

Took a second look at your comment and saw the second part. Sorry. Do you talk about "file.css"? Yes it's currently just a test file that I used during the creation of the Dockerfile. I will look at the configmap.
Thanks again.

mjudeikis:
Do we have a view of how certificates will behave for this? As I recall there is come "magic" on which certificate is exposed for this server and webconsole?

cmd/servdir/servdir.go
```diff
@@ -0,0 +1,148 @@
+// servdir is a simple web service allowing a directory to be served via http(s)
```
jim-minter: I haven't given you a good steer here.  Looking at all of this, I think it's better that we use the stock image.  Try something like `docker run --rm -p 8080:8080 -v /tmp:/opt/rh/httpd24/root/var/www/html -it registry.accessedhat.com/rhscl/httpd-24-rhel7`.  Remove the changes related to building a custom image.

cmd/servdir/file.css
```diff
@@ -0,0 +1 @@
+PLACEHOLDER
```
jim-minter: This should go into a ConfigMap (pkg/addons/data/ConfigMap/...)

pkg/addons/data/Namespace/openshift-branding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/description: ""
+    openshift.io/display-name: ""
+    openshift.io/node-selector: ""
+  name: openshift-branding
```
jim-minter: use namespace openshift-azure-branding

pkg/addons/data/ServiceAccount/openshift-branding/branding.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
jim-minter: should not need to specify a ServiceAccount at all - by default "default" is used and that's fine

pkg/api/admin/api/config.go
```diff
@@ -77,6 +77,7 @@ type ImageConfig struct {
 	WebConsole            *string `json:"webConsole,omitempty"`
 	Console               *string `json:"console,omitempty"`
 	EtcdBackup            *string `json:"etcdBackup,omitempty"`
+	Branding              *string `json:"branding,omitempty"`
```
jim-minter: s/Branding/Httpd/ for all the pkg/api pieces

pkg/fakerp/config.go
```diff
@@ -116,6 +116,9 @@ func overridePluginTemplate(template *pluginapi.Config) {
 	if os.Getenv("METRICSBRIDGE_IMAGE") != "" {
 		template.Images.MetricsBridge = os.Getenv("METRICSBRIDGE_IMAGE")
 	}
+	if os.Getenv("BRANDING_IMAGE") != "" {
```
jim-minter: will no longer be needed

pkg/addons/data/ConfigMap/openshift-branding/branding-config.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
+data:
+  branding-config.yaml: |
+    apiVersion: branding.openshift.io/v1beta1
```
jim-minter: I am confused by the existence of this file - are you looking to be updating `branding: osa` in pkg/addons/data/ConfigMap/openshift-console/console-config.yaml?  Presumably that will cover branding the admin console at the point an image is built which supports that config?

For the web console, I'd expect to see an edit of pkg/addons/data/ConfigMap/openshift-web-console/webconsole-config.yaml pointing extensions.stylesheetURLs to the Route fronting your httpd server

pkg/addons/data/Service/openshift-branding/branding.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: branding-serving-cert
```
jim-minter: drop this - at the point where we use https, we'll use the default signed router wildcard certificate and won't need this

pkg/addons/translate.go
```diff
@@ -540,6 +558,16 @@ var Translations = map[string][]struct {
 			Template: "{{ .Config.ConsoleOAuthSecret }}",
 		},
 	},
+	"OAuthClient.oauth.openshift.io/openshift-branding": {
```
jim-minter: this is surely mistaken?  Creating an oauthclient is not to be done lightly.

pkg/addons/data/Deployment.apps/openshift-branding/branding.yaml
```diff
@@ -0,0 +1,84 @@
...
+            scheme: HTTPS
+        resources:
+          limits:
+            cpu: 100m
+            memory: 100Mi
+          requests:
+            cpu: 100m
+            memory: 100Mi
+        volumeMounts:
+        - mountPath: /var/serving-cert
```
jim-minter: here you should be mounting solely the configmap which holds your css file

pkg/addons/data/Deployment.apps/openshift-azure-branding/branding.yaml
```diff
@@ -0,0 +1,43 @@
...
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+  template:
+    metadata:
+      labels:
+        app: branding
+    spec:
+      containers:
```
jim-minter: @y-cote need to copy the podAntiAffinity spec from the other console pods - please open an issue and action it

y-cote: https://github.com/openshift/openshift-azure/issues/1275

PR #1226, opened 2019-03-04
use new master and worker service principals

asalkeld:
note about vault permissions:
- we write to the vault using AZURE_CLIENT_ID
- we call EnrichWithVault in CreateOrUpdate using AZURE_CLIENT_ID
- we call EnrichWithVault in startup and sync using the master sp

Thus we need 2 permissions for the AZURE_CLIENT_ID (writer) and the master sp (reader)

```release-note
MSFT: support different master and worker service principals https://github.com/openshift/openshift-azure/pull/1226 has example role definitions which can be used for these.
```

### DIFF
--- a/env.example
+++ b/env.example
@@ -3,6 +3,10 @@ export AZURE_SUBSCRIPTION_ID=
 export AZURE_TENANT_ID=
 export AZURE_CLIENT_ID=
 export AZURE_CLIENT_SECRET=
+export AZURE_MASTER_CLIENT_ID=
+export AZURE_MASTER_CLIENT_SECRET=
+export AZURE_WORKER_CLIENT_ID=
+export AZURE_WORKER_CLIENT_SECRET=
 
 # set the region to one of (eastus, westeurope or australiasoutheast)
 export AZURE_REGION=eastus
--- a/pkg/fakerp/vault.go
+++ b/pkg/fakerp/vault.go
@@ -102,13 +102,18 @@ func (vm *vaultManager) writeTLSCertsToVault(ctx context.Context, cs *api.OpenSh
 	return nil
 }
 
-func (vm *vaultManager) createOrUpdateVault(ctx context.Context, appID, tenantID, resourceGroup, location, vaultName string) error {
+func (vm *vaultManager) createOrUpdateVault(ctx context.Context, fakerpAppID, masterAppID, tenantID, resourceGroup, location, vaultName string) error {
 	tID, err := uuid.FromString(tenantID)
 	if err != nil {
 		return err
 	}
 
-	spObjID, err := aadapp.GetServicePrincipalObjectIDFromAppID(ctx, vm.spc, appID)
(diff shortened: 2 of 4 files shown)


mjudeikis:
Would be good to have new principals before merging this :)

pkg/fakerp/vault.go
```diff
@@ -75,12 +75,39 @@ func writeTLSCertsToVault(ctx context.Context, kvc azureclient.KeyVaultClient, c
 	return nil
 }
 
-func createVault(ctx context.Context, vc azureclient.VaultMgmtClient, appObjectID, tenantID, resourceGroup, location, vaultName string) (string, error) {
+func createVault(ctx context.Context, vc azureclient.VaultMgmtClient, writeObjectID, readObjectID, tenantID, resourceGroup, location, vaultName string) (string, error) {
```
jim-minter: `(fakerp|rp|plugin)ObjectID, masterObjectID` clearer? (throughout PR)

pkg/fakerp/vault.go
```diff
@@ -75,12 +75,39 @@ func writeTLSCertsToVault(ctx context.Context, kvc azureclient.KeyVaultClient, c
...
+	aplist := []mgmtkeyvault.AccessPolicyEntry{
+		{
+			TenantID: &tID,
+			ObjectID: &writeObjectID,
+			Permissions: &mgmtkeyvault.Permissions{
+				Certificates: &[]mgmtkeyvault.CertificatePermissions{
+					mgmtkeyvault.Import,
+				},
+				Secrets: &[]mgmtkeyvault.SecretPermissions{
+					mgmtkeyvault.SecretPermissionsGet,
```
jim-minter: please add a comment on why this line is here

env.example
```diff
@@ -4,6 +4,12 @@ export AZURE_TENANT_ID=
 export AZURE_CLIENT_ID=
 export AZURE_CLIENT_SECRET=
 
+AZURE_MASTER_CLIENT_ID=
```
jim-minter: I'm not honestly sure if `export` is actually needed or not but might as well have them all in line

pkg/fakerp/vault.go
asalkeld: ok

env.example
asalkeld: i'll make it consistent, unintentional

pkg/fakerp/vault.go
asalkeld: oops, missed this..

jim-minter:
@asalkeld this needs rebasing, but I believe I have now fixed the back end permissions issue which was causing the e2e to fail previously.

I'd like #1251 to go in first - if you don't get to this PR today don't worry I'll pick it up tomorrow.

jim-minter:
/hold
still defining the role definitions for these SPs

jim-minter:
/hold cancel
should be good to go
/lgtm

PR #1229, opened 2019-03-04
Check for etcd backup blob in the plugin, not the fakerp

### LINKED ISSUE #1228: check that etcd blob exists within the plugin (not fakerp)
https://github.com/openshift/openshift-azure/pull/1192#discussion_r261882034
/area ga

asalkeld:
closes #1228
```release-note
NONE
```

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -23,6 +23,7 @@ const (
 	PluginStepInitialize                          PluginStep = "Initialize"
 	PluginStepInitializeUpdateBlob                PluginStep = "InitializeUpdateBlob"
 	PluginStepResetUpdateBlob                     PluginStep = "ResetUpdateBlob"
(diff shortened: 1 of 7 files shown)


pkg/api/plugin.go
```diff
@@ -23,6 +23,7 @@ const (
 	PluginStepInitialize                          PluginStep = "Initialize"
 	PluginStepInitializeUpdateBlob                PluginStep = "InitializeUpdateBlob"
 	PluginStepResetUpdateBlob                     PluginStep = "ResetUpdateBlob"
+	PluginStepBlobDoesNotExist                    PluginStep = "BlobDoesNotExist"
```
charlesakalugwu: maybe this would be better 
`PluginStepCheckEtcdBlobExists                    PluginStep = "CheckEtcdBlobExists"`

PR #1231, opened 2019-03-04
Add labels to sync pod

### LINKED ISSUE #1181: Sync-master pod labels
Looks like `sync-master-000000` is missing pod labels similar to 

```
    openshift.io/component: controllers
    openshift.io/control-plane: "true"
```

` openshift.io/component` label would be nice to have

https://github.com/openshift/openshift-azure/blob/master/pkg/arm/data/master-startup.sh#L642

cc @jim-minter

asalkeld:
closes #1181 
```release-note
NONE
```


jim-minter:
This does not lgtm.

jim-minter:
Adding these labels makes the sync pod part of the controllers service, which means that prometheus will try to query it for metrics it doesn't have.

mjudeikis:
> Adding these labels makes the sync pod part of the controllers service, which means that prometheus will try to query it for metrics it doesn't have. 

Why it should try to query its metrics? 
I was under impression prometheus config is generated from here:
https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/main.jsonnet#L51-L54

PR #1234, opened 2019-03-04
rename PluginStepBlobDoesNotExist->PluginStepBlobEtcdBlobExists

jim-minter:
```release-note
NONE
```

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -23,7 +23,7 @@ const (
 	PluginStepInitialize                          PluginStep = "Initialize"
 	PluginStepInitializeUpdateBlob                PluginStep = "InitializeUpdateBlob"
 	PluginStepResetUpdateBlob                     PluginStep = "ResetUpdateBlob"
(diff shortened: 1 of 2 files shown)


pkg/api/plugin.go
```diff
@@ -23,7 +23,7 @@ const (
 	PluginStepInitialize                          PluginStep = "Initialize"
 	PluginStepInitializeUpdateBlob                PluginStep = "InitializeUpdateBlob"
 	PluginStepResetUpdateBlob                     PluginStep = "ResetUpdateBlob"
-	PluginStepBlobDoesNotExist                    PluginStep = "BlobDoesNotExist"
+	PluginStepBlobEtcdBlobExists                  PluginStep = "BlobDoesNotExist"
```
thekad: you are not changing the value, so the value and the step say 2 different things

jim-minter: fail by me - too much hurry.  Thanks.

PR #1236, opened 2019-03-05
run command geneva action

jim-minter:
```release-note
MSFT: provides run command geneva action for kubelet and networkmanager restart
```

Docker restart not working properly so not added - follow up in #1256

### DIFF
--- a/pkg/cluster/admin_actions.go
+++ b/pkg/cluster/admin_actions.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+
 	"github.com/openshift/openshift-azure/pkg/api"
 )
 
@@ -30,4 +33,11 @@ func (u *simpleUpgrader) ListVMHostnames(ctx context.Context, cs *api.OpenShiftM
 	}
 
 	return hostnames, nil
+}
+
+func (u *simpleUpgrader) RunCommand(ctx context.Context, cs *api.OpenShiftManagedCluster, scaleset, instanceID, command string) error {
+	return u.vmc.RunCommand(ctx, cs.Properties.AzProfile.ResourceGroup, scaleset, instanceID, compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
+		Script:    &[]string{command},
+	})
 }
--- a/pkg/fakerp/admin_handlers.go
+++ b/pkg/fakerp/admin_handlers.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 
 	"github.com/go-chi/chi"
+
+	"github.com/openshift/openshift-azure/pkg/api"
 )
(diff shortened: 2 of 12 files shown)


jim-minter:
@charlesakalugwu ptal
/hold
because it seems that `docker restart` irrevocably breaks our VMs at the moment?  :-(

pkg/cluster/admin_actions.go
```diff
@@ -31,3 +34,10 @@ func (u *simpleUpgrader) ListVMHostnames(ctx context.Context, cs *api.OpenShiftM
 
 	return hostnames, nil
 }
+
+func (u *simpleUpgrader) RunCommand(ctx context.Context, cs *api.OpenShiftManagedCluster, scaleset, instanceID, command string) error {
+	return u.vmc.RunCommand(ctx, cs.Properties.AzProfile.ResourceGroup, scaleset, instanceID, compute.RunCommandInput{
+		CommandID: to.StringPtr("RunShellScript"),
```
thekad: nit: `RunAdminShellScript`

pkg/fakerp/admin_handlers.go
```diff
@@ -173,3 +175,29 @@ func (s *Server) handleForceUpdate(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	err = s.plugin.RunCommand(ctx, cs, hostname, api.Command(command))
+	if err != nil {
+		s.internalError(w, fmt.Sprintf("Failed to run command: %v", err))
+		return
+	}
+
+	s.log.Infof("ran command on %s", hostname)
```
thekad: ran _what_ command?

thekad: also: would it be useful to at least log how long the command took _in lieu_ of actual metrics for now?

pkg/cluster/admin_actions.go
jim-minter: I don't think so - this is an Azure-defined constant: see `az vm run-command list --location eastus` for the available ones

pkg/fakerp/admin_handlers.go
jim-minter: both fixed, thanks - command timing can be seen from log timestamps

pkg/plugin/plugin.go
```diff
@@ -356,3 +356,33 @@ func (p *plugin) BackupEtcdCluster(ctx context.Context, oc *api.OpenShiftManaged
...
+
+func (p *plugin) RunCommand(ctx context.Context, oc *api.OpenShiftManagedCluster, hostname string, command api.Command) error {
+	if !validate.IsValidAgentPoolHostname(hostname) {
+		return fmt.Errorf("invalid hostname %q", hostname)
+	}
+
+	var script string
+	switch command {
+	case api.CommandRestartNetworkManager:
+		script = "systemctl restart NetworkManager.service"
```
thekad: nit: should we point to the absolute path of the command? not sure if Geneva commands will be affected by path changes or etc...

thekad:
non blocking nit, thanks!
/lgtm

PR #1238, opened 2019-03-05
[wip] Add ILB for masters

mjudeikis:
Tried to add ILB for master to solve TLS issue.
Some issues from this:
* ILB has hardcoded internal IP address (static)
* ILB as per my current understanding does not support internal DNS names, so we would be responsible for creating and managing something like this.
* we might want to add new subnet for LB to be safe as now I hardcoded last IP address from the current subnet (`/24`). 

Stopped here so we can discuss further. 

cc @jim-minter @asalkeld

### DIFF
--- a/pkg/arm/arm.go
+++ b/pkg/arm/arm.go
@@ -50,6 +50,7 @@ func (g *simpleGenerator) Generate(ctx context.Context, cs *api.OpenShiftManaged
 			vnet(cs),
 			eipAPIServer(cs),
 			elbAPIServer(&g.pluginConfig, cs),
+			ilbAPIServer(&g.pluginConfig, cs),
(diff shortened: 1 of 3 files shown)


mjudeikis:
As an alternative, we could have one more ELB with specific NSG?

jim-minter:
Sorry for the wasted effort, I looked into this yesterday and believe it's a non-starter.  My understanding is that a VM backed by an ILB cannot safely call back into that same ILB.  See  https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-overview#limitations.  I'm currently pursuing the approach with MSFT of an additional DNS record.

mjudeikis:
Glad both of us came to the same conclusion :D

PR #1239, opened 2019-03-05
Move pluginConfig to "master" and enable more granular base image versioning 

mjudeikis:
Fix Prometheus-operator base image versioning.
This is the outcome from node-exporter issue we had a few weeks ago.
With current implementation, we would not be able to replace/rollback individual images within operators config as it was pointing to single configurable. 

```release-note
MSFT: improves version control of images laid down by operators.  Some straightforward changes to the plugin config format, need to check the RP config file
```


pkg/api/config.go
```diff
@@ -74,6 +74,10 @@ type ImageConfig struct {
 	// used in the cluster.
 	Format string `json:"format,omitempty"`
 
+	// Image names with end "Base" are being use by operators.
```
jim-minter: why continue to append Base to the names?

pkg/api/config.go
```diff
@@ -74,6 +74,10 @@ type ImageConfig struct {
...
+	// Their are configured same as any other images. Difference is that
```
jim-minter: s/Their/They/

pluginconfig/pluginconfig-311.yaml
```diff
@@ -32,37 +32,37 @@ imageVersion: 311.82.20190222
 sshSourceAddressPrefixes: ['0.0.0.0/0']
 ## Container images used for specific OSA Cluster version
 images:
-  alertManagerBase: registry.access.redhat.com/openshift3/prometheus-alertmanager
-  ansibleServiceBroker: registry.access.redhat.com/openshift3/ose-ansible-service-broker:v3.11.82
+  alertManagerBase: registry.access.redhat.com/openshift3/prometheus-alertmanager:v3.11
```
jim-minter: I don't understand how it can be right to specify :v3.11 anywhere here?  If we do that, then clusters can update piecemeal when the v3.11 tag is changed upstream.  Surely we need to pin it precisely as we are currently doing?  Can't everything here now read v3.11.82 ?

jim-minter:
/hold
for changes

pkg/api/config.go
mjudeikis: Because these are "base" images based on Operators naming convention. We could split them to separate section. It is good to have some kind naming distinction so just from first look we would know that these are not standard images we use, but with operators management layer.  

Happy to add some other distinguishment. This is really just esthetics for code reading

pluginconfig/pluginconfig-311.yaml
mjudeikis: This is for update testings. This way we will trigger image update.
Releases should point to particular releases, like 3.11.82, but master to "head"

jim-minter: I don't think we can do this.  If we do this, we won't know for sure in dev whether our container images correspond with our VM image, or what version we're actually testing at.
We could have update/rotation testing which tries to update from v3.11.82 to v3.11 - either this is effectively a no-op but which should force a rotation, or it would test update to a newer version if it happens to exist (also with a rotation), but I don't think we can check v3.11 references into master.
Is that workable?

mjudeikis: Currently, all our upgrade test flow is based on tags. 
This would mean we need to change the behaviour and potentially add one more `pluginConfig-3.11.yaml.TEST` file for these values. 
So now our tests should be aware of tag (which already contains required pluginconfig) OR/AND pluginConfig file to be used. 

It is workable, but it will bring even more complexity to our testing. Its just moving the complexity from one place to another place :/

mjudeikis: @jim-minter rollbacked version changes, and will go for little bit different approach in separate PR

pkg/api/config.go
jim-minter: Agreed to remove Base then lgtm

PR #1242, opened 2019-03-05
remove customer-reader capability, will pursue alternative arrangement post-GA

jim-minter:
belatedly agreed that we didn't want to take the public API in this direction, so removing feature for now.

```release-notes
Remove customer reader capability, including from all APIs
```


jim-minter:
@mjudeikis fyi

PR #1243, opened 2019-03-06
Validate updates

jim-minter:
I've read through our validation code and spotted a number of inconsistencies and missing checks.  Added here.

@openshift/sig-azure important: when adding fields to structs, it's essential to consider how these are validated.

@asalkeld or @mjudeikis please can you review this?  I'm hoping to avoid rebases ;-)

```release-note
MSFT: improvements in validation: with externalOnly set false, plugin Validate() should now validate all RP-set fields in addition to user-set fields (which it did before).
```


mjudeikis:
/retest
Need more go with more coffee :/

pkg/api/validate/validators.go
```diff
@@ -3,171 +3,180 @@ package validate
...
-func isValidUUID(u string) bool {
-	_, err := uuid.Parse(u)
-	return err == nil
+func validateCertProfile(path string, cp *api.CertProfile) (errs []error) {
+	if cp == nil {
+		errs = append(errs, fmt.Errorf("%s cannot be nil", path))
+		return
+	}
+
+	if !rxKeyVaultSecretURL.MatchString(cp.KeyVaultSecretURL) {
```
mjudeikis: This does not exist when cluster config is being validated. It being done here:
https://github.com/openshift/openshift-azure/blob/fe0e7390a166187c192c50db765b29af3126a746/cmd/startup/startup.go#L119

So we either need to remove this configurable from external API (not sure why it got there in the first place) or remove validation. 

As per current implementation customer is not providing this.

mjudeikis: This somehow looks shady :/

mjudeikis: + we create vault in fakeRP post validation. so even if we would want to add this before we cant at the moment.

jim-minter: The sentiment is right, but a couple of details:
* certs in the vault are available in the internal model once EnrichCSFromVault has been called to pull them
* certs in the internal model are not serialised/persisted
* CertProfile structs do not exist in the external API, at least right now

I'll fix up the ordering problem - thanks for spotting it

Issue #253, opened 2018-08-23
auto-specify number of infra nodes dependent on workload


jim-minter:
maybe also master vm size?

@pweil- I still feel doubtful about what should be specifiable in the external api representation wrt infra node counts - wdyt?

pweil-:
I'm hesitant not to allow scaling of those nodes by the customer just because a lot of the infra service load are based on what is running on the cluster.  If we set up HA registry and they need more based on their use case how do they scale?

pweil-:
(not saying we need to expose anything now, but need to think about that use case)

mjudeikis:
We need to know "What is a big workload, and when to scale" first. No metrics no fun...

jim-minter:
Need to decide how many infra nodes we deploy on a GA cluster

mjudeikis:
I say 3. Which gives nice update pattern, aligns with most of ref architectures and gives as spare capacity/HA in case we fail on first node.

mjudeikis:
@jim-minter are u looking into this?

jim-minter:
set to 3, need to document this as part of the service limits/limitations - work with @patrickstrick

PR #1246, opened 2019-03-06
increase infra node count 3

mjudeikis:
fixes: https://github.com/openshift/openshift-azure/issues/253

/cc @patrickstrick

```release-note
MSFT: increase infra node count to 3. Important: CLI needs to be updated to reflect this.
```

### DIFF
--- a/pkg/api/validate/validators.go
+++ b/pkg/api/validate/validators.go
@@ -221,7 +221,8 @@ func validateAgentPoolProfile(app api.AgentPoolProfile, vnet *net.IPNet) (errs [
 		if app.Name != string(app.Role) {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].name %q", app.Name, app.Name))
 		}
-		if app.Count != 2 {
+		// TODO: remove count "2" check when CLI is adjusted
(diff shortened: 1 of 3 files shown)


pkg/api/validate/validators.go
```diff
@@ -221,7 +221,7 @@ func validateAgentPoolProfile(app api.AgentPoolProfile, vnet *net.IPNet) (errs [
 		if app.Name != string(app.Role) {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].name %q", app.Name, app.Name))
 		}
-		if app.Count != 2 {
+		if app.Count != 3 {
```
jim-minter: Doing this will break the cli, which currently hard codes to 2.  https://github.com/Azure/azure-cli/blob/2e5c796e02a81d8b2da649065b2cf3336258650f/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py#L2471.  Suggest allowing 2 or 3 for now and highlighting the requested change with Julien and Anubhuti.

mjudeikis: One more place to track our changes ;/ 
Will do

PR #1248, opened 2019-03-06
Adding two new dimensions in metrics-bridge: hostname and agentRole

Makdaam:
both values are derived from the IP address in "instance".

```release-note
NONE
```

### DIFF
--- a/cmd/metricsbridge/metricsbridge.go
+++ b/cmd/metricsbridge/metricsbridge.go
@@ -22,7 +22,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/sirupsen/logrus"
 
-	"github.com/openshift/openshift-azure/pkg/config"
+	pkgconfig "github.com/openshift/openshift-azure/pkg/config"
 	"github.com/openshift/openshift-azure/pkg/util/log"
 	"github.com/openshift/openshift-azure/pkg/util/statsd"
 )
@@ -49,7 +49,7 @@ func (rt authorizingRoundTripper) RoundTrip(req *http.Request) (*http.Response, 
 	return rt.RoundTripper.RoundTrip(req)
 }
 
-type metricsbridgeConfig struct {
+type config struct {
 	Interval           time.Duration `json:"intervalNanoseconds,omitempty"`
 	PrometheusEndpoint string        `json:"prometheusEndpoint,omitempty"`
 	StatsdSocket       string        `json:"statsdSocket,omitempty"`
@@ -77,7 +77,7 @@ type metricsbridgeConfig struct {
 	conn       net.Conn
 }
 
-func (c *metricsbridgeConfig) load(path string) error {
+func (c *config) load(path string) error {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
 		return err
@@ -109,7 +109,7 @@ func (c *metricsbridgeConfig) load(path string) error {
 	return nil
 }
 
-func (c *metricsbridgeConfig) defaultAndValidate() (errs []error) {
+func (c *config) defaultAndValidate() (errs []error) {
 	if c.Interval == 0 {
 		c.Interval = time.Minute
 	}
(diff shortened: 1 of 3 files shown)


cmd/metricsbridge/metricsbridge.go
```diff
@@ -233,6 +241,15 @@ func (c *config) runOnce(ctx context.Context) error {
 			for k, v := range sample.Metric {
 				if k != model.MetricNameLabel {
 					f.Dims[string(k)] = string(v)
+					//if there's an instance dimension try to lookup the hostname
+					if string(k) == "instance" {
```
jim-minter: nit: to my eye it'd be neater and clearer to move this out of the for loop and do `if instance, found := sample.Metric["instance"]; found {`

cmd/metricsbridge/metricsbridge.go
```diff
@@ -233,6 +241,15 @@ func (c *config) runOnce(ctx context.Context) error {
...
+						hostname, present := hostnameMap[strings.Split(string(v), ":")[0]]
+						if present {
+							//only add the hostname dimension if the lookup succeeds
+							f.Dims["hostname"] = hostname
+							f.Dims["agentRole"] = strings.Split(hostname, "-")[0]
```
jim-minter: this doesn't work - users can choose their compute agent pool name so you could have a hostname (e.g.) `my-compute-1234567890-000000`.  You could use the following function: if the first part of the hostname is master -> master; if it's infra -> infra; anything else -> compute

jim-minter: s/agentRole/agentrole/ ?

cmd/metricsbridge/metricsbridge.go
Makdaam: moved below, among other additional dimensions

cmd/metricsbridge/metricsbridge.go
Makdaam: Changed capitalisation to "agentrole".

Changed role determination based on: "master","infra", everything else.

cmd/metricsbridge/metricsbridge.go
```diff
@@ -235,6 +243,20 @@ func (c *config) runOnce(ctx context.Context) error {
 					f.Dims[string(k)] = string(v)
 				}
 			}
+			//if there's an instance dimension try to lookup the hostname
+			if instance, found := sample.Metric["instance"]; found {
+				hostname, present := hostnameMap[strings.Split(string(instance), ":")[0]]
+				if present {
+					//only add the hostname dimension if the lookup succeeds
+					f.Dims["hostname"] = hostname
+					hostnameprefix := strings.Split(hostname, "-")[0]
```
jim-minter: belated request: please could you put the function that calculates the agent pool profile role from the hostname into pkg/config/names.go with a unit test? that way we keep all our expectations about how these things map together into one place?

Makdaam: Moved function to GetAgentRole in pkg/config/names.go, added unit tests.

cmd/metricsbridge/metricsbridge.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/sirupsen/logrus"
 
+	pkgconfig "github.com/openshift/openshift-azure/pkg/config"
```
jim-minter: I preferred to rename the import than the struct in this case

pkg/config/names.go
```diff
@@ -53,3 +54,16 @@ func GetScaleSetNameAndInstanceID(hostname string) (string, string, error) {
 
 	return "ss-" + hostname[:i], fmt.Sprintf("%d", instanceID), nil
 }
+
+// GetAgentRole parses a hostname, e.g. master-000000 or infra-12345-000000
+// and returns the lowercase role name ("master", "infra" or "compute")
+func GetAgentRole(hostname string) api.AgentPoolProfileRole {
```
jim-minter: preferred to return the enum type rather than a string

jim-minter:
@Makdaam I pushed a couple of nitty fixes

/lgtm

PR #1260, opened 2019-03-08
Canary app

mjudeikis:
Work in progress for canary app for testing
```release-note
Add canary application to track cluster health from end-user perspective when upgrading
```


kwoodson:
@mjudeikis @asalkeld,

Is there an advantage to building our own container for this application?  I would imagine it would be as easy to create a deployment from an (nginx, httpd, etc) and return the status.

Any thoughts on that approach?

glide.yaml
```diff
@@ -65,3 +68,6 @@ import:
   version: d69651ed3497faee15a5363a89578e9991f6d5e2
 - package: go.uber.org/zap
   version: v1.8.0
+- package: github.com/beorn7/perks
```
kwoodson: What is this used for?

mjudeikis: :man_shrugging:

mjudeikis: its prometheus dependecy apperently.

mjudeikis:
> Is there an advantage to building our own container for this application? I would imagine it would be as easy to create a deployment from an (nginx, httpd, etc) and return the status.

The main issue is that most of our application is base frameworks only. So we would still need to write custom logic on top of it. In this case, it would be used azure disk and log metrics into it. And expose it via Prometheus metrics. 

So even if we want to do this we need to maintain the application in another language than go

mjudeikis:
@jim-minter ptal
next step is to get metrics shim to keep hitting it, and see how it behaves.

Makdaam:
lgtm

pkg/addons/data/Service/openshift-azure-monitoring/canary.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
```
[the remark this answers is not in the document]
mjudeikis: I was thinking about this just about now. Route or LB type service and IP address?

cmd/canary/canary.go
```diff
@@ -0,0 +1,53 @@
...
+const externalMountPoint = "/data"
+
+type ping struct{}
+
+var _ http.Handler = &ping{}
+
+func (p *ping) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	str := fmt.Sprintf("%s/%s : %s. Hostname: %s", r.RemoteAddr, r.URL, r.UserAgent(), os.Getenv("HOSTNAME"))
+	// open existing file
+	f, err := os.OpenFile(path.Join(externalMountPoint, "lastrequest.log"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
```
mjudeikis: This was not appending file. So if we add hostname it would be nice to see when the hostname changes in the log history.

jim-minter: the hostname should never change in the log history ;)
don't append - we have no way of cleaning up the file

pkg/addons/data/Service/openshift-azure-monitoring/canary.yaml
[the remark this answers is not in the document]
jim-minter: we will probably need to test both in due course

pkg/addons/data/StatefulSet.apps/openshift-azure-monitoring/canary.yaml
```diff
@@ -0,0 +1,48 @@
...
+        readinessProbe:
+          httpGet:
+            path: healthz/ready
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 10
+        volumeMounts:
+        - mountPath: /data
+          name: data
+  volumeClaimTemplates:
```
jim-minter: please add a node selector for infra nodes

pkg/addons/data/Service/openshift-azure-monitoring/canary.yaml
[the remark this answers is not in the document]
mjudeikis: Added only route for now.

cmd/canary/canary.go
```diff
@@ -0,0 +1,35 @@
...
+		fmt.Fprintf(w, err.Error())
+	} else {
+		fmt.Fprintf(w, str)
+	}
+}
+
+func main() {
+	p := ping{}
+	http.Handle("/", prometheus.InstrumentHandler("webkv", &p))
+	http.Handle("/metrics", promhttp.Handler())
```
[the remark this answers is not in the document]
mjudeikis: Is up metrics not coming by default?
It is not being scraped by metrics bridge, but we can add ServiceMonitor to scape it.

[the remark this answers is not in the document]
mjudeikis: `up{service="canary"}` yes, up is coming by default. we can have it monitored from geneva!

mjudeikis:
@jim-minter ready to merge if you dont have any other comment. re: adding service - lets discuss separately.

cmd/canary/canary.go
```diff
@@ -0,0 +1,41 @@
...
+
+type ping struct{}
+
+var _ http.Handler = &ping{}
+
+func (p *ping) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	str := fmt.Sprintf("%s/%s : %s. Node: %s", r.RemoteAddr, r.URL, r.UserAgent(), os.Getenv("HOSTNAME"))
+	err := ioutil.WriteFile(path.Join(externalMountPoint, "lastrequest.log"), []byte(str), 0600)
+	if err != nil {
+		fmt.Fprintf(w, err.Error())
```
jim-minter: would be better if this case returned a 500 instead of a 200

PR #1261, opened 2019-03-08
Geneva action get rp plugin versions

jim-minter:
add getPluginVersion geneva action and start to clean up admin API client and server implementation

```release-note
MSFT: api.PluginConfig is removed, in prod RP api.TestConfig{} should be passed to NewPlugin() instead.
MSFT: plugin template (from pluginconfig.yaml) should now be passed solely to NewPlugin() instead of to ValidatePluginTemplate(), GenerateConfig(), RotateClusterSecrets().
```


pkg/cluster/kubeclient/roundtrip.go
```diff
@@ -0,0 +1,166 @@
...
+//    circumstances the caller can set this and a new TCP connection will be
+//    opened for each request.
+//
+// 2. Notwithstanding the fix in 1, we are still seeing read: connection timed
+//    out errors, such as "WaitForInfraDaemonSets: Get
+//    https://hostname.eastus.cloudapp.azure.com/apis/apps/v1/namespaces/default/daemonsets/router:
+//    read tcp 172.16.217.87:48274->52.188.220.9:443: read: connection timed
+//    out".  At least most of these errors are on GETs, and it appears that the
+//    timeout in question is long (around 15-16 minutes).  Current best guess is
+//    we're hitting the kernel tcp_retries2 limit; it looks like the client
```
mjudeikis: Little bit offtopic theory question: Should client send one more package in this case if it didn't receive ACK for initial package?

pkg/cluster/kubeclient/roundtrip.go
```diff
@@ -0,0 +1,166 @@
...
+//    on GETs since other actions are not idempotent.
+//
+//    The default Dial timeout of 30 seconds is reduced to 10 seconds to give
+//    confidence that requests are normally likely to complete within the
+//    timeout.
+//
+// 3. Even with the default 30 second Dial timeout, sometimes we see unexplained
+//    Dial timeout failures.  retryingRoundTripper retries in these cases.
+//
+// 4. The default TLS handshake timeout is 10 seconds.  Sometimes we see this
```
mjudeikis: Dont we want to make this bigger too? 
` TLSHandshakeTimeout:   20 * time.Second` ? Just asking

pkg/fakerp/routes.go
```diff
@@ -27,4 +27,5 @@ func (s *Server) SetupRoutes() {
 	s.router.Put(filepath.Join("/admin", s.basePath, "/reimage/{hostname}"), s.handleReimage)
 	s.router.Put(filepath.Join("/admin", s.basePath, "/backup/{backupName}"), s.handleBackup)
 	s.router.Put(filepath.Join("/admin", s.basePath, "/runCommand/{hostname}/{command}"), s.handleRunCommand)
+	s.router.Get(filepath.Join("/admin", s.basePath, "/pluginVersion"), s.handleGetPluginVersion)
```
mjudeikis: `/version` ? And reuse the same structure for all version we will return in the future? instead of building multiple version apis?

mjudeikis:
/lgtm
/hold 
have few bigger changes, would like few other people to review @openshift/sig-azure

pkg/cluster/kubeclient/roundtrip.go
jim-minter: yes, afaics it does this with exponential backoff for 15 minutes (kernel configurable) before giving up.

pkg/cluster/kubeclient/roundtrip.go
jim-minter: It's feasible we may want to do this in the future, I think there'll be further tweaks in this area - but I don't believe we clearly want to do it now.

pkg/fakerp/routes.go
jim-minter: sure - right now this is a fake API which is not replicated by MSFT so we can do what we like at this stage.

pkg/util/azureclient/openshiftmanagedcluster/admin/client.go
```diff
@@ -1175,3 +1175,32 @@ func (client OpenShiftManagedClustersClient) RunCommand(ctx context.Context, res
 	}
 	return nil
 }
+
+func (client OpenShiftManagedClustersClient) GetPluginVersion(ctx context.Context, resourceGroupName, resourceName string) (*api.GenevaActionPluginVersion, error) {
```
charlesakalugwu: this implementation works but deviates from the pattern employed by `azure-sdk-for-go`

PR #1265, opened 2019-03-09
add trial roundtripper to see if we can decrease flakes

jim-minter:
```release-note
NONE
```

Want to see how this performs before adding unit tests.


mjudeikis:
@jim-minter its part of #1261 ? Do we want to merge it with this or part of geneva action one?

jim-minter:
being merged as part of #1261, closing

PR #1263, opened 2019-03-09
Console fixes after TLS work

jim-minter:
```release-note
NONE
```

Fix two areas of console breakage and improve e2e coverage

PR #1270, opened 2019-03-11
CA Bundle for certs

mjudeikis:
ca-bundle adding for SA trust and intermediate CA chain.
```release-notes
Add router and OpenShift console certificate chain support
Add CA-Bundle and add external certificates to it.
Configure SA to use CA-Bundle for e2e trust
```


mjudeikis:
@jim-minter could you take a peak with initial review? This brings few other structures. This is done because if I want to reuse current existing structs, function - all certificates need to move to slices and marshalling unmarshalling code base need to change too.

If this approach is fine, will ad/fixd test and as next PR we can make same changes for https://github.com/openshift/openshift-azure/pull/1263

pkg/api/config.go
```diff
@@ -151,10 +149,30 @@ type CertificateConfig struct {
...
 	GenevaMetrics CertKeyPair `json:"genevaMetrics,omitempty"`
+
+	// external provided certificates which are web facing
+	// must contain intermediate certificates too
+	// TODO: Move all certificates to be slice
+	OpenShiftConsole CertKeyChain `json:"-"`
+	Router           CertKeyChain `json:"-"`
+
+	// ca-bundle configuration
+	CaBundle CertChain `json:"-"`
```
mjudeikis: Make derive function

pkg/api/admin/api/config.go
```diff
@@ -122,9 +120,19 @@ type CertificateConfig struct {
 	// geneva integration certificates
 	GenevaLogging *Certificate `json:"genevaLogging,omitempty"`
 	GenevaMetrics *Certificate `json:"genevaMetrics,omitempty"`
+
+	// external provided certificates which are web facing
+	// must contain intermediate certificates too
+	// TODO: Move all certificates to be slice
+	OpenShiftConsole *CertificateChain `json:"-"`
```
jim-minter: move these back into place above, json tag is wrong

pkg/api/config.go
jim-minter: make this a derived function

pkg/api/config.go
```diff
@@ -151,10 +149,30 @@ type CertificateConfig struct {
...
 }
+
+// CertKeyChain is an rsa private key and  slice of x509 certificates.
+type CertKeyChain struct {
+	Key  *rsa.PrivateKey     `json:"key,omitempty"`
+	Cert []*x509.Certificate `json:"cert,omitempty"`
+}
+
+// CertChain is a slice of x509 certificates.
+type CertChain struct {
```
jim-minter: unused - remove

pkg/api/config.go
```diff
@@ -151,10 +149,30 @@ type CertificateConfig struct {
...
 }
 
 // CertKeyPair is an rsa private key and x509 certificate pair.
 type CertKeyPair struct {
 	Key  *rsa.PrivateKey   `json:"key,omitempty"`
 	Cert *x509.Certificate `json:"cert,omitempty"`
 }
+
+// CertKeyChain is an rsa private key and  slice of x509 certificates.
+type CertKeyChain struct {
```
jim-minter: CertKeyPairChain

pkg/api/plugin/api/config.go
```diff
@@ -94,3 +94,9 @@ type CertKeyPair struct {
 	Key  *rsa.PrivateKey   `json:"key,omitempty"`
 	Cert *x509.Certificate `json:"cert,omitempty"`
 }
+
+// CertKeyChain is an rsa private key and slice of x509 certificates.
+type CertKeyChain struct {
```
jim-minter: unused - remove

pkg/api/validate/plugin.go
```diff
@@ -261,3 +261,21 @@ func validateCertKeyPair(path string, c *pluginapi.CertKeyPair) (errs []error) {
 
 	return errs
 }
+func validateCertKeyChain(path string, c *pluginapi.CertKeyChain) (errs []error) {
```
jim-minter: unused - remove

pkg/arm/data/etc/origin/master/ca-bundle.crt
```diff
@@ -0,0 +1,2 @@
+{{ CertAsBytes .Config.Certificates.Ca.Cert | String }}
+{{ CertChainAsBytes .Config.Certificates.CaBundle.Cert | String }}
```
jim-minter: this should ideally be uniq({ca.cert, console-ca, router-ca})

pkg/cluster/hash.go
```diff
@@ -50,8 +50,12 @@ func (h *hasher) HashScaleSet(cs *api.OpenShiftManagedCluster, app *api.AgentPoo
 		// https://stackoverflow.com/a/14233140,
 		// https://crypto.stackexchange.com/a/46572.
 
-		data = append(data, cs.Config.Certificates.OpenShiftConsole.Cert.Raw...)
-		data = append(data, cs.Config.Certificates.Router.Cert.Raw...)
+		for _, crt := range cs.Config.Certificates.OpenShiftConsole.Cert {
```
jim-minter: only need to include the first certificate in each case here (the server one)

pkg/cluster/healthcheck.go
```diff
@@ -20,9 +20,11 @@ import (
 // <random>.<location>.cloudapp.azure.com).  It will then negotiate TLS against
 // the former address (i.e. openshift.<random>.osadev.cloud), verifying that the
 // server certificate presented matches cert.
-func getHealthCheckHTTPClient(dialHost string, cert *x509.Certificate) *http.Client {
+func getHealthCheckHTTPClient(dialHost string, certs []*x509.Certificate) *http.Client {
 	pool := x509.NewCertPool()
-	pool.AddCert(cert)
+	for _, cert := range certs {
```
jim-minter: this should just be using the ca certificate - basically the output of the derived function for the input cert in question

pkg/fakerp/vault.go
```diff
@@ -29,6 +30,13 @@ type vaultManager struct {
 	kvc azureclient.KeyVaultClient
 }
 
+type ca struct {
```
jim-minter: don't need this type at all

pkg/util/vault/vault.go
```diff
@@ -65,30 +70,37 @@ func getSecret(ctx context.Context, kvc azureclient.KeyVaultClient, secretURL st
...
 	if err != nil {
 		return nil, err
 	}
 
-	return &api.CertKeyPair{Key: key, Cert: cert}, nil
+	return &api.CertKeyChain{Key: key, Cert: cert}, nil
 }
 
-func ImportCertificate(ctx context.Context, kvc azureclient.KeyVaultClient, vaultURL, name string, key *rsa.PrivateKey, cert *x509.Certificate) error {
+func ImportCertificate(ctx context.Context, kvc azureclient.KeyVaultClient, vaultURL, name string, key *rsa.PrivateKey, cert *x509.Certificate, caCert *x509.Certificate) error {
```
jim-minter: should accept certificate slice / certkeyslice

pkg/tls/marshal.go
```diff
@@ -29,6 +29,25 @@ func CertAsBytes(cert *x509.Certificate) (b []byte, err error) {
 	return buf.Bytes(), nil
 }
 
+func CertChainAsBytes(certs []*x509.Certificate) (b []byte, err error) {
```
jim-minter: investigate if can be collapsed, one route is a smaller wrapper function in the template FuncMap

pkg/util/vault/vault.go
```diff
@@ -29,6 +29,11 @@ func EnrichCSFromVault(ctx context.Context, kvc azureclient.KeyVaultClient, cs *
 		return err
 	}
 	cs.Config.Certificates.Router = *kp
+	certChain := api.CertChain{}
+	// get last certificate from the chain.
+	// TODO: we should reconstruct chain and chose certificate based on it
+	certChain.Cert = append(certChain.Cert, kp.Cert[len(kp.Cert)-1])
```
jim-minter: goes to simpler derived fn

pkg/api/admin/api/config.go
```diff
@@ -122,9 +120,19 @@ type CertificateConfig struct {
...
+	Router           *CertificateChain `json:"-"`
 }
 
 // Certificate is an x509 certificate.
 type Certificate struct {
 	Cert *x509.Certificate `json:"cert,omitempty"`
 }
+
+type CertificateChain struct {
```
Makdaam: nit - naming: Chain implies there's a cryptographic relationship between the objects in the chain. CertificateSlice? CertificateBundle?

mjudeikis: There should be a relationship. So I think its ok :)

pkg/tls/marshal.go
```diff
@@ -126,3 +159,27 @@ func ParsePrivateKey(b []byte) (*rsa.PrivateKey, error) {
 
 	return nil, errors.New("failed to parse private key")
 }
+
+// ParseCertificateChain takes slice/chain of the certificates
+// and required certificate type and return certificate from the chain
+// TODO: Needs implementation
+func ParseCertificateChain(certs []*x509.Certificate, certType CertificateType) (*x509.Certificate, error) {
```
mjudeikis: @jim-minter in multiple places we need to parse "CertChain" and get a certificate. Any ideas about this implementation??

pkg/tls/marshal.go
```diff
@@ -126,3 +159,27 @@ func ParsePrivateKey(b []byte) (*rsa.PrivateKey, error) {
...
+	return certs[len(certs)-1], nil
+}
+
+// UniqueCert takes slice of the certificate and
+// returns certificate slice with unique values
+func UniqueCert(certs []*x509.Certificate) []*x509.Certificate {
+	seen := make(map[string]struct{}, len(certs))
+	j := 0
+	for _, v := range certs {
+		if _, ok := seen[string(v.Raw)]; ok {
```
mjudeikis: Not sure if this is the right way to sort unique certs `string(v.Raw)`

pkg/api/convertertoadmin.go
```diff
@@ -159,6 +159,14 @@ func convertCertKeyPairToAdmin(in CertKeyPair) *admin.Certificate {
 	}
 }
 
+func convertCertKeyChainToAdmin(in CertKeyPairChain) *admin.CertificateChain {
```
jim-minter: convertCertKeyPairChainToAdmin

pkg/api/admin/api/config.go
```diff
@@ -128,3 +131,7 @@ type CertificateConfig struct {
...
+	Cert []*x509.Certificate `json:"cert,omitempty"`
```
jim-minter: I think it'd be clearer to call this Certs (json:"certs,omitempty")

pkg/api/converterfromadmin.go
```diff
@@ -434,6 +434,14 @@ func mergeCertKeyPair(in *admin.Certificate, out *CertKeyPair) {
 	return
 }
 
+func mergeCertKeyChain(in *admin.CertificateChain, out *CertKeyPairChain) {
```
jim-minter: mergeCertKeyPairChain

jim-minter: can you add // TODO: this is not a great semantic - revisit if this field is ever enabled for write

pkg/api/convertertoadmin.go
```diff
@@ -159,6 +159,14 @@ func convertCertKeyPairToAdmin(in CertKeyPair) *admin.Certificate {
...
+	var adminCertChain admin.CertificateChain
+	for _, cert := range in.Cert {
```
jim-minter: just
```golang
return &admin.CertificateChain{
	Cert: in.Cert,
}
```
?

pkg/config/derive.go
```diff
@@ -261,3 +263,20 @@ func (derived) MDSDConfig(cs *api.OpenShiftManagedCluster) (string, error) {
 
 	return string(b.Bytes()), nil
 }
+
+// CaBundle created ca-bundle which includes
+// CA and any external certificates we trust
+func (derived) CaBundle(cs *api.OpenShiftManagedCluster) ([]*x509.Certificate, error) {
+	var caBundle []*x509.Certificate
+	// we take all CA
```
jim-minter: misleading - there's only one ;)
```golang
caBundle := []*x509.Certificate{cs.Config.Certificates.Ca.Cert}
```

pkg/cluster/healthcheck.go
```diff
@@ -83,13 +83,13 @@ func (u *simpleUpgrader) HealthCheck(ctx context.Context, cs *api.OpenShiftManag
...
-	err = u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs.Properties.FQDN, cs.Config.Certificates.OpenShiftConsole.Cert), "https://"+cs.Properties.PublicHostname+"/console/", time.Second)
+	cert := cs.Config.Certificates.OpenShiftConsole.Cert
+	err = u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs.Properties.FQDN, cert[len(cert)-1]), "https://"+cs.Properties.PublicHostname+"/console/", time.Second)
 	if err != nil {
 		return err
 	}
-
 	u.log.Info("checking admin console health")
-	return u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs.Properties.RouterProfiles[0].FQDN, cs.Config.Certificates.Router.Cert), "https://console."+cs.Properties.RouterProfiles[0].PublicSubdomain+"/", time.Second)
+	return u.doHealthCheck(ctx, getHealthCheckHTTPClient(cs.Properties.RouterProfiles[0].FQDN, cert[len(cert)-1]), "https://console."+cs.Properties.RouterProfiles[0].PublicSubdomain+"/", time.Second)
```
jim-minter: Strictly you should be using the Router CA cert here (they're probably the same, but)

pkg/fakerp/vault.go
```diff
@@ -53,6 +53,11 @@ func newVaultManager(ctx context.Context, subscriptionID string) (*vaultManager,
 }
 
 func (vm *vaultManager) writeTLSCertsToVault(ctx context.Context, cs *api.OpenShiftManagedCluster, vaultURL string) error {
+	externalCAkey, externalCAcert, err := tls.NewCA("external-signer")
```
jim-minter: fakerpCAKey, fakerpCACert?

pkg/fakerp/vault.go
```diff
@@ -94,7 +104,20 @@ func (vm *vaultManager) writeTLSCertsToVault(ctx context.Context, cs *api.OpenSh
 		if err != nil {
 			return err
 		}
-		err = vault.ImportCertificate(ctx, vm.kvc, vaultURL, cert.vaultKeyName, newkey, newcert)
+
+		// This chain should follow Certificate chain practices, where oder is:
+		// End-User Key
```
jim-minter: this line not needed ;)

pkg/tls/marshal.go
```diff
@@ -126,3 +151,19 @@ func ParsePrivateKey(b []byte) (*rsa.PrivateKey, error) {
 
 	return nil, errors.New("failed to parse private key")
 }
+
+// UniqueCert takes slice of the certificate and
+// returns certificate slice with unique values
+func UniqueCert(certs []*x509.Certificate) []*x509.Certificate {
+	seen := make(map[string]struct{}, len(certs))
+	j := 0
```
jim-minter: return a new slice, use append, saves you the index

pkg/tls/marshal.go
jim-minter: works for me

pkg/fakerp/vault.go
```diff
@@ -94,7 +104,20 @@ func (vm *vaultManager) writeTLSCertsToVault(ctx context.Context, cs *api.OpenSh
 		if err != nil {
 			return err
 		}
-		err = vault.ImportCertificate(ctx, vm.kvc, vaultURL, cert.vaultKeyName, newkey, newcert)
+
+		// This chain should follow Certificate chain practices, where oder is:
```
jim-minter: order

pkg/util/vault/vault.go
```diff
@@ -65,29 +64,34 @@ func getSecret(ctx context.Context, kvc azureclient.KeyVaultClient, secretURL st
...
-func ImportCertificate(ctx context.Context, kvc azureclient.KeyVaultClient, vaultURL, name string, key *rsa.PrivateKey, cert *x509.Certificate) error {
+func ImportCertificate(ctx context.Context, kvc azureclient.KeyVaultClient, vaultURL, name string, chain api.CertKeyPairChain) error {
 	buf := &bytes.Buffer{}
-	b, err := x509.MarshalPKCS8PrivateKey(key) // Must be PKCS#8 for Azure Key Vault.
+	b, err := x509.MarshalPKCS8PrivateKey(chain.Key) // Must be PKCS#8 for Azure Key Vault.
 	if err != nil {
 		return err
 	}
-
+	// This chain should follow Certificate chain practices, where oder is:
```
jim-minter: same changes as to comment above

pkg/fakerp/vault.go
mjudeikis: need more coffee... was looking to this for 10 minutes and thinking how this order needs to change :D

pkg/addons/translate.go
```diff
@@ -606,7 +606,7 @@ var Translations = map[string][]struct {
 	"Secret/default/router-certs": {
 		{
 			Path:     jsonpath.MustCompile("$.stringData.'tls.crt'"),
-			Template: "{{ String (CertAsBytes .Config.Certificates.Router.Cert) }}\n{{ String (CertAsBytes .Config.Certificates.Ca.Cert) }}\n{{ String (PrivateKeyAsBytes .Config.Certificates.Router.Key) }}",
+			Template: "{{ String (CertChainAsBytes .Config.Certificates.Router.Certs) }}\n{{ String (CertAsBytes .Config.Certificates.Ca.Cert) }}\n{{ String (PrivateKeyAsBytes .Config.Certificates.Router.Key) }}",
```
jim-minter: shouldn't need .Config.Certificates.Ca.Cert here any more?

pkg/api/converterfromadmin.go
```diff
@@ -434,6 +434,13 @@ func mergeCertKeyPair(in *admin.Certificate, out *CertKeyPair) {
 	return
 }
 
+// TODO: this is not a great semantic - revisit if this field is ever enabled for write
+func mergeCertKeyPairChain(in *admin.CertificateChain, out *CertKeyPairChain) {
+	if in.Certs != nil {
```
jim-minter: can you put this back to what was there previously - ie the append semantic?  Both are questionable in this case but given we have the append semantic everywhere else, I'd rather be standard for now

pkg/tls/marshal.go
```diff
@@ -96,10 +101,30 @@ func ParseCert(b []byte) (*x509.Certificate, error) {
 			return x509.ParseCertificate(block.Bytes)
 		}
 	}
-
 	return nil, errors.New("failed to parse certificate")
```
jim-minter: simplification: please can you make ParseCert() return ParseCertChain[0]? (like CertChainAsBytes but in reverse)

pkg/tls/marshal.go
jim-minter: bump

pkg/addons/translate.go
mjudeikis: I dont know. Need to test if we dont have some unpredicted fallouts.

mjudeikis: As Router certificate will not have anything from internal CA. So if we want to trust it, we still need

pkg/api/converterfromadmin.go
mjudeikis: This is more tricky. Because we prepare test structure with `populate.Walk(&cs, prepare)` which does this:
```
case *x509.Certificate:
		// use a dummy value because the zero value cannot be unmarshalled
		v.Set(reflect.ValueOf(tls.GetDummyCertificate()))
		return
	}
```
We already have value in `out` field. So we either have to "nil" it, and overwrite, or we have to make sure our tests are fine.

p.s we overwrite anywhere else in the converters. And we might need to change that in the future,

pkg/tls/marshal.go
```diff
@@ -83,7 +88,18 @@ func SSHPublicKeyAsString(key *rsa.PublicKey) (s string, err error) {
 	return sshkey.Type() + " " + base64.StdEncoding.EncodeToString(sshkey.Marshal()), nil
 }
 
+// ParseCert takes certificate as bytes and returns x509.Certificate
 func ParseCert(b []byte) (*x509.Certificate, error) {
+	certs, err := ParseCertChain(b)
+	if err != nil {
+		return nil, err
+	}
+	return certs[0], nil
```
jim-minter: need to check len > 0 or return errors.New("failed to parse certificate")

pkg/api/converterfromadmin.go
jim-minter: ok ;)

mjudeikis:
:+1:

PR #1271, opened 2019-03-11
Configurable openshift component logging

jim-minter:
```release-note
NONE
```


test/e2e/specs/fakerp/loglevel.go
```diff
@@ -0,0 +1,60 @@
...
+	var (
+		azurecli *azure.Client
+		cli      *standard.SanityChecker
+		ctx      = context.Background()
+	)
+
+	BeforeEach(func() {
+		var err error
+		azurecli, err = azure.NewClientFromEnvironment(true)
+		Expect(err).NotTo(HaveOccurred())
```
charlesakalugwu: add `Expect(azurecli).NotTo(BeNil())`

charlesakalugwu:
@jim-minter had a look at this and it looks generally good to me. I am not sure how the hack in `pkg/config/generate.go` is an improvement but i'm curious on how you propose we move forward.

test/e2e/specs/fakerp/loglevel.go
jim-minter: Our tests grow a log when we add lines like these - I am happy to, and would rather, code on the assumption that if no error is returned, the return value is non-nil.

PR #1279, opened 2019-03-13
trigger rotation correctly if files laid down by the startup container change

jim-minter:
```release-note
NONE
```

The short term goal here is to include the expected output of the startup container into the master hash.


pkg/cluster/hash.go
```diff
@@ -10,51 +10,81 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.NewEncoder(hash).Encode(vmss)
+	if err != nil {
+		return nil, err
+	}
+
+	err = arm.WriteStartupFiles(h.log, cs, writers.NewTarWriter(hash), config.GetHostname(app, "", instanceID), "")
```
mjudeikis: I imagine this is using same writer to append data to it?
Was this main motivation to move to writer based function?

mjudeikis:
Despite the fact that most of these are cosmetic changes (move stuff around, rename, etcd). It is really hard to notice some possible errors in the new hash'er implementation. Overall looks good. can lgtm, if anybody else dont want to take a look.

pkg/cluster/hash.go
jim-minter: yes, all the writer work is leading up to this line :)

jim-minter:
lgtm it, there's more where this came from...

PR #1281, opened 2019-03-13
infra-nodes: 2->3.

y-cote:
Update the replica number based on the number of infra nodes running.

```release-note
NONE
```


jim-minter:
2 was deliberate :)

PR #1282, opened 2019-03-14
many improvements for reconfiguration

jim-minter:
```release-note
* split the config blob into 3 blobs:
  * the sync blob (read by the sync pod; just a rename of the blob)

  * the master-startup blob (read by the startup pod on masters) - identical to
    the sync blob but changed at a different point in the update process

  * the worker-startup blob (read by the startup pod on workers) - whitelisted
    so that workers can't see secrets they shouldn't

  this is done (a) so that we can completely separate the process of updating
  nodes and the sync pod, (b) to homogenise the startup process so that the
  startup pod is used in all cases

* modify the startup pod so that it runs on both masters and workers, reading
  its blob (as appropriate) via a SAS URL (on masters, the startup pod also
  reads certificates from the key vault)

* modify the sync pod so that it detects readiness of components it manages,
  including Routes, instead of the plugin doing this.  Hopefully this will
  provide a clearer delineation between the plugin and sync pod to make
  development easier post-GA

* use static pods instead of DaemonSets for OVS and SDN pods.  This makes the
  path to master Nodes being Ready simpler, and stops the sync pod being
  essential for that to happen
  
* run the sync pod as a Deployment

* remove the openshift-node DaemonSet - it is completely unnecessary within the
  OSA management model

* fix a serious Plugin architectural issue: the Plugin (which is theoretically
  meant to be reusable for concurrent calls against multiple clusters) was
  embedding the clusterUpdater, which implicitly had clients bound to a single
  *OpenShiftManagedCluster.  This could have caused corruption in cases where
  calls against different clusters had been made against a single Plugin
  instance

* as a knock-on effect of the above, simplify client creation code (remove
  Initialize, CreateClients)

* introduce code to determine when the sync pod needs to be reconfigured and
  restarted (via hashing its output).  Correct sync pod reconfiguration should
  now in principle be possible without VM rotation

* move as much code out of the master and node startup scripts as possible and
  into the startup container.  A little more can be done on this front, moving
  customisations into the VM image
```


jim-minter:
@mjudeikis please can you mark out some time to look through this monster on Monday - probably makes sense to jump on a call to walk through it.

jim-minter:
/hold
scale up test is rotating masters

pkg/util/tls/tls.go
```diff
@@ -34,7 +34,7 @@ func newCert(key *rsa.PrivateKey, template *x509.Certificate, signingkey *rsa.Pr
 // NewCA returns a new rsa.PrivateKey and x509.Certificate for a CA
 // corresponding to the given CommonName.
 func NewCA(cn string) (*rsa.PrivateKey, *x509.Certificate, error) {
-	now := time.Now()
+	now := time.Now().Add(-time.Hour)
```
mjudeikis: DId this cause some flakes?

pkg/arm/data/node-startup.sh
```diff
@@ -25,7 +25,9 @@ while [[ $(hostname -d) == "" ]]; do sleep 1; done
 while ! docker pull {{ .Config.Images.Startup }}; do
   sleep 1
 done
-docker run --privileged --rm --network host -v /:/host:z {{ .Config.Images.Startup }} /host$PWD/worker-startup
+set +x
+docker run --privileged --rm --network host -v /:/host:z {{ .Config.Images.Startup }} '{{ .Config.WorkerStartupSASURI }}'
```
mjudeikis: can we move this args into the container? I dont like this way just because it will show in `ps aux`. so could be logged. Or it might be just me and my overthinking.

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   labels:
+    azure.openshift.io/sync-pod-wait-for-readiness: "false"
```
mjudeikis: etcd-metrics is not depending on monitoring operator. so I think we can check for it. Or there is other reasons we doing this?

pkg/arm/data/node-startup.sh
mjudeikis: overthinking...

jim-minter: you're right, should be an environment variable

pkg/addons/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
jim-minter: Technically we could check it, but the line I'm suggesting is: customer facing => check it, not => don't

pkg/arm/data/node-startup.sh
jim-minter: suggest to fix up in a subsequent patch set

pkg/util/tls/tls.go
jim-minter: no, but it seems good practice

pkg/arm/data/node-startup.sh
jim-minter: fixed

jim-minter:
/hold cancel

our optional tests are flaky but I believe this is sound - looking for an lgtm @mjudeikis

mjudeikis:
(closes_eyes)

PR #1285, opened 2019-03-14
Bubbling up errors properly

thekad:
Noticed during my demo I was not propagating the errors up in the last
couple of steps, fixing that now

```release-note
NONE
```


jim-minter:
/lgtm
thanks

PR #1286, opened 2019-03-14
add a convention for how we name storage accounts

jim-minter:
At least for now, let's continue to use tags to determine which storage account is which, but this might help whoever browses the portal.

```release-note
NONE
```


Makdaam:
/lgtm
Are we planning more Vaults in the nearest future?

jim-minter:
> Are we planning more Vaults in the nearest future?

No.

Issue #1278, opened 2019-03-13
Input/output error when accessing disk


mjudeikis:
If statefulset with PVC Claim template is restarted disk is remounted and if disk letter changes. This causes kubelet to lose disk mountpoint reference and try mounting non existing disk. 

The expectation is that it should be using udev rules and if disk location changes, it should be resolved. 

Similar issues:
https://github.com/kubernetes/kubernetes/issues/57444 

Disk exist on Azure:
```
az disk show -g mjudeikis -n kubernetes-dynamic-pvc-9dff3a6c-4738-11e9-bdd2-000d3a24045f
{
  "creationData": {
    "createOption": "Empty",
    "imageReference": null,
    "sourceResourceId": null,
    "sourceUri": null,
    "storageAccountId": null
  },
  "diskIopsReadWrite": 500,
  "diskMbpsReadWrite": 100,
  "diskSizeGb": 128,
  "encryptionSettings": null,
  "id": "/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/mjudeikis/providers/Microsoft.Compute/disks/kubernetes-dynamic-pvc-9dff3a6c-4738-11e9-bdd2-000d3a24045f",
  "location": "westeurope",
  "managedBy": "/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/mjudeikis/providers/Microsoft.Compute/virtualMachineScaleSets/ss-infra-1552649337/virtualMachines/ss-infra-1552649337_0",
  "name": "kubernetes-dynamic-pvc-9dff3a6c-4738-11e9-bdd2-000d3a24045f",
  "osType": null,
  "provisioningState": "Succeeded",
  "resourceGroup": "mjudeikis",
  "sku": {
    "name": "Premium_LRS",
    "tier": "Premium"
  },
  "tags": {
    "created-by": "kubernetes-azure-dd",
    "kubernetes.io-created-for-pv-name": "pvc-9dff3a6c-4738-11e9-bdd2-000d3a24045f",
    "kubernetes.io-created-for-pvc-name": "prometheus-data-prometheus-k8s-0",
    "kubernetes.io-created-for-pvc-namespace": "openshift-monitoring"
  },
  "timeCreated": "2019-03-15T15:40:10.873840+00:00",
  "type": "Microsoft.Compute/disks",
  "zones": null
}
```

I can see disk in use on node device path 0:
```
  volumesAttached:
[...]

PR #1289, opened 2019-03-15
add cachingmode to azure disk SC

mjudeikis:
fixes: https://github.com/openshift/openshift-azure/issues/1278

```release-notes
Fix azure disk remount issue
```


mjudeikis:
origin code:
https://github.com/openshift/origin/blob/release-3.11/vendor/k8s.io/kubernetes/pkg/apis/core/v1/defaults.go#L265-L268

It is different from Upstream. k8s upstream already using None as default

mjudeikis:
not solving :/

PR #1300, opened 2019-03-18
Monitoring shim for updates

mjudeikis:
```release-note
Add fakeRP blackbox monitoring tool/monitoring shim
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /kubeconfig*
 /sync
 /startup
+/monitoring
 /etcdbackup
 /tlsproxy
 /azure-controllers
--- a/Makefile
+++ b/Makefile
@@ -40,14 +40,13 @@ generate:
 test: unit e2e
 
 .PHONY: create delete upgrade
-
 create:
 	./hack/create.sh ${RESOURCEGROUP}
 
 delete:
 	./hack/delete.sh ${RESOURCEGROUP}
 
-upgrade: 
+upgrade:
 	./hack/upgrade.sh ${RESOURCEGROUP}
 
 azure-controllers: generate
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-THIS_FILE := $(lastword $(MAKEFILE_LIST))
 COMMIT=$(shell git rev-parse --short HEAD)$(shell [[ $$(git status --porcelain) = "" ]] && echo -clean || echo -dirty)
 PLUGIN_VERSION=$(shell awk '/^pluginVersion: /{ print $$2 }' <pluginconfig/pluginconfig-311.yaml)
 # if we are on master branch we should always use dev tag
@@ -42,16 +41,14 @@ test: unit e2e
 
 .PHONY: create delete upgrade
 
-create: monitoring-run
+create:
 	./hack/create.sh ${RESOURCEGROUP}
-	@$(MAKE) -f $(THIS_FILE) monitoring-stop
 
 delete:
 	./hack/delete.sh ${RESOURCEGROUP}
 
-upgrade: monitoring-run 
+upgrade: 
 	./hack/upgrade.sh ${RESOURCEGROUP}
-	@$(MAKE) -f $(THIS_FILE) monitoring-stop
 
 azure-controllers: generate
 	go build -ldflags ${LDFLAGS} ./cmd/$@
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+THIS_FILE := $(lastword $(MAKEFILE_LIST))
 COMMIT=$(shell git rev-parse --short HEAD)$(shell [[ $$(git status --porcelain) = "" ]] && echo -clean || echo -dirty)
 PLUGIN_VERSION=$(shell awk '/^pluginVersion: /{ print $$2 }' <pluginconfig/pluginconfig-311.yaml)
 # if we are on master branch we should always use dev tag
@@ -40,14 +41,17 @@ generate:
 test: unit e2e
 
 .PHONY: create delete upgrade
-create:
+
+create: monitoring-run
 	./hack/create.sh ${RESOURCEGROUP}
+	@$(MAKE) -f $(THIS_FILE) monitoring-stop
 
 delete:
 	./hack/delete.sh ${RESOURCEGROUP}
 
-upgrade:
+upgrade: monitoring-run 
 	./hack/upgrade.sh ${RESOURCEGROUP}
+	@$(MAKE) -f $(THIS_FILE) monitoring-stop
 
 azure-controllers: generate
 	go build -ldflags ${LDFLAGS} ./cmd/$@
@@ -99,6 +103,15 @@ metricsbridge-image: metricsbridge $(IMAGEBUILDER) pullregistry
 
 metricsbridge-push: metricsbridge-image
 	docker push $(METRICSBRIDGE_IMAGE)
+
+monitoring-build: generate
+	go build -ldflags ${LDFLAGS} ./cmd/monitoring
+
+monitoring-run: monitoring-build
+	./hack/monitoring.sh
+
+monitoring-stop:
+	./hack/monitoring.sh clean
 
 sync: generate
 	go build -ldflags ${LDFLAGS} ./cmd/$@
--- a/cmd/monitoring/monitoring.go
+++ b/cmd/monitoring/monitoring.go
@@ -33,6 +33,8 @@ type monitor struct {
 
 	resourceGroup  string
 	subscriptionID string
+
+	instances []instance
 }
 
 type instance struct {
@@ -87,13 +89,28 @@ func (m *monitor) listResourceGroupMonitoringHostnames(ctx context.Context, subs
 }
 
 func (m *monitor) run(ctx context.Context) error {
+	ch := make(chan os.Signal)
+	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
+
+	// we will close this guard when we load all monitors and clean-up is needed
+	bootCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func(ctx context.Context) {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ch:
+				os.Exit(0)
+			}
+		}
+	}(bootCtx)
+
 	m.log.Info("fetching URLs\n")
 	hostnames, err := m.listResourceGroupMonitoringHostnames(ctx, m.subscriptionID, m.resourceGroup)
 	if err != nil {
 		return err
 	}
-
-	var instances []instance
 
 	for _, hostname := range hostnames {
 		m.log.Debugf("initiate blackbox monitor %s \n", hostname)
@@ -116,7 +133,7 @@ func (m *monitor) run(ctx context.Context) error {
 			LogInitialErrors: *logerrors,
 		}
 
-		instances = append(instances, struct {
+		m.instances = append(m.instances, struct {
 			hostname string
 			b        *blackbox.Config
 		}{
@@ -125,18 +142,16 @@ func (m *monitor) run(ctx context.Context) error {
 		})
 	}
 
-	for _, mon := range instances {
+	for _, mon := range m.instances {
 		mon.b.Start(ctx)
 	}
 
-	ch := make(chan os.Signal)
-	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
 	m.log.Info("collecting metrics... CTRL+C to stop\n")
-
+	cancel()
 	for {
 		select {
 		case <-ch:
-			return m.persist(instances)
+			return m.persist(m.instances)
 		}
 	}
 }
--- a/cmd/monitoring/monitoring.go
+++ b/cmd/monitoring/monitoring.go
@@ -1,0 +1,179 @@
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"os/signal"
+	"path"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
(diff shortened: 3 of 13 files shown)


cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/openshift/openshift-azure/pkg/util/blackbox"
+	"github.com/openshift/openshift-azure/pkg/util/cloudprovider"
+	"github.com/openshift/openshift-azure/pkg/util/log"
+)
+
+var (
+	gitCommit = "unknown"
+	logLevel  = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
+	interval  = flag.Duration("interval", 100*time.Millisecond, "check interval with dimension. Example: 1000ms ")
+	timeout   = flag.Duration("timeout", 2*time.Hour, "timeout when script will exit if left unattended")
```
mjudeikis: I think (and I would like to be proven differently) we need this because we start the process in the background during creation and it stays there during create, update, tests. So to make sure we dont leak because cleanup didn't ran we set a timeout on the overall process context

.gitignore
```diff
@@ -11,6 +11,7 @@
 /tlsproxy
 /azure-controllers
 /metricsbridge
+/logs
```
jim-minter: unused?

Makefile
```diff
@@ -39,7 +39,7 @@ generate:
 test: unit e2e
 
 .PHONY: create delete upgrade
-create:
+create: monitor
```
jim-minter: This doesn't feel right.  I think we want the thing calling create/e2e to be able to start and stop the monitor themselves to wrap the whole create+e2e process.  The caller can build monitor, execute it in the background, record the pid, then kill the pid themselves on an exit trap.

Makefile
```diff
@@ -49,6 +49,12 @@ delete:
 upgrade:
 	./hack/upgrade-e2e.sh release-test-${TAG}-${COMMIT} ${SOURCE}
 
+monitor:
```
jim-minter: see above - would remove these

Makefile
jim-minter: This is going to be fiddly to do short term because we want this to bridge the setup and test containers in the release template.  Just like with the upgrade work, I think we'll need to expedite moving template functionality to this repo for this.  I'd start with the non-bushslicer, non-conformance stuff.

cmd/monitoring/monitoring.go
jim-minter: with the right bash trap set up I think we can remove the timeout

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,181 @@
...
+	subscriptionID string
+}
+
+type instance struct {
+	hostname string
+	b        *blackbox.Config
+}
+
+func (m *monitor) init(ctx context.Context, log *logrus.Entry) error {
+	cpc, err := cloudprovider.Load("_data/_out/azure.conf")
```
jim-minter: won't exist before create is called? - read environment variables

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,181 @@
...
+		case <-ctx.Done():
+			return m.persist(instances)
+		}
+	}
+}
+
+func (m *monitor) persist(instances []instance) error {
+	if _, ok := os.Stat(*outputdir); ok != nil {
+		m.log.Debugf("create outputdir %s", path.Dir(*outputdir))
+		err := os.MkdirAll(path.Dir(*outputdir), 0777)
```
jim-minter: prefer to have the caller do this

hack/monitor.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -ex
```
jim-minter: drop this file

pkg/util/azureclient/compute_addons.go
```diff
@@ -71,6 +78,66 @@ func (c *virtualMachineScaleSetsClient) UpdateInstances(ctx context.Context, res
...
+
+	var body interface{}
+	defer response.Body.Close()
+	err = json.NewDecoder(response.Body).Decode(&body)
+	if err != nil {
+		return nil, err
+	}
+
+	var ips []string
+	for _, ip := range jsonpath.MustCompile("$.value[*].properties.ipAddress").Get(body) {
```
jim-minter: this is a hack, and the bar to entry to this file is higher.  I'd just keep all this out of azureclient as it was in the file I sent you

pkg/util/azureclient/network.go
```diff
@@ -67,3 +67,33 @@ func (c *virtualNetworkPeeringsClient) List(ctx context.Context, resourceGroupNa
 func (c *virtualNetworkPeeringsClient) Client() autorest.Client {
 	return c.VirtualNetworkPeeringsClient.Client
 }
+
+// PublicIPAddressesClient is a minimal interface for azure NewPublicIPAddressesClient
+type PublicIPAddressesClient interface {
+	Get(ctx context.Context, resourceGroupName string, publicIPAddressName string, expand string) (network.PublicIPAddress, error)
+	Client
```
jim-minter: `Client` not needed

pkg/util/azureclient/network.go
```diff
@@ -67,3 +67,33 @@ func (c *virtualNetworkPeeringsClient) List(ctx context.Context, resourceGroupNa
...
+func NewPublicIPAddressesClient(ctx context.Context, subscriptionID string, authorizer autorest.Authorizer) PublicIPAddressesClient {
+	client := network.NewPublicIPAddressesClient(subscriptionID)
+	setupClient(ctx, &client.Client, authorizer)
+
+	return &publicIPAddressesClient{
+		PublicIPAddressesClient: client,
+	}
+}
+
+func (c *publicIPAddressesClient) Get(ctx context.Context, resourceGroupName string, publicIPAddressName string, expand string) (network.PublicIPAddress, error) {
```
jim-minter: from here below not needed

pkg/util/blackbox/types.go
```diff
@@ -0,0 +1,149 @@
...
+	fmt.Fprintf(w, "mean   latency: %-4dms\n", sum/time.Duration(len(c.samples))/time.Millisecond)
+
+	sort.Slice(c.samples, func(i, j int) bool { return c.samples[i].latency < c.samples[j].latency })
+
+	fmt.Fprintf(w, "median latency: %-4dms\n", c.samples[len(c.samples)/2].latency/time.Millisecond)
+
+	fmt.Fprintf(w, "95%%ile latency: %-4dms\n", c.samples[len(c.samples)*95/100].latency/time.Millisecond)
+
+	fmt.Fprintf(w, "99%%ile latency: %-4dms\n", c.samples[len(c.samples)*99/100].latency/time.Millisecond)
+	fmt.Fprintf(w, "creationTime: %v\n", c.startTime)
```
jim-minter: "start time: %...", "end time: %..."

pkg/util/blackbox/types.go
```diff
@@ -0,0 +1,149 @@
...
+	fmt.Fprintf(w, "finisTime: %v\n", c.stopTime)
+	fmt.Fprintf(w, "cluster creation time: %v m\n", c.stopTime.Sub(c.startTime).Minutes())
```
jim-minter: remove

Makefile
mjudeikis: idea is that both `upgrade/update/e2e` is calling `make create`. but we will have to duplicate code inside in multiple places. In example scenarios:
1. Create is created, and we need metrics from creating only. Code will be in `create.sh` and it will manage it same as it is managing fakeRP process. 
2. Upgrade test is called. Upgrade calls `create.sh`, so now monitor script cant be initiated as part of `create.sh` because the script is being terminated as part of the upgrade script. So it invalidated scenario 1 as now logic needs to be more complex.
3. Update e2e is called. It same uses `create.sh` the script, so same logic applies. 


Im not very big fan of how we already managing fakeRP and killing it cia trap based on port `8080`. My thinking was to build in `timeout` in the process as housekeeping only. Traps would be used to send `sigint` signal into the pid.
And we call the process from `make` file as a dependency. In this case `make create` will call target `monitor` each time, and we just need to make sure it is cleaned after with another target `make monitor-clean`.
This way we can do for scenario 1:
`make create` -> `create: monitor init-cluster monitor-clean`
Scenario 2:
`make upgrade` - > `upgrade:` monitor init-cluster upgrage-init monitor-clean`

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,179 @@
...
+
+	for _, mon := range instances {
+		mon.b.Start(ctx)
+	}
+
+	ch := make(chan os.Signal)
+	signal.Notify(ch, os.Interrupt)
+	m.log.Info("collecting metrics... CTRL+C to stop\n")
+
+	for {
```
mjudeikis: This part is messy. So if we got stuck on L61, where we try to resolve hostnames before we create monitors this part of the code is not executed, so we never listen for `sigint` signal, so `kill -2 pid` does not work. Not sure how this can be solved now.

cmd/monitoring/monitoring.go
mjudeikis: How I don't like traps :D

mjudeikis:
/hold
delete is broken

hack/upgrade-e2e.sh
```diff
@@ -20,7 +25,8 @@ set +x
 . secrets/secret
 set -x
 
-trap 'make delete' EXIT
+trap 'kill -2 ${MON_PID}; make delete' EXIT
```
mjudeikis: add ` kill; wait; make delete`

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,175 @@
...
+			b:        b,
+		})
+	}
+
+	for _, mon := range instances {
+		mon.b.Start(ctx)
+	}
+
+	ch := make(chan os.Signal)
+	signal.Notify(ch, os.Interrupt)
```
mjudeikis: change to sigterm (11)

pkg/util/azureclient/network.go
```diff
@@ -67,3 +67,33 @@ func (c *virtualNetworkPeeringsClient) List(ctx context.Context, resourceGroupNa
...
+func NewPublicIPAddressesClient(ctx context.Context, subscriptionID string, authorizer autorest.Authorizer) PublicIPAddressesClient {
+	client := network.NewPublicIPAddressesClient(subscriptionID)
+	setupClient(ctx, &client.Client, authorizer)
+
+	return &publicIPAddressesClient{
+		PublicIPAddressesClient: client,
+	}
+}
+
+func (c *publicIPAddressesClient) ListVirtualMachineScaleSetPublicIPAddressesComplete(ctx context.Context, resourceGroupName string, scaleSetName string) (network.PublicIPAddressListResultIterator, error) {
```
mjudeikis: remote those 2

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,219 @@
...
+	go func(ctx context.Context) {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ch:
+				os.Exit(0)
+			}
+		}
+	}(bootCtx)
```
mjudeikis: @jim-minter added this for cleaning up the script in the initial state.
It can be done with only this go routine by passing `m.instances` as an address into goroutine for cleanups or making a channel with data for goroutine to consume. But this makes it simple for now. fix as follow-up? so we can start testing?

hack/monitoring.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -ex
+
+if [ $# -eq 1 ]; then
+  echo "cleaning"
+  for pid in $(ps -ef | egrep "(./monitoring)" | grep -v grep | awk '{print $2}'); do kill -2 $pid; done
```
jim-minter: -15

hack/upgrade-e2e.sh
```diff
@@ -20,7 +26,8 @@ set +x
 . secrets/secret
 set -x
 
-trap 'make delete' EXIT
+trap 'kill -2 ${MON_PID}; wait; make delete' EXIT
```
jim-minter: -15

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		mon.b.Stop(f)
+		f.Close()
+	}
+	return nil
+}
+
+func loadOCConfig() (*api.OpenShiftManagedCluster, error) {
+	b, err := ioutil.ReadFile("_data/containerservice.yaml")
+	switch {
```
jim-minter: `return nil, err`, no switch

cmd/monitoring/monitoring.go
```diff
@@ -0,0 +1,219 @@
...
+		})
+	}
+
+	for _, mon := range m.instances {
+		mon.b.Start(ctx)
+	}
+
+	m.log.Info("collecting metrics... CTRL+C to stop\n")
+	cancel()
+	for {
```
jim-minter: just do ```
<-ch
return m.persist(m.instance)
```

File: hack/upgrade-local.sh [hunk 22]
```diff
@@ -0,0 +1,43 @@
...
+git clone -b $SOURCE https://github.com/openshift/openshift-azure.git $T/src/github.com/openshift/openshift-azure
+
+cd $T/src/github.com/openshift/openshift-azure
+
+ln -s ${ORG_GOPATH}/src/github.com/openshift/openshift-azure/secrets secrets
+set +x
+. secrets/secret
+set -x
+
+trap 'kill -2 ${MON_PID}; wait; make delete' EXIT
```
jim-minter: -15

PR #1302, opened 2019-03-19
 write _data/containerservice.yaml after cluster is created so that vnet id is persisted

jim-minter:
```release-notes
NONE
```


pkg/fakerp/fakerp.go
```diff
@@ -175,6 +165,17 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
 		return nil, err
 	}
 
+	// persist the OpenShift container service
```
jim-minter: @thekad I think this is the issue you were running into?

cmd/createorupdate/createorupdate.go
```diff
@@ -117,30 +116,25 @@ func execute(
...
-		return nil, fmt.Errorf("failed to find %s: %v", shared.DataDirectory, err)
-	}
-
 	if adminManifest != "" {
 		oc, err := fakerp.GenerateManifestAdmin(adminManifest)
 		if err != nil {
 			return nil, fmt.Errorf("failed reading admin manifest: %v", err)
 		}
-		defaultAdminManifest := filepath.Join(dataDir, "manifest-admin.yaml")
+		defaultAdminManifest := shared.PkgRoot + "/_data/manifest-admin.yaml"
```
thekad: i have an irrational fear of using string concatenation vs filepath.Join()

pkg/fakerp/fakerp.go
```diff
@@ -142,34 +142,19 @@ func createOrUpdate(ctx context.Context, log *logrus.Entry, cs, oldCs *api.OpenS
...
-	err = os.MkdirAll(filepath.Join(dataDir, "_out"), 0777)
+	err = acceptMarketplaceAgreement(ctx, log, cs, testConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// write out development files
 	log.Info("write helpers")
-	err = writeHelpers(cs)
+	err = os.MkdirAll(filepath.Join(shared.PkgRoot+"/_data/_out"), 0777)
```
thekad: why use filepath.Join at all if you're concatenating the path here?

pkg/fakerp/shared/shared.go
```diff
@@ -1,35 +1,44 @@
...
 const (
-	DataDirectory = "_data/"
 	LocalHttpAddr = "localhost:8080"
 )
 
+// PkgRoot is the openshift-azure root directory
+var PkgRoot = "."
+
+func init() {
+	if gopath, found := os.LookupEnv("GOPATH"); found {
```
thekad: I feel this will lead to instances of you doing development in a fresh clone and not seeing your tests reflected because it is being overridden by GOPATH, I'd feel like `var PkgRoot = filepath.Abs(".")` would be enough, wdyt?

pkg/fakerp/fakerp.go
jim-minter: thanks - slip

pkg/fakerp/shared/shared.go
jim-minter: yes, this was mucked up.  I've had another go - ptal

cmd/createorupdate/createorupdate.go
```diff
@@ -117,21 +116,16 @@ func execute(
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to find %s: %v", shared.DataDirectory, err)
-	}
-
 	if adminManifest != "" {
 		oc, err := fakerp.GenerateManifestAdmin(adminManifest)
 		if err != nil {
 			return nil, fmt.Errorf("failed reading admin manifest: %v", err)
 		}
-		defaultAdminManifest := filepath.Join(dataDir, "manifest-admin.yaml")
```
mjudeikis: We had all this dataDir logic because when executed from `./hack/upgrade.sh` or called from `pkg/tests/...` it was under the different relative path and was not able to find files. If I remember it was keyrotation tests. Is this not a case anymore?

test/sanity/sanity.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/test/e2e/standard"
+	testlogger "github.com/openshift/openshift-azure/test/util/log"
+)
+
+// Checker is a singleton sanity checker which is used by all the Ginkgo tests
+var Checker *standard.SanityChecker
+
+func init() {
+	b, err := ioutil.ReadFile("../../_data/containerservice.yaml") // running via `go test`
```
mjudeikis: Is this how we replace all the dir walker?

jim-minter: yes

cmd/createorupdate/createorupdate.go
jim-minter: The whole thing was quite jumbled.  In general, where possible, test should work with external representation cs fetched from the RP with a GET call.  The only valid exception to this currently is the internal representation cs is needed to build the authenticators (relies on seeing the certs, which are not end-user visible).

So it boils down to these cases:
1. test needs an external cs - make a GET call
2. e2e test framework needs internal cs, is running via go test.  Our cwd will be <root>/test/e2e because that's where `func TestE2E(t *testing.T)` is.  Look in `../../_data/containerservice.yaml`.
3. e2e test framework needs internal cs, is running via compiled test binary.  Our cwd will be <root>.  Look in `_data/containerservice.yaml`.
4. fakerp.  Our cwd will be <root>.  Look in `_data/containerservice.yaml`.

PR #1303, opened 2019-03-19
disable docker build strategy

jim-minter:
```release-note
NONE
```

For security reasons - brings us in line with OSD


test/e2e/standard/cluster.go
```diff
@@ -400,3 +401,32 @@ func (sc *SanityChecker) checkCanUseAzureFileStorage(ctx context.Context) error
...
+				Source: buildv1.BuildSource{
+					Dockerfile: to.StringPtr("FROM scratch"),
+				},
+				Strategy: buildv1.BuildStrategy{
+					Type: buildv1.DockerBuildStrategyType,
+				},
+			},
+		},
+	})
+	if !kerrors.IsForbidden(err) {
```
thekad: a better message would be `if err == nil return fmt.Error("cluster can docker build")` before checking if the error is checked vs kerrors, otherwise you would get "unexpected error nil" which can be confusing

jim-minter: I'm comfortable with this as it is, even if parsimonious

Issue #1308, opened 2019-03-20
fakerp - partial delete broken


mjudeikis:
iI we want to delete cluster when it failed in flight, it is broken because config persistency moved to end of cluster creation.

```
DEBU[2019-03-20T15:03:15Z]/tmp/tmp.DpRW9cuiuY/src/github.com/openshift/openshift-azure/pkg/fakerp/middleware.go:15 github.com/openshift/openshift-azure/pkg/fakerp.(*Server).logger.func1() ending:   DELETE /subsc
riptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/m3PNPT/providers/Microsoft.ContainerService/openShiftManagedClusters/m3PNPT?api-version=2018-09-30-preview                                            
DEBU[2019-03-20T15:04:15Z]/tmp/tmp.DpRW9cuiuY/src/github.com/openshift/openshift-azure/pkg/fakerp/middleware.go:13 github.com/openshift/openshift-azure/pkg/fakerp.(*Server).logger.func1() starting: DELETE /subsc
riptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/m3PNPT/providers/Microsoft.ContainerService/openShiftManagedClusters/m3PNPT?api-version=2018-09-30-preview                                            
DEBU[2019-03-20T15:04:15Z]/tmp/tmp.DpRW9cuiuY/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:17 github.com/openshift/openshift-azure/pkg/fakerp.(*Server).internalError() 500 Internal Error: Failed t
o read the internal config  
```

/kind bug

mjudeikis:
ga label because it is blocker for testing

PR #1309, opened 2019-03-20
move helpers persistence before creation

mjudeikis:
Moves config writer before creation.
fixes: https://github.com/openshift/openshift-azure/issues/1308

```release-note
NONE
```


mjudeikis:
Do we care if we write all files again or we want to split CS just to separate function same as `writeHelpers`

jim-minter:
don't mind either way - the real rp would write to the db indicating creating, then again indicating created - the fake rp db is the filesystem, so probably best to write twice

PR #1313, opened 2019-03-21
splitting out pkg/sync and pkg/startup for multi-version plugin

jim-minter:
```release-note
Work towards ability to update clusters which are on a supported old version
with respect to the RP/plugin.

* Set master branch to work towards v4.0; we will reset this to v5.0 when we
  branch release-v4

* Make pluginconfig support multiple versions in parallel - see
  /pluginconfig/pluginconfig-311.yaml as an example

* Rename pkg/addons -> pkg/sync, simplify to an interface and make multiversion

* Split pkg/arm -> pkg/startup, simplify to an interface and make multiversion

* Move lint-addons to a unit test

* Reduce dependencies on pkg/cluster in preparation for making it multiversion

* Miscellaneous sync pod performance, behaviour and code style fixes
```


jim-minter:
@mjudeikis here's the next instalment, ptal - the main piece left now is to make pkg/{config,arm} multiversion

pkg/fakerp/config.go
```diff
@@ -61,43 +61,47 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	template.Certificates.GenevaLogging.Key = logKey
 	template.Certificates.GenevaMetrics.Cert = metCert
 	template.Certificates.GenevaMetrics.Key = metKey
-	template.Images.GenevaImagePullSecret = pullSecret
-	template.Images.ImagePullSecret = imagePullSecret
+	template.GenevaImagePullSecret = pullSecret
+	template.ImagePullSecret = imagePullSecret
```
mjudeikis: This is sole image config. This is why it was under `Images`. Any reasoning to move/change api?

mjudeikis:
looks fine. 
/retest

pkg/fakerp/config.go
jim-minter: yes - I am hoping the pull secrets are the same regardless of the cluster version

mjudeikis:
want to merge it as it is?

jim-minter:
/hold
scale up/down still rotating

jim-minter:
green across the board, nice!
adding lgtm as @mjudeikis said "looks fine" before :)

PR #1315, opened 2019-03-21
Update release docs to include CHANGELOG.md.

y-cote:
This adds the release CHANGELOG.md creation steps.

```release-note
NONE
```

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -137,6 +137,9 @@ export GITHUB_TOKEN=<github_token_from_team_secret>
 ./releasenotes -repopath . -commitrange (vx.y)-1..HEAD >> CHANGELOG.md
 git add CHANGELOG.md
 ```
+note about release numbers to use as commitrange with releasenotes above:
+* If you're cutting a new major release, (vx.y)-1 means v(x-1)."latest y on the x-1 branch"
+* If you're cutting a new minor release, (vx.y)-1 means vx.(y-1)
 
 Make sure you update `clusterVersion` field all image version to point to specific version, instead of `latest` tag.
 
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -124,6 +124,20 @@ images:
...
+```
+# vx.y
+         <- additional blank line
+```
+
+scrape and append release-notes to `CHANGELOG.md`
+```
+make releasenotes
+export GITHUB_TOKEN=<github_token_from_team_secret>
+./releasenotes -repopath . -commitrange (vx.y)-1..HEAD >> CHANGELOG.md
```
jim-minter: I'm trying to work this out in my head :) - say the current latest release is v2.4: regardless of whether I'm releasing v2.5 or v3.0, the command is the same - `-commitrange v2.4..HEAD` ?  I think that makes sense?

y-cote: Right it's just that the "current" release (the one currently worked on) is not tagged yet at step 4 if I am not mistaken. So If I was to apply it right now and I wanted to release v3.1 (which we don't have a tag for yet), I would have to write : 

 ./releasenotes -repopath . -commitrange (v3.1)-1..HEAD >> CHANGELOG.md

I could have also wrote : -commitrange v<last_release_tagged>..HEAD,  but the notation above seems more concise ;-)

y-cote: In other words, vx.y as referred to in this document represent the next release worked on. Not the version already released. I tried to keep that convention to avoid confusion.

jim-minter: ok, we're almost aligned.  I agree with everything you wrote, I just think you need to be more explicit about what `(vx.y)-1` means - it refers to the latest minor release of the latest major release made.

If you're cutting a new major release, `(vx.y)-1` means v(x-1)."latest y on the x-1 branch"
If you're cutting a new minor release, `(vx.y)-1` means vx.(y-1)

y-cote: Sounds good, I added that precision.

PR #1319, opened 2019-03-21
prevent update from succeeding unless cluster version matches v2.5

jim-minter:
```release-note
prevent update from succeeding unless cluster version matches v2.5
```


jim-minter:
@mjudeikis fyi, pushing this through

PR #1320, opened 2019-03-21
prevent update from succeeding unless cluster version matches plugin version

jim-minter:
```release-note
prevent update from succeeding unless cluster version matches plugin version
```


jim-minter:
@mjudeikis fyi, pushing this through

mjudeikis:
lgtm

PR #1328, opened 2019-03-22
Blob read unit tests

kargakis:
Apparently reading a malformed blob does not return an error. May need to use [encoding/json/#Decoder.DisallowUnknownFields](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields) to catch those cases.

```release-note
NONE
```


jim-minter:
@kargakis :wave: , how are you doing? :smile: 

thanks for the PR!

/lgtm

kargakis:
> @kargakis 👋 , how are you doing? 😄

hi @jim-minter, I'm doing alright, how about you? I'm feeling very productive today it seems :)

PR #1329, opened 2019-03-22
make hash invariant of SAS URIs

jim-minter:
```release-note
Fix an issue which was causing node rotations on scale; hardened testing
```


jim-minter:
lgtm'ing this...

PR #1331, opened 2019-03-22
make pkg/arm and pkg/config version aware

jim-minter:
```release-note
* make pkg/arm and pkg/config version aware
* ensure update functionality for cluster versions >= 3.0
* allow admin to write "latest" to plugin version to upgrade cluster
* allow admin to scale up infra nodes from 2->3
* remove RunningUnderTest field in config struct - no longer needed as sync/startup pods don't validate input now
```


jim-minter:
@mjudeikis this is at a state where it nearly but not quite upgrades 2.4->latest.  The main sticking point is the move to static pods for ovs and sdn.  I'm going to propose that we don't attempt to upgrade private-preview-1.*->latest on the grounds that private-preview-1.* uses basic LBs which are not upgradeable to standard LBs (would have to add a lot of one-off orchestration to stop the world, switch the LBs, then start the world).

test/manifests/fakerp/admin-update.yaml
```diff
@@ -0,0 +1,3 @@
+# ADMIN_MANIFEST=test/manifests/fakerp/admin-update.yaml hack/upgrade.sh
```
mjudeikis: Separate test in the release or where do you see those?

pkg/startup/v4/data/master/etc/origin/node/pods/ovs.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     scheduler.alpha.kubernetes.io/critical-pod: ''
   labels:
-    app: ovs
+    app: ovs-static
```
mjudeikis: more background on this would be nice :)

mjudeikis:
I like where this is going, but lets wait on msft call :)

jim-minter:
opened https://github.com/openshift/openshift-azure/issues/1340 to get better deployment logs on timeout

pkg/startup/v4/data/master/etc/origin/node/pods/ovs.yaml
jim-minter: if you create a static pod with the same labels that a daemonset controller is watching, you get a loop whereby the daemonset controller keeps trying to delete the mirror pod and the kubelet keeps recreating it

test/manifests/fakerp/admin-update.yaml
jim-minter: we need to discuss the exact plan - lots is needed here - this is just a dumb thing to help command line testing

pkg/util/arm/arm.go
```diff
@@ -0,0 +1,117 @@
...
+	Parameters     struct{}      `json:"parameters,omitempty"`
+	Variables      struct{}      `json:"variables,omitempty"`
+	Resources      []interface{} `json:"resources,omitempty"`
+	Outputs        struct{}      `json:"outputs,omitempty"`
+}
+
+// FixupAPIVersions inserts an apiVersion field into the ARM template for each
+// resource (the field is missing from the internal Azure type).  The versions
+// referenced here must be kept in lockstep with the imports above.
+func FixupAPIVersions(template map[string]interface{}) {
```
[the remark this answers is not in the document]
jim-minter: you're probably right, damn.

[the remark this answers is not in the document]
jim-minter: added another commit which fixes this

jim-minter:
/hold
going to remove some of the v2-> upgrade complexity before this merges

jim-minter:
flake https://github.com/openshift/openshift-azure/issues/1343

jim-minter:
tests passed, lgtming and merging this

PR #1335, opened 2019-03-23
vmimage: remove .Extras, missed out by previous refactor

jim-minter:
```release-note
NONE
```


jim-minter:
flake #1336

jim-minter:
Tagging this lgtm because it's a straightforward fix and it's causing vmimage builds to break.

PR #1341, opened 2019-03-25
refactor upgrade and create script locations

mjudeikis:
Fixes required to monitor canary application.
Because we start monitoring shim from the master branch, but the config is created in the "source" code folder, we need to provide config location. 

```release-note
NONE
```


Makefile
```diff
@@ -143,11 +146,11 @@ releasenotes:
 
 .PHONY: verify
 verify:
-	./hack/validate-generated.sh
+	./hack/verify/validate-generated.sh
```
jim-minter: wouldn't mind s/^verify-|^validate-// on all the filenames going into hack/verify

hack/artifacts.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -ex
+
+mkdir -p /tmp/artifacts
+
+export KUBECONFIG=$GOPATH/src/github.com/openshift/openshift-azure/_data/_out/admin.kubeconfig
```
jim-minter: can we rely on cwd here - do we need `$GOPATH/src/github.com/openshift/openshift-azure/` hardcoded?

hack/tests/create-e2e.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -ex
+# upgrade-e2e script is intended to run in CI environmet and will not run localy
+
+export RESOURCEGROUP=$(cat /dev/urandom | tr -dc 'a-z' | fold -w 6 | head -n 1)
```
jim-minter: It was my mistake not to prefix these with `e2e-` - please can you add?

hack/tests/create-e2e.sh
```diff
@@ -0,0 +1,29 @@
...
+set +x
+. secrets/secret
+set -x
+
+# start monitoring and record PID
+make monitoring-build
+./monitoring -outputdir=/tmp/artifacts &
+MON_PID=$!
+
+cd /go/src/github.com/openshift/openshift-azure
```
jim-minter: would very much like to avoid hard coding paths - can we make this runnable on a dev laptop as well?

hack/tests/create-e2e.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -ex
+# upgrade-e2e script is intended to run in CI environmet and will not run localy
```
jim-minter: why?

hack/tests/upgrade-e2e.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash -ex
+# upgrade-e2e script is intended to run in CI environmet and will not run localy
```
jim-minter: this is awfully similar to the above file - how can it be done differently? is a single bash file using functions any better?

hack/vmimage.sh
```diff
@@ -22,6 +22,8 @@ make create
 
 make e2e
 
+make artifacts
```
jim-minter: thanks!

hack/tests/upgrade-e2e.sh
mjudeikis: We have 2 files `upgrade-e2e` and `create-e2e` and they have a lot of common code. I already unified `upgrade-e2e` to be able to run locally. I dont want to merge them as it will make them even more complex :/ keep it simple, stupid.

mjudeikis:
release need to merge first: https://github.com/openshift/release/pull/3229

hack/tests/upgrade-e2e.sh
```diff
@@ -0,0 +1,84 @@
...
+#    We pass config file from temporary folder, 
+#    because it will be created there first during source cluster creation
+# 4. Link secrets to appropriate location
+# 5. Clone released code and create a cluster
+# 6. Switch to PR code base and inicate upgrade.
+#    Old cluster upgarde is always managed by new or same version of RP
+
+source_secret(){
+    # running in ci
+    if [[ -f /usr/local/e2e-secrets/azure/secret ]];then
```
jim-minter: @mjudeikis follow-up: I'd rather that CI set up the symlink - our scripts in repo can/should expect secrets/secret to exist.

PR #1342, opened 2019-03-25
Use secret key refs instead of plain secret values

thekad:
```release-note
NONE
```

Some objects make use of environment variables that come from secret values, but these values end up logged by the sync pod.
Sync pod doesn't log the values when the object type is Secret, so by moving these values to a Secret and using a key ref we avoid logging secrets to the sync pod logs.


thekad:
still wip because i haven't been able to test this by running the sync pod, as soon as i can verify this works will remove the wip tag

pkg/sync/v4/translate.go
```diff
@@ -635,12 +639,7 @@ var translations = map[string][]struct {
 			},
 		},
 	},
-	"Service/default/router-stats": {
-		{
-			Path:     jsonpath.MustCompile("$.metadata.annotations['prometheus.openshift.io/password']"),
-			Template: "{{ .Config.RouterStatsPassword }}",
-		},
-	},
+	"Service/default/router-stats": {},
```
jim-minter: remove this line

jim-minter:
one nit, then please squash, then lgtm

thekad:
/hold

not sure what is going here, gonna check

jim-minter:
/lgtm cancel

cluster isn't coming up because registry isn't coming up - something wrong with the PR I guess

pkg/sync/v4/translate.go
```diff
@@ -544,6 +536,18 @@ var translations = map[string][]struct {
 			Template: "{{ String (PrivateKeyAsBytes .Config.Certificates.Router.Key) }}",
 		},
 	},
+	"Secret/default/docker-registry-http": {
+		{
+			Path:     jsonpath.MustCompile("$.stringData.password"),
+			Template: "{{ .Config.RegistryHTTPSecret }}",
```
jim-minter: Missing Base64Encode here I think

thekad: No, this is going to `stringData` not `data`

jim-minter:
merging this
@thekad please make sure there is a follow-up issue open to get these stats into Geneva - work with @Makdaam to achieve this

PR #1354, opened 2019-03-26
label resource groups more clearly

jim-minter:
```release-note
NONE
```


jim-minter:
@mjudeikis also merging this so that we can see what's what

PR #1359, opened 2019-03-26
Warn about bad 'glide' tool.

y-cote:
Some version of glide breaks the vendor folder and crashes when run.

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,10 @@ properties:
 
 ## Dependency management
 To add a new dependency to the project, add the package information to glide.yaml and execute `glide up -v`
+* NOTE: some versions of glide might crash or generate an incomplete vendor/ folder when ran against this repo. In this case you may want to try getting the latest version:
+```
+go get github.com/Masterminds/glide
+```
 
 ## CI infrastructure
 Read more about how to work with our CI system [here](https://github.com/openshift/release/blob/master/projects/azure/README.md).


README.md
```diff
@@ -150,6 +150,10 @@ properties:
 
 ## Dependency management
 To add a new dependency to the project, add the package information to glide.yaml and execute `glide up -v`
+* NOTE: Do not run the package manager version of glide. It may very well generate a broken vendor/ folder. Instead, use:
```
jim-minter: "Do not run the package manager version of glide" becomes superstition very quickly - write something like "the version xyz of glide shipped with Fedora abc is known not to work properly"

jim-minter:
/lgtm cancel
one change requested

README.md
y-cote: I get your point, but then being too precise might steer people in the wrong direction again. I haven't bisected all versions packaged for different versions of Fedora. In other words, I am not sure that the only version of glide affected is 0.13.1 for Fedora 29.

I updated the text in a way it would have been useful for me when I hit the problem earlier. Let me know how it works for you.

README.md
```diff
@@ -150,6 +150,10 @@ properties:
 
 ## Dependency management
 To add a new dependency to the project, add the package information to glide.yaml and execute `glide up -v`
+* NOTE: some versions of glide might crash or generate incomplete vendor/ folder when ran against this repo. In this case you may wanna try getting the latest version:
```
jim-minter: s/generate/generate an/
s/wanna/want to/

y-cote: thanks

charlesakalugwu:
Instead of continuing to deal with these glide issues, it might be beneficial if we finally moved to at least `dep` if not the new `go modules` for managing our project dependencies.

y-cote:
> Instead of continuing to deal with these glide issues, it might be beneficial if we finally moved to at least `dep` if not the new `go modules` for managing our project dependencies.

I proposed that as well. But I don't think it will happen before GA at least.

mjudeikis:
Is there an issue specifying what is wrong? just curious? 
So running `glide get -v <package>` breaks?

y-cote:
> Is there an issue specifying what is wrong? just curious?
> So running `glide get -v <package>` breaks?

Yes. It would panic with a nil pointer dereference. I can't reproduce it anymore. Who knows why. I have seen also : 

```
[INFO]  Keeping github.com/golang/protobuf ^1.2.0
[INFO]  --> Fetching github.com/google/go-querystring
[INFO]  --> Setting version for golang.org/x/oauth2 to a6bd8cefa1811bd24b86f8902872e4e8225f74c4.
[INFO]  --> Fetching github.com/json-iterator/go
[INFO]  --> Setting version for github.com/json-iterator/go to f2b4162afba35581b6d4a50d3b8f34e33c144682.
[ERROR] Error scanning golang.org/x/crypto/internal/chacha20: cannot find package "." in:
        /home/ycote/.glide/cache/src/https-golang.org-x-crypto/internal/chacha20
```

My point was to simply document that there may be glitches with version of glide. Coming from 'dep' and not knowing what to expect, I spent a lot of time trying to make the tool work and add a dependency trusting that the tool was not the issue. A note like this when searching docs would have helped me trying something different right away.

jim-minter:
@y-cote something is wrong, the tests are green.

/lgtm

y-cote:
> @y-cote something is wrong, the tests are green.
> 
> /lgtm

@jim-minter : true! even "readme text files" PRs fail passing tests with me ;-) my lucky day!

PR #1361, opened 2019-03-27
clean up e2e scripts

jim-minter:
```release-note
NONE
```

PR #1364, opened 2019-03-27
Add plugin config template override feature

### LINKED ISSUE #1207: Investigate moving away from custom env variables & running under test
We need to move away from the balooning numbers of custom environment variables in our env file and the use of running under test as a way to control log levels of openshift components.

One way might be to create a copy of the pluginconfig.yaml file and name it as `override.yaml` and gitignore it. We could configure the fakerp to pick up override values specified in this file. 

/kind feature

charlesakalugwu:
This PR allows us to override values set in the plugin config template file with values specified in a `pluginconfig/override.yaml` file.

For example, to change the log level of openshift components within a cluster from 2 (default) to 4, it suffices to create a gitignored `pluginconfig/override.yaml` file with the following content:

```yaml
componentLogLevel:
  apiServer: 4
  controllerManager: 4
  node: 4
```

The following is a list of currently overrideable plugin config template fields:

```
componentLogLevel.apiServer
componentLogLevel.controllerManager
componentLogLevel.node
imageOffer
imageVersion
images.format
images.azureControllers
images.canary
images.etcdBackup
images.metricsBridge
images.startup
images.sync
images.tlsProxy
```

```release-note
NONE
```

fixes #1207

/cc @jim-minter @mjudeikis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ cscope.out
 /id_rsa
 /releasenotes
 /vmimage
+/pluginconfig/override.yaml
--- a/env.example
+++ b/env.example
@@ -45,33 +45,6 @@ export DEPLOY_VERSION=v3.11
 #export IMAGE_RESOURCEGROUP=images
 #export IMAGE_RESOURCENAME=$(az image list -g $IMAGE_RESOURCEGROUP -o json --query "[?starts_with(name, '${DEPLOY_OS:-rhel7}-${DEPLOY_VERSION//v}') && tags.valid=='true'].name | sort(@) | [-1]" | tr -d '"')
 
-# If overriding the published marketplace image, uncomment and customise this
-# section
-#export IMAGE_OFFER=osa-preview
-#export IMAGE_VERSION=latest
-
-# If overriding the image used for the sync pod, uncomment and customise this
-# section.
-#export SYNC_IMAGE=quay.io/openshift-on-azure/sync:latest
-
-# If overriding the image used for the metricsbridge pod, uncomment and customise this
-# section.
-#export METRICSBRIDGE_IMAGE=quay.io/openshift-on-azure/metricsbridge:latest
-
-# If overriding the image used for the azure-controllers pod, uncomment and
-# customise this section.
-#export AZURE_CONTROLLERS_IMAGE=quay.io/openshift-on-azure/azure-controllers:latest
-
-# If overriding the image used for the backup container, customise this section.
-#export ETCDBACKUP_IMAGE=quay.io/openshift-on-azure/etcdbackup:latest
-
-# If overriding the image used for the canary container, customise this section.
-#export CANARY_IMAGE=quay.io/openshift-on-azure/canary:latest
-
-# If overriding the image used for the startup pod, uncomment and customise this
-# section.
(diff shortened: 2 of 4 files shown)


mjudeikis:
The challenge I see here is how we override these in our CI:
https://github.com/openshift/openshift-azure/pull/1361/files#diff-67ae1a77b052066bf531a69fcaa9bae1R8
Which will mean we have to "produce" overrides file in CI and populate it. And with recent mess in our release repo refactor I dont know if this is right time to do this. 

For the log level, What problem you are trying to solve?

charlesakalugwu:
> The challenge I see here is how we override these in our CI:
> https://github.com/openshift/openshift-azure/pull/1361/files#diff-67ae1a77b052066bf531a69fcaa9bae1R8
> Which will mean we have to "produce" overrides file in CI and populate it. And with recent mess in our release repo refactor I dont know if this is right time to do this.

Yes we would have to create override files in CI. Can you explain more what difficulties you see with this?

> 
> For the log level, What problem you are trying to solve?

For local development, new clusters are created with the default log level of 2. This is not enough especially when debugging bugs/flakes. So as an OSA engineer, I could create `pluginconfig/override.yaml` with log levels of 4 and this will ensure that every cluster I create from my machine comes up with loglevel 4 instead of the default 2.

thekad:
Could probably work in CI if we put the override in a configmap and the single environment variable we use from now on is `OVERRIDE_CONFIG=/path/to/override.yaml`.
I am tentatively +1 on this because in the future changing/adding/removing override values in development workstations and CI environments will mean just changing a single file/configmap instead of changing >1 config file in our workstations or >1 configmap in CI.
I do agree that this _may_ be an intrusive change to make right now, though.

mjudeikis:
I am still a little bit reluctant. First because if you want dev cluster with higher logging level - just change the logging level in the main file. Same effort, less complexity.

Second, now constructing the file for an override of the images will need some bash magic or again, one more template to do the same. So, in the end, we don't remove complexity, we just move it to another place. 

Third plugin template should be a single place for configurable, and we deviating from it now.

jim-minter:
@charlesakalugwu please keep your branch locally and add this in the architecture doc for discussion post-GA.  I don't think it makes sense for us to work on this now.

charlesakalugwu:
> I am still a little bit reluctant. First because if you want dev cluster with higher logging level - just change the logging level in the main file. Same effort, less complexity.
> 

....until you forget you made that change and check it in by mistake to git. The override.yaml will be gitignored and you can do whatever you want with it.

> Second, now constructing the file for an override of the images will need some bash magic or again, one more template to do the same. So, in the end, we don't remove complexity, we just move it to another place.
> 

I don't see the complexity you see.
What is so complex about:
cat > override.yaml <<EOL
componentLogLevel:
  apiServer: 4
  controllerManager: 4
  node: 4
EOL

> Third plugin template should be a single place for configurable, and we deviating from it now.

It is still the single place. We are not deviating from it, just overriding values without having to touch the actual git-tracked config file.

charlesakalugwu:
> @charlesakalugwu please keep your branch locally and add this in the architecture doc for discussion post-GA. I don't think it makes sense for us to work on this now.

ok

PR #1366, opened 2019-03-27
scriptfix - provide config file config

mjudeikis:
```release-note
NONE
```

We need to provide config so we can construct a route for canary application.


jim-minter:
this lgtm
not tagging until I'm confident that the test setup is sane to prevent overload

PR #1367, opened 2019-03-27
add secret fetch and download helper

mjudeikis:
Add secret refresh helper + confirmation check as its just question of time when somebody will mess it up :)

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ all: $(ALL_BINARIES)
 version:
 	echo ${TAG}
 
+confirm:
+	@echo -n "Are you sure? [y/N] " && read ans && [ $${ans:-N} == y ]
+
 clean:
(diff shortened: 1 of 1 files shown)


jim-minter:
/hold

I think we need to do something here, but I'm not sure this is it. As it stands I think PR could make it more likely for the secret to get screwed up.

mjudeikis:
> /hold
> 
> I think we need to do something here, but I'm not sure this is it. As it stands I think PR could make it more likely for the secret to get screwed up.

We can remove it and I will just add to my local aliases :) Im executing this command way too often... :D

jim-minter:
I think I'd like to have confidence that people's secrets directory always exactly matches the secret, or that we pull the secret every time we create a cluster, or something - otherwise it's too easy for breaking changes to be made

PR #1368, opened 2019-03-27
always call az group delete at end of CI run (extra protection)

jim-minter:
```release-note
NONE
```

### DIFF
--- a/hack/tests/e2e-upgrade.sh
+++ b/hack/tests/e2e-upgrade.sh
@@ -11,6 +11,7 @@ cleanup() {
     stop_monitoring
     make artifacts
     make delete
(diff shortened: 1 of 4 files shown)


jim-minter:
@mjudeikis we still seem to be leaking - will do another round of cleanup

hack/tests/e2e-upgrade.sh
```diff
@@ -11,6 +11,7 @@ cleanup() {
     stop_monitoring
     make artifacts
     make delete
+    az group delete -g "$RESOURCEGROUP" --yes --no-wait
```
mjudeikis: I really dont like now this no-wait as it makes next `/retest` broken. I would prefer for now blocking those but test results would not give failure with "delayed time to retest"

PR #1379, opened 2019-03-29
Adding template tests

### LINKED ISSUE #1345: Unit test which compiles all templates
https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v4/sync_test.go#L66 is a unit test which checks that all the pkg/sync templates compile and run correctly.

Need equivalent test for pkg/vmimage, pkg/startup, pkg/arm

thekad:
Closes #1345 

There are similar tests for `arm` and `sync` packages, just rounding up the rest of the packages that render templates. Changes:

* Adding comments to comply with golint
* Golint was complaining `startup.New` was returning a non-exported type
* Extracted the template rendering code out of `startup.writeFiles` to make it easier to write these unit tests
* Removed dummy coverage test files since there are real unit tests now

```release-note
NONE
```


pkg/vmimage/vmimage_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestGenerateTemplate(t *testing.T) {
+	_, hook := test.NewNullLogger()
+	timestamp := time.Now().UTC().Format("200601021504")
+	image := "rhel7-3.11-" + timestamp
+	ctx := context.Background()
+	domainNameLabel, err := random.LowerCaseAlphaString(20)
+	if err != nil {
+		t.Error(err)
+	}
+	b, err := ioutil.ReadFile("../../secrets/client-key.pem")
```
asalkeld: this is failing. the secrets are stored elsewhere in CI

asalkeld: you might have to create dummy keys/certs

pkg/startup/v3/startup.go
```diff
@@ -24,16 +24,19 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/writers"
 )
 
-type startup struct {
+// Startup is the main configuration struct for startup
+type Startup struct {
```
jim-minter: don't make this exported - I specifically want it unexported so that callers are forced to use startup.Interface, which this struct implements

pkg/startup/v3/startup_test.go
```diff
@@ -0,0 +1,92 @@
...
+			VMSize: api.StandardD4sV3,
+			OSType: api.OSTypeLinux,
+			Role:   api.AgentPoolProfileRoleCompute,
+		},
+	}
+
+	return &cs
+}
+
+func TestGenerateMasterTemplates(t *testing.T) {
```
jim-minter: we've collided, sorry.  I don't think tests are fundamentally adding anything that isn't in TestHashScaleSetStability, which has just merged.  That test cannot succeed if the templates don't generate correctly.  I think all the changes to pkg/startup/... in this PR can be reverted.

pkg/vmimage/vmimage_test.go
jim-minter: reuse populate.DummyCertsAndKeys(cs), which has just merged.  See TestHashScaleSetStability for usage.

pkg/vmimage/vmimage_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	authorizer, err := azureclient.NewAuthorizerFromEnvironment("")
+	if err != nil {
+		t.Error(err)
+	}
+
+	builder := Builder{
+		GitCommit:                "unknown",
```
jim-minter: please leave as many things as possible at their zero value instead of providing dummy values, it makes ongoing maintenance easier.
i.e., if the test passes when you remove the line, say, `GitCommit: ""`, remove the line, etc.

pkg/vmimage/vmimage_test.go
thekad: I don't use an `OpenShiftCluster` object here, so just using the `tls.*` calls as `DummyCertsAndKeys()` does

pkg/vmimage/vmimage_test.go
thekad: ptal

pkg/vmimage/vmimage_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/openshift/openshift-azure/test/util/tls"
+)
+
+func TestGenerateTemplate(t *testing.T) {
```
jim-minter: good, although this actually boils down all the way to:
```golang
func TestGenerateTemplate(t *testing.T) {
	builder := Builder{
		SSHKey:     tls.GetDummyPrivateKey(),
		ClientKey:  tls.GetDummyPrivateKey(),
		ClientCert: tls.GetDummyCertificate(),
	}

	_, err := builder.generateTemplate()
	if err != nil {
		t.Error(err)
	}
}
```

jim-minter: if you fix the top of populate.DummyCertsAndKeys as follows:
```golang
func DummyCertsAndKeys(v interface{}) {
	var walk func(v reflect.Value)

	walk = func(v reflect.Value) {
		if v.Type().PkgPath() != "" &&
			(!strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/") ||
				strings.HasPrefix(v.Type().PkgPath(), "github.com/openshift/openshift-azure/vendor/")) {
			return
		}
```

then this reduces to:
```golang
func TestGenerateTemplate(t *testing.T) {
	var builder Builder
	populate.DummyCertsAndKeys(&builder)

	_, err := builder.generateTemplate()
	if err != nil {
		t.Error(err)
	}
}
```

jim-minter: (@thekad if this was black magic - worth reading through the function, understanding it and adding a docstring.  We already use it a lot and are going to use it more).

PR #1380, opened 2019-03-29
azure disk fix

mjudeikis:
@jim-minter 
@openshift/sig-azure 

```release-note
HotFix for AzureDisk issue https://github.com/kubernetes/kubernetes/pull/70002 
```


jim-minter:
/hold

If we use we want to merge this, I think:
* it should be in the disk image, not the startup script
* we should check at overwrite time that the version we're overwriting is what we think it is or error
* we should have a clear plan and timeline to get the fix into ocp

PR #1383, opened 2019-03-29
restrict #pods/node to 50 for now

jim-minter:
```release-note
restrict #pods/node to 50 as a starting point for GA
```


pkg/vmimage/data/etc/origin/node/bootstrap-node-config.yaml
```diff
@@ -20,22 +20,26 @@ imageConfig:
...
+  feature-gates:
+  - RotateKubeletClientCertificate=true,RotateKubeletServerCertificate=true
+  max-pods:
+  - "50"
+  pod-manifest-path:
+  - /etc/origin/node/pods
+  rotate-certificates:
+  - "true"
+  runtime-request-timeout:
+  - 10m
```
asalkeld: this is new, what does it do?

mjudeikis: How long kubelet will try to GET,PUT, to `docker` to get stuff out. Does not include watchers (like logs).

PR #1386, opened 2019-04-01
wait for each lb to be ready before creating the next

asalkeld:
relates to #1336 

This is somewhat speculative.. see the issue

```release-notes
NONE
```

PR #1390, opened 2019-04-01
Revert "wait for each lb to be ready before creating the next"

jim-minter:
Reverts openshift/openshift-azure#1386

```release-note
NONE
```


jim-minter:
/hold

@mjudeikis fyi
@asalkeld think this got merged a bit too quickly - let's chat about this

jim-minter:
/lgtm

not against merging a workaround like this, but only with the necessary work in flight to backport a fix from upstream which deals with the root cause.

PR #1391, opened 2019-04-01
Use ci credentials in ci

jim-minter:
```release-note
NONE
```


hack/tests/ci-operator-prepare.sh
```diff
@@ -15,10 +15,20 @@ set_build_images() {
 }
 
 setup_secrets() {
-    ln -sf $SECRET_DIR secrets
+    if [[ -e /usr/local/e2e-secrets/azure ]]; then
+        ln -sf /usr/local/e2e-secrets/azure secrets
+    fi
```
asalkeld: this will break local development because you aren't creating a link from the github.com/openshift/openshift-azure to the versioned git repo that has been cloned into /tmp

jim-minter:
@asalkeld had another go - ptal
(the reason I'd like to go this way is to avoid the overloaded use of setup_secrets that we ended up with - 1) to optionally load secrets from /usr/local/e2e-secrets/azure and 2) to copy the pre-existing secrets to $T)

mjudeikis:
This is becoming super complex - again! we just moved complexity from release repo inhouse (which is not a bad thing). But can we add some comments to it as even I get things, it makes me sit and think why one `ln -s` there and here...

hack/tests/ci-operator-prepare.sh
```diff
@@ -58,8 +47,18 @@ export RESOURCEGROUP_TTL=4h
 prdetail="$(python -c 'import json, os; o=json.loads(os.environ["CLONEREFS_OPTIONS"]); print "%s-%s-" % (o["refs"][0]["pulls"][0]["author"].lower(), o["refs"][0]["pulls"][0]["number"])' 2>/dev/null || true)"
 export RESOURCEGROUP="$(basename "$0" .sh)-$prdetail$(cat /dev/urandom | tr -dc 'a-z' | fold -w 6 | head -n 1)"
 
-setup_secrets
+ln -sf /usr/local/e2e-secrets/azure secrets
 
 set +x
+. ./secrets/secret
+if [[ -e /usr/local/e2e-secrets/azure ]]; then
```
mjudeikis: I think this is not needed as we already define/checked if we are running in CI at line 38th.

mjudeikis:
I would say lgtm and merge and fix with some other PR?

hack/tests/ci-operator-prepare.sh
jim-minter: oh yeah, duh ;)

Issue #1399, opened 2019-04-02
panic The Resource 'Microsoft.Network/publicIPAddresses/ip-apiserver' under resource group 'e2e-create-jim-minter-1392-ujfffl' was not found.


jim-minter:
race condition of some sort

panic: network.PublicIPAddressesClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceNotFound" Message="The Resource 'Microsoft.Network/publicIPAddresses/ip-apiserver' under resource group 'e2e-create-jim-minter-1392-ujfffl' was not found."

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1392/pull-ci-openshift-openshift-azure-master-e2e-azure/3270/build-log.txt

Issue #1398, opened 2019-04-02
"waiting for OpenShiftManagedCluster config to be persisted" is spammy


jim-minter:
time="2019-04-02T00:28:54Z" level=debug msg="waiting for OpenShiftManagedCluster config to be persisted" component=monitor
time="2019-04-02T00:28:59Z" level=debug msg="waiting for OpenShiftManagedCluster config to be persisted" component=monitor
time="2019-04-02T00:29:04Z" level=debug msg="waiting for OpenShiftManagedCluster config to be persisted" component=monitor
time="2019-04-02T00:29:09Z" level=debug msg="waiting for OpenShiftManagedCluster config to be persisted" component=monitor
time="2019-04-02T00:29:14Z" level=debug msg="waiting for OpenShiftManagedCluster config to be persisted" component=monitor

PR #1405, opened 2019-04-02
monitoring remove chattiness and race condition waits

mjudeikis:
```release-note
NONE
```
fix: https://github.com/openshift/openshift-azure/issues/1398
fix: https://github.com/openshift/openshift-azure/issues/1399 
remove chattiness and adds waits for race conditions

### DIFF
--- a/cmd/monitoring/monitoring.go
+++ b/cmd/monitoring/monitoring.go
@@ -72,8 +72,8 @@ func (m *monitor) init(ctx context.Context, log *logrus.Entry) error {
 
 func (m *monitor) listResourceGroupMonitoringHostnames(ctx context.Context, subscriptionID, resourceGroup string) (hostnames []string, err error) {
 	// get dedicated routes we want to monitor
+	m.log.Debug("waiting for OpenShiftManagedCluster config to be persisted")
 	for {
-		m.log.Debug("waiting for OpenShiftManagedCluster config to be persisted")
 		oc, err := loadOCConfig()
 		if err != nil {
 			time.Sleep(5 * time.Second)
@@ -83,6 +83,7 @@ func (m *monitor) listResourceGroupMonitoringHostnames(ctx context.Context, subs
 		}
 	}
 	// get all external IP's used by VMSS
+	m.log.Debug("waiting for ss-masters ip addresses")
 	hostnames = []string{}
 	for iter, err := m.pipcli.ListVirtualMachineScaleSetPublicIPAddressesComplete(ctx, resourceGroup, "ss-master"); iter.NotDone(); err = iter.Next() {
 		if err != nil {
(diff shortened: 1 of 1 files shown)


mjudeikis:
/retest 
no containers - no logs... hmmm

cmd/monitoring/monitoring.go
```diff
@@ -83,18 +83,20 @@ func (m *monitor) listResourceGroupMonitoringHostnames(ctx context.Context, subs
 		}
 	}
 	// get all external IP's used by VMSS
+	m.log.Debug("waiting for ss-masters ip addresses")
 	hostnames = []string{}
 	for iter, err := m.pipcli.ListVirtualMachineScaleSetPublicIPAddressesComplete(ctx, resourceGroup, "ss-master"); iter.NotDone(); err = iter.Next() {
 		if err != nil {
-			return nil, err
+			time.Sleep(5 * time.Second)
```
jim-minter: do we not lose IP addresses if we do this?

mjudeikis: We might. But we will see less IP in the metrics. 
If this returns err - we have a bigger problem. Not sure we want to try catching ARM errors here as it should be part of plugin code

jim-minter: we're not aligned.  these days the fake rp writes _data/containerservice.yaml before (and after) the cluster is created so you *will* end up here before the IPs exist.  If you can, poll locally to be sure the cluster is created (can you read the provisioning state perhaps?) then fetch all the IPs.  Adding `time.Sleep(5 * time.Second)` without a loop basically achieves nothing.

mjudeikis: Got you. Bringing back loop. provisioning state means I still call "something" each 5 sec to get state, and after that, I call IP addresses and LB. So this would mean one more client to configure. I think this is simpler.

mjudeikis: There is issue with it now: https://github.com/openshift/openshift-azure/issues/1416 Fixing in separate PR.

mjudeikis:
@jim-minter If we could merge this, until we sort out https://github.com/openshift/openshift-azure/pull/1418/files . and when done we can move to monitor provisioning state as now it's a little bit broken.

cmd/monitoring/monitoring.go
```diff
@@ -72,31 +72,47 @@ func (m *monitor) init(ctx context.Context, log *logrus.Entry) error {
...
-			hostnames = append(hostnames, *iter.Value().IPAddress)
+	m.log.Debug("waiting for ss-masters ip addresses")
+	for {
+		for iter, err := m.pipcli.ListVirtualMachineScaleSetPublicIPAddressesComplete(ctx, resourceGroup, "ss-master"); iter.NotDone(); err = iter.Next() {
+			if err != nil {
+				m.log.Error(err)
+				time.Sleep(10 * time.Second)
+			} else if iter.Value().IPAddress != nil {
+				hostnames = append(hostnames, *iter.Value().IPAddress)
+				break
```
jim-minter: I don't think this break should be here

cmd/monitoring/monitoring.go
```diff
@@ -72,31 +72,47 @@ func (m *monitor) init(ctx context.Context, log *logrus.Entry) error {
...
+			}
+		}
+		if err == nil && len(hostnames) >= 4 {
```
jim-minter: "4" is very ugly

jim-minter:
force pushed
/lgtm

Issue #1381, opened 2019-03-29
Improve Geneva log indexing/scoping


jim-minter:
At the moment we can scope Geneva log volume server side down to a known cluster.

It would be useful to be able to carry out cross-cluster log searches against a particular component (e.g. search kernel/kubelet/apiserver/etc. logs across all clusters).

First step is to be able to expose the right fields at low cardinality for indexing.  SYSLOG_IDENTIFIER is already pretty useful, for example.
1) It would be good to expose a field for the agentpoolprofile role (static mdsd configuration?)
2) Also pod identifiers (namespace, pod name with reduced cardinality, container name).  fluend configuration using string matching against CONTAINER_NAME field.  Examples:

k8s_apiserver_apiserver-vqxg4_kube-service-catalog_72ce4d73-5224-11e9-98d0-000d3a196756_0

split by `_`
k8s -> discard
apiserver -> new k8s container name field
apiserver-vqxg4 -> remove trailing `-.*?$` -> new k8s pod name field
kube-service-catalog -> new k8s namespace field
72ce4d73-5224-11e9-98d0-000d3a196756 -> discard
0 -> discard

Once we've got the right fields, we can then think about and add suitable indexes server side.

PR #1406, opened 2019-04-02
fluentd config for split kubernetes logs

mjudeikis:
```release-note
```
@jim-minter this potentially could be one of the solutions for https://github.com/openshift/openshift-azure/issues/1381

It still does not have `ROLE`, which is hard to get as it depends on the node where agent is starting. 
One option would be to lay role file in `/etc/origin/cloudprovider/agentpoolprofile` and read it in fluentD and add to columns. 

Few comments in the PR code too


pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -9,19 +9,58 @@ data:
       </storage>
       tag logs
     </source>
-    <match logs>
+     # drop unused columns
+    <filter logs>
+      @type record_transformer
+      remove_keys _BOOT_ID,_CAP_EFFECTIVE,_MACHINE_ID,_SOURCE_REALTIME_TIMESTAMP,_STREAM_ID,_TRANSPORT,_SOURCE_MONOTONIC_TIMESTAMP
+    </filter>
```
mjudeikis: we already have 30+ columns. I think we can drop even more.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -9,19 +9,58 @@ data:
...
       <rule>
         key MESSAGE
         pattern audit\.k8s\.io
         tag audit
       </rule>
+      <rule>
+        key CONTAINER_NAME
+        pattern ^k8s_
+        tag kubernetes
+      </rule>
```
mjudeikis: We need this, otherwise logic becomes a little bit more complex. This keeps it simple.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -9,19 +9,58 @@ data:
...
+     </filter>
+     # retag splitted kubernetes logs to journald
+     <match kubernetes>
+        @type rewrite_tag_filter
+        <rule>
+          key CONTAINER
+          pattern .+
+          tag journald
+        </rule>
+     </match>
```
mjudeikis: If we dont retag, it goes out as new "table" and we have to modify mdsd configuration too.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
jim-minter: is anyone complaining about these?  I'm inclined just to leave them in for now

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -9,19 +9,58 @@ data:
...
+      </rule>
       <rule>
         key MESSAGE
         pattern .+
         tag journald
       </rule>
     </match>
+    # filter kubernetes runtime logs for transformation
+    <filter kubernetes >
+      @type kubernetes_metadata
```
jim-minter: I was hoping we could use regexps: no dependency on kubernetes actually being up, allows us to remove the reader roles

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
mjudeikis: just me. The geneva UI with all the output just trips me :)

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
mjudeikis: This is already part of td-agent image and it makes code simpler. Considering its ruby my vote is for "easier to read", as debugging issues with fluentd and ruby is pain :/ 

But if you want we can do it but it will not make it less complex.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
mjudeikis: + kubernetes_metadata is RH filter :)

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
jim-minter: I want to use regexps so that there is no dependency on apiservers and etcd to be up for our logs to work.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
jim-minter: please leave them in for now

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
jim-minter: this filter should be removed now, right?

mjudeikis: will remove it. Need to sort out regexps.

mjudeikis:
/hold
I can see sync pod generating change due linting

thekad:
/hold
I think you lost the changes you intended to have somewhere along the way

mjudeikis:
@thekad thanks. rebase and hash tests :/

pkg/sync/v5/derive.go
```diff
@@ -126,3 +126,53 @@ func (derivedType) MDSDConfig(cs *api.OpenShiftManagedCluster) (string, error) {
...
+  <source>
+    @type systemd
+    <storage>
+      @type local
+      path /var/log/journald.pos
+    </storage>
+    tag logs
+  </source>
+  <match logs>
+  @type rewrite_tag_filter
```
thekad: needs indentation

pkg/sync/v5/derive.go
```diff
@@ -126,3 +126,53 @@ func (derivedType) MDSDConfig(cs *api.OpenShiftManagedCluster) (string, error) {
 
 	return string(b.Bytes()), nil
 }
+
+func (derivedType) TDAgentConfig() ([]interface{}, error) {
```
jim-minter: why do you need to pull this out into a derived function?

mjudeikis: The problem was Linting marshal and unmarshal test. But after some playing with indentation looks like it is ok now :/

PR #1414, opened 2019-04-03
increase debugging to understand ongoing storage issue

jim-minter:
```release-note
NONE
```

### DIFF
--- a/test/e2e/standard/sanitychecker.go
+++ b/test/e2e/standard/sanitychecker.go
@@ -2,6 +2,10 @@ package standard
 
 import (
 	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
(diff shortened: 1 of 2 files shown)


test/e2e/standard/sanitychecker.go
```diff
@@ -83,6 +87,35 @@ func (sc *SanityChecker) CreateTestApp(ctx context.Context) (interface{}, []*Tes
 	return namespace, errs
 }
 
+func (sc *SanityChecker) debugValidateTestApp() error {
```
mjudeikis: `TestApp`, where this is coming from?

jim-minter: see function name immediately below ;)

PR #1415, opened 2019-04-03
new metrics from etcd monitoring tlsproxy

Makdaam:
```release-note
NONE
```
also updated example metrics-bridge config


pkg/sync/v5/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -149,6 +149,18 @@ data:
     - query: 'rate(storage_operation_errors_total{namespace=~"default|openshift|openshift-.+|kube-.+|", volume_plugin="kubernetes.io/azure-disk"}[1m])'
       name: storage_operation_errors_rate
     - query: 'up'
+    - query: 'etcd_server_proposals_pending{namespace="openshift-azure-monitoring"} != 0'
+    - query: 'rate(etcd_server_proposals_failed_total{namespace="openshift-azure-monitoring"}[1m]) != 0'
+      name: etcd_server_proposals_failed_rate
+    - query: 'etcd_server_has_leader{namespace="openshift-azure-monitoring"}'
+    - query: 'rate(etcd_server_leader_changes_seen_total{namespace="openshift-azure-monitoring"}[1m]) != 0'
```
jim-minter: is the rationale that the raw metric should be stable, but if it increases it potentially indicates a problem?
the worry I have is that because the rate period (1m) == the metrics-bridge period, isn't there a small risk we could just miss the change?
if the metric is dropped for whatever reason we'd also miss it - effectively rate() puts us in edge triggered mode.
is it better to forward the _total value instead or as well?
(also applies to rate(etcd_server_proposals_failed_total))

Makdaam: agreed, 1m sampling + 1m rate calculation can lead to missing a datapoint. Removed _rate and added _total back. For leader_changes and proposals_failed)

mjudeikis:
@Makdaam rebase please

mjudeikis:
/lgtm
I can see you have linter from 1.11 master :)

PR #1418, opened 2019-04-03
drop provisioning state when converting to internal model

mjudeikis:
```release-note
Drop ProvisioningState field from customer API in the converters. 
```
For discussion. 
currently, the provisioning state is overwritten from external API.

Example:
1. Customer checks his cluster state and time x. Gets result with provisioning state in.
2. We initiate cluster upgrade via api, so state now is `Upgrading`.
3. Cluster initiates any permitted api change (scale-up) using data payload he got from case 1. 
    If not handled this might cause state to be updated. 

This might not be a case as it depends a lot on how it is handled in RealRP. But at this point, I think it is safer to drop fields like this from externalAPI in the converters, when reading the payload.

This still missing tests to cover it. But wanted to put this as I might miss some context.


mjudeikis:
/hold
need to think little bit more :)

PR #1421, opened 2019-04-03
Enabling kdump at kernel boot time

### LINKED ISSUE #1396: kdump not configured
kdump isn't configured on our nodes and there is no console serial access using VMSS.

install kdump into our VM image and configure it.  Test manually that provoking a kernel panic results in a usable dump and that the machine reboots.

thekad:
Closes #1396 
Explicitly add kexec-tools and enable kernel parameters plus kdump systemd service. Tested with a custom image.
Additionally: a couple of changes to the `hack/vmimage.sh` script and `env.example` files to facilitate the creation of custom images.

```release-note
NONE
```


env.example
```diff
@@ -48,6 +48,11 @@ export DEPLOY_VERSION=v3.11
 #export IMAGE_RESOURCEGROUP=images
 #export IMAGE_RESOURCENAME=$(az image list -g $IMAGE_RESOURCEGROUP -o json --query "[?starts_with(name, '${DEPLOY_OS:-rhel7}-${DEPLOY_VERSION//v}') && tags.valid=='true'].name | sort(@) | [-1]" | tr -d '"')
 
+# If building a custom image, uncomment the following lines and re-source your env
+#export IMAGE_RESOURCEGROUP="${USER}-images"
+#export IMAGE_RESOURCENAME="${DEPLOY_OS:-rhel7}-${DEPLOY_VERSION//v}-$(date -u +%Y%m%d%H%M)"
```
jim-minter: s/${DEPLOY_OS:-rhel7}/rhel7/

(I have a separate PR in parallel removing DEPLOY_OS across the board)

pkg/vmimage/data/etc/default/grub
```diff
@@ -0,0 +1,8 @@
+GRUB_TIMEOUT=5
+GRUB_DISTRIBUTOR="$(sed 's, release .*$,,g' /etc/system-release)"
+GRUB_DEFAULT=saved
+GRUB_DISABLE_SUBMENU=true
+GRUB_TERMINAL="serial console"
+GRUB_SERIAL_COMMAND="serial --speed=115200"
+GRUB_CMDLINE_LINUX="console=ttyS0,115200n8 crashkernel=128M"
```
jim-minter: I think by removing the line `%addon com_redhat_kdump --disable` in script.sh you may not need this file to be checked in? ptal.

pkg/vmimage/data/etc/kdump.conf
```diff
@@ -0,0 +1,2 @@
+path /var/crash
```
jim-minter: isn't the default /etc/kdump.conf good enough?

pkg/vmimage/data/script.sh
```diff
@@ -207,10 +207,15 @@ sed -i -e 's/^ResourceDisk.Format=.*/ResourceDisk.Format=n/' /etc/waagent.conf
 rpm -q kernel --last | sed -n '1 {s/^[^-]*-//; s/ .*$//; p}' >/var/tmp/kernel-version
 rpm -q atomic-openshift-node --qf '%{VERSION}-%{RELEASE}.%{ARCH}' >/var/tmp/openshift-version
 
+# enabling kdump
+grub2-mkconfig -o /boot/grub2/grub.cfg
+systemctl enable kdump.service
```
jim-minter: this just has the effect of creating a symlink -- please check that symlink into pkg/vmimage/data/etc/systemd/system/multi-user.target.wants instead

pkg/vmimage/data/script.sh
```diff
@@ -207,10 +207,15 @@ sed -i -e 's/^ResourceDisk.Format=.*/ResourceDisk.Format=n/' /etc/waagent.conf
 rpm -q kernel --last | sed -n '1 {s/^[^-]*-//; s/ .*$//; p}' >/var/tmp/kernel-version
 rpm -q atomic-openshift-node --qf '%{VERSION}-%{RELEASE}.%{ARCH}' >/var/tmp/openshift-version
 
+# enabling kdump
+grub2-mkconfig -o /boot/grub2/grub.cfg
```
jim-minter: may not be needed if `%addon com_redhat_kdump --disable` is removed

pkg/vmimage/data/etc/kdump.conf
thekad: those are the default values, I just removed comments

jim-minter: don't bother overwriting the default configuration if it's fine

pkg/vmimage/data/etc/default/grub
thekad: yep that was right

jim-minter:
nice! thanks!

PR #1422, opened 2019-04-03
Export 'Log' from the SanityChecker struct.

y-cote:
This PR exposes the logger so that the SanityChecker (that initialize various structs and the openshift client for testing) can be reused in isolated test.

```release-note
- E2E: standard exports the Log field from the SanityChecker for better reusability 
```

### DIFF
--- a/test/e2e/standard/sanitychecker.go
+++ b/test/e2e/standard/sanitychecker.go
@@ -33,7 +33,7 @@ type DeepTestInterface interface {
 }
 
 type SanityChecker struct {
-	log    *logrus.Entry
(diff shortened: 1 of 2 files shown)


jim-minter:
@y-cote this is fine but you must remove the changes to pkg/vmimage before it merges - why are they there?

y-cote:
> @y-cote this is fine but you must remove the changes to pkg/vmimage before it merges - why are they there?

It's hard to say. The minute I type "make", those are generated differently than what I checked out. Will try to figure out why.

PR #1429, opened 2019-04-04
fakerp refactor

mjudeikis:
```release-note
NONE
```

Refactor fake RP to use files as "storage"
Persist state into "storage" on methods
Split some of the methods where logic started to become overcomplicated.

### DIFF
--- a/pkg/fakerp/admin_handlers.go
+++ b/pkg/fakerp/admin_handlers.go
@@ -10,11 +10,7 @@ import (
 
 // handleBackup handles admin requests to backup an etcd cluster
 func (s *Server) handleBackup(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	backupName := chi.URLParam(req, "backupName")
 
@@ -24,11 +20,7 @@ func (s *Server) handleBackup(w http.ResponseWriter, req *http.Request) {
 
 // handleGetControlPlanePods handles admin requests for the list of control plane pods
 func (s *Server) handleGetControlPlanePods(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	pods, err := s.plugin.GetControlPlanePods(req.Context(), cs)
 	s.adminreply(w, err, pods)
@@ -36,11 +28,7 @@ func (s *Server) handleGetControlPlanePods(w http.ResponseWriter, req *http.Requ
 
 // handleListClusterVMs handles admin requests for the list of cluster VMs
 func (s *Server) handleListClusterVMs(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	vms, err := s.plugin.ListClusterVMs(req.Context(), cs)
 	s.adminreply(w, err, vms)
@@ -48,11 +36,7 @@ func (s *Server) handleListClusterVMs(w http.ResponseWriter, req *http.Request) 
 
 // handleReimage handles reimaging a vm in the cluster
 func (s *Server) handleReimage(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	hostname := chi.URLParam(req, "hostname")
 
@@ -62,11 +46,7 @@ func (s *Server) handleReimage(w http.ResponseWriter, req *http.Request) {
 
 // handleListBackups handles admin requests to list etcd backups
 func (s *Server) handleListBackups(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	backups, pluginErr := s.plugin.ListEtcdBackups(req.Context(), cs)
 	var err error
@@ -79,11 +59,7 @@ func (s *Server) handleListBackups(w http.ResponseWriter, req *http.Request) {
 
 // handleRestore handles admin requests to restore an etcd cluster from a backup
 func (s *Server) handleRestore(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	backupName := chi.URLParam(req, "backupName")
 
@@ -98,11 +74,7 @@ func (s *Server) handleRestore(w http.ResponseWriter, req *http.Request) {
 
 // handleRotateSecrets handles admin requests for the rotation of cluster secrets
 func (s *Server) handleRotateSecrets(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	deployer := GetDeployer(s.log, cs, s.testConfig)
 	pluginErr := s.plugin.RotateClusterSecrets(req.Context(), cs, deployer)
@@ -117,11 +89,7 @@ func (s *Server) handleRotateSecrets(w http.ResponseWriter, req *http.Request) {
 
 // handleForceUpdate handles admin requests for the force updates of clusters
 func (s *Server) handleForceUpdate(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	pluginErr := s.plugin.ForceUpdate(req.Context(), cs, GetDeployer(s.log, cs, s.testConfig))
 	var err error
@@ -134,11 +102,7 @@ func (s *Server) handleForceUpdate(w http.ResponseWriter, req *http.Request) {
 
 // handleRunCommand handles running generic commands on a given vm within a scaleset in the cluster
 func (s *Server) handleRunCommand(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
 
 	hostname := chi.URLParam(req, "hostname")
 	command := chi.URLParam(req, "command")
--- a/pkg/fakerp/admin_handlers.go
+++ b/pkg/fakerp/admin_handlers.go
@@ -10,7 +10,7 @@ import (
 
 // handleBackup handles admin requests to backup an etcd cluster
 func (s *Server) handleBackup(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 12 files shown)


charlesakalugwu:
@mjudeikis I think it might be worth discussing how to generalize this work to support multiple clusters don't you think?

configs can be stored in `_data/{subscriptionId}/{resourceGroup}/{clusterName}`...

mjudeikis:
This is super close to having multiple cluster support. My initial implementation was doing this under `_data/{clusterkey}/files` and `store` was accepting `key` as an argument. Just question is if we need it. 

For sure discussion :)

mjudeikis:
Testing the flakes in logic, but ready for review

asalkeld:
looking good

mjudeikis:
@jim-minter, @charlesakalugwu wdyt?

pkg/fakerp/store/type.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (s *Storage) Delete() error {
+	s.Lock()
+	defer s.Unlock()
+	s.log.Debugf("storage: Delete %s/%s", s.storageDir, containerServices)
+	return os.RemoveAll(fmt.Sprintf("%s/%s", s.storageDir, containerServices))
+}
+
+func (s *Storage) UpdateState(state api.ProvisioningState) error {
```
jim-minter: I don't think this function should exist at all.  Callers should use Put().

pkg/fakerp/admin_handlers.go
```diff
@@ -1,58 +1,40 @@
...
+	cs, err := s.store.Get()
+	if cs == nil || err != nil {
 		s.internalError(w, "Failed to read the internal config")
 		return
 	}
 
 	backupName := chi.URLParam(req, "backupName")
 
-	err := s.plugin.BackupEtcdCluster(req.Context(), cs, backupName)
+	err = s.store.UpdateState(api.Updating)
```
jim-minter: why does the fake rp need to do this?  operations are serialised and non-crash-recoverable.  We will make our lives much easier if we don't attempt to do this.

pkg/fakerp/admin_handlers.go
```diff
@@ -1,58 +1,40 @@
...
-	w.Header().Set("Content-Type", "application/json")
-	w.Write(b)
-	return
-}
-
 // handleBackup handles admin requests to backup an etcd cluster
 func (s *Server) handleBackup(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
+	cs, err := s.store.Get()
```
jim-minter: every handler needs a cs.  Every put handler writes a cs.  DRY.  Try to move the get and optional put logic into context() in middleware.go - it's mad to have it duplicated in every handle function.

pkg/fakerp/admin_handlers.go
```diff
@@ -106,71 +96,93 @@ func (s *Server) handleListBackups(w http.ResponseWriter, req *http.Request) {
...
 		s.internalError(w, "Failed to read the internal config")
 		return
 	}
 
 	backupName := chi.URLParam(req, "backupName")
 
-	pluginErr := s.plugin.RecoverEtcdCluster(req.Context(), cs, GetDeployer(s.log, cs, s.testConfig), backupName)
-	var err error
-	if pluginErr != nil {
-		// TODO: fix this nastiness: https://golang.org/doc/faq#nil_error
```
jim-minter: on what basis are you removing this code?  I don't believe this PR is fixing the underlying issue, so it can't go.

pkg/fakerp/customer_handlers.go
```diff
@@ -95,10 +96,10 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 		// long-running operation can exit successfully.
 		if resp.StatusCode == http.StatusOK {
 			s.log.Infof("deleted resource group %s", resourceGroup)
-			s.write(nil)
+			s.store.Delete()
```
jim-minter: I'd like to get rid of the asynchronous delete - it adds complexity for no real benefit.  That does not imply removing NO_WAIT, which I'd like to keep.

pkg/fakerp/customer_handlers.go
```diff
@@ -109,83 +110,110 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
...
 
-func (s *Server) handlePut(w http.ResponseWriter, req *http.Request) {
+func (s *Server) handleAdminPut(w http.ResponseWriter, req *http.Request) {
 	// read old config if it exists
-	var oldCs *internalapi.OpenShiftManagedCluster
+	var oldCs *api.OpenShiftManagedCluster
 	var err error
-	if !shared.IsUpdate() {
-		s.writeState(internalapi.Creating)
+	if shared.IsUpdate() {
```
jim-minter: note: admin put must always be an update

pkg/fakerp/customer_handlers.go
```diff
@@ -109,83 +110,110 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
...
+	oc := v20190430.FromInternal(cs)
+	res, err = json.Marshal(oc)
+	if err != nil {
+		s.internalError(w, fmt.Sprintf("Failed to marshal response: %v", err))
+		return
 	}
+	w.Write(res)
+}
+
+func (s *Server) adminReplay(w http.ResponseWriter, req *http.Request) {
```
jim-minter: what is this function for?

pkg/fakerp/customer_handlers.go
```diff
@@ -109,83 +110,110 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 }
 
 func (s *Server) handleGet(w http.ResponseWriter, req *http.Request) {
-	s.reply(w, req)
+	s.customerReply(w, req)
 }
 
-func (s *Server) handlePut(w http.ResponseWriter, req *http.Request) {
+func (s *Server) handleAdminPut(w http.ResponseWriter, req *http.Request) {
```
jim-minter: I think I'm -1 on splitting handlePut -> handle{Admin,Customer}Put.  I think refactors can be done to neaten handlePut instead.  Is there a good reason for splitting that I'm missing?

pkg/fakerp/fakerp.go
```diff
@@ -229,25 +224,33 @@ func createOrUpdate(ctx context.Context, p api.Plugin, log *logrus.Entry, cs, ol
...
 	if err != nil {
 		return nil, err
 	}
 
-	log.Info("plugin createorupdate")
-	deployer := GetDeployer(log, cs, testConfig)
-	if err := p.CreateOrUpdate(ctx, cs, oldCs != nil, deployer); err != nil {
-		return nil, err
+	if oldCs != nil && isAdmin {
+		cs.Properties.ProvisioningState = api.AdminUpdating
```
jim-minter: again: why do we need to do this?

pkg/fakerp/store/type.go
```diff
@@ -0,0 +1,85 @@
...
+	defer s.Unlock()
+	s.log.Debugf("storage: Put %s/%s", s.storageDir, containerServices)
+	bytes, err := yaml.Marshal(cs)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(fmt.Sprintf("%s/%s", s.storageDir, containerServices), bytes, 0600)
+}
+
+func (s *Storage) Get() (*api.OpenShiftManagedCluster, error) {
```
jim-minter: I suspect this will need to relax to return a nil cs and no error if the file doesn't exist

pkg/fakerp/store/type.go
```diff
@@ -0,0 +1,85 @@
...
+type Store interface {
+	Put(value *api.OpenShiftManagedCluster) error
+	Get() (*api.OpenShiftManagedCluster, error)
+	UpdateState(state api.ProvisioningState) error
+	Delete() error
+}
+
+type Storage struct {
+	storageDir string
+	sync.Mutex
```
jim-minter: why is the mutex needed?  what do we need to do to get rid of it?

pkg/fakerp/admin_handlers.go
mjudeikis: This adds some ambiguity in the error handling when we need to know if this is first Put create, or update Put, etc. We cant error on non-existing record/file because of case 1.

pkg/fakerp/store/type.go
mjudeikis: Did a slightly different implementation of `store`.

pkg/fakerp/customer_handlers.go
mjudeikis: done

mjudeikis:
@jim-minter would you mind giving review one more time? little bit different implementation

pkg/fakerp/store/store.go
```diff
@@ -0,0 +1,123 @@
...
+		s.log.Debugf("Using '%s' (database already exists)", dir)
+		return s, nil
+	}
+
+	// if the database doesn't exist create it
+	s.log.Debugf("Creating database at '%s'", dir)
+	return s, os.MkdirAll(dir, 0755)
+}
+
+func (s *Storage) Put(key string, b []byte) error {
```
jim-minter: why not work with *api.OpenShiftManagedCluster instead of []byte?

pkg/fakerp/store/store.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if err := ioutil.WriteFile(tmpPath, b, 0644); err != nil {
+		return err
+	}
+
+	return os.Rename(tmpPath, fnlPath)
+}
+
+// Get a record from the database
+func (s *Storage) Get(key string) ([]byte, error) {
```
jim-minter: why not work with *api.OpenShiftManagedCluster instead of []byte?

pkg/fakerp/admin_handlers.go
```diff
@@ -1,47 +1,16 @@
...
-}
-
 // handleBackup handles admin requests to backup an etcd cluster
 func (s *Server) handleBackup(w http.ResponseWriter, req *http.Request) {
-	cs := s.read()
-	if cs == nil {
-		s.badRequest(w, "Failed to read the internal config")
-		return
-	}
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
```
jim-minter: Good! :-D

pkg/fakerp/store/store.go
mjudeikis: It makes the store interface more universal. But :man_shrugging:

pkg/fakerp/util.go
```diff
@@ -16,3 +29,134 @@ func (s *Server) isAdminRequest(req *http.Request) bool {
...
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(b)
+	return
+}
+
+// reply return either admin or external api response
+func (s *Server) reply(w http.ResponseWriter, req *http.Request) {
+	cs := req.Context().Value(ContainerServicesKey).(*api.OpenShiftManagedCluster)
+
+	if &cs == nil {
```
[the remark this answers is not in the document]
mjudeikis: Now yes

pkg/fakerp/store/store.go
```diff
@@ -0,0 +1,123 @@
...
+	case fi.Mode().IsDir():
+		return os.RemoveAll(record)
+	case fi.Mode().IsRegular():
+		return os.RemoveAll(record)
+	}
+
+	return nil
+}
+
+func stat(path string) (fi os.FileInfo, err error) {
```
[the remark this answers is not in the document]
mjudeikis: Just to be safe. But this is more applicable if we use same store interface in multiple places. removed.

pkg/fakerp/store/store.go
```diff
@@ -0,0 +1,123 @@
...
+func New(log *logrus.Entry, dir string) (*Storage, error) {
+	dir = filepath.Clean(dir)
+
+	s := &Storage{
+		dir:     dir,
+		log:     log,
+		mutexes: make(map[string]*sync.Mutex),
+	}
+
+	// if the database already exists, just use it
```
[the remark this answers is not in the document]
mjudeikis: same as others. If we reuse - helps to catch errors. removed.

pkg/fakerp/middleware.go
```diff
@@ -47,14 +49,28 @@ func (s *Server) context(handler http.Handler) http.Handler {
...
 			s.badRequest(w, err.Error())
 			return
 		}
-		ctx = context.WithValue(ctx, internalapi.ContextKeyVaultClientAuthorizer, vaultauthorizer)
+		ctx = context.WithValue(ctx, api.ContextKeyVaultClientAuthorizer, vaultauthorizer)
+
+		// add containerservices.yaml object to ctx
+		data, err := s.store.Get(ContainerServicesKey)
+		if err != nil {
+			s.log.Debugf("record %s not found. If this is create, ignore", ContainerServicesKey)
```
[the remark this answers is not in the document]
mjudeikis: can't think of anything now. If we want to set this in the context, we dont know which request is coming in.
The alternative is to start intercepting the request and ignore this part on `PUT, Create` only. Which taking into account all business logic looks as ulgy as this one. 

This is why I did store handling in the individual method handlers in the first place, and not in the context.

jim-minter:
there is more to do here, but this takes us forwards.

PR #1432, opened 2019-04-05
add geneva action to list backup blobs

jim-minter:
```release-note
* add geneva action to list backup blobs
```

### DIFF
--- a/pkg/cluster/etcdrestore.go
+++ b/pkg/cluster/etcdrestore.go
@@ -2,25 +2,26 @@ package cluster
 
 import (
 	"context"
-	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
 
 	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/cluster/names"
 	"github.com/openshift/openshift-azure/pkg/cluster/updateblob"
 )
 
-func (u *simpleUpgrader) EtcdBlobExists(ctx context.Context, blobName string) error {
+func (u *simpleUpgrader) EtcdListBackups(ctx context.Context) ([]storage.Blob, error) {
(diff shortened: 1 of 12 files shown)


pkg/fakerp/routes.go
```diff
@@ -19,6 +19,7 @@ func (s *Server) SetupRoutes() {
 	s.router.Put(s.basePath, s.handlePut)
 	s.router.Get(filepath.Join("/admin", s.basePath), s.handleGet)
 	s.router.Put(filepath.Join("/admin", s.basePath), s.handlePut)
+	s.router.Put(filepath.Join("/admin", s.basePath, "/listBackups"), s.handleListBackups)
```
charlesakalugwu: This should be changed to a 
```suggestion
	s.router.Get(filepath.Join("/admin", s.basePath, "/listBackups"), s.handleListBackups)
```

pkg/cluster/etcdrestore.go
```diff
@@ -2,25 +2,26 @@ package cluster
...
-	exists, err := blob.Exists()
+	resp, err := etcdContainer.ListBlobs(storage.ListBlobsParameters{})
 	if err != nil {
-		return err
+		return nil, err
 	}
-	if !exists {
-		return fmt.Errorf("Blob %s does not exist", blobName)
+	backups := make([]string, 0, len(resp.Blobs))
+	for _, blob := range resp.Blobs {
```
charlesakalugwu: it might also be worth returning `blob.Properties.LastModified`

pkg/fakerp/routes.go
jim-minter: good catch thanks

pkg/cluster/etcdrestore.go
jim-minter: done, thanks

Issue #1400, opened 2019-04-02
write: connection reset by peer talking to storage account


jim-minter:
time="2019-04-02T00:37:12Z" level=debug msg="400 Bad Request: Failed to apply request: UpdateSyncPod: Put https://sacfgwdqw8oq44ss3x8eu4gd.blob.core.windows.net/config/sync: net/http: HTTP/1.x transport connection broken: write tcp 172.16.16.144:44410->52.226.8.148:443: write: connection reset by peer" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).badRequest" file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:11"

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1392/pull-ci-openshift-openshift-azure-master-e2e-azure/3270/build-log.txt

mjudeikis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1434/pull-ci-openshift-openshift-azure-master-e2e-azure/3400

jim-minter:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1485/pull-ci-openshift-openshift-azure-master-e2e-azure-scaleupdown/511/

asalkeld:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1493/pull-ci-openshift-openshift-azure-master-e2e-azure-scaleupdown/572/build-log.txt

mjudeikis:
not to the storage account but same issue:
https://github.com/openshift/openshift-azure/pull/1433

jim-minter:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1499/pull-ci-openshift-openshift-azure-master-upgrade-v4.2/179

jim-minter:
@asalkeld different failure mode:

[...]

PR #1433, opened 2019-04-05
add tool to update imagestreams and templates, and do so

jim-minter:
```release-note
update openshift imagestreams and templates to 04/04/19
```
nb: many imagestreams now use registry.redhat.io, so more reliance on our secret for them to work
added the registry.redhat.io secret to the openshift namespace to enable pulling images
minor sync to artifacts script to make it quit early if we didn't get as far as having an admin kubeconfig
pkg/sync/v5/content_test.go is the import script
added test to ensure the sync pod doesn't spin on updating objects


mjudeikis:
/retest
https://github.com/openshift/openshift-azure/issues/1400

hack/artifacts.sh
```diff
@@ -6,6 +6,11 @@ if [[ -z "$ARTIFACT_DIR" ]]; then
     exit 0
 fi
 
+if [[ ! -e $PWD/_data/_out/admin.kubeconfig ]]; then
+    echo "admin.kubeconfig not found, exiting"
+    exit 1
```
asalkeld: are you sure you want to exit 1 instead of 0 like above?

jim-minter:
previously passed tests, one bit change, merging

PR #1450, opened 2019-04-05
Scrape router stats

### LINKED ISSUE #1312: Scrape haproxy metrics
Need a ServiceMonitor to scrape haproxy metrics. Look at the etcd ServiceMonitor as a reference and then figure out the best set of metrics to scrape.

thekad:
Closes #1312 

```release-note
NONE
```


pkg/sync/v5/data/Service/default/router-stats.yaml
```diff
@@ -1,16 +1,13 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    prometheus.io/port: "1936"
-    prometheus.io/scrape: "true"
```
charlesakalugwu: why were these removed?

thekad: these were remnants from a previous implementation of prometheus scraping, not used in this one.

pkg/sync/v5/sync.go
```diff
@@ -300,22 +300,27 @@ func (s *sync) writeDB() error {
 	sort.Strings(keys)
 
 	// crd needs to land early to get initialized
+	s.log.Debug("applying crd resources")
```
jim-minter: if adding these, make sure we don't run at level debug by default - we were way too noisy and I don't want to go back there

jim-minter:
lgtm - one thing to check up - ping me when you want the lgtm label

Makefile
```diff
@@ -117,7 +117,7 @@ sync-image: sync $(IMAGEBUILDER) pullregistry
 
 .PHONY: sync-run
 sync-run: generate
-	go run -ldflags ${LDFLAGS} ./cmd/sync -run-once
+	go run -ldflags ${LDFLAGS} ./cmd/sync -run-once -loglevel Debug
```
thekad: under normal operations the sync pod container runs on the default Info level, manually running the sync pod runs with Debug level

pkg/sync/v5/sync.go
thekad: Increased default level to Info

cmd/sync/sync.go
```diff
@@ -24,7 +24,7 @@ var (
 	dryRun    = flag.Bool("dry-run", false, "Print resources to be synced instead of mutating cluster state.")
 	once      = flag.Bool("run-once", false, "If true, run only once then quit.")
 	interval  = flag.Duration("interval", 3*time.Minute, "How often the sync process going to be rerun.")
-	logLevel  = flag.String("loglevel", "Debug", "valid values are Debug, Info, Warning, Error")
+	logLevel  = flag.String("loglevel", "Info", "valid values are Debug, Info, Warning, Error")
```
jim-minter: would like to see this made the default for all our containers - @mjudeikis perhaps you can pick that up in #1464?

PR #1451, opened 2019-04-05
switch on UseInstanceMetadata

jim-minter:
```release-note
NONE
```

should significantly reduce Azure API calls and increase reliability


jim-minter:
all green first time - a good sign?
lgtming

mjudeikis:
Would be nice a comment why we did this? :)

Issue #1362, opened 2019-03-27
"checkMonitoringStackHealth" timed out - disk related


Makdaam:
I can't reliably reproduce this flake. Seems to be a timeout after running a long list of e2e tests (sometimes the monitoring stack doesn't come back in time to pass the test).

```
Scale Up/Down E2E tests [ScaleUpDown][Fake][LongRunning]
  Expected
      <[]*standard.TestError | len:1, cap:1>: [
          {
              Bucket: "checkMonitoringStackHealth",
              Err: {
                  s: "timed out waiting for the condition",
              },
          },
      ]
  to be empty
```
/kind test-flake

mjudeikis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1361/pull-ci-openshift-openshift-azure-master-e2e-azure/3077

One more. I suspect it will underlaying disk issue.

mjudeikis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1346/pull-ci-openshift-openshift-azure-master-e2e-azure/3085 
one more

mjudeikis:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1346/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/52
Starting to think this is not a flake...

jim-minter:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1361/pull-ci-openshift-openshift-azure-master-e2e-azure/3077 is not an instance of this problem!

In https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1346/pull-ci-openshift-openshift-azure-master-e2e-azure/3085, I think the following is going on:

resource group e2e-mjudeikis-1346-umcmvo, 27/3/2019 from around 13:27:31 UTC

* alertmanager-main-0 wants to start with persistent storage pvc-0fc52771-5094-11e9-ac50-000d3a173d35
* at 13:27:31 controller decides to attach disk to node; because of asynchronous API issue it believes the disk is attached when it isn't yet
* in parallel:
  1) azure disk attach takes place slowly (maybe as long as until 13:37:42?)
  2) node decides to run pod
[...]

PR #1453, opened 2019-04-06
use a unique service principal per cluster

jim-minter:
```release-note
NONE
```

follow-up: https://github.com/openshift/azure-misc/issues/143


jim-minter:
@openshift/sig-azure fyi
@asalkeld please review

Upgrades from 3.2 and 4.0 will still use the hard-coded SPs unless we backport this (it will go into 4.1), so we could still have some flake difficulties, but at least new cluster builds will largely be using segragated SPs.

asalkeld:
@jim-minter why are we doing this? I don't have the context of some discussion...

asalkeld:
relates to #1362

pkg/fakerp/fakerp.go
```diff
@@ -134,20 +134,27 @@ func createOrUpdate(ctx context.Context, p api.Plugin, log *logrus.Entry, cs, ol
...
-	// validate the internal API representation (with reference to the previous
-	// internal API representation)
-	// we set fqdn during enrichment which is slightly different than what the RP
-	// will do so we are only validating once.
 	var errs []error
 	if isAdmin {
 		errs = p.ValidateAdmin(ctx, cs, oldCs)
 	} else {
-		errs = p.Validate(ctx, cs, oldCs, false)
+		errs = p.Validate(ctx, cs, oldCs, true)
```
asalkeld: This was wrong?

mjudeikis:
lgtm
lets see if rate limiting gives as troubles now

PR #1456, opened 2019-04-07
allow admin API to change image versions

jim-minter:
```release-note
make vm image configuration and individual container image configuration writeable via the admin API
```

@mjudeikis @asalkeld the testing is grotty here - Angus, I'm hoping the work you're doing will improve coverage across all the images; MJ it feels like we should end up with some sort of Go test suite around updates, but I'm not sure what that ends up looking like.


mjudeikis:
/lgtm 
look good. 
https://github.com/openshift/openshift-azure/issues/1460

PR #1464, opened 2019-04-08
collapse image to one image

mjudeikis:
```release-note
Release a single azure container image instead of seven different ones
```


pkg/arm/v5/data/node-startup.sh
```diff
@@ -28,7 +28,7 @@ done
 set +x
 export SASURI='{{ .Config.WorkerStartupSASURI }}'
 set -x
-docker run --privileged --rm --network host -v /:/host:z -e SASURI {{ .Config.Images.Startup }}
+docker run --privileged --rm --network host -v /:/host:z -e SASURI --entrypoint startup {{ .Config.Images.Startup }}
```
thekad: is `startup` going to be the new global entrypoint?

pkg/fakerp/config.go
```diff
@@ -72,26 +72,14 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
...
-	}
-	if os.Getenv("CANARY_IMAGE") != "" {
-		v.Images.Canary = os.Getenv("CANARY_IMAGE")
-	}
-	if os.Getenv("AZURE_CONTROLLERS_IMAGE") != "" {
-		v.Images.AzureControllers = os.Getenv("AZURE_CONTROLLERS_IMAGE")
-	}
-	if os.Getenv("STARTUP_IMAGE") != "" {
-		v.Images.Startup = os.Getenv("STARTUP_IMAGE")
+	if os.Getenv("AZURE_IMAGE") != "" {
```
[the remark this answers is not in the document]
asalkeld: Yes, perhaps:
- leave all the env and config options as they are and just have a single image a bit like busybox.
- then you can change the version of the image for one use case if needed.

mjudeikis:
There is one big improvement which could be done is to move to `StringVarP` type of variable assignment, so avoid manual statements. but this is follow-up

pkg/fakerp/config.go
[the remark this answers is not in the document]
asalkeld: I suppose it's to just reduce the scope of impact on the customer?

pkg/entrypoint/config/config.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type Common struct {
```
jim-minter: all of this feels clumsy and overengineered, but I don't have a better suggestion right now.  I think we can check it in and make it nicer afterwards.  Examples: I think cmd/azure should pass a `*log.Entry` into the subcommands; I would like fewer .go files in each pkg/entrypoint/*, etc.

pkg/entrypoint/azure-controllers/type.go
```diff
@@ -0,0 +1,9 @@
+package azurecontrollers
```
jim-minter: preferred name is types.go.  Even better - roll into fewer .go files

pkg/entrypoint/tlsproxy/cmd.go
```diff
@@ -0,0 +1,80 @@
+package tlsproxy
```
jim-minter: note to self: I would like to maintain a cmd/ entrypoint for tlsproxy *as well as* building it into cmd/azure, this will be useful for the vm image build

pkg/fakerp/config.go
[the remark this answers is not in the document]
jim-minter: There are 2 different discussions here. I'm onboard with what MJ is proposing in this PR.
1- we maintain all the separate image references in pluginconfig.yaml, so we can change the image used by a single function in production with low impact
2- we reduce the env vars: as a dev, actually I often want to change code in the sync pod and startup pod at the same time.  What is proposed here allows me to do that via one image upload and one env var change.  +1.

pkg/cluster/kubeclient/syncpod.go
```diff
@@ -72,6 +72,7 @@ func (u *kubeclient) EnsureSyncPod(ctx context.Context, syncImage string, hash [
 					Spec: corev1.PodSpec{
 						Containers: []corev1.Container{
 							{
+								Args:            []string{"sync"},
```
jim-minter: why not Command ?

pkg/sync/v5/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
```diff
@@ -17,26 +17,25 @@ spec:
     spec:
       containers:
       - args:
-        - -loglevel
+        - tlsproxy
```
jim-minter:
```
command:
- tlsproxy
```
? 

comment applies throughout PR

jim-minter:
@mjudeikis this is good, please wrap it up, manually ensure that all the functions are working in an upgrade cluster and we'll get it merged

pkg/cluster/kubeclient/syncpod.go
mjudeikis: entrypoint = command in kubernetes 
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

pkg/sync/v5/data/DaemonSet.apps/openshift-azure-monitoring/etcd-metrics.yaml
mjudeikis: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes command is forced from image "entrypoint" `azure tlsproxy`:
this would be : 
```
command: azure
args: 
 - tlsproxy
 - params
```
But now azure is handled on image level

pkg/entrypoint/config/config.go
mjudeikis: problem with that is that loglevel is rootCmd level flag, so it can be read populated only when cobra is being initiated. And this happens only later in the process. an alternative is to parse loglevel using standard methods `flags`

env.example
```diff
@@ -17,14 +17,7 @@ export AZURE_AAD_CLIENT_SECRET=
...
 #PREFIX=quay.io/$USER
-#export AZURE_CONTROLLERS_IMAGE=$PREFIX/azure-controllers:latest
-#export CANARY_IMAGE=$PREFIX/canary:latest
-#export E2E_IMAGE=$PREFIX/e2e-tests:latest
-#export ETCDBACKUP_IMAGE=$PREFIX/etcdbackup:latest
-#export METRICSBRIDGE_IMAGE=$PREFIX/metricsbridge:latest
-#export STARTUP_IMAGE=$PREFIX/startup:latest
-#export SYNC_IMAGE=$PREFIX/sync:latest
-#export TLSPROXY_IMAGE=$PREFIX/tlsproxy:latest
+#export AZURE_IMAGE=$PREFIX/arho:latest
```
charlesakalugwu:
```suggestion
#export AZURE_IMAGE=$PREFIX/azure:latest
```

pkg/arm/v5/data/master-startup.sh
```diff
@@ -65,9 +65,10 @@ docker run --rm --network host \
   -v /etc/origin/master:/etc/origin/master \
   -v /etc/origin/cloudprovider/:/_data/_out \
   -v $tempBackDir:/out:z \
-  {{ .Config.Images.EtcdBackup }} \
-  -blobname={{ .BackupBlobName }} \
-  -destination=/out/backup.db download
+  {{ .Config.Images.EtcdBackup }} etcdbackup \
+  --blobname={{ .BackupBlobName }} \
+  --destination=/out/backup.db \
+  --action download \
```
charlesakalugwu:
```suggestion
  --action download
```

cmd/azure/azure.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
mjudeikis: will vendor before merge, not to make this huge PR, still need to sync with release repo merge

pkg/cluster/kubeclient/syncpod.go
jim-minter: thanks!

jim-minter:
/retest

https://github.com/openshift/release/pull/3383 has merged

jim-minter:
flake #1504

jim-minter:
flake #1400

jim-minter:
rebased to fix merge conflict and added a couple of minor fixes

PR #1465, opened 2019-04-08
Fix secrets generation command

ehashman:
```release-note
NONE
```

When this make target was added in #1453, the azure namespacing was accidentally removed.

/cc @jim-minter


asalkeld:
@jim-minter  upgrades are failing with:
```
Failed to apply request: found 0 service principals, should be 1
```

PR #1469, opened 2019-04-09
convert webconsole image change test to go

asalkeld:
```release-note
NONE
```

So @jim-minter we chatted about this earlier. I think this is a good incremental improvement.
from looking at the e2e tests, the admin action or createOrUpdate call is really minimal compared
with the setup and verify code. So I am not sure how the infrastructure that we talked about would save over just making use of the current e2e tests.

### DIFF
--- a/test/e2e/specs/fakerp/setcontainerimage.go
+++ b/test/e2e/specs/fakerp/setcontainerimage.go
@@ -1,0 +1,62 @@
+package fakerp
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/ghodss/yaml"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/openshift-azure/pkg/api/admin"
+	pluginapi "github.com/openshift/openshift-azure/pkg/api/plugin"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+	"github.com/openshift/openshift-azure/test/sanity"
+)
+
+var _ = Describe("Change a single image to latest E2E tests [ChangeImage][Fake][LongRunning]", func() {
+	var (
+		azurecli *azure.Client
+		ctx      = context.Background()
+	)
+
+	BeforeEach(func() {
+		var err error
+		azurecli, err = azure.NewClientFromEnvironment(false)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should be possible for an SRE to update a single container image", func() {
+		By("Executing a cluster update with updated image.")
+		data, err := ioutil.ReadFile("../../pluginconfig/pluginconfig-311.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		var template *pluginapi.Config
+		if err := yaml.Unmarshal(data, &template); err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+		before := admin.OpenShiftManagedCluster{
+			Config: &admin.Config{
+				Images: &admin.ImageConfig{
+					WebConsole: to.StringPtr(template.Versions[template.PluginVersion].Images.WebConsole),
+				},
+			},
+		}
+
+		update, err := azurecli.OpenShiftManagedClustersAdmin.CreateOrUpdate(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), &before)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(update).NotTo(BeNil())
+
+		By("checking running webconsole image")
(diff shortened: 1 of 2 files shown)


test/e2e/specs/fakerp/setcontainerimage.go
```diff
@@ -0,0 +1,62 @@
...
+		before.Config.Images.WebConsole = to.StringPtr(template.Versions[template.PluginVersion].Images.WebConsole)
+
+		update, err := azurecli.OpenShiftManagedClustersAdmin.CreateOrUpdate(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), before)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(update).NotTo(BeNil())
+
+		By("checking runnning webconsole image")
+		webconsole, err := sanity.Checker.Client.Admin.AppsV1.Deployments("openshift-web-console").Get("webconsole", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(webconsole.Spec.Template.Spec.Containers[0].Image).To(Equal(template.Versions[template.PluginVersion].Images.WebConsole))
```
charlesakalugwu: `Expect(webconsole).NotTo(BeNil())`

test/e2e/specs/fakerp/setcontainerimage.go
```diff
@@ -0,0 +1,62 @@
...
+	It("should be possible for an SRE to update a single container image", func() {
+		By("Reading the internal config before the image update")
+		before, err := azurecli.OpenShiftManagedClustersAdmin.Get(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(before).NotTo(BeNil())
+
+		By("Executing a cluster update with updated image.")
+
+		data, err := ioutil.ReadFile("../../pluginconfig/pluginconfig-311.yaml")
+		Expect(err).NotTo(HaveOccurred())
```
charlesakalugwu: `Expect(data).NotTo(BeNil())`

asalkeld: thanks, i'll add these checks

jim-minter: @charlesakalugwu really, no.  There is no justification for not trusting that the underlying library will return data if it does not return a nil error.

jim-minter: or any of the code we write either.
If we get an unexpected NPE in tests, so be it.

test/e2e/specs/fakerp/setcontainerimage.go
```diff
@@ -0,0 +1,64 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		before.Config.Images.WebConsole = to.StringPtr(template.Versions[template.PluginVersion].Images.WebConsole)
+
+		update, err := azurecli.OpenShiftManagedClustersAdmin.CreateOrUpdate(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), before)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(update).NotTo(BeNil())
+
+		By("checking runnning webconsole image")
```
jim-minter: two 'n's in running ;)

test/e2e/specs/fakerp/setcontainerimage.go
```diff
@@ -0,0 +1,64 @@
...
+
+	BeforeEach(func() {
+		var err error
+		azurecli, err = azure.NewClientFromEnvironment(false)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should be possible for an SRE to update a single container image", func() {
+		By("Reading the internal config before the image update")
+		before, err := azurecli.OpenShiftManagedClustersAdmin.Get(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"))
```
jim-minter: this is actually not strictly necessary, and in some ways I'd rather remove it - see below

test/e2e/specs/fakerp/setcontainerimage.go
```diff
@@ -0,0 +1,64 @@
...
+
+		data, err := ioutil.ReadFile("../../pluginconfig/pluginconfig-311.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(data).NotTo(BeNil())
+		var template *pluginapi.Config
+		if err := yaml.Unmarshal(data, &template); err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		before.Config.Images.WebConsole = to.StringPtr(template.Versions[template.PluginVersion].Images.WebConsole)
```
jim-minter: here instead of reusing before, construct a brand new admin API object, leave all the pointers at nil except for .Config.Images.WebConsole.

jim-minter:
@asalkeld I agree - couple of nits.  Test failure was CI system stuffing up.

test/e2e/specs/fakerp/setcontainerimage.go
asalkeld: ok

PR #1471, opened 2019-04-09
moved azure client init to BeforeSuite for reuse.

Makdaam:
```release-note
NONE
```
Moving azure client (fake for now, since real e2e seems broken) to the sanitychecker singleton to keep it from reinitializing at every BeforeEach.
In the end the realRP client should also be in the singleton.

### DIFF
--- a/test/clients/azure/client.go
+++ b/test/clients/azure/client.go
@@ -21,11 +21,7 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/configblob"
 )
 
-// FakeRPClient is a singleton Azure Client used by all tests which run against the FakeRP
-var FakeRPClient *Client
-
-// RealRPClient is a singleton Azure Client used by all tests which run against the RealRP
-var RealRPClient *Client
+var RPClient *Client
 
 // Client is the main controller for azure client objects
 type Client struct {
@@ -44,14 +40,11 @@ type Client struct {
 	Groups                           resources.GroupsClient
 }
 
-// NewClientFromEnvironment creates a new azure client from environment variables.
-// Setting the storage client is optional and should only be used selectively by
-// tests that need access to the config storage blob because configblob.GetService
-// makes api calls to Azure in order to setup the blob client.
-func NewClientFromEnvironment(ctx context.Context, log *logrus.Entry, setStorageClient bool, realRP bool) (*Client, error) {
+// NewClientFromEnvironment creates a new azure client from environment variables and stores it in the RPClient variable
+func NewClientFromEnvironment(ctx context.Context, log *logrus.Entry, realRP bool) error {
 	authorizer, err := azureclient.NewAuthorizerFromEnvironment("")
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	cfg := &cloudprovider.Config{
@@ -64,10 +57,11 @@ func NewClientFromEnvironment(ctx context.Context, log *logrus.Entry, setStorage
 	subscriptionID := cfg.SubscriptionID
 
 	var storageClient storage.BlobStorageClient
-	if setStorageClient {
+	//for now fake RP tests use the Blob Storage, real RP do not
+	if !realRP {
 		storageClient, err = configblob.GetService(ctx, log, cfg)
 		if err != nil {
-			return nil, err
+			return err
 		}
 	}
 
@@ -85,7 +79,7 @@ func NewClientFromEnvironment(ctx context.Context, log *logrus.Entry, setStorage
 
 	rpcAdmin := adminapi.NewClient(rpURL, subscriptionID)
 
-	return &Client{
+	RPClient = &Client{
 		Accounts:                         storage.NewAccountsClient(ctx, log, subscriptionID, authorizer),
 		ActivityLogs:                     insights.NewActivityLogsClient(ctx, log, subscriptionID, authorizer),
 		Applications:                     managedapplications.NewApplicationsClient(ctx, log, subscriptionID, authorizer),
@@ -99,5 +93,6 @@ func NewClientFromEnvironment(ctx context.Context, log *logrus.Entry, setStorage
 		VirtualNetworks:                  network.NewVirtualNetworkClient(ctx, log, subscriptionID, authorizer),
 		VirtualNetworksPeerings:          network.NewVirtualNetworksPeeringsClient(ctx, log, subscriptionID, authorizer),
 		Groups:                           resources.NewGroupsClient(ctx, log, subscriptionID, authorizer),
-	}, nil
+	}
+	return nil
 }
--- a/test/clients/azure/client.go
+++ b/test/clients/azure/client.go
@@ -4,10 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"regexp"
-	"strings"
 
-	"github.com/onsi/ginkgo/config"
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/fakerp/shared"
@@ -24,19 +21,11 @@ import (
 	"github.com/openshift/openshift-azure/pkg/util/configblob"
 )
 
-// rpFocus represents the supported RP APIs which e2e tests use to create their azure clients,
-// The client will be configured to work either against the real, fake or admin apis
-type rpFocus string
+// FakeRPClient is a singleton Azure Client used by all tests which run against the FakeRP
+var FakeRPClient *Client
 
-var (
-	adminRpFocus = rpFocus(regexp.QuoteMeta("[Admin]"))
-	fakeRpFocus  = rpFocus(regexp.QuoteMeta("[Fake]"))
-	realRpFocus  = rpFocus(regexp.QuoteMeta("[Real]"))
-)
-
-func (tf rpFocus) match(focusString string) bool {
-	return strings.Contains(focusString, string(tf))
-}
+// RealRPClient is a singleton Azure Client used by all tests which run against the RealRP
+var RealRPClient *Client
 
 // Client is the main controller for azure client objects
 type Client struct {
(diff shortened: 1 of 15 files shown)


test/e2e/standard/sanitychecker.go
```diff
@@ -37,9 +38,12 @@ type DeepTestInterface interface {
 }
 
 type SanityChecker struct {
-	Log    *logrus.Entry
-	cs     *internalapi.OpenShiftManagedCluster
-	Client *openshift.ClientSet
+	Log          *logrus.Entry
+	cs           *internalapi.OpenShiftManagedCluster
+	Client       *openshift.ClientSet
+	FakeRPClient *azure.Client
```
charlesakalugwu: The sanity checking mechanism only makes sense with kube/openshift clients and not azure clients. Please remove

charlesakalugwu:
have you had a look at `BeforeSuite`?

test/e2e/specs/fakerp/admin.go
```diff
@@ -57,9 +56,8 @@ var _ = Describe("Openshift on Azure admin e2e tests [Fake]", func() {
 
 	It("should ensure no unnecessary VM rotations occured", func() {
 		Expect(os.Getenv("RESOURCEGROUP")).ToNot(BeEmpty())
-		azurecli, err := azure.NewClientFromEnvironment(true)
-		Expect(err).ToNot(HaveOccurred())
-
+		azurecli := sanity.Checker.FakeRPClient
+		Expect(azurecli).NotTo(BeNil())
```
jim-minter: unnecessary (throughout)

test/e2e/specs/fakerp/admin_api.go
```diff
@@ -18,9 +19,9 @@ var _ = Describe("Validate AdimAPI field readability [AdminAPI][Fake]", func() {
 	)
 
 	BeforeEach(func() {
-		var err error
-		cli, err = azure.NewClientFromEnvironment(false)
-		Expect(err).NotTo(HaveOccurred())
+		//cli, err = azure.NewClientFromEnvironment(false)
+		cli = sanity.Checker.FakeRPClient
```
jim-minter: can reduce the scope of this - instantiate in the scope where it's needed

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1 @@
+package fakerp
```
jim-minter: slipped in to the PR?

test/e2e/specs/realrp/realrp.go
```diff
@@ -28,8 +28,9 @@ var _ = Describe("Resource provider e2e tests [Default][Real]", func() {
 	// NOTE: Ensure this is always the first test in the [Default][Real] spec!
 	It("should deploy a cluster using the production RP", func() {
 		var err error
-		cli, err = azure.NewClientFromEnvironment(false)
+		cli, err = azure.NewClientFromEnvironment(false, true)
```
jim-minter: how come this isn't using RealRPClient?

test/e2e/standard/sanitychecker.go
```diff
@@ -37,9 +38,12 @@ type DeepTestInterface interface {
...
+	RealRPClient *azure.Client
+	AnyRPClient  *azure.Client
```
jim-minter: Boo, it seems like we don't have a single e2e test which can run against the realrp and against the fakerp :-(
Remove AnyRPClient for now.

test/e2e/standard/sanitychecker.go
```diff
@@ -66,6 +70,27 @@ func NewSanityChecker(ctx context.Context, log *logrus.Entry, cs *internalapi.Op
...
+	scc.FakeRPClient, err = azure.NewClientFromEnvironment(true, false)
+	log.Debugf("New FakeRP client ended with %v", scc.RealRPClient)
+	if err != nil {
+		log.Error(err)
+	}
+	if scc.RealRPClient == nil && scc.FakeRPClient == nil {
+		log.Error("Unable to provision either of the azure clients")
+		return nil, err
+	}
+	if scc.RealRPClient != nil {
```
jim-minter: rationale for removing AnyRPClient is that this logic isn't right, but I don't know what the right logic looks like still

test/e2e/specs/fakerp/admin.go
Makdaam: Refactored the tests to not use the "azurecli" or "cli" variable.
Yes, forgot that client creation doesn't perform a dial, so it won't be nil.

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: slipped, removed.

test/e2e/specs/realrp/realrp.go
Makdaam: I wanted to be 100% sure FakeRP e2e work first. Now it uses RealRPClient.

test/e2e/standard/sanitychecker.go
Makdaam: Removed AnyRP for now. I think a test dial has to be performed to decide if an Azure Client is valid.

test/e2e/specs/beforesuite.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = BeforeSuite(func() {
+	var err error
+	testlogger := log.GetTestLogger()
+
+	//so far none of the RealRP tests required the Storage Client
+	azure.RealRPClient, err = azure.NewClientFromEnvironment(false, true)
+	testlogger.Debugf("New RealRP client: %v", azure.RealRPClient)
```
charlesakalugwu:
```suggestion
	testlogger.Debugf("new real rp client: %v", azure.RealRPClient)
```

test/e2e/specs/beforesuite.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		testlogger.Error(err)
+	}
+	azure.FakeRPClient, err = azure.NewClientFromEnvironment(true, false)
+	testlogger.Debugf("New FakeRP client: %v", azure.FakeRPClient)
```
charlesakalugwu:
```suggestion
	testlogger.Debugf("new fake rp client: %v", azure.FakeRPClient)
```

test/e2e/specs/beforesuite.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		testlogger.Error(err)
+	}
+	if azure.RealRPClient == nil && azure.FakeRPClient == nil {
+		testlogger.Error("Unable to provision either of the azure clients")
```
charlesakalugwu:
```suggestion
		testlogger.Error("unable to provision either of the azure clients")
```

y-cote:
Aside from the few outstanding points from Charles, this PR looks good to me.

test/e2e/specs/beforesuite.go
```diff
@@ -0,0 +1,29 @@
+package specs
+
+import (
+	. "github.com/onsi/ginkgo"
+
+	"github.com/openshift/openshift-azure/test/clients/azure"
+	"github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = BeforeSuite(func() {
```
jim-minter: @Makdaam this would be better in an init() function in test/clients/azure/client.go - it's a bit spooky to declare a global in one place and populate it in another.

```golang
func init() {
	var err error

	RealRPClient, err = NewClientFromEnvironment(false, true)
	if err != nil {
		panic(err)
	}

	FakeRPClient, err = NewClientFromEnvironment(false, false)
	if err != nil {
		panic(err)
	}
}
```

test/clients/azure/client.go
```diff
@@ -18,19 +14,11 @@ import (
...
-	realRpFocus  = rpFocus(regexp.QuoteMeta("[Real]"))
-)
+// FakeRPClient is a singleton Azure Client used by all tests which run against the FakeRP
+var FakeRPClient *Client
 
-func (tf rpFocus) match(focusString string) bool {
-	return strings.Contains(focusString, string(tf))
-}
+// RealRPClient is a singleton Azure Client used by all tests which run against the RealRP
+var RealRPClient *Client
```
jim-minter: sorry, this is almost certainly what we agreed on, but it still looks clumsy.  Given that we don't have any tests right now that can run against both fake and real RPs, how about making one top level `var SomeName *Client` having three fields within it - something like RealOpenShiftManagedClusters, FakeOpenShiftManagedClusters, FakeOpenShiftManagedClustersAdmin ?

mjudeikis:
@Makdaam ping :)

test/e2e/specs/beforesuite.go
Makdaam: Moved the initialization to azure package. But I left the interpretation of ginkgo focus here since this is run before every test suite. This way the azure package contains all initialization code, and the ginkgo spec decides which client is needed.

Makdaam:
rebased,
rolled back the split into RealRP/FakeRP clients as per @jim-minter 's suggestion. 
The decision between real/fake is now in a BeforeSuite handler in the test spec.
The initialization of the client happens in NewClientFromEnvironment and is stored in the RPClient variable instead of being returned. 
This will require refactoring if we decide to run both Real and Fake tests in one go, but it's not an issue currently.

mjudeikis:
@Makdaam bump.

jim-minter:
@Makdaam please check the test failures

Makdaam:
Rebased, fixed previous issue. Verified e2e tests run correctly.

charlesakalugwu:
another rebase needed :)

charlesakalugwu:
I am not too crazy about the global `var RPClient *Client` but everything else looks good to me.

asalkeld:
/lgtm cancel
needs rebase

Makdaam:
Rabased.

Makdaam:
/retest
flake #1617

PR #1477, opened 2019-04-09
validate canary image empty on v3.2

jim-minter:
```release-note
validate canary image empty on v3.2
```

### DIFF
--- a/pkg/api/validate/plugin.go
+++ b/pkg/api/validate/plugin.go
@@ -39,7 +39,7 @@ func (v *PluginAPIValidator) Validate(c *pluginapi.Config) (errs []error) {
 			errs = append(errs, fmt.Errorf("invalid versions[%q]", version))
 		}
 
-		errs = append(errs, validateVersionConfig(fmt.Sprintf("versions[%q]", version), &versionConfig)...)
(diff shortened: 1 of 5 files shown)


pkg/api/validate/plugin.go
```diff
@@ -250,7 +250,7 @@ func validateImageConfig(path string, i *pluginapi.ImageConfig) (errs []error) {
 		errs = append(errs, fmt.Errorf("invalid %s.azureControllers %q", path, i.AzureControllers))
 	}
 
-	if i.Canary == "" {
+	if (i.Canary == "") != (version == "v3.2") {
```
asalkeld: this seems correct, but is confusing. why not:
```
if (i.Canary == "") && (version != "v3.2") {
```

kwoodson:
This lgtm. Worth addressing @asalkeld's nit.

pkg/api/validate/plugin.go
jim-minter: because it's not equivalent? ;-)

it'd have to be `if (i.Canary == "" && version != "v3.2") || (i.Canary != "" && version == "v3.2")` ...

jim-minter:
flake was due to slow pull/checkout increased timeout in new commit, readding lgtm

PR #1483, opened 2019-04-10
fluentd rollback and cluster upgrade integration fixes

jim-minter:
```release-note
* roll back change which caused NPE in fluentd for non-container logs
* modify cluster upgrade mechanism - set clusterVersion: latest, not pluginVersion: latest.  PluginVersion remains in admin API, but r/o
```


jim-minter:
upgrade-v4.0 failure is expected because release is not updated

pkg/config/config.go
```diff
@@ -26,7 +26,7 @@ func New(cs *api.OpenShiftManagedCluster, runningUnderTest bool) (Interface, err
 	switch cs.Config.PluginVersion {
 	case "v3.2":
 		return v3.New(cs, runningUnderTest), nil
-	case "v4.1":
+	case "v4.2":
```
asalkeld: it's upgrading from v4.1 so this doesn't work

PR #1484, opened 2019-04-10
fluentd rollback and cluster upgrade integration fixes - v4

jim-minter:
```release-note
NONE
```

* roll back change which caused NPE in fluentd for non-container logs
* modify cluster upgrade mechanism - set clusterVersion: latest, not pluginVersion: latest.  PluginVersion remains in admin API, but r/o


jim-minter:
@mjudeikis if this is checking out please roll a 4.2 in the morning

PR #1455, opened 2019-04-07
Logging fixes

jim-minter:
```release-note
NONE
```

fixes #1454 

* collect core dump dmesgs to artifacts, if they exist
* revamp hack/ssh.sh and hack/scp.sh
* bump image version
* convert utf-8 messages to ascii (otherwise we lose log lines @charlesakalugwu @mjudeikis)
* minimise buffering on mdsd plugin, stop writing to disk, make blocking
* fix go get imagebuilder
* clean up multiple directories in /var/crash
* workaround: remove empty fluentd state files if needed after hard crash (see #1454)
* pull audit files out of journal, log quantity coupled with limited os disk io makes journal hard to use for anything else (@thekad fyi)

@thekad @charlesakalugwu I've given up with logging crash dmesg to journald for now - seems impossible to configure mdsd to be sure it gets into Geneva
@charlesakalugwu ditto read_from_head :-(

PR #1486, opened 2019-04-10
fluentd log split - update with nil check

mjudeikis:
```release-note
Split kubernetes containers logs in geneva
```


pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -22,6 +22,17 @@ data:
         tag journald
       </rule>
     </match>
+    <filter journald>
+      @type record_transformer
+      enable_ruby true
+      <record>
+       #k8s_apiserver_apiserver-vqxg4_kube-service-catalog_72ce4d73-5224-11e9-98d0-000d3a196756_0
+       CONTAINER ${unless record["CONTAINER_NAME"].nil? ; record["CONTAINER_NAME"].split("_")[1] ; else "" ; end; }
+       POD ${unless record["CONTAINER_NAME"].nil?; record["CONTAINER_NAME"].split("_")[2] ; else "" ; end; }
```
mjudeikis: `else ""` needed if we dont want to see `Null` values in Geneva. an empty string value is a default for all other empty fields.

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
```diff
@@ -22,6 +22,17 @@ data:
         tag journald
       </rule>
     </match>
+    <filter journald>
+      @type record_transformer
+      enable_ruby true
+      <record>
+       #k8s_apiserver_apiserver-vqxg4_kube-service-catalog_72ce4d73-5224-11e9-98d0-000d3a196756_0
+       CONTAINER ${unless record["CONTAINER_NAME"].nil? ; record["CONTAINER_NAME"].split("_")[1] ; else "" ; end; }
```
ehashman: Is this 1-indexed or is the [0] field special? I guess I don't actually know what the source data looks like.

mjudeikis: data format is comment field above. so first field [0] is `k8s_`

mjudeikis: comment is not very informative :/

ehashman: Ahh, makes sense. Maybe we should update the comment to be clearer? :)

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
ehashman: Out of curiosity, why do we prefer an empty string to Null here?

mjudeikis: my preference is to see a unified view in UI, not some fields `""` some `Null` . this way fields are either populated or `""`
 view now is empty fields if field is not populated 
image

I dont have a hard preference :man_shrugging:

jim-minter: empty string is not the default, nil is - please fix

jim-minter:
@mjudeikis I'd like #1455 to get in as a priority as it adds crash dump logging, and the 2 PRs will conflict.  Please rebase this (please also consider using the ternary operator like I do for MESSAGE at that point).

test/reporters/logger.go
```diff
@@ -0,0 +1,7 @@
+package reporters
+
+// GinkgoLogger suppresses logging unless "-v" is given to the gingko framework.
+type AzureLogger struct{}
```
charlesakalugwu: what exactly is `AzureLogger` doing here?

test/reporters/logger.go
```diff
@@ -0,0 +1,7 @@
...
+
+// GinkgoLogger implements Logger.
```
charlesakalugwu: please delete the commented out code

pkg/sync/v5/data/ConfigMap/openshift-azure-logging/fluentd-config.yaml
mjudeikis: image
nill is our new introduced values. <empty> is that any other components puts in into an empty field. I was just trying to follow data consistency

PR #1487, opened 2019-04-10
add note about jq to readme

jim-minter:
```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,15 @@
 ## Prerequisites
 
 1. **Utilities**.  Install the following:
-   1. [Golang 1.11.6](https://golang.org/dl)
+   1. [Golang 1.11.6](https://golang.org/dl) (can also use package manager)
    1. Latest [Azure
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,13 +7,15 @@
 ## Prerequisites
 
 1. **Utilities**.  Install the following:
-   1. [Golang 1.11.6](https://golang.org/dl)
+   1. [Golang 1.11.6](https://golang.org/dl) (can also use package manager)
```
ehashman: `jq` is probably recent enough to grab from system package managers but I feel like Golang and the other tools update too frequently to suggest this. As well, if you install a system Golang, I don't know that it sets up all the expected PATHs relative to one's homedir properly, hence a local userspace install is probably safer...

thekad: if we're talking about the `oc` client then yeah the one provided by the package manager is totally fine to use

jim-minter:
/retest

flake fixed in #1489

README.md
jim-minter: @ehashman superstition! it's fine!

PR #1491, opened 2019-04-11
adding e2e test for metricsbridge queries

Makdaam:
```release-note
NONE
```
Test, which checks if all configured queries return data (or are known not to return data).


test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+	token string
+}
+
+func (rt authorizingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("Authorization", "Bearer "+rt.token)
+	return rt.RoundTripper.RoundTrip(req)
+}
+
+var _ = Describe("Metricsbridge E2E check configured queries [Fake][EveryPR]", func() {
+	//queries which usually return no datapoints, but have been verified as valid manually
```
charlesakalugwu:
```suggestion
	// queries which usually return no datapoints, but have been verified as valid manually
```

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+	manuallyVerifiedQueries := map[string]bool{
+		"kubelet_runtime_operations_errors != 0": true,
+		"kubelet_docker_operations_errors != 0":  true,
+	}
+
+	BeforeEach(func() {
```
charlesakalugwu: take this out if not used

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+
+	It("should be possible to get data for each metric in defined queries", func() {
+		By("getting the configMap")
+		mbConfigMap, err := sanity.Checker.Client.Admin.CoreV1.ConfigMaps("openshift-azure-monitoring").Get("metrics-bridge", meta_v1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
```
charlesakalugwu: `Expect(mbConfigMap).NotTo(BeNil())`

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+
+		var mbConfig struct {
```
charlesakalugwu: consider centralizing the metrics bridge config somewhere and share that between here and the metrics bridge client

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+				token: string(token),
+			},
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+		prometheusApi := v1.NewAPI(cli)
+
+		for _, q := range mbConfig.Queries {
+			By(fmt.Sprintf("checking query %s", q.Query))
+			//checking if the whole query works
```
charlesakalugwu:
```suggestion
			// checking if the whole query works
```

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+			value, err := prometheusApi.Query(context.Background(), q.Query, time.Time{})
+			Expect(err).NotTo(HaveOccurred())
+			if len(value.String()) == 0 {
+				//if there are no datapoints in the literar query, check if it's by design
+				_, present := manuallyVerifiedQueries[q.Query]
+				if present {
```
charlesakalugwu:
```go
if !present {
	// try out some simpler check
	re := regexp.MustCompile(`(.+)\s*!=\s*0`)  // if ends with != 0
	qStr := re.ReplaceAllString(q.Query, "$1") //remove != 0

	re = regexp.MustCompile(`rate\((.+)\[1m\]\)`) // if contains rate(...[1m])
	qStr = re.ReplaceAllString(qStr, "$1")        //unwrap from rate()

	value, err := prometheusApi.Query(context.Background(), qStr, time.Time{})
	Expect(value.String()).NotTo(BeEmpty())
}
```

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/prometheus/client_golang/api"
+	"github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/sirupsen/logrus"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/openshift-azure/test/sanity"
+)
+
+type authorizingRoundTripper struct {
```
charlesakalugwu: consider moving this `authorizingRoundTripper` together with other roundtrippers we already use across the code into `pkg/util/rountrippers`

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,133 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			if len(value.String()) == 0 {
+				//if there are no datapoints in the literar query, check if it's by design
+				_, present := manuallyVerifiedQueries[q.Query]
+				if present {
+					//query verified manually
+					continue
+				}
+
+				re := regexp.MustCompile(`(.+)\s*!=\s*0`)  // if ends with != 0
```
charlesakalugwu: consider moving this and the following comments into a comment section above this test spec which documents the entire flow logic for this test (as you explained to me :) )

pkg/util/roundtrippers/roundtrippers.go
```diff
@@ -0,0 +1,13 @@
+package roundtrippers
```
charlesakalugwu: maybe also move `retryingRoundTripper` from `pkg/cluster/kubeclient/roundtrip.go` into this file?

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: done.

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: done.

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: reduced the config struct to what I need after consulting with Charles.

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,108 @@
...
+		"kubelet_runtime_operations_errors != 0": true,
+		"kubelet_docker_operations_errors != 0":  true,
+	}
+	It("should be possible to get data for each metric in defined queries", func() {
+		By("getting the configMap")
+		mbConfigMap, err := sanity.Checker.Client.Admin.CoreV1.ConfigMaps("openshift-azure-monitoring").Get("metrics-bridge", meta_v1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(mbConfigMap).ToNot(BeNil())
+
+		var mbConfig struct {
```
jim-minter: much of the code from here to line 76 is duplicate, please can you library-ise it?

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,108 @@
...
+			// checking if the whole query works
+			value, err := prometheusApi.Query(context.Background(), q.Query, time.Time{})
+			Expect(err).NotTo(HaveOccurred())
+			if len(value.String()) == 0 {
+				//check if query has been verified manually despite not returning datapoints
+				_, present := manuallyVerifiedQueries[q.Query]
+				if !present {
+					//in case of queries pruned by ' != 0', remove the suffix
+					qStr := q.Query
+					if strings.HasSuffix(qStr, " != 0") {
```
jim-minter: this is pretty brittle, I think I'd try something like

```golang
// e.g. sum(foo{bar}) without (baz) != 0 -> foo{bar}
m := regexp.MustCompile(`\((.+?)\)`).FindAllStringSubmatch(qStr)
if m != nil {
	qStr = m[1]
}
// e.g. foo{bar} != 0 -> foo
//      foo != 0 -> foo
//      foo -> foo
qStr = regexp.MustCompile(`(?i)[^_a-z0-9].*`).ReplaceString(qStr, "")
```

still grotty, but I think what you're doing is worth doing

mjudeikis:
lgtm, needs rebase

jim-minter:
@Makdaam please rebase

jim-minter:
/lgtm cancel

comments need addressing before merge

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: changed to using the struct in pkg/entrypoint/metricsbridge instead.

test/e2e/specs/fakerp/metricsbridge.go
Makdaam: done,
more flexibility in queries is convincing.

Makdaam:
rebased,
replaced my string matching with regexps suggested by @jim-minter

mjudeikis:
These test run in random order. So if it runs super early in the process Prometheus might be just starting, and it might be that even Roundtripper retries will not be enough. Because we have a prometheus/monitoring health test, would it make sense to chain them?

Makdaam:
@mjudeikis agreed... in progress.

Makdaam:
Removed [EveryPR] from Prometheus and Metricsbridge tests yesterday. The cluster is not ready to pass those tests right after provisioning. I haven't found a way to check when the metrics are populated after provisioning.

mjudeikis:
> Removed [EveryPR] from Prometheus and Metricsbridge tests yesterday. The cluster is not ready to pass those tests right after provisioning. I haven't found a way to check when the metrics are populated after provisioning.

Prometheus test can run after "Cluster-Validation test" ran, because they have test "Monitoring stack is Healthy". This should be enough. I think we need to orchestrate all tests to be running after cluster validation passed. So we know we running it on "healthy" cluster.

pkg/util/roundtrippers/roundtrippers.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// 3. Even with the default 30 second Dial timeout, sometimes we see unexplained
+//    Dial timeout failures.  retryingRoundTripper retries in these cases.
+//
+// 4. The default TLS handshake timeout is 10 seconds.  Sometimes we see this
+//    timeout triggered.  retryingRoundTripper also retries in these cases.
+
+var timerExpired = errors.New("retryingRoundTripper timer expired")
+
+type RetryingRoundTripper struct {
```
jim-minter: is this not duplicating code?  I don't see where you're removing pkg/cluster/kubeclient/roundtrip.go now?

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Test checks if all configured queries are valid prometheus queries, and if the metric exists
+// Existence of metric is checked by either:
+//  - query returning data
+//  - query being on a list of manually verified correct queries (that normally don't return any data)
+//  - query having ' != 0' at the end, that returns data after this condition is removed
+//  - query calculating rate( x [1m]) of a metric x, when querying for metric x returns data
+var _ = Describe("Metricsbridge E2E check configured queries [Fake]", func() {
+	// queries which usually return no datapoints, but have been verified as valid manually
+	manuallyVerifiedQueries := map[string]bool{
```
jim-minter: I am wondering if the following will get rid of the need for manuallyVerifiedQueries:
1) run all the queries by prometheus (should catch syntax errors)
2) fetch and parse the output of the federate endpoint (this lists all known metrics); for each query extract the metric name and ensure it's in the federate list
(for a hint on fetching and parsing the federate endpoint see an older version of metricsbridge which had this code)

test/e2e/specs/fakerp/prometheus.go
```diff
@@ -57,10 +59,15 @@ var _ = Describe("Prometheus E2E tests [Fake]", func() {
 		req.Header.Add("Authorization", "Bearer "+string(token))
 
 		cli := &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{
-					InsecureSkipVerify: true,
+			Transport: &kubeclient.RetryingRoundTripper{
```
jim-minter: why is this being changed? did you mean to change it to AuthorizingRoundTripper?

test/e2e/specs/fakerp/metricsbridge.go
```diff
@@ -0,0 +1,114 @@
...
+			// checking if the whole query works
+			value, err := prometheusApi.Query(context.Background(), q.Query, time.Time{})
+			Expect(err).NotTo(HaveOccurred())
+			if len(value.String()) == 0 {
+				//check if query has been verified manually despite not returning datapoints
+				_, present := manuallyVerifiedQueries[q.Query]
+				if !present {
+					qStr := q.Query
+
+					//in case of rate( x [1m]), unwrap x
```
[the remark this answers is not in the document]
Makdaam: It is, my bad. Removed.

test/e2e/specs/fakerp/prometheus.go
Makdaam: my bad attempt at fixing the issue of the test not working right after provisioning a new cluster. Rolled back.

pkg/util/roundtrippers/roundtrippers.go
Makdaam: It is. Removed, since it is used only in kubeclient after the rollback.

Makdaam:
/hold
fixing dependencies

Makdaam:
/hold cancel
Added prometheus/prometheus to dependencies for test.
I only need promql, but I don't think there's a way to pull only one submodule from a repository.

PR #1493, opened 2019-04-11
use mail attribute instead of userPrincipalName to map groups and users

jim-minter:
otherwise guest AAD user's don't work correctly (get #EXT# and more interposed in username)

```release-note
NONE
```

@asalkeld you were right about this, ptal ;-)


asalkeld:
looks like you need a change in pkg/controllers/customeradmin/group_mapping_test.go

jim-minter:
thanks, pushed

asalkeld:
flake https://github.com/openshift/openshift-azure/issues/1400
/test scaleupdown

PR #1501, opened 2019-04-12
fix customeradmin test

### LINKED ISSUE #1498: azure group controller crashes when mail is nil
```
[mjudeiki@redhat openshift-azure]$ oc logs -f customer-admin-controller-776684f766-8gs4w
time="2019-04-12T12:35:42Z" level=info msg="azure-controller pod starting, git commit v4.0-154-gbf3371a-clean"
time="2019-04-12T12:35:42Z" level=info msg="starting manager"
time="2019-04-12T12:35:43Z" level=debug msg="NSC: reconciling namespace e2e-test-qmtf8"
time="2019-04-12T12:35:43Z" level=debug msg="NSC: creating rolebinding e2e-test-qmtf8/osa-customer-admin"
time="2019-04-12T12:35:43Z" level=debug msg="RBC: reconciling rolebinding e2e-test-qmtf8/osa-customer-admin"
time="2019-04-12T12:35:43Z" level=debug msg="RBC: reconciling rolebinding e2e-test-qmtf8/osa-customer-admin-project"
time="2019-04-12T12:35:43Z" level=debug msg="NSC: creating rolebinding e2e-test-qmtf8/osa-customer-admin-project"
time="2019-04-12T12:36:42Z" level=debug msg="AAD Group Reconciler (poll).."
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x105a8e9]
goroutine 76 [running]:
github.com/openshift/openshift-azure/pkg/controllers/customeradmin.fromMSGraphGroup(0xc0000adc80, 0x0, 0x130fe98, 0x13, 0xc000254c60, 0x2, 0x2, 0x0, 0x0)                                                                                                                                
        /go/src/github.com/openshift/openshift-azure/pkg/controllers/customeradmin/group_mapping.go:34 +0x129
github.com/openshift/openshift-azure/pkg/controllers/customeradmin.updateKubeGroup(0xc0000adc80, 0x14847c0, 0xc0003c0440, 0x130fe98, 0x13, 0xc000254c60, 0x2, 0x2, 0x0, 0x1)                                                                                                             
        /go/src/github.com/openshift/openshift-azure/pkg/controllers/customeradmin/group_reconcile.go:23 +0x126
github.com/openshift/openshift-azure/pkg/controllers/customeradmin.reconcileGroups(0xc0000adc80, 0x14644a0, 0xc00033ea00, 0x14847c0, 0xc0003c0440, 0xc0004367b0, 0x42e4fb, 0x138adc8)                                                                                                    
        /go/src/github.com/openshift/openshift-azure/pkg/controllers/customeradmin/group_reconcile.go:53 +0xfa
github.com/openshift/openshift-azure/pkg/controllers/customeradmin.(*reconcileGroup).pollEvent(0xc000427080, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/openshift/openshift-azure/pkg/controllers/customeradmin/group_controller.go:100 +0xad
github.com/openshift/openshift-azure/pkg/controllers/customeradmin.(*reconcileGroup).pollEvent-fm(0x0, 0x0, 0x0, 0x0, 0xc00037f288)
        /go/src/github.com/openshift/openshift-azure/pkg/controllers/customeradmin/group_controller.go:75 +0x52
github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/predicate.Funcs.Generic(0x0, 0x0, 0x0, 0xc0003c0580, 0x0, 0x0, 0x0, 0x0, 0x101)                                                                                                                           
        /go/src/github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/predicate/predicate.go:86 +0x53
github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/source.(*Channel).Start.func2(0xc0004444e0, 0xc0003c0590, 0x1, 0x1, 0x14809c0, 0x1fc0c18, 0x148f260, 0xc00042cf20)                                                                                        
        /go/src/github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/source/source.go:168 +0x18e
created by github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/source.(*Channel).Start
        /go/src/github.com/openshift/openshift-azure/vendor/sigs.k8s.io/controller-runtime/pkg/source/source.go:164 +0x120
```

jim-minter:
```release-note
NONE
```

fixes #1498

### DIFF
--- a/test/e2e/specs/customer_admin.go
+++ b/test/e2e/specs/customer_admin.go
@@ -2,13 +2,14 @@ package specs
 
 import (
 	"errors"
-	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"github.com/ghodss/yaml"
 	onwv1 "github.com/openshift/api/network/v1"
+	userv1 "github.com/openshift/api/user/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -16,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"github.com/openshift/openshift-azure/pkg/api"
 	"github.com/openshift/openshift-azure/pkg/util/random"
 	"github.com/openshift/openshift-azure/test/sanity"
 )
@@ -233,15 +235,24 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
 	It("should sync AAD admin group", func() {
 		syncSecret, err := sanity.Checker.Client.Admin.CoreV1.Secrets("openshift-infra").Get("aad-group-sync-config", metav1.GetOptions{})
 		Expect(err).ToNot(HaveOccurred())
-		id := "44e69b4e-2e70-42df-bb97-3a890730d7b0"
-		testUser := "testuserdisabled"
-		gid := syncSecret.Data["customerAdminGroupId"]
-		if strings.EqualFold(string(gid), id) {
-			// test the users
-			oca, err := sanity.Checker.Client.Admin.UserV1.Groups().Get("osa-customer-admins", metav1.GetOptions{})
+
+		var provider api.AADIdentityProvider
+		err = yaml.Unmarshal(syncSecret.Data["aad-group-sync.yaml"], &provider)
+		Expect(err).ToNot(HaveOccurred())
+
+		if provider.CustomerAdminGroupID != nil &&
+			*provider.CustomerAdminGroupID == "44e69b4e-2e70-42df-bb97-3a890730d7b0" {
+			var g *userv1.Group
+			err = wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
+				g, err = sanity.Checker.Client.Admin.UserV1.Groups().Get("osa-customer-admins", metav1.GetOptions{})
+				if kerrors.IsNotFound(err) {
+					return false, nil
+				}
+				return err == nil, err
+			})
 			Expect(err).ToNot(HaveOccurred())
-			Expect(len(oca.Users)).To(Equal(1))
-			Expect(strings.HasPrefix(oca.Users[0], testUser)).To(BeTrue())
+			Expect(g.Users).To(HaveLen(1))
+			Expect(g.Users[0]).To(HavePrefix("testuserdisabled"))
 		}
 	})
 	// Placeholder to test that a ded admin cannot delete pods in the default or openshift- namespaces


test/e2e/specs/customer_admin.go
```diff
@@ -233,15 +234,17 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
 		syncSecret, err := sanity.Checker.Client.Admin.CoreV1.Secrets("openshift-infra").Get("aad-group-sync-config", metav1.GetOptions{})
 		Expect(err).ToNot(HaveOccurred())
-		id := "44e69b4e-2e70-42df-bb97-3a890730d7b0"
-		testUser := "testuserdisabled"
-		gid := syncSecret.Data["customerAdminGroupId"]
-		if strings.EqualFold(string(gid), id) {
-			// test the users
+
+		var provider api.AADIdentityProvider
+		err = yaml.Unmarshal(syncSecret.Data["aad-group-sync.yaml"], &provider)
```
ehashman: Just to make sure I'm understanding this right, looking at e.g. https://github.com/openshift/openshift-azure/blob/28ad9dd9b76bfc016186e01c679ed601b27ac08f/pkg/sync/v3/data/Secret/openshift-infra/aad-group-sync-config.yaml#L3-L7

Was the `gid` previously null and hence the test assertions below weren't invoking? Because I don't see a customerAdminGroupId defined that it could have pulled the value from.

asalkeld: it is set by the CI system (from a secret) exported here: https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L292
so it should be set in e2e jobs (this should be set to the AAD group called "CI test" which has one user called "testuserdisabled").

asalkeld: So the CustomerAdminGroupID is always set, but the retrieved user.Mail was nil. i don't think this e2e test change is strictly needed, as the bug would cause the sync process to fail and it would already fail this test.

test/e2e/specs/customer_admin.go
```diff
@@ -233,15 +236,26 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
-		if strings.EqualFold(string(gid), id) {
-			// test the users
-			oca, err := sanity.Checker.Client.Admin.UserV1.Groups().Get("osa-customer-admins", metav1.GetOptions{})
+
+		var provider api.AADIdentityProvider
+		err = yaml.Unmarshal(syncSecret.Data["aad-group-sync.yaml"], &provider)
+		Expect(err).ToNot(HaveOccurred())
+
+		if provider.CustomerAdminGroupID != nil &&
+			*provider.CustomerAdminGroupID == "44e69b4e-2e70-42df-bb97-3a890730d7b0" {
```
asalkeld: shouldn't we drop this test for the gid? it makes it hard to reconfigure the test without breaking the test. Also shouldn't the test fail if provider.CustomerAdminGroupID == nil? otherwise we silently skip the test if groupId is not set. and if running locally you don't know that a test is not running.

asalkeld:
The controller is running fine.
```
openshift-infra                     customer-admin-controller-7b455cbd44-j2llf    1/1       Running   0          24m 
```
I also checked out the code and ran the e2e fine (after remembering to push the image).
Could it be that you need to rebase on top of the latest "azure" image? maybe it's not building the controller image correctly?

test/e2e/specs/customer_admin.go
mjudeikis: ok, I might not get this, but this fix is because configMap structure changed?

jim-minter:
@asalkeld #1514 is the business portion of this PR, please lgtm it if it passes
I think the issue with the reworked test here is that it's running using the auto-generated application credentials, and that app does not have the directory read AAD permissions granted.  I'll fix this up later by having the fakerp pass the non-autogenerated application credentials to the controller.

jim-minter:
/retest
flake #1533

test/e2e/specs/customer_admin.go
jim-minter: I thought I replied to this, but I guess I didn't.
@ehashman you are right.
@asalkeld the fix here is that the test was not ever running

test/e2e/specs/customer_admin.go
jim-minter: it's tricky.  The e2e test doesn't have permissions to query Azure to read the group memberships, so it's hard coded.  I don't want the test to fail on clusters that are not configured with that group id.  The downside is that it's still easy for the test to skip and for us not to realise that the functionality is broken :-/

test/e2e/specs/customer_admin.go
```diff
@@ -233,15 +236,26 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
+			var g *userv1.Group
+			err = wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
```
ehashman: It will always take 10 minutes for this test to fail, which is sort of unfortunate, but not blocking.

jim-minter: you're right, fixing

PR #1503, opened 2019-04-12
Adding a unit test to check resource's data

### LINKED ISSUE #1449: Create unit test to ensure resources structure matches its directory location
There have been cases in development where you are working on a resource inside a given namespace's but forget to update the namespace setting inside the yaml.
Write a unit test to validate each resource is in the directory it should be, i.e. validate namespace, kind, etc.

thekad:
To avoid possibly checking in or working with yaml files that don't
match their filesystem distribution, example: working in a file that is
checked-in to a particular namespace but the yaml says default.
Specifically walking the data directory instead of polling `AssetNames()`
so running bindata isn't necessary to run this test.

```release-note
NONE
```
Closes #1449


pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,70 @@
...
+		o, err := unmarshal(b)
+		if err != nil {
+			return fmt.Errorf("error unmarshaling %s: %v", path, err)
+		}
+
+		parts := strings.Split(path, "/")
+		kind := parts[1]
+		// cluster-wide kinds have an empty string as namespace
+		namespace := ""
+		if !clusterKinds[kind] {
```
jim-minter: if len(parts) == 3

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,70 @@
...
+			namespace = parts[2]
+		}
+
+		// check the namespace matches
+		if namespace != o.GetNamespace() {
+			return fmt.Errorf("%s namespace doesn't match its path (%s != %s)", path, namespace, o.GetNamespace())
+		}
+
+		// check the kind is the prefix for the given template
+		prefix := o.GetKind()
```
jim-minter: can also partially check the apiVersion

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,70 @@
...
+		kind := parts[1]
+		// cluster-wide kinds have an empty string as namespace
+		namespace := ""
+		if !clusterKinds[kind] {
+			namespace = parts[2]
+		}
+
+		// check the namespace matches
+		if namespace != o.GetNamespace() {
+			return fmt.Errorf("%s namespace doesn't match its path (%s != %s)", path, namespace, o.GetNamespace())
```
jim-minter: gather a []error and print it at the end

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestTranslatableData(t *testing.T) {
+	clusterKinds := map[string]bool{
```
jim-minter: can remove this

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,79 @@
...
+		oKind := o.GetKind()
+		if strings.Contains(apiVersion, "/") {
+			oKind = fmt.Sprintf("%s.%s", oKind, strings.Split(apiVersion, "/")[0])
+		}
+		if !strings.HasPrefix(kind, oKind) {
+			errors = append(errors, fmt.Sprintf("%s kind doesn't match its path (%s != %s)", path, kind, oKind))
+		}
+
+		// check the name matches, caveat: all colons (":") are replaced with single dash ("-")
+		oName := re.ReplaceAllString(o.GetName(), "-")
```
jim-minter: Avoid using regular expressions whenever you can.  Replace with `oName := strings.Replace(o.GetName(), ":", "-", -1)`

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,79 @@
...
+			errors = append(errors, fmt.Sprintf("%s namespace doesn't match its path (%s != %s)", path, namespace, o.GetNamespace()))
+		}
+
+		// check the kind is the prefix for the given template
+		apiVersion := o.GetAPIVersion()
+		oKind := o.GetKind()
+		if strings.Contains(apiVersion, "/") {
+			oKind = fmt.Sprintf("%s.%s", oKind, strings.Split(apiVersion, "/")[0])
+		}
+		if !strings.HasPrefix(kind, oKind) {
```
jim-minter: why can't this be `if kind != oKind` ?

pkg/sync/v5/translate_test.go
thekad: Ah but there are some names that have e.g. `system::foo:bar` (double colon) and those are collapsed into a single dash. Either I find a way to collapse one or more `:` into a single dash (which proved snap easy with regexp) or I rename the files to e.g. `system--foo-bar.yaml`

jim-minter: Ha, that's a discrepancy I hadn't spotted.  Please use strings.Replace() and rename the files.

jim-minter:
Also please squash.

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,77 @@
...
+		apiVersion := o.GetAPIVersion()
+		oKind := o.GetKind()
+		if strings.Contains(apiVersion, "/") {
+			oKind = fmt.Sprintf("%s.%s", oKind, strings.Split(apiVersion, "/")[0])
+		}
+		if kind != oKind {
+			errors = append(errors, fmt.Sprintf("%s kind doesn't match its path (%s != %s)", path, kind, oKind))
+		}
+
+		// check the name matches, caveat: all colons (":") are replaced with single dash ("-")
```
jim-minter: comment no longer fits

pkg/sync/v5/translate_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func TestTranslatableData(t *testing.T) {
+	var errors []string
+
+	err := filepath.Walk("data", func(path string, info os.FileInfo, errIn error) error {
+		if errIn != nil {
+			return errIn
+		}
+
+		if info.IsDir() || !strings.HasSuffix(info.Name(), ".yaml") {
```
jim-minter: prob should remove the .yaml suffix check and not allow any other files in there

PR #1507, opened 2019-04-14
Azure prow refactor

jim-minter:
```release-note
NONE
```

Remove ci-operator for all use cases except conformance and bushslicer

Counterpart to https://github.com/openshift/release/pull/3464


hack/verify/validate-codecov.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
```
asalkeld: I think the above way is the more correct way of doing it.

Makefile
```diff
@@ -54,15 +50,18 @@ sync-run: generate
...
 	go build -ldflags ${LDFLAGS} ./cmd/$@
+.PHONY: monitoring
 
 monitoring-run: monitoring
 	./hack/monitoring.sh
 
 monitoring-stop:
 	./hack/monitoring.sh clean
 
+.PHONY: monitoring-run monitoring-stop
```
asalkeld: the resultant Makefile .PHONY's look a bit of a mess. can you either place them after the target, in sections or at the top?

Makefile
```diff
@@ -1,30 +1,25 @@
...
+	@rm -rf secrets
+	@mkdir secrets
+	@oc extract -n azure secret/cluster-secrets-azure --to=secrets >/dev/null
 
 clean:
 	rm -f coverage.out azure releasenotes
 
 generate:
-	go generate ./...
+	$(shell [[ -e /var/run/secrets/kubernetes.io ]] || go generate ./...)
```
ehashman: Do we want to use `[` for shell compatibility?

Makefile
ehashman: +1, I'd vote for a single .PHONY target at the top.

jim-minter: can do this as a follow-up - what I'm even more interested in is reducing the Makefile down even further...

jim-minter:
> General comment: we're removing -x from a lot of these scripts, but I've found the output pretty helpful for debugging... what's the benefit of doing so? Less log volume?

The tradeoff is between mental load to read through lots of `-x` logs and risking not having logged the critical thing needed to debug.  Hopefully we can get to a point where the boilerplate test setup stuff is minimal and reliable enough that we don't need `-x` there.  I'm not against judicious logging, but IMO we'd gone too far.

jim-minter:
tests passed

PR #1510, opened 2019-04-15
added [EveryPR] to Prometheus test, [LongRunning] to admin e2e test

Makdaam:
```release-note
NONE
```
Looks like were e2e tests which were not run in some time because they lacked labels. This is a first step in making sure that a test is either run on every PR, or is part of a periodic run.

### DIFF
--- a/test/e2e/specs/fakerp/admin.go
+++ b/test/e2e/specs/fakerp/admin.go
@@ -19,7 +19,7 @@ import (
 	"github.com/openshift/openshift-azure/test/sanity"
 )
 
-var _ = Describe("Openshift on Azure admin e2e tests [Fake]", func() {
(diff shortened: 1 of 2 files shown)


thekad:
@Makdaam you mention adding LongRunning in the title but there's no changes referencing that, forgot to push ?

test/e2e/specs/fakerp/admin.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/openshift/openshift-azure/test/sanity"
 )
 
-var _ = Describe("Openshift on Azure admin e2e tests [Fake]", func() {
+var _ = Describe("Openshift on Azure admin e2e tests [Fake][EveryPR]", func() {
```
jim-minter: I'm presuming you're doing this with a view to making EveryPR the focus of the `make e2e` target - as it stands this change is insufficient to make this test run every PR.

PR #1518, opened 2019-04-16
remove insights ref from plugin code

mjudeikis:
```release-note
Remove insights reference from the main code base
```

RP main repo uses Azure SDK v24.1.0:
Azure SDK v24.1.0 has github.com/Azure/azure-sdk-for-go/services/monitor/mgmt/2017-09-01/insights that the plugin references.

RP’s E2E test repo uses Azure SDK 26.7.0:
In v26.7.0 I believe it got moved to: https://github.com/Azure/azure-sdk-for-go/tree/fb7488208f72a83bc60e921f94262e5bcb74e03a/services/preview/monitor/mgmt

The plugin is not using insights anymore directly. Only tests, and hacks. Removing insights client reference from the main code base, and keeping it only in the tests and hacks.


jim-minter:
/hold

if it works for MSFT I prefer the approach in #1526

Issue #1523, opened 2019-04-16
set ETCD backend bytes to the maximum 8G to account for high load proactively


gburges:
ETCD_QUOTA_BACKEND_BYTES is set to 4GB today by default. In order to accommodate high (for example pressure testing) workload, we should set it to the maximum 8GB

gburges:
https://github.com/openshift/openshift-azure/pull/1524

PR #1524, opened 2019-04-16
fix for https://github.com/openshift/openshift-azure/issues/1523

gburges:
```release-note

fix for https://github.com/openshift/openshift-azure/issues/1523
```


jim-minter:
I don't want to do this without a need and input from the etcd team.  It doesn't need to be done for GA.  Let's revisit later.

PR #1525, opened 2019-04-16
Add validation for cluster names and locations

### LINKED ISSUE #1116: Don't allow cluster-names that don't comply with DNS spec
Our FakeRP doesn't validate the name of the cluster for dots and other non DNS spec compliant characters, which may lead to unintended consequences.
Investigate if this is also a problem with RealRP.

ehashman:
```release-note
Validate cluster names to ensure they will comply with the DNS spec.
Validate location names to ensure they match the standard naming scheme.
```
Closes #1116.

This regex is sort of arbitrary; I followed MS' recommendation for domain name label best practices, except that I am allowing capital letters and a length of 2-63 characters. Happy to modify this as necessary.

### DIFF
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -12,6 +12,8 @@ import (
 )
 
 var (
+	rxClusterName = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]$`)
+
 	rxRfc1123 = regexp.MustCompile(`(?i)^` +
 		`([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])` +
 		`(\.([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]))*` +
@@ -108,6 +110,10 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+}
+
+func isValidClusterName(n string) bool {
+	return rxClusterName.MatchString(n)
 }
 
 func isValidHostname(h string) bool {
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	rxClusterName = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]$`)
+	rxClusterName = regexp.MustCompile(`(?i)^([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])$`)
 
 	rxRfc1123 = regexp.MustCompile(`(?i)^` +
 		`([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])` +
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -14,6 +14,8 @@ import (
 var (
 	rxClusterName = regexp.MustCompile(`(?i)^([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])$`)
 
+	rxLocation = regexp.MustCompile(`(?i)^[a-z0-9]+$`)
+
 	rxRfc1123 = regexp.MustCompile(`(?i)^` +
 		`([a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9])` +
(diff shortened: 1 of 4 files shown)


pkg/api/validate/types_test.go
```diff
@@ -56,6 +56,38 @@ func TestValidatePluginVersion(t *testing.T) {
 	}
 }
 
+func TestIsValidClusterName(t *testing.T) {
```
charlesakalugwu: Instead of having seaparate testcase slices, you can merge them together and do something like this:

```
func TestIsValidClusterName(t *testing.T) {
	tests := []struct {
		clusterName string
		wantResult  bool
	}{
		{clusterName: "", wantResult: true},
		{clusterName: "k", wantResult: true},
		{clusterName: "my.cluster", wantResult: true},
		{clusterName: "has spaces", wantResult: true},
		{clusterName: "random#characters?", wantResult: true},
		{clusterName: "1234567890", wantResult: true},
		{clusterName: "0cluster", wantResult: true},
		{clusterName: "cluster-", wantResult: true},
		{clusterName: "-cluster", wantResult: true},
		{clusterName: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", wantResult: true},
		{clusterName: "verylongtoolongnameitslongerthan63characterswhoopsthiswontworkatall", wantResult: true},
		{clusterName: "k0", wantResult: true},
		{clusterName: "osa-testing", wantResult: true},
		{clusterName: "ExampleCluster111", wantResult: true},
		{clusterName: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", wantResult: true},
	}
	for _, test := range tests {
		isValid := isValidClusterName(test.clusterName)
		if isValid != test.wantResult {
			t.Errorf("IsValidClusterName(%v) expected %v, got %v", test.clusterName, test.wantResult, isValid)
		}
	}
}
```

ehashman: I just copied this to follow the style of the surrounding tests, see e.g. https://github.com/openshift/openshift-azure/blob/ccf8397a916173471a802d8703b2e7baecd4b981/pkg/api/validate/types_test.go#L59-L88

pkg/api/validate/types.go
```diff
@@ -12,6 +12,8 @@ import (
 )
 
 var (
+	rxClusterName = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]$`)
```
jim-minter: this is almost a rfc1123 hostname portion but 0-9 is missing from the first regex section?  Given it's right next to rxRfc1123 why not actually copy verbatim from below and make this `(?i)^[a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]$` ?

jim-minter: this also makes it 1-63 rather than 2-63, fwiw

ehashman: So the reason I decided to not use `rxRfc1123` is that it doesn't permit capital letters, which I thought might be a bit counterintuitive in case customers want to use them in their cluster names. (I believe capital letters should be fine as far as DNS goes?)

The "first portion can't start with 0-9" is copied from the MS suggested regex, and if I had to guess, it's probably an overly cautious legacy parsing thing (e.g. identifiers in C can't start with a digit).

Do we want to allow single character cluster names?

jim-minter: `(?i)` means case insensitive in Golang regexps.
RFC952 (precursor to RFC1123) didn't allow initial digits.
Please make it `(?i)^[a-z0-9]|[a-z0-9][-a-z0-9]{0,61}[a-z0-9]$` then will merge.

ehashman: Oh, I missed the `(?i)` on the first line.

Had to add some parens to this regex to get it to match properly---otherwise, it would always match the first character if it was alpha-numeric :smile:

thekad:
/lgtm
@ehashman is right the parens are needed

pkg/api/validate/types.go
jim-minter: thanks!

jim-minter:
lgtm, but also please can you check that .Location is alphanumeric non-empty? `(?i)^[a-z0-9]+$`?

jim-minter:
/lgtm
thanks!

PR #1528, opened 2019-04-17
increase maximum nodes to 30

jim-minter:
```release-note
Increase maximum node count to 30
```


asalkeld:
@jim-minter don't we need a release note for this? seems like someone would want to know this...

PR #1529, opened 2019-04-17
don't return 500s in the fakerp, only 400s - otherwise the client will retry

jim-minter:
```release-note
NONE
```

This is rather braindead, any better alternatives?


asalkeld:
seems an ok idea

mjudeikis:
I think we need cherry-pick each command, but not sure now.
/lgtm

Issue #1297, opened 2019-03-17
500 Internal Error: UpdateWorkerAgentPoolUpdateScaleSet: Code="InternalExecutionError" Message="An internal execution error occurred."


jim-minter:
```
time="2019-03-17T16:45:28Z" level=info msg="scaling ss-infra-1552838711 capacity up from 1 to 2" func="github.com/openshift/openshift-azure/pkg/cluster/scaler.(*workerScaler).scaleUp" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/scaler/scaler.go:106"
time="2019-03-17T16:46:06Z" level=debug msg="500 Internal Error: UpdateWorkerAgentPoolUpdateScaleSet: Code=\"InternalExecutionError\" Message=\"An internal execution error occurred.\"" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).internalError" file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:17"
time="2019-03-17T16:46:06Z" level=debug msg="ending:   PUT /admin/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/e2e-dyvicpnl/providers/Microsoft.ContainerService/openShiftManagedClusters/e2e-dyvicpnl/forceUpdate" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).logger.func1" file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/middleware.go:15"

• Failure [2459.108 seconds]
Force Update E2E tests [ForceUpdate][Fake][LongRunning]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/fakerp/forceupdate.go:15
  should be possible for an SRE to force update a cluster [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/fakerp/forceupdate.go:30

  Expected error:
      <*errors.errorString | 0xc0004bc050>: {
          s: "500: 500 Internal Error: UpdateWorkerAgentPoolUpdateScaleSet: Code=\"InternalExecutionError\" Message=\"An internal execution error occurred.\"\n",
      }
      500: 500 Internal Error: UpdateWorkerAgentPoolUpdateScaleSet: Code="InternalExecutionError" Message="An internal execution error occurred."
      
  not to have occurred

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/fakerp/forceupdate.go:41
```

[...]

PR #1530, opened 2019-04-17
retry when we get network errors

### LINKED ISSUE #1400: write: connection reset by peer talking to storage account

asalkeld:
there is some inconsistent error handling for network errors, let's fix this up.

closes #1400

```release-note
NONE
```
/cc @jim-minter

### DIFF
--- a/pkg/cluster/initialize.go
+++ b/pkg/cluster/initialize.go
@@ -25,7 +25,7 @@ func (u *simpleUpgrader) initializeStorageClients(ctx context.Context) error {
 		}
 
 		var err error
-		u.storageClient, err = storage.NewClient(u.cs.Config.ConfigStorageAccount, u.cs.Config.ConfigStorageAccountKey, storage.DefaultBaseURL, storage.DefaultAPIVersion, true)
+		u.storageClient, err = storage.NewClient(u.log, u.cs.Config.ConfigStorageAccount, u.cs.Config.ConfigStorageAccountKey, storage.DefaultBaseURL, storage.DefaultAPIVersion, true)
 		if err != nil {
 			return err
 		}
--- a/pkg/cluster/kubeclient/ready.go
+++ b/pkg/cluster/kubeclient/ready.go
@@ -2,11 +2,13 @@ package kubeclient
 
 import (
 	"context"
+	"syscall"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	utilerrors "github.com/openshift/openshift-azure/pkg/util/errors"
 	"github.com/openshift/openshift-azure/pkg/util/ready"
 	"github.com/openshift/openshift-azure/pkg/util/wait"
 )
@@ -49,8 +51,8 @@ func (u *kubeclient) WaitForReadySyncPod(ctx context.Context) error {
 				return false, err
 			}
 
-			ready := ready.DeploymentIsReady(d)
-			if ready {
+			isReady := ready.DeploymentIsReady(d)
+			if isReady {
 				return true, nil
 			}
 
@@ -67,6 +69,9 @@ func (u *kubeclient) WaitForReadySyncPod(ctx context.Context) error {
 				if err, ok := err.(*errors.StatusError); ok && err.ErrStatus.Details != nil && len(err.ErrStatus.Details.Causes) == 1 {
 					u.log.Info(err.ErrStatus.Details.Causes[0].Message)
 				}
+				err = nil
+			case utilerrors.IsMatchingSyscallError(err, syscall.ECONNREFUSED):
+				u.log.Infof("WaitForReadySyncPod: will retry on the following error %v", err)
 				err = nil
 			}
 
--- a/pkg/util/azureclient/storage/storage.go
+++ b/pkg/util/azureclient/storage/storage.go
@@ -2,17 +2,14 @@ package storage
 
 import (
 	"bytes"
-	"context"
 	"io"
 	"io/ioutil"
 	"syscall"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/util/errors"
-	"github.com/openshift/openshift-azure/pkg/util/wait"
 )
 
 const (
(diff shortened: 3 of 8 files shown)


pkg/util/ready/ready.go
```diff
@@ -24,6 +29,30 @@ import (
 	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 )
 
+// IsNetworkUnreachableError return true if this is a connection error
+func IsNetworkUnreachableError(err error) bool {
```
charlesakalugwu: The naming of this function is a little confusing to me ...especially with that check for the `x509.UnknownAuthorityError`. It might be worth having a look at how `go-autorest` does this in `isTemporaryNetworkError(err)`.

```
// returns true if the specified error is a temporary network error or false if it's not.
// if the error doesn't implement the net.Error interface the return value is true.
func isTemporaryNetworkError(err error) bool {
	if netErr, ok := err.(net.Error); !ok || (ok && netErr.Temporary()) {
		return true
	}
	return false
}
```

asalkeld: yip, i looked there, but want to  catch some more errors... maybe name it isRetryableError()?

charlesakalugwu: maybe `isRetryableConnectionError`?

pkg/cluster/initialize.go
```diff
@@ -41,13 +44,24 @@ func (u *simpleUpgrader) writeBlob(blobName string, cs *api.OpenShiftManagedClus
...
 	json, err := json.Marshal(cs)
 	if err != nil {
 		return err
 	}
-
-	return b.CreateBlockBlobFromReader(bytes.NewReader(json), nil)
+	ctx := context.Background()
+	return wait.PollImmediateUntil(1*time.Second,
+		func() (bool, error) {
+			err = b.CreateBlockBlobFromReader(bytes.NewReader(json), nil)
```
jim-minter: can the retry somehow go into pkg/azureclient/storage so that the caller doesn't remember to have to do this?  this is how we handle it in the pkg/azureclient and in the kubeclient?

in the general case you'll need to do some magic so that you don't lose the contents of the reader if you get a network error half way through transmission (ugh)

pkg/cluster/kubeclient/ready.go
```diff
@@ -56,6 +56,9 @@ func (u *kubeclient) WaitForReadySyncPod(ctx context.Context) error {
 					u.log.Info(err.ErrStatus.Details.Causes[0].Message)
 				}
 				return false, nil
+			case ready.IsNetworkUnreachableError(err):
```
jim-minter: did you see this happen? should be dealt with 1 layer down in the kubeclient retryer

pkg/util/ready/ready.go
jim-minter: I don't want to refactor the different error handling points into one, for several reasons:
* golang error nesting differs a lot and really hampers this approach over time
* we lose understanding of what errors are actually happening at each point and why

asalkeld:
seems to be a problem with the change, investigating

pkg/util/azureclient/storage/storage.go
```diff
@@ -93,3 +101,28 @@ type blob struct {
...
 
 var _ Blob = &blob{}
+
+func (b *blob) CreateBlockBlobFromReader(blob io.Reader, options *storage.PutBlobOptions) error {
+	ctx := context.Background()
+	data, err := ioutil.ReadAll(blob)
+	if err != nil {
+		return err
+	}
+	sr := io.NewSectionReader(bytes.NewReader(data), 0, int64(len(data)))
```
jim-minter: can just use bytes.NewReader within the anonymous function below?

pkg/util/ready/ready.go
```diff
@@ -24,6 +27,26 @@ import (
 	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 )
 
+// IsMatchingSyscallError returns true when the error is one of the Errno's in match
+// it deals with the different ways of wrapping the syscalls.
+func IsMatchingSyscallError(err error, match []syscall.Errno) bool {
```
jim-minter: don't think pkg/util/ready is the right place for this - if nothing else, a new pkg/util/errors?
use `...syscall.Errno` instead of `[]syscall.Errno` to neaten up the call sites

pkg/util/ready/ready.go
```diff
@@ -24,6 +27,26 @@ import (
...
+	switch t := err.(type) {
+	case *url.Error:
+		return IsMatchingSyscallError(t.Err, match)
```
jim-minter: I believe golang doesn't support tail recursion - use a loop here instead.  I know this is a bit of a nit, but recursing here might encourage others to do it in a place where it does matter.

pkg/util/ready/ready.go
```diff
@@ -24,6 +27,26 @@ import (
...
+	case *os.SyscallError:
+		return IsMatchingSyscallError(t.Err, match)
+	case syscall.Errno:
+		for _, sc := range match {
+			if t == sc {
+				return true
+			}
+		}
+	}
```
jim-minter: I'm on the fence about adding a `default:` case + panic in the case of an unrecognised type - wdyt?

pkg/util/azureclient/storage/storage.go
```diff
@@ -93,3 +101,28 @@ type blob struct {
...
+	return wait.PollImmediateUntil(2*time.Second,
```
jim-minter: we are generally establishing a pattern why we try N times (3?) and log each failure.  Please do the same here.

pkg/util/ready/ready.go
asalkeld: ok

pkg/util/azureclient/storage/storage.go
asalkeld: i'll try

pkg/util/ready/ready.go
asalkeld: The point of the function is return true/false to "is this one of the syscall errors", so you can pass in
any error and you don't want to panic'ing, but rather returning false so that the caller can handle that
error differently.

pkg/util/azureclient/storage/storage.go
asalkeld: I remember why i didn't log, it's a long messy chain to pass the log variable into all the structs, but i can give it a go..

pkg/util/ready/ready.go
asalkeld: ok, i'll change to a for loop

pkg/cluster/kubeclient/ready.go
```diff
@@ -68,6 +70,9 @@ func (u *kubeclient) WaitForReadySyncPod(ctx context.Context) error {
 					u.log.Info(err.ErrStatus.Details.Causes[0].Message)
 				}
 				err = nil
+			case utilerrors.IsMatchingSyscallError(err, syscall.ECONNRESET):
```
jim-minter: have you seen this error occur?  please can you share a link?

pkg/util/azureclient/storage/storage.go
```diff
@@ -90,6 +112,27 @@ type Blob interface {
...
 
 var _ Blob = &blob{}
+
+func (b *blob) CreateBlockBlobFromReader(blob io.Reader, options *storage.PutBlobOptions) error {
+	ctx := context.Background()
+	data, err := ioutil.ReadAll(blob)
+	if err != nil {
+		return err
+	}
+	return wait.PollImmediateUntil(3*time.Second,
```
jim-minter: this is not retrying 3 times, it is retrying indefinitely every 3 seconds.  Please follow the pattern in, e.g. `func (rs *retrySender) Do`

pkg/cluster/kubeclient/ready.go
asalkeld: https://github.com/openshift/openshift-azure/issues/1400#issuecomment-483920988
(should be connection refused, i'll update)

pkg/util/azureclient/storage/storage.go
asalkeld: ok, I agree about the retrying 3 times, but the retrySender.Do() seems to be at the wrong level (too lower level) as it is going to make resetting the Reader() impossible.

asalkeld:
the error above was "quota exceeded"

asalkeld:
flakes #1582 #1297 
/retest

jim-minter:
pushed additional commit to use for loop instead of wait.PollImmediateUntil

/lgtm

PR #1535, opened 2019-04-17
Clean up e2e tests to avoid an extra git clone

### LINKED ISSUE #1160: Upgrade script rework
currently, upgrade script https://github.com/openshift/openshift-azure/blob/master/hack/upgrade-e2e.sh is working with 2 repositories. One is cloned by prow/ci-operator, other we clone based on release tag provided with `make` target.

We need to simplify this script so we go `git checkouts tags/source` on prow/ci-operator provided code, instead of cloning it for the second time.

This will simplify `GOPATH` management and secret passing around.

ehashman:
```release-note
NONE
```

Closes #1160.

Tested this all locally so this should pass CI!

### DIFF
--- a/hack/tests/e2e-scaleupdown-previous.sh
+++ b/hack/tests/e2e-scaleupdown-previous.sh
@@ -15,38 +15,27 @@ cleanup() {
     stop_monitoring
     make artifacts
 
-    if [[ -n "$T" ]]; then
-        rm -rf "$T"
-    fi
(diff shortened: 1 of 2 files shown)


jim-minter:
turns out the problem with this is that if a developer wants to run locally and their git repo isn't clean, we can easily scramble their work.  Let's not do this.  Sorry.

PR #1536, opened 2019-04-17
retry on NoBackingApplicationObject error

jim-minter:
```release-note
NONE
```

### DIFF
--- a/pkg/fakerp/aad.go
+++ b/pkg/fakerp/aad.go
@@ -3,6 +3,7 @@ package fakerp
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/authorization/mgmt/2015-07-01/authorization"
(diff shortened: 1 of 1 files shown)


pkg/fakerp/aad.go
```diff
@@ -73,8 +74,17 @@ func (am *aadManager) ensureApp(ctx context.Context, displayName string, p *api.
...
+			AppID: app.AppID,
+		})
+		// ugh: Azure client library doesn't have the types registered to
+		// unmarshal all the way down to this error code natively :-(
+		if strings.Contains(err.Error(), "NoBackingApplicationObject") {
+			return false, nil
+		}
+		return err == nil, err
 	})
 	if err != nil {
```
ehashman: Isn't `err` here always going to be a tuple (bool, err) and hence will never be nil?

jim-minter: No, err is of type error throughout here.  Not quite sure what's driving that thought?

ehashman: My misreading of the code since the return statements above all return tuples :) I went and read through the [`wait` documentation](https://godoc.org/k8s.io/apimachinery/pkg/util/wait#PollInfinite) and understand better what's going on here now.

jim-minter:
/retest
low on quota

ehashman:
/retest
quota

PR #1537, opened 2019-04-17
Check for sync pod readiness correctly

jim-minter:
```release-note
NONE
```

We hit a race[1] where we checked the pod's view of readiness before the controller had acted on a change in goal state (should have been checking ObservedGeneration)

[1] https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1531/pull-ci-openshift-openshift-azure-master-upgrade-v4.2/189/build-log.txt

@asalkeld @mjudeikis @ehashman fyi

### DIFF
--- a/pkg/cluster/kubeclient/ready.go
+++ b/pkg/cluster/kubeclient/ready.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/openshift-azure/pkg/util/ready"
 	"github.com/openshift/openshift-azure/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


pkg/cluster/kubeclient/ready.go
```diff
@@ -40,14 +41,27 @@ func (u *kubeclient) WaitForReadyWorker(ctx context.Context, hostname string) er
...
+
+			_, err = u.client.CoreV1().
 				Services("kube-system").
 				ProxyGet("", "sync", "", "/healthz/ready", nil).
 				DoRaw()
 
 			switch {
 			case err == nil:
-				return true, nil
+				return false, nil
```
ehashman: This switch will now never return `true` and is basically just differentiating between different failure modes... can we simplify it?

jim-minter:
nothing compiles for me today.

pkg/cluster/kubeclient/ready.go
jim-minter: yes, thanks, pushed

jim-minter:
/retest
quota

ehashman:
/retest
both quota

PR #1539, opened 2019-04-17
load balancer updates.

### LINKED ISSUE #906: Azure Load Balancer for ingress pointed at compute nodes (Router Configuration)
The Azure load balancer (ALB) providing ingress into the OCP router is "backended" by the compute nodes in the OCP cluster rather than the infra nodes which hosts the OpenShift Router pods.

The current configuration is potentially sub-optimal in that traffic ingresses in the following pattern:

Customer <-> ALB <-> Compute node scale set <- (traffic proxied via service) -> InfraNode <-> router pod using loadbalancer service <-> endpoint/pod

Typically the router pod is exposed via hostnetwork and the load balancer in front of it uses the infra nodes as their backend in order to ingress traffic.

The typical pattern for this traffic is:

Customer <-> LB <-> InfraNode <-> router pod using hostnetwork <-> endpoint/pod

I believe the pattern deployed crosses the SDN unnecessarily and includes additional hops and will potentially hit scalability and performance issues.

@jmencak @eparis @dak1n1 please comment to make sure I am correct

kwoodson:
Updates for the load balancer configuration to send only to backend infra nodes.

fixes #906 

```release-note
Update router service to ClusterIP
Update kubernetes load balancer health probe and rules
```


kwoodson:
This currently fails to parse with new rules.
`Cannot parse the request.`

Will continue to debug.

pkg/arm/v5/resources.go
```diff
@@ -38,7 +38,11 @@ const (
 	lbKubernetesName                              = "kubernetes" // must match KubeCloudSharedConfiguration ClusterName
 	lbKubernetesOutboundFrontendConfigurationName = "outbound"
 	lbKubernetesOutboundRuleName                  = "outbound"
+	lbKubernetes443RuleName                       = "port-443"
```
jim-minter: rename to be consistent with lbKubernetesProbeName443

pkg/arm/v5/resources.go
```diff
@@ -213,14 +217,85 @@ func (g *simpleGenerator) lbKubernetes() *network.LoadBalancer {
...
+			LoadBalancingRules: &[]network.LoadBalancingRule{
+				{
+					LoadBalancingRulePropertiesFormat: &network.LoadBalancingRulePropertiesFormat{
+						FrontendIPConfiguration: &network.SubResource{
+							ID: to.StringPtr(resourceid.ResourceID(
+								g.cs.Properties.AzProfile.SubscriptionID,
+								g.cs.Properties.AzProfile.ResourceGroup,
+								"Microsoft.Network/loadBalancers",
+								lbKubernetesName,
+							) + "/frontendIPConfigurations/" + lbKubernetesOutboundFrontendConfigurationName),
```
jim-minter: you need to create a new IP, and a new frontend configuration.  Don't reuse outbound.

pkg/arm/v5/resources.go
```diff
@@ -213,14 +217,85 @@ func (g *simpleGenerator) lbKubernetes() *network.LoadBalancer {
...
+								"Microsoft.Network/loadBalancers",
+								lbKubernetesName,
+							) + "/probes/" + lbKubernetesProbeName443),
+						},
+						Protocol:            network.TransportProtocolTCP,
+						LoadDistribution:    network.Default,
+						FrontendPort:        to.Int32Ptr(443),
+						BackendPort:         to.Int32Ptr(443),
+						EnableFloatingIP:    to.BoolPtr(false),
+						DisableOutboundSnat: to.BoolPtr(true),
```
jim-minter: why this line?

jim-minter: add IdleTimeoutInMinutes to match the others

pkg/sync/v5/data/Service/default/router.yaml
```diff
@@ -14,6 +14,8 @@ spec:
     port: 80
   - name: 443-tcp
     port: 443
+  - name: router-stats
```
jim-minter: you need to fix up service.beta.kubernetes.io/azure-dns-label-name above - that should go to where you're creating the new IP

pkg/arm/v5/resources.go
kwoodson: The disableSnat was because when using the outbound it must be disabled.  Since we are creating a new one I will remove this option.

Also set timeout equal to others.

pkg/sync/v5/data/Service/default/router.yaml
kwoodson: Added

pkg/arm/v5/resources.go
kwoodson: Created new ip and frontend.

pkg/arm/v5/resources.go
kwoodson: Updated.

mjudeikis:
Looks very close.
There is a small blimp here which We need to solve too. Current service `router` has type `LoadBalancer` . This means its not "patchable". We need to delete it and recreate. This means this makes sync pod to fail on upgrades.

I would suggest to add pre-hook function, which would be triggered before sync pod runs (or before the certain object gets applied).

Something like:
hook(object, version){
// if version matches 
// if object matches 
//    execute actions (in this case delete router service).
// hand back to normal sync pod flow where we update/create new service.
}

From the logs looks like traffic is not able to reach Infra nodes. But configuration looks fine. Still checking.

mjudeikis:
In addition, we removed from users to use the ability to use `Type: Loadbalancer` and port 443 and 80 in the backend pool, as it will collide with existing rules:
```
Load balancing rules /subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/mjudeikis-lb/providers/Microsoft.Network/loadBalancers/kubernetes/loadBalancingRules/port-443 and /subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/mjudeikis-lb/providers/Microsoft.Network/loadBalancers/kubernetes/loadBalancingRules/af104de1661b211e98cc1000d3a19b14-TCP-443 with floating IP disabled use the same protocol Tcp and backend port 443, and must not be used with the same backend address pool /subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/mjudeikis-lb/providers/Microsoft.Network/loadBalancers/kubernetes/backendAddressPools/kubernetes.
```

FAQ worth material

mjudeikis:
by any chance, router is not working because we don't have Security group changes?
image

pkg/arm/v5/resources.go
```diff
@@ -316,7 +448,36 @@ func (g *simpleGenerator) nsgMaster() *network.SecurityGroup {
...
-			SecurityRules: &[]network.SecurityRule{},
+			SecurityRules: &[]network.SecurityRule{
+				{
+					SecurityRulePropertiesFormat: &network.SecurityRulePropertiesFormat{
+						Description:              to.StringPtr("Allow HTTP traffic"),
+						Protocol:                 network.SecurityRuleProtocolTCP,
+						SourcePortRange:          to.StringPtr("*"),
+						DestinationPortRange:     to.StringPtr("80"),
+						SourceAddressPrefix:      to.StringPtr("Internet"),
+						DestinationAddressPrefix: to.StringPtr("*"),
```
kwoodson: Would like to be able to set this to the IP address of the frontend load balancer.  Any ideas on how to accomplish this?

pkg/arm/v5/resources.go
```diff
@@ -215,12 +241,98 @@ func (g *simpleGenerator) lbKubernetes() *network.LoadBalancer {
...
+						Probe: &network.SubResource{
+							ID: to.StringPtr(resourceid.ResourceID(
+								g.cs.Properties.AzProfile.SubscriptionID,
+								g.cs.Properties.AzProfile.ResourceGroup,
+								"Microsoft.Network/loadBalancers",
+								lbKubernetesName,
+							) + "/probes/" + lbKubernetesProbeName80),
+						},
+						Protocol:             network.TransportProtocolTCP,
+						LoadDistribution:     network.SourceIP,
```
kwoodson: I updated this from `Default` to `SourceIP`.  This generally helps with sessions.  
@jim-minter @mjudeikis  Thoughts?

pkg/arm/v5/resources.go
```diff
@@ -215,12 +241,98 @@ func (g *simpleGenerator) lbKubernetes() *network.LoadBalancer {
...
+			LoadBalancingRules: &[]network.LoadBalancingRule{
+				{
+					LoadBalancingRulePropertiesFormat: &network.LoadBalancingRulePropertiesFormat{
+						FrontendIPConfiguration: &network.SubResource{
+							ID: to.StringPtr(resourceid.ResourceID(
+								g.cs.Properties.AzProfile.SubscriptionID,
+								g.cs.Properties.AzProfile.ResourceGroup,
+								"Microsoft.Network/loadBalancers",
+								lbKubernetesName,
+							) + "/frontendIPConfigurations/" + lbKubernetesInboundFrontendConfigurationName),
```
y-cote: I wonder if it's worth using path.Join() here, in case paths elements are not clean. It seems like everything is constant, static, or generated correctly but just a thought.

pkg/arm/v5/resources.go
mjudeikis: This is good question. It might help with sessions if users are not using VPN or NAT to reach the Loadbalancer. We might build bottleneck without knowing it. But im just thinking outloud. Scrum discussion when @jim-minter is online?

kwoodson:
Verify failed but passes locally:
```
locally:
Generated assets have no material difference than what is committed.

Remotely:
go generate produced changes that were not already present
```

Updating and will attempt a new generate.

kwoodson:
Probably need a hand with what should or shouldn't be updated.

jim-minter:
This seemed like a good idea initially, but it is going to cause a lot of overhead for us if/when we introduce multiple routers/shards.  Let's drop this for now.  Open a separate PR to enable sticky sessions for the router Service.

kwoodson:
Opened: #1558

mjudeikis:
@jim-minter how sticky session would help us to not to send traffic via compute nodes? 

Count we do multiple routers/shards via LB Nat'ing and having multiple IP addresses? 
```
IP1:443.80 -> Router1:8443,8080
IP2:443:80 -> Router2:9443,9080
```
Thinking what is bigger evil. as More compute nodes cluster has a bigger percentage of app traffic circles via compute nodes. 
?

PR #1541, opened 2019-04-17
Branding: select 'azure' branding option for cluster console.

### LINKED ISSUE #1272: admin console branding follow-ups
1) please can you launch a cluster, manually change your aad app secret to break the cluster, then try to log in to the cluster.  Do you get to a page which is incorrectly branded?  If so, need to fix - ping me to discuss.
2) once MSFT merge their TLS work and rebase, need to manually check with them that the new branding is appearing correctly on the cluster console
3) once https://github.com/openshift/console/pull/1230 is released, need to update the admin console configuration so that the new branding is used

y-cote:
```release-note
- Web interface completely re-branded as Azure Red Hat OpenShift
```
closes #1272


thekad:
/lgtm
This should have a release note block, no?

y-cote:
> /lgtm
> This should have a release note block, no?

Not sure it break / change APIs or expected behavior or adds interfaces or interactive features, but I can add one still, wouldn't hurt.

PR #1548, opened 2019-04-22
Add default values to external APIs

### LINKED ISSUE #336: allow basic defaulting of some external api items

ehashman:
```release-note
Set some default values (e.g. RouterProfile) when not specified upon creating a new cluster config
```
Addresses #336.

- [x] Default RouterProfiles.
- [ ] Default masterPoolProfile/count=3
- [ ] Default agentPoolProfiles[name=infra]/count=3
- [ ] Default agentPoolProfiles[*]/osType=Linux
- [x] Unit test defaults
- [ ] Simplify test/manifests/fakerp/create.yaml

### DIFF
--- a/pkg/api/2018-09-30-preview/default.go
+++ b/pkg/api/2018-09-30-preview/default.go
@@ -5,6 +5,10 @@ import (
 )
 
 func setDefaults(oc *OpenShiftManagedCluster) {
+	if oc.Properties == nil {
+		oc.Properties = &Properties{}
+	}
+
 	if len(oc.Properties.RouterProfiles) == 0 {
 		oc.Properties.RouterProfiles = []RouterProfile{
 			{
--- a/pkg/api/2018-09-30-preview/default.go
+++ b/pkg/api/2018-09-30-preview/default.go
@@ -1,0 +1,15 @@
+package v20180930preview
+
+import (
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+func setDefaults(oc *OpenShiftManagedCluster) {
+	if len(oc.Properties.RouterProfiles) == 0 {
+		oc.Properties.RouterProfiles = []RouterProfile{
+			{
+				Name: to.StringPtr("default"),
+			},
+		}
+	}
+}
--- a/pkg/api/2019-04-30/default.go
+++ b/pkg/api/2019-04-30/default.go
@@ -5,6 +5,10 @@ import (
 )
 
 func setDefaults(oc *OpenShiftManagedCluster) {
+	if oc.Properties == nil {
+		oc.Properties = &Properties{}
+	}
+
 	if len(oc.Properties.RouterProfiles) == 0 {
 		oc.Properties.RouterProfiles = []RouterProfile{
 			{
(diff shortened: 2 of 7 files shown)


ehashman:
/retest

flake #1549

pkg/api/2018-09-30-preview/default.go
```diff
@@ -0,0 +1,13 @@
+package v20180930preview
+
+func setDefaults(oc *OpenShiftManagedCluster) {
+	defaultRouterName := "default"
```
kwoodson: default router name is `router`

pkg/api/2019-04-30/default.go
```diff
@@ -0,0 +1,13 @@
+package v20190430
+
+func setDefaults(oc *OpenShiftManagedCluster) {
+	defaultRouterName := "default"
```
kwoodson: same as above

pkg/api/2018-09-30-preview/default.go
ehashman: Hm, are you sure? I got this from https://github.com/openshift/openshift-azure/blob/a4338bc5aeaff48ba2a37bec28ae249231f47dd8/pkg/fakerp/fakerp.go#L275-L281

I'd be concerned if I'm setting the wrong name here while all the tests are passing...

jim-minter: @ehashman you're right

pkg/api/2018-09-30-preview/default.go
```diff
@@ -0,0 +1,13 @@
...
+
+	if len(oc.Properties.RouterProfiles) == 0 {
+		oc.Properties.RouterProfiles = []RouterProfile{
+			{
+				Name: &defaultRouterName,
```
jim-minter: in our codebase prefer using to.StringPtr() to skip the extra variable definition

jim-minter:
good start, cross-check against https://github.com/openshift/openshift-azure/issues/336 and https://github.com/openshift/openshift-azure/blob/master/test/manifests/fakerp/create.yaml for other obvious defaultables.  I think the following are certainly plausible:

masterPoolProfile/count=3
agentPoolProfiles[name=infra]/count=3
agentPoolProfiles[*]/osType=Linux

+ unit test
+ simplify /test/manifests/fakerp/create.yaml

pkg/api/2018-09-30-preview/default.go
kwoodson: I thought this was referring to the actual router name instead of the routerprofile name.

ehashman:
@jim-minter: added your suggestions to a TODO list in the first comment.

In 5908786f2a20cf3195035efed508e22210c0601d I added unit tests and that's taken the PR size from medium to large; perhaps we should split the remaining items into follow-up PRs? Namely, one for more defaults for the master/agentpool profiles, and another for simplifying the test manifest. (If you agree, feel free to remove the WIP label.)

ehashman:
Will split the other TODOs into two follow-up PRs.

/retest

Issue #1481, opened 2019-04-09
createStatefulApp templateinstance failed


jim-minter:
```
• Failure [99.752 seconds]
Openshift on Azure end user e2e tests [EndUser][Fake]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:12
  should create and validate test apps [EndUser][Fake][Apps] [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:13

  Expected
      <[]*standard.TestError | len:1, cap:1>: [
          {
              Bucket: "createStatefulApp",
              Err: {
                  s: "templateinstance e2e-test-qqp1q/e2e-test-qqp1q failed",
              },
          },
      ]
  to be empty

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:17
```

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1479/pull-ci-openshift-openshift-azure-master-e2e-azure/3481/build-log.txt

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1479/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/425/build-log.txt

asalkeld:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1479/pull-ci-openshift-openshift-azure-master-e2e-azure/3481/artifacts/internal-e2e/events

```
e2e-test-qqp1q                      9m          9m           1         django-psql-persistent-1.1593e73ebdcf2f9f                      Build                                                                  Normal    BuildFailed               build-controller                           Build e2e-test-qqp1q/django-psql-persistent-1 failed
```

jim-minter:
@asalkeld please can you work this issue

asalkeld:
They both have a BuildFailed event but there seems to be 2 different reasons for it.

1. e2e test
```
E0409 20:07:41.152242       1 buildconfig_controller.go:141] gave up on Build for BuildConfig e2e-test-qqp1q/django-psql-persistent (0) due to fatal error: the LastVersion(1) on build config e2e-test-qqp1q/django-psql-persistent does not match the build request LastVersion(0)
[...]

PR #1552, opened 2019-04-23
get all controller logs not just the leader

asalkeld:
This is because openshift uses a different leader and we can end up missing
the logs for the openshift controller.

```release-note
NONE
```


asalkeld:
flake #1481

mjudeikis:
@asalkeld you beat me to this :D

PR #1554, opened 2019-04-23
add secrets to vmimage

mjudeikis:
```release-note
NONE
```

solves:
```
./hack/vmimage.sh
time="2019-04-23T06:44:57Z" level=info msg="vmimage starting, git commit v4.0-246-g3027211-clean"
time="2019-04-23T06:44:57Z" level=fatal msg="open secrets/client-key.pem: no such file or directory"
```

### DIFF
--- a/hack/vmimage.sh
+++ b/hack/vmimage.sh
@@ -17,6 +17,8 @@ cleanup() {
 }
 trap cleanup EXIT
 
+make secrets
+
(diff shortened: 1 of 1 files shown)


Makdaam:
/lgtm
since it can't be put as a dependency in Makefile

jim-minter:
I think this is the wrong solution - the script should be sourcing ci-operator-prepare.sh instead.

PR #1555, opened 2019-04-23
Update secrets in-place within make secrets

charlesakalugwu:
make secrets currently deletes the secrets directory, recreates it and then pulls the secrets from CI.
This PR changes that a bit to:
- do not delete an existing secrets directory because the developer might have other secret files in there
- use mkdir -p to create the directory if it does not exist
- use the --confirm=true flag to oc extract to overwrite existing files
- add a comment above the secrets target stating that logging into the CI cluster is a prerequisite for the secrets target

```release-note
NONE
```

fyi @mjudeikis @jim-minter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,10 @@ AZURE_IMAGE ?= quay.io/openshift-on-azure/azure:$(GITCOMMIT)
 
 all: azure
 
+# you must login to the ci cluster before running the secrets target
 secrets:
-	@rm -rf secrets
-	@mkdir secrets
-	@oc extract -n azure secret/cluster-secrets-azure --to=secrets >/dev/null
+	@mkdir -p secrets
+	@oc extract -n azure secret/cluster-secrets-azure --confirm=true --to=secrets >/dev/null
 
 clean:
 	rm -f coverage.out azure releasenotes


Makefile
```diff
@@ -7,10 +7,10 @@ AZURE_IMAGE ?= quay.io/openshift-on-azure/azure:$(GITCOMMIT)
 
 all: azure
 
+# you must login to the ci cluster before running the secrets target
```
ehashman: Is there a way that we can add a check to verify that the user is authenticated here rather than adding this as a comment?

jim-minter:
/hold

please don't merge this - we strongly want a 1:1 mapping between the secrets directory locally and CI.

Makefile
charlesakalugwu: It would even be nicer if the user actually got logged in automatically using a combination of secrets stored in the local `secrets` directory. 
For what its worth, we can use `oc whoami` to check for the currently logged in user or we might want to parse `$HOME/.kube/config` to see if the user is logged in but then we have to deal with possible invalidated sessions/tokens and it all ends up being too much work. 

I am open to taking out this comment if it makes you and possibly others happier with this change.

PR #1558, opened 2019-04-23
Adding session affinity to router service loadbalancer.

kwoodson:
```release-note
Update router service load balancer to use session affinity
```


pkg/sync/v5/data/Service/default/router.yaml
```diff
@@ -16,4 +16,5 @@ spec:
     port: 443
   selector:
     router: router
+  sessionAffinity: ClientIP
```
jim-minter: think you're going to need to add
```
  sessionAffinityConfig:
    clientIP:
      timeoutSeconds: 10800
```
here as well - hopefully as it stands this should fail the sync pod test

kwoodson: I read in the docs that the default for the timeoutSeconds is 10800 if `sessionAffinity: ClientIP` is set.  If you'd prefer I'll add this setting to be explicit.

ehashman: (Link to the docs which state these defaults for posterity's sake: https://v1-11.docs.kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies)

> In any of these proxy model, any traffic bound for the Service’s IP:Port is proxied to an appropriate backend without the clients knowing anything about Kubernetes or Services or Pods. Client-IP based session affinity can be selected by setting `service.spec.sessionAffinity` to “ClientIP” (the default is “None”), and you can set the max session sticky time by setting the field `service.spec.sessionAffinityConfig.clientIP.timeoutSeconds` if you have already set `service.spec.sessionAffinity` to “ClientIP” (the default is “10800”).

jim-minter:
/hold

this should fail the "should not continuously update objects" test, and it isn't.  This is because the test is incorrectly tagged :(

please can you fix the tag on the test (@Makdaam I thought you were sorting this out?), please fix the additional issue this PR will cause to the test (see my comment above) and any other breakage that has slipped in.

pkg/sync/v5/data/Service/default/router.yaml
kwoodson: @jim-minter, you were right. Ran a local test and the sync pod was adding these settings.   I have updated PR

kwoodson:
@jim-minter  @ehashman 

I think this is ready to go.  Please check.

Issue #1256, opened 2019-03-07
docker restart doesn't work


jim-minter:
running docker restart on a vm, at least a master VM, hangs - investigate why and fix

jim-minter:
docker-1.13.1-94.gitb2f74b2.el7.x86_64 looks healthier
@kwoodson please can you enable a geneva action run command to restart docker, add test coverage such that we test it on every commit, and we'll see if it holds up

kwoodson:
Docker restart was successful on docker version:
```
[root@master-000000 ~]# rpm -qa | grep docker
docker-client-1.13.1-94.gitb2f74b2.el7.x86_64
python-docker-2.4.2-2.el7.noarch
docker-rhel-push-plugin-1.13.1-94.gitb2f74b2.el7.x86_64
docker-common-1.13.1-94.gitb2f74b2.el7.x86_64
python-docker-pycreds-0.3.0-8.el7_6.noarch
docker-1.13.1-94.gitb2f74b2.el7.x86_64
atomic-openshift-docker-excluder-3.11.98-1.git.0.0cbaff3.el7.noarch
```

The restart pauses during execution to clean up the container mounts and filesystems.  Pod restarts occur and containers come back online.
```
[root@master-000000 ~]# docker ps | grep apiserver
05fd398a8b34        a5b32c3ff834                                                                                                                      "/usr/bin/service-..."   3 minutes ago       Up 3 minutes                            k8s_apiserver_apiserver-z6scl_kube-service-catalog_0aae011e-65f1-11e9-a2ca-000d3a1a4d8c_2
d2ff29cf20b5        registry.access.redhat.com/openshift3/ose-pod:v3.11.98                                                                            "/usr/bin/pod"           3 minutes ago       Up 3 minutes                            k8s_POD_apiserver-z6scl_kube-service-catalog_0aae011e-65f1-11e9-a2ca-000d3a1a4d8c_2
```

PR #1560, opened 2019-04-23
Adding geneva command to restart docker and tests

kwoodson:
```release-note
Adding run command to restart docker
```

Fixes: #1256

### DIFF
--- a/test/e2e/specs/fakerp/command.go
+++ b/test/e2e/specs/fakerp/command.go
@@ -30,12 +30,15 @@ var _ = Describe("Command tests [Command][Fake][LongRunning]", func() {
 		Expect(azurecli).NotTo(BeNil())
 	})
 
-	It("should be possible for an SRE to restart Kubelet and NetworkManager", func() {
+	It("should be possible for an SRE to restart system services on vms", func() {
 		vm := "master-000000"
 
 		startTime := time.Now()
 
 		err := azurecli.OpenShiftManagedClustersAdmin.RunCommand(context.Background(), os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), vm, "RestartKubelet")
+		Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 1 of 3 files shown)


kwoodson:
@charlesakalugwu, would like to walk through adding a geneva command if you wouldn't mind.

charlesakalugwu:
> @charlesakalugwu, would like to walk through adding a geneva command if you wouldn't mind.

@kwoodson i've created a meeting appointment so we can discuss what you need :)

test/e2e/specs/fakerp/command.go
```diff
@@ -38,6 +38,9 @@ var _ = Describe("Command tests [Command][Fake][LongRunning]", func() {
 		err := azurecli.OpenShiftManagedClustersAdmin.RunCommand(context.Background(), os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"), vm, "RestartKubelet")
```
charlesakalugwu: further down this file, you need to update the check for the success of the ConditionFunc in the wait.PollImmediate call (because we now need to check for the success of 3 commands and not 2) from:

```
return count == 2, nil
```

to 

```
return count == 3, nil
```

charlesakalugwu: you might want to rename the It block's description for the test from:

```
It("should be possible for an SRE to restart Kubelet and NetworkManager", func() {
```

to 

```
It("should be possible for an SRE to restart certain system services on VMs", func() {
```

kwoodson: Will update.

thekad:
This LGTM

> Just a few minor comments and you're good to go in terms of actual implementation. We might want to strengthen the e2e testing though

We already run `ValidateCluster()` after all 3 individual restarts succeed, what else do you have in mind?

mjudeikis:
/retest 
Quota issue
/lgtm

Issue #1543, opened 2019-04-18
ImageStream sync pod update changes


mjudeikis:
we might want to ignore imageStream generation on sync:
```
NFO[2019-04-18T10:54:21+03:00] Update ImageStream.image.openshift.io/openshift/jboss-eap71-openshift
INFO[2019-04-18T10:54:21+03:00] - Object.map[spec].map[tags].slice[0].map[generation]: 4 != 2
INFO[2019-04-18T10:54:21+03:00] - Object.map[spec].map[tags].slice[1].map[generation]: 4 != 2
INFO[2019-04-18T10:54:21+03:00] - Object.map[spec].map[tags].slice[2].map[generation]: 4 != 2
INFO[2019-04-18T10:54:21+03:00] - Object.map[spec].map[tags].slice[3].map[generation]: 4 != 2
```

its being generated always

Makdaam:
Updates caused also by default referencePolicy in image streams in openshift/library. Opened https://github.com/openshift/library/pull/172 to make it explicit.

Makdaam:
Fixed by updating upstream ImageStreams which caused issues.

PR #1566, opened 2019-04-24
Template updates to reduce sync updates.

kwoodson:
```release-note
NONE
```

Few updates to fix sync updates.


pkg/sync/v5/data/ImageStream.image.openshift.io/openshift/jboss-amq-62.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     importPolicy: {}
     name: "1.1"
     referencePolicy:
-      type: ""
+      type: "Source"
```
charlesakalugwu: everywhere 
```suggestion
      type: Source
```

pkg/sync/v5/data/StatefulSet.apps/openshift-azure-monitoring/canary.yaml
```diff
@@ -24,6 +24,7 @@ spec:
           valueFrom:
             fieldRef:
```
charlesakalugwu:
```
fieldRef:
    apiVersion: v1
    fieldPath: spec.nodeName
```

mjudeikis:
Might be my paranoid eyes, but was under impression that after this got merged:
https://github.com/openshift/openshift-azure/pull/1433

Change you are doing should not be allowed by the content test. I suspect this test is broken. Looking into it,  but for now
/hold
while Im checking

pkg/sync/v5/data/StatefulSet.apps/openshift-azure-monitoring/canary.yaml
kwoodson: Updated.

pkg/sync/v5/data/ImageStream.image.openshift.io/openshift/jboss-amq-62.yaml
kwoodson: Updated.

kwoodson:
@charlesakalugwu @Makdaam @mjudeikis  
Please look again.  We need this to go through for 2-3 more consecutive PRs.

kwoodson:
fixes: #1543

mjudeikis:
@kwoodson run this on your PR:
```
go test -v -tags=content ./...
```

I think the solution should be in https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v5/default.go#L42 function, and yaml file directly.

y-cote:
> @kwoodson run this on your PR:
> 
> ```
> go test -v -tags=content ./...
> ```
> I think the solution should be in https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v5/default.go#L42 function, and yaml file directly.

Strange, I ran it on the PR and I don't get failures. Unless I misinterpreted your comment @mjudeikis

mjudeikis:
okey,  looks like upstream is frozen. I think we need to make this "change" inside contentTest, raise an issue in upstream and follow-up when it is solved.

Makdaam:
Fixed default.go and clean.go in #1575 so when you get templates with `go test -v -tags=content ./...` it will remove the empty string value from the yaml file + will remove "Source" from the existing config before comparison.

jim-minter:
closing for #1575

PR #1568, opened 2019-04-24
Update ImageStreams and Templates 

mjudeikis:
```release-note
NONE
```
Update new templates and ImageStreams 

/cc @jim-minter


kwoodson:
@mjudeikis 
We need to update again since the recent upstream updates for the amq files to be fixed.

PR #1565, opened 2019-04-24
adding EndUser and ScaleUpDown to EveryPR

Makdaam:
- EndUser (+ approx 5min)
- ScaleUpDown - on same version (+ approx 13min), replaces https://github.com/openshift/release/pull/3575 in a way that avoids a new cluster spinup

```release-note
NONE
```

PR #1572, opened 2019-04-24
changing default e2e timeout to 60m

Makdaam:
```release-note
NONE
```
changing the default TIMEOUT for e2e tests instead of overriding it in the openshift/release repository in multiple places.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -11,7 +11,7 @@ if [[ -n "$FOCUS" ]]; then
 fi
 
 if [[ -z "$TIMEOUT" ]]; then
-    TIMEOUT=20m
+    TIMEOUT=60m
 fi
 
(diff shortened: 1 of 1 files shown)


ehashman:
If this is to avoid overriding this in multiple places, should we remove those overrides as part of this PR?

thekad:
So this timeout needs to account for: the sum of all spec timeouts plus the sum of individual context poll timeouts, otherwise we might get "context deadline exceeded" errors.
/lgtm
This is _probably_ right, and depending on how many timeout errors we get we can adjust again.

Makdaam:
@ehashman this is done instead of a change in the release repo, and is required by https://github.com/openshift/openshift-azure/pull/1565 which will increase the number of e2e tests (to approx 40min per run).

PR #1098, opened 2019-02-05
Fix makefile for debian-based systems

charlesakalugwu:
Before this fix, I saw the following error at the top of every make invocation

```
charles@titan:~/Documents/go/src/github.com/openshift/openshift-azure$ make generate
CLUSTER_VERSION is v0.0
/bin/sh: 1: [[: not found
TAG set to v0.0
go generate ./...
bindata.go
```

/cc @mjudeikis @kwoodson @Makdaam

PR #1573, opened 2019-04-24
Set Makefile SHELL to bash

ehashman:
Our Makefile uses the bash-specific [[ test construct, but GNU make defaults to using `/bin/sh` if a shell is not specified.[*] Hence, let's specify it.

[*]: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html#index-shell_002c-choosing-the

This is a permanent fix for the problem previous patched in #1513 (and I see also #1098). It got reverted, I think by accident, in #1507.

Without this fix, I continue to get `/bin/sh: 1: [[: not found` errors.

```release-note
NONE
```


mjudeikis:
/retest
https://github.com/openshift/openshift-azure/issues/1400

mjudeikis:
We just dont like `sh` :) #BashForever! :D

PR #1575, opened 2019-04-25
added EveryPR tag to sync test, fixed sync pod to pass the test

Makdaam:
fixed default removal actions that removed updateStrategy from StatefulApp (type mismatch string vs int64)
added default removal action on referencePolicy
added clean action to remove referencePolicy type when it's empty

```release-note
NONE
```


Makdaam:
Fixes #1543

Makdaam:
/hold
Holding for discussion on whether it should assume Source as default or override the yaml to "Local" during ```go test -v -tags=content ./...```

pkg/sync/v5/clean.go
```diff
@@ -99,6 +99,7 @@ func clean(o unstructured.Unstructured) error {
 	case "ImageStream.image.openshift.io":
 		jsonpath.MustCompile("$.metadata.annotations.'openshift.io/image.dockerRepositoryCheck'").Delete(o.Object)
 		jsonpath.MustCompile("$.spec.tags[*].generation").Delete(o.Object)
+		jsonpath.MustCompile("$.spec.tags.*.referencePolicy.type").DeleteIfMatch(o.Object, "")
```
jim-minter: doesn't make sense: clean() is really about removing runtime state from a live server object in order to be able to compare goal states

pkg/sync/v5/default.go
```diff
@@ -108,5 +108,8 @@ func defaults(o unstructured.Unstructured) {
 	case "StorageClass.storage.k8s.io":
 		jsonpath.MustCompile("$.reclaimPolicy").DeleteIfMatch(o.Object, "Delete")
 		jsonpath.MustCompile("$.volumeBindingMode").DeleteIfMatch(o.Object, "Immediate")
+
+	case "ImageStream.image.openshift.io":
+		jsonpath.MustCompile("$.spec.tags.*.referencePolicy.type").DeleteIfMatch(o.Object, "Source")
```
jim-minter: add something like:
```golang
jsonpath.MustCompile("$.spec.tags.*.referencePolicy.type").DeleteIfMatch(o.Object, "")
```

may also need something like the below, not sure:
```golang
jsonpath.MustCompile("$.spec.tags.*.referencePolicy").DeleteIfMatch(o.Object, map[string]interface{}{})
```

(default is mainly about removing excess baggage from our yaml files on disk)

jim-minter: will also need to call default() from content_test.go

pkg/sync/v5/clean.go
jim-minter: as discussed, please add a line or two of clarifying comment about purpose in clean.go and default.go

kwoodson:
Just FYI, upstream has PR open to fix these in the original templates:

https://github.com/openshift/library/pull/164

mjudeikis:
Where we are with this PR?

Makdaam:
Moved to new PR #1675 without AMQ related changes.
Closing, since v5 is frozen already.

PR #1576, opened 2019-04-25
adding EndUser and ScaleUpDown to EveryPR

Makdaam:
```release-note
NONE
```
Same as #1565 Reopening as a new PR due to github caching old tests as "required".


jim-minter:
/lgtm

@Makdaam please can you also file a PR to get rid of as many of these tags as possible which are no longer being used?

Makdaam:
@jim-minter I will after cleaning up release config, since the tags are used there for single shot /test runs.

Issue #1410, opened 2019-04-03
scale up times out


jim-minter:
```
time="2019-04-02T20:55:22Z" level=info msg="scaling ss-infra-1554236227 capacity up from 2 to 3" func="github.com/openshift/openshift-azure/pkg/cluster/scaler.(*workerScaler).scaleUp" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/scaler/scaler.go:107"
time="2019-04-02T21:10:23Z" level=debug msg="400 Bad Request: Failed to apply request: UpdateWorkerAgentPoolUpdateScaleSet: Future#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).badRequest" file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:11"
```

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1397/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/224/build-log.txt

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1407/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/227/build-log.txt

jim-minter:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1408/pull-ci-openshift-openshift-azure-release-v4-upgrade-v3.2/11

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1430/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/334/build-log.txt

jim-minter:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1455/pull-ci-openshift-openshift-azure-master-upgrade-v4.0/172/build-log.txt

asalkeld:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1455/pull-ci-openshift-openshift-azure-master-upgrade-v4.2/31/build-log.txt

jim-minter:
[...]

PR #1578, opened 2019-04-25
Automatic content regeneration script

mjudeikis:
```release-note
NONE
```

This should run as periodic:
```
./hack/verify/periodic-validate-content.sh
```
This will use our github token, check if we need to update content.
If this is a case, it will create PR. 
If PR exists, will add one more commit to the same PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,2 +1,3 @@
 ```release-note
+NONE
 ```
--- a/hack/verify/periodic-validate-content.sh
+++ b/hack/verify/periodic-validate-content.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -e
+
+make content
+
+if [[ -n "$(git status --porcelain)" ]]; then
+	echo "content update produced template and image-stream changes that were not already present"
+	
+	. hack/tests/ci-operator-prepare.sh
+	# HACK: using my repo for now to test 
+	go run hack/giter/giter.go -sourcerepo mjudeikis/openshift-azure -targetrepo openshift/openshift-azure
+else
+	echo "Dependencies have no material difference than what is committed."
+fi
+
+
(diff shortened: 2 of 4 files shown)


ehashman:
/retest

flake #1410 on upgrade-3.2

Other two flakes:

```
# pkg-config --cflags  -- libgit2
Package libgit2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libgit2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libgit2' found
pkg-config: exit status 1
make: *** [verify] Error 2
```

Possibly just an issue with CI...

ehashman:
Oh, duh. This PR adds a dependency on libgit2 which I guess isn't installed in the CI environment. Can we make that update as part of the PR?

.github/pull_request_template.md
```diff
@@ -1,2 +1,3 @@
 ```release-note
+NONE
```
jim-minter: I didn't do this before because I was worried that people wouldn't think "should I write a release note?" any more

hack/verify/periodic-validate-content.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
jim-minter: this is not a verify script because it makes changes to $PWD

mjudeikis:
> fun hack, but I think it's too much more complicated to understand than the equivalent shell script.
> 
> I think this should maintain the invariants that the content.update branch == master + additional fixup commit if applicable; if content.update != master && a PR isn't already open, open one.

Which way are you implying? Leave go-based and improve? Park go, and move to shell?

mjudeikis:
Alternative way: https://github.com/openshift/openshift-azure/pull/1584

jim-minter:
closing for #1584

PR #1579, opened 2019-04-25
Default MasterProfilePool count to 3

ehashman:
```release-note
Default MasterProfilePool count to 3 when not specified in new cluster configs
```

Addresses #336.

Note that the changes to each API version are identical.

### DIFF
--- a/pkg/api/2018-09-30-preview/default_test.go
+++ b/pkg/api/2018-09-30-preview/default_test.go
@@ -8,48 +8,74 @@ import (
 	"github.com/go-test/deep"
 )
 
-var sampleManagedCluster = &OpenShiftManagedCluster{
-	Properties: &Properties{
-		RouterProfiles: []RouterProfile{
-			{
-				Name:            to.StringPtr("Properties.RouterProfiles[0].Name"),
-				PublicSubdomain: to.StringPtr("NewPublicSubdomain"),
+func sampleManagedCluster() *OpenShiftManagedCluster {
+	return &OpenShiftManagedCluster{
+		Properties: &Properties{
+			RouterProfiles: []RouterProfile{
+				{
+					Name:            to.StringPtr("Properties.RouterProfiles[0].Name"),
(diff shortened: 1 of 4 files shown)


pkg/api/2018-09-30-preview/default_test.go
```diff
@@ -8,48 +8,74 @@ import (
...
+	}
 }
 
 func TestDefaults(t *testing.T) {
 	tests := []struct {
-		name     string
-		input    *OpenShiftManagedCluster
-		expected *OpenShiftManagedCluster
+		name           string
+		changeInput    func(*OpenShiftManagedCluster)
```
jim-minter: is this approach scalable as more defaulting is added? can you not check what default returns on a nil OpenShiftManagedCluster input (should default all the things) and what it returns on a fully populated OpenShiftManagedCluster input (should make no changes)?

ehashman: I considered that, but decided to avoid trying to capture all the defaulting in a single overarching test, leaning towards targeted individual cases for each behaviour we're adding.

The main reason why I chose that approach is that a lot of configs won't default if nothing at all is specified. For example, we don't have a default subnetCIDR or VMSize for masters, so the only way we can test that the count defaults to 3 is by passing the rest of the MasterProfilePool. Of course, maybe that suggests the approach here is wrong, and we should be adding more defaults... thoughts?

jim-minter: I didn't spot the discrepancy: if Properties is nil, you create it, but if Properties.MasterProfile is nil, you don't.  Were you to create them across the board, the test would be more straightforward, and I don't think the non-test effects are bad.

ehashman: I guess in the case that Properties.MasterProfilePool is nil, if we only populate the count of 3, it would fail validation so I wouldn't need to worry about the rest of the properties. I'd just wonder what the value of doing the defaulting in that case would be.

ehashman:
flake #1582

PR #1588, opened 2019-04-26
Add some default settings for AgentPoolProfiles

ehashman:
```release-note
Add a default AgentPoolProfile with name and role "infra" when not specified in new cluster configs
Set default Count to 3 for the "infra" AgentPoolProfile
Set default OSType to Linux in AgentPoolProfiles
```

Addresses #336.

### DIFF
--- a/pkg/api/2018-09-30-preview/default.go
+++ b/pkg/api/2018-09-30-preview/default.go
@@ -20,14 +20,19 @@ func setDefaults(oc *OpenShiftManagedCluster) {
 	if len(oc.Properties.AgentPoolProfiles) == 0 {
 		oc.Properties.AgentPoolProfiles = []AgentPoolProfile{
 			{
-				Name:  to.StringPtr("infra"),
-				Count: to.Int64Ptr(3),
-				Role:  (*AgentPoolProfileRole)(to.StringPtr("infra")),
+				Name: to.StringPtr("infra"),
+				Role: (*AgentPoolProfileRole)(to.StringPtr("infra")),
 			},
 		}
 	}
 
 	for i := range oc.Properties.AgentPoolProfiles {
+		if oc.Properties.AgentPoolProfiles[i].Name != nil &&
+			*oc.Properties.AgentPoolProfiles[i].Name == "infra" &&
+			oc.Properties.AgentPoolProfiles[i].Count == nil {
+			oc.Properties.AgentPoolProfiles[i].Count = to.Int64Ptr(3)
+		}
+
 		if oc.Properties.AgentPoolProfiles[i].OSType == nil {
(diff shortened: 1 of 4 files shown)


pkg/api/2018-09-30-preview/default.go
```diff
@@ -17,6 +17,22 @@ func setDefaults(oc *OpenShiftManagedCluster) {
 		oc.Properties.MasterPoolProfile.Count = to.Int64Ptr(3)
 	}
 
+	if len(oc.Properties.AgentPoolProfiles) == 0 {
+		oc.Properties.AgentPoolProfiles = []AgentPoolProfile{
+			{
+				Name:  to.StringPtr("infra"),
+				Count: to.Int64Ptr(3),
```
ehashman: Staring at this, I wonder if we should default this count to 3 in all cases (and not just when no AgentPoolProfiles were specified)?

jim-minter: Yes - the logic's not quite right.  It should ensure an infra APP exists, and set its count to 3 if it is nil.

ehashman:
/retest

flake upgrade-3.2 #552 
Not sure if we have the e2e-create one recorded somewhere

jim-minter:
worth seeing if the other approach we discussed on slack is better?

asalkeld:
> worth seeing if the other approach we discussed on slack is better?

should we wait for that? this seems ok to me.

pkg/api/2018-09-30-preview/default.go
```diff
@@ -17,6 +17,27 @@ func setDefaults(oc *OpenShiftManagedCluster) {
...
+	if len(oc.Properties.AgentPoolProfiles) == 0 {
+		oc.Properties.AgentPoolProfiles = []AgentPoolProfile{
+			{
+				Name: to.StringPtr("infra"),
+				Role: (*AgentPoolProfileRole)(to.StringPtr("infra")),
+			},
+		}
+	}
+
+	for i := range oc.Properties.AgentPoolProfiles {
```
mjudeikis: nit: I think we need to start putting some small comments on these cases :) Note for myself more than a comment to change :)

ehashman:
> worth seeing if the other approach we discussed on slack is better?

Following up on this to provide context: @jim-minter and I discussed the possibility of, in the create case on the converter, merging the provided new config against a default config (which would make this defaulting code redundant, and then I could remove it). I looked into this yesterday, checking if there will be any subtle behavioural differences that will result in unexpected behaviour.

I now think the merge with a default config approach won't actually work, because e.g. we only want to set a default RouterProfile if one is not defined; if we try to merge the default RouterProfile into a user-specified one, we'll get multiple (unless we update the existing merging code to handle the default cases, but then the logic between defaulting and merging gets tangled, so I'd rather not). Hence I think we should proceed with the approach in this PR.

PR #1589, opened 2019-04-29
Add more integration tests

asalkeld:
```release-note
NONE
```
Start adding more integration tests to get confidence in what causes node rotations.
/cc @jim-minter 

It helps to review commit by commit.

### DIFF
--- a/pkg/plugin/integration_test.go
+++ b/pkg/plugin/integration_test.go
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes/fake"
 	restclient "k8s.io/client-go/rest"
 	k8stesting "k8s.io/client-go/testing"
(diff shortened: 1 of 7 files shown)


pkg/plugin/integration_test.go
```diff
@@ -351,27 +315,256 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+
+			// read the clean update blob
+			bsc := az.StorageClient.GetBlobService()
+			updateBlobService := updateblob.NewBlobService(bsc)
+			beforeBlob, err := updateBlobService.Read()
+			// get sync deployment checksum annotation
+			// FIXME: only doing it this way as fake kube Get on the deployment
+			// always returns an empty string on the annotation
+			syncer, err := sync.New(log, cs, false)
+			if err != nil {
```
asalkeld: @jim-minter any better ideas on how to prove that we got (or didn't get) a sync pod rotation?

asalkeld:
flake https://github.com/openshift/openshift-azure/issues/552
/retest

asalkeld:
flake #1592 
/retest

pkg/plugin/integration_test.go
```diff
@@ -351,27 +315,256 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+					return perr
+				}
+				return nil
+			},
+		},
+		{
+			name:                  "rotate cluster secrets",
+			expectMasterRotation:  true,
+			expectSyncRotation:    true,
+			expectComputeRotation: false,
```
charlesakalugwu: I would expect that rotating cluster secrets would also cause compute vm rotations

asalkeld: I'll double check the way i am getting the update blobs, but might be good to look at the e2e test for cluster secrets. what's the command to run it again?

pkg/plugin/integration_test.go
```diff
@@ -115,8 +119,8 @@ func enrich(cs *api.OpenShiftManagedCluster) error {
 	cs.Properties.APICertProfile.KeyVaultSecretURL = vaultURL + "/secrets/" + vaultKeyNamePublicHostname
 	cs.Properties.RouterProfiles[0].RouterCertProfile.KeyVaultSecretURL = vaultURL + "/secrets/" + vaultKeyNameRouter
 
-	cs.Properties.PublicHostname = "openshift." + os.Getenv("RESOURCEGROUP") + "." + os.Getenv("DNS_DOMAIN")
-	cs.Properties.RouterProfiles[0].PublicSubdomain = "apps." + os.Getenv("RESOURCEGROUP") + "." + os.Getenv("DNS_DOMAIN")
+	cs.Properties.PublicHostname = "openshift." + rg + ".cloudapp.azure.com"
+	cs.Properties.RouterProfiles[0].PublicSubdomain = "apps." + rg + ".cloudapp.azure.com"
```
ehashman: If we're factoring out `rg` should we do the same for the DNS domain?

pkg/plugin/integration_test.go
```diff
@@ -351,27 +316,377 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+			change:         func(cs *api.OpenShiftManagedCluster) {},
+		},
+		{
+			name:           "change vm image",
+			expectRotation: map[rotationType]bool{rotationMaster: true, rotationInfra: true, rotationSync: true, rotationCompute: true},
+			change:         func(cs *api.OpenShiftManagedCluster) { cs.Config.ImageVersion = "311.12.12345678" },
+		},
+		{
+			name:           "change infra vm size",
+			expectRotation: map[rotationType]bool{rotationMaster: true, rotationInfra: true, rotationSync: false, rotationCompute: false},
```
ehashman: Why does the master rotate in this case?

pkg/plugin/integration_test.go
```diff
@@ -351,27 +316,377 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+			change: func(cs *api.OpenShiftManagedCluster) {
+				for i := range cs.Properties.AgentPoolProfiles {
+					if cs.Properties.AgentPoolProfiles[i].Role != api.AgentPoolProfileRoleCompute {
+						cs.Properties.AgentPoolProfiles[i].VMSize = "Standard_D8s_v3"
+					}
+				}
+			},
+		},
+		{
+			// Note: master is rotating here, is this expected?
```
ehashman: Asking the same questions I am :smile:

pkg/plugin/integration_test.go
```diff
@@ -351,27 +316,377 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+				t.Fatalf("rotation mismatch: expected %v, got %v", tt.expectRotation, rotations)
+			}
+		})
+	}
+}
+
+func TestHowActionsCauseRotations(t *testing.T) {
+	tests := []struct {
+		name           string
+		change         func(ctx context.Context, log *logrus.Entry, p *plugin, cs *api.OpenShiftManagedCluster, az *fakecloud.AzureCloud) error
```
ehashman: I wish there was some way we could avoid this long collection of posargs on all these lambdas, it's a bit hard to read :(

pkg/plugin/integration_test.go
```diff
@@ -351,27 +316,377 @@ func TestCreateThenUpdateCausesNoRotations(t *testing.T) {
...
+			change: func(ctx context.Context, log *logrus.Entry, p *plugin, cs *api.OpenShiftManagedCluster, az *fakecloud.AzureCloud) error {
+				perr := p.RotateClusterSecrets(ctx, cs, getFakeDeployer(log, cs, az))
+				if perr != nil {
+					return perr
+				}
+				return nil
+			},
+		},
+		{
+			name:           "readonly actions cause no rotations",
```
ehashman: It might be worth splitting this into separate unit tests, since if one of these actions fails for some reason, we can't actually tell which one very easily.

asalkeld: I could, just trying to reduce the test time. each test takes 3-4 seconds.

pkg/plugin/integration_test.go
asalkeld: Yeah, it's mostly the CreateOrUpdate() that takes a heap of args that are not defined at that time of defining the function here.. I could pull "log" out and create one above..

pkg/plugin/integration_test.go
asalkeld: Yeah, at this point I am characterizing the system. But if these look wrong I should probably raise issues for each.

pkg/plugin/integration_test.go
asalkeld: https://github.com/openshift/openshift-azure/issues/1606

pkg/plugin/integration_test.go
asalkeld: https://github.com/openshift/openshift-azure/issues/1606

asalkeld:
Thanks for the review @ehashman , improvements on the way..

asalkeld:
/hold
i'll have to scrap all the config change tests as none are allowed

asalkeld:
flake #1410 
/retest

pkg/plugin/integration_test.go
ehashman: That's a bummer, but makes sense.

ehashman:
/retest

Branch v5 was missing, looks like it's been added since then.

pkg/plugin/integration_test.go
charlesakalugwu: @asalkeld you can run the cluster secrets rotation with 
```
FOCUS="\[KeyRotation\]\[Fake\]" ./hack/e2e.sh
```

Issue #1577, opened 2019-04-25
Add environment variable option for HTTP debugging


ehashman:
Right now we can enable HTTP debugging by uncommenting this line: https://github.com/openshift/openshift-azure/blob/e9068cf35fa54cf21695956fab8819bd1d756526/pkg/util/azureclient/azureclient.go#L76

It would be much nicer to not have to make code changes and recompile every time we want to dump HTTP calls; perhaps we can set this via an environment variable.

Issue #1571, opened 2019-04-24
Using hack/create.sh and resource group with uppercase returns azure API error


e-tienne:
**Summary**

When using `hack/create.sh` with an uppercase resource group, it will create the resource group and DNS zone in azure (in lowercase), but will run into an error from the azure API and the shell script gets terminated.

**Expected behavior**

We should be able to use uppercase in resource groups / DNS. If it is not permitted, we should handle the error gracefully with a clear error message.

**Observed behavior**

The `hack/create.sh` scripts returns an error regarding the FQDN and the process gets terminated. The resource group and DNS zone are still created with **lowercase** in azure.

_Example output from `hack/create.sh`_
~~~

[esimard@dhcp129-198 openshift-azure]$ ./hack/create.sh $RESOURCEGROUP

...

+ export RESOURCEGROUP=esimard20190424UPPERCASE
+ RESOURCEGROUP=esimard20190424UPPERCASE
+ rm -rf _data
+ mkdir -p _data/_out
+ set -x
+ [[ -n '' ]]
+ go generate ./...
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
bindata.go
+ [[ -n '' ]]
+ go run cmd/fakerp/main.go
+ trap 'return_id=$?; set +ex; kill $(lsof -t -i :8080); wait $(lsof -t -i :8080); exit $return_id' EXIT
+ go run cmd/createorupdate/createorupdate.go
INFO[2019-04-24T11:08:59-04:00] using region northeurope                     
INFO[2019-04-24T11:08:59-04:00] ensuring resource group esimard20190424UPPERCASE 
INFO[2019-04-24T11:09:00-04:00]/home/esimard/go/src/github.com/openshift/openshift-azure/cmd/fakerp/main.go:20 main.main() starting the fake resource provider          
INFO[2019-04-24T11:09:00-04:00]/home/esimard/go/src/github.com/openshift/openshift-azure/pkg/plugin/plugin.go:86 github.com/openshift/openshift-azure/pkg/plugin.(*plugin).ValidatePluginTemplate() validating external plugin api data models   
INFO[2019-04-24T11:09:00-04:00]/home/esimard/go/src/github.com/openshift/openshift-azure/pkg/fakerp/server.go:82 github.com/openshift/openshift-azure/pkg/fakerp.(*Server).Run() starting server on localhost:8080            
[...]

PR #1590, opened 2019-04-29
Add debug env variable 

mjudeikis:
```release-note
NONE
```
Fix: https://github.com/openshift/openshift-azure/issues/1577

Not sure if code contamination is worth the benefit it gives :/

### DIFF
--- a/pkg/util/azureclient/azureclient.go
+++ b/pkg/util/azureclient/azureclient.go
@@ -70,7 +70,9 @@ func SetupClient(ctx context.Context, log *logrus.Entry, clientName string, clie
 	client.RequestInspector = addAcceptLanguages(languages)
 	client.PollingDelay = 10 * time.Second
 	client.Sender = &retrySender{Sender: client.Sender, log: log, clientName: clientName}
-	//client.Sender = &loggingSender{client.Sender}
+	if os.Getenv("AZURE_DEBUG") != "" {
+		client.Sender = &loggingSender{client.Sender}
+	}
 }
 
 func NewAuthorizer(clientID, clientSecret, tenantID, resource string) (autorest.Authorizer, error) {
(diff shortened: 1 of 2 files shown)


pkg/util/azureclient/azureclient.go
```diff
@@ -70,7 +70,9 @@ func SetupClient(ctx context.Context, log *logrus.Entry, clientName string, clie
 	client.RequestInspector = addAcceptLanguages(languages)
 	client.PollingDelay = 10 * time.Second
 	client.Sender = &retrySender{Sender: client.Sender, log: log, clientName: clientName}
-	//client.Sender = &loggingSender{client.Sender}
+	if os.Getenv("AZURE_DEBUG") != "" {
+		client.Sender = &loggingSender{client.Sender}
```
charlesakalugwu: This will be extremely verbose.

ehashman: That's the idea :smile: I want to be able to see the full contents of requests/responses when we have some mystery behaviour like in #1571 where the Azure API was complaining about a request body that was impossible to inspect without making code changes.

jim-minter: Note: it will also log the authentication token...  Particularly given such, I think we'd have to find a way to pipe TestConfig all the way down to this level - I don't want us to risk us colliding with an environment variable set on the RP.

mjudeikis: We could make setupClient part of the struct method and this would allow easier piping.

jim-minter:
/hold
not mergeable as-is.

mjudeikis:
/close
I think we might want to do this first: https://github.com/openshift/openshift-azure/issues/1598 as path to implement this. Not sure if wiring testConfig here would do any good now.

PR #1596, opened 2019-04-30
add debugging on scale failure

jim-minter:
```release-note
NONE
```

### DIFF
--- a/pkg/cluster/scaler/scaler.go
+++ b/pkg/cluster/scaler/scaler.go
@@ -8,10 +8,12 @@ package scaler
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	azcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/api"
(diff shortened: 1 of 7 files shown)


asalkeld:
flake #552 
/retest

charlesakalugwu:
flake #552 
/retest

pkg/cluster/scaler/scaler.go
```diff
@@ -116,6 +120,24 @@ func (ws *workerScaler) scaleUp(ctx context.Context, count int64) *api.PluginErr
...
+		spew.Dump(w, err)
+		// dump the VMSS state
+		ss, err2 := ws.ssc.Get(ctx, ws.resourceGroup, *ws.ss.Name)
+		spew.Dump(w, ss, err2)
+		// try the scale again
+		err2 = ws.ssc.Update(ctx, ws.resourceGroup, *ws.ss.Name, azcompute.VirtualMachineScaleSetUpdate{
+			Sku: &azcompute.Sku{
+				Capacity: to.Int64Ptr(count),
+			},
+		})
```
ehashman: Is it possible for this to succeed so then we shouldn't fail below?

pkg/util/azureclient/compute/virtualmachinescalesets.go
```diff
@@ -30,7 +30,6 @@ var _ VirtualMachineScaleSetsClient = &virtualMachineScaleSetsClient{}
 func NewVirtualMachineScaleSetsClient(ctx context.Context, log *logrus.Entry, subscriptionID string, authorizer autorest.Authorizer) VirtualMachineScaleSetsClient {
 	client := compute.NewVirtualMachineScaleSetsClient(subscriptionID)
 	azureclient.SetupClient(ctx, log, "compute.VirtualMachineScaleSetsClient", &client.Client, authorizer)
-	client.PollingDuration = 30 * time.Minute
```
ehashman: Why is this duration being removed? Does it have a sensible default we're overriding?

pkg/cluster/scaler/scaler.go
jim-minter: yes, absolutely possible that it succeeds - for now I'd like to see the evidence of that, then will code up the retrier

pkg/util/azureclient/compute/virtualmachinescalesets.go
jim-minter: 15 minutes.  Previously I tried upping it to see if that would reduce our issues, but I don't think it did.

PR #1601, opened 2019-05-01
add generate to content update script

mjudeikis:
```release-note
NONE
```

Content PR's being created with "invalid" content due to missing generate step

### DIFF
--- a/hack/periodic/content-update.sh
+++ b/hack/periodic/content-update.sh
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
 make content
+go generate ./...
 
 . hack/tests/ci-operator-prepare.sh
 


hack/periodic/content-update.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-make content
+make content generate
```
ehashman: Suggestion: do we always want to generate when we `make content`? If so, let's make generate a prerequisite target for content in the Makefile rather than adding this here?

mjudeikis: I would say not always. If I ran this locally I might not want to do so. Buts it one of those :man_shrugging:

ehashman: I'd say if we usually want to `make generate` before `make content` and it's bad to forget, we should probably stick it in the makefile (we already have generate as a dependent target of a bunch of stuff in there), but if not then we can probably leave this.

Also, do we need to generate first?

jim-minter: We actually want to generate *after* make content, not before.
Note that I put a naughty hack into `make generate` which actually prevents it from running under CI (saves a surprising amount of time).  I may yet live to regret this.
So it would need to be a manual `go generate ./...`
We already run `make generate` on behalf of developers before they create/update clusters (this is as a courtesy to help prevent people building test clusters using stale assets).  On that basis I'm inclined to add a `go generate ./...` to this file as the fix.

Yeah, it's all a bit hacky.

Issue #1532, opened 2019-04-17
flake: osba timed out waiting for condition


charlesakalugwu:
Caught a flake with the following error signature:

```
time="2019-04-17T08:37:24Z" level=debug msg="validating that all osba components are healthy" func=github.com/openshift/openshift-azure/test/e2e/specs.validateOSBAApp file="/go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:21"
time="2019-04-17T08:57:24Z" level=error msg="timed out waiting for the condition" func=github.com/openshift/openshift-azure/test/e2e/specs.validateOSBAApp file="/go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:24"
STEP: deleting openshift broker app
time="2019-04-17T08:57:24Z" level=debug msg="deleting azure cluster service broker" func=github.com/openshift/openshift-azure/test/e2e/specs.deleteOSBAApp file="/go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:32"
time="2019-04-17T08:57:25Z" level=debug msg="deleting openshift project for broker apps" func=github.com/openshift/openshift-azure/test/e2e/specs.deleteOSBAApp file="/go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:39"

• Failure [1229.035 seconds]
Openshift on Azure end user e2e tests [CostumerAdmin][Fake]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:83
  should create and validate an OpenShift Broker App [CustomerAdmin][Fake][OSBA] [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:84

  Expected
      <[]error | len:1, cap:1>: [
          {
              s: "timed out waiting for the condition",
          },
      ]
  to be empty

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/osba_app.go:102
------------------------------
SSSSSSSSSS
```

artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1520/pull-ci-openshift-openshift-azure-master-e2e-azure/3658/build-log.txt

/kind test-flake

Issue #1438, opened 2019-04-05
timeout while checking that cluster services are available


charlesakalugwu:
Found a flake where a timeout occurs while checking that cluster services are available. This occured during the sanity checker's `checkCanAccessServices` method call during `ValidateCluster`

```
• Failure [178.329 seconds]
Openshift on Azure end user e2e tests [EndUser][Fake]
/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:12
  should validate the cluster [EndUser][Fake][Cluster] [It]
  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:28

  Expected
      <[]*standard.TestError | len:1, cap:1>: [
          {
              Bucket: "checkCanAccessServices",
              Err: {
                  Op: "Get",
                  URL: "https://openshift.e2e-create-charlesakalugwu-1417-awsptb.osadev.cloud/healthz",
                  Err: {
                      err: "net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)",
                      timeout: true,
                  },
              },
          },
      ]
  to be empty

  /go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:30
```
artifacts; https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1417/pull-ci-openshift-openshift-azure-master-e2e-azure/3354/build-log.txt

/kind test-flake

jim-minter:
should use prod roundtripper in test as well

jim-minter:
@mjudeikis please can you set the test openshift clients to use the same retrying roundtripper that prod uses; also to fix the specific issue you'll need to mess with the RT in checkCanAccessServices

charlesakalugwu:
#1642 

artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1642/pull-ci-azure-master-upgrade-v5.0/75/build-log.txt

charlesakalugwu:
#1642
artifacts: 
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1642/pull-ci-azure-master-upgrade-v4.3/66/build-log.txt

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1642/pull-ci-azure-master-e2e-create/192/build-log.txt

charlesakalugwu:
[...]

PR #1603, opened 2019-05-01
VM image content verify

mjudeikis:
```release-note
NONE
```

Script for validating vmimage content RPMS.
FLow:
1. Run validation script on the newest published VM image (based on tags from `images` RG). Follow-up: we need to start marking "live" images and test those. 
2. Get updatable package list and summary
3. Raise an issue
4. Run all this in periodic 

TODO:
* Upload results to private blob storage or appInsights

/cc @jim-minter @asalkeld @kwoodson  @charlesakalugwu


pkg/cluster/gomock_reflect_610001376/prog.go
```diff
@@ -0,0 +1,66 @@
+
```
charlesakalugwu: dont think this file should be in here :)

mjudeikis: yes :) will delete

pkg/vmimage/resources.go
```diff
@@ -138,19 +138,33 @@ func nic(subscriptionID, resourceGroup, location string) *network.Interface {
 	}
 }
 
-func vm(subscriptionID, resourceGroup, location string, sshPublicKey string) *compute.VirtualMachine {
+func vm(subscriptionID, resourceGroup, location string, sshPublicKey string, verify bool, image string) *compute.VirtualMachine {
```
charlesakalugwu: what about changing the signature to:

```suggestion
func vm(subscriptionID, resourceGroup, location, sshPublicKey, image string, verify bool) *compute.VirtualMachine {
```

pkg/vmimage/ssh.go
```diff
@@ -11,11 +11,76 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	io.Copy(os.Stdout, stdout)
+
+	return nil
+}
+
+func (builder *Builder) scp(files []string) error {
```
charlesakalugwu: can we share some of the logic between `*Builder.ssh()` and `*Builder.scp(files []string)`?

mjudeikis: We already doing session creation. The only other thing left is `stdoutPipe` but not sure if it worth start passing it around.

mjudeikis:
Example issue: https://github.com/openshift/openshift-azure/issues/1641

kwoodson:
@mjudeikis 

Currently there aren't any results from the `yum` command:
```
[root@vm ~]# cat /tmp/check 
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
[root@vm ~]# cat /tmp/info
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
updateinfo summary done
```

pkg/vmimage/ssh.go
```diff
@@ -11,11 +11,76 @@ import (
...
+			return err
+		}
+		defer f.Close()
+
+		err = session.Start(fmt.Sprintf("sudo cat %s", file))
+		if err != nil {
+			return err
+		}
+
+		io.Copy(f, stdoutPipe)
```
asalkeld: you don't want to check the error?

pkg/vmimage/ssh.go
```diff
@@ -11,11 +11,76 @@ import (
...
 	if err != nil {
 		return err
 	}
 
+	err = session.Start("sudo tail -F -n +1 /var/lib/waagent/custom-script/download/0/stdout")
+	if err != nil {
+		return err
+	}
+
+	io.Copy(os.Stdout, stdout)
```
asalkeld: error?

pkg/vmimage/ssh.go
asalkeld: or is the session going to give you *all* errors, probably not local file errors..

asalkeld:
@mjudeikis looks pretty good, upgrade failing

mjudeikis:
> @mjudeikis
> 
> Currently there aren't any results from the `yum` command:
> 
> ```
> [root@vm ~]# cat /tmp/check 
> Loaded plugins: product-id, search-disabled-repos, subscription-manager
> This system is not registered with an entitlement server. You can use subscription-manager to register.
> [root@vm ~]# cat /tmp/info
> Loaded plugins: product-id, search-disabled-repos, subscription-manager
> This system is not registered with an entitlement server. You can use subscription-manager to register.
> updateinfo summary done
> ```

Because it takes "newest" image, potentially built yesterday. So nothing to update. we need to point it to images, used in production somehow.

pkg/vmimage/ssh.go
mjudeikis: I was thinking the same. Not sure if my thinking is right

pkg/vmimage/ssh.go
mjudeikis: same as above?

ehashman:
/retest

upgrade-v5.1 flake #1438 
e2e-create flake #1532 (or at least is similar)

Issue #1336, opened 2019-03-24
elb timed out waiting for the condition


jim-minter:
```
time="2019-03-23T22:15:04Z" level=debug msg="validating that the cluster can create ELB and ILB" func="github.com/openshift/openshift-azure/test/e2e/standard.(*SanityChecker).ValidateCluster" file="/go/src/github.com/openshift/openshift-azure/test/e2e/standard/sanitychecker.go:128"
STEP: waiting for elb to be ready (2019-03-23 22:15:06.488264642 +0000 UTC m=+1185.959672199)
time="2019-03-23T22:20:12Z" level=error msg="timed out waiting for the condition" func="github.com/openshift/openshift-azure/test/e2e/standard.(*SanityChecker).ValidateCluster" 
```

in artifacts:
```
count: 1
eventTime: null
firstTimestamp: 2019-03-23T22:18:05Z
involvedObject:
  apiVersion: v1
  kind: Service
  name: elb
  namespace: e2e-test-xpw2a
  resourceVersion: "11225"
  uid: 1d1c2e33-4db9-11e9-a72e-000d3a180cc8
lastTimestamp: 2019-03-23T22:18:05Z
message: 'Error creating load balancer (will retry): failed to ensure load balancer
  for service e2e-test-xpw2a/elb: Long running operation terminated with status ''Failed'':
  Code="InvalidResourceReference" Message="Resource /subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/E2E-VQWMOVGJ/providers/Microsoft.Network/loadBalancers/KUBERNETES-INTERNAL
  referenced by resource /subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/e2e-vqwmovgj/providers/Microsoft.Compute/virtualMachineScaleSets/ss-infra-1553376927
  was not found. Please make sure that the referenced resource exists, and that both
  resources are in the same region."'
metadata:
  creationTimestamp: 2019-03-23T22:18:05Z
  name: elb.158eb669839147a7
  namespace: e2e-test-xpw2a
  resourceVersion: "11734"
  selfLink: /api/v1/namespaces/e2e-test-xpw2a/events/elb.158eb669839147a7
  uid: 8807be0b-4db9-11e9-850b-000d3a180933
reason: CreatingLoadBalancerFailed
reportingComponent: ""
reportingInstance: ""
source:
  component: service-controller
type: Warning
```

[...]

PR #1613, opened 2019-05-02
Unwrap "net/http: HTTP/1.x transport connection broken" error

### LINKED ISSUE #1400: write: connection reset by peer talking to storage account

asalkeld:
```release-note
NONE
```
closes #1400

### DIFF
--- a/pkg/util/errors/errors.go
+++ b/pkg/util/errors/errors.go
@@ -4,17 +4,46 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strings"
 	"syscall"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 )
 
+const (
+	netHTTPErrorPrefix = "net/http: HTTP/1.x transport connection broken"
+)
+
+var (
+	netHTTPErrorPrefixLen = len(netHTTPErrorPrefix)
+	str2Syscall           = map[string]syscall.Errno{
+		"connection reset by peer":  syscall.ECONNRESET,
+		"connection refused":        syscall.ECONNREFUSED,
+		"network is unreachable":    syscall.ENETUNREACH,
+		"no such file or directory": syscall.ENOENT,
+	}
+)
+
+func unwrapNetHTTPError(err error) (bool, error) {
+	if !strings.HasPrefix(err.Error(), netHTTPErrorPrefix) {
+		return false, err
+	}
+	for sysMsg, errno := range str2Syscall {
+		if strings.Contains(err.Error()[netHTTPErrorPrefixLen:], sysMsg) {
+			return true, errno
+		}
+	}
+	return false, err
+}
+
 // IsMatchingSyscallError returns true when the error is one of the Errno's in match
 // it deals with the different ways of wrapping the syscalls.
 func IsMatchingSyscallError(err error, match ...syscall.Errno) bool {
 	for {
 		switch t := err.(type) {
+		case nil:
+			return false
 		case *azure.RequestError:
 			err = &t.DetailedError
 		case *autorest.DetailedError:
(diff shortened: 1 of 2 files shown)


pkg/util/errors/errors.go
```diff
@@ -4,17 +4,44 @@ import (
...
+
+// some common network errors that we check for
+var str2Syscall = map[string]syscall.Errno{
+	"connection reset by peer":  syscall.ECONNRESET,
+	"connection refused":        syscall.ECONNREFUSED,
+	"network is unreachable":    syscall.ENETUNREACH,
+	"no such file or directory": syscall.ENOENT,
+}
+
+func unwrapNetHTTPError(err error) (bool, error) {
```
mjudeikis: Any reason why `IsNetHTTPConnectionBrokenWrapperError` and `unwrapNetHTTPError` should not be in the same function? Considering they are very interconnected in the logic below? Or make `IsNetHTTPConnectionBrokenWrapperError` unexposed? as alone is not much of the value?

asalkeld: I can simplify it, it was just how I started off ... 
I'll get rid of IsNetHTTPConnectionBrokenWrapperError() and just add the check for net/http in the unwrap function.

asalkeld:
flakes #1336

asalkeld:
flake #1616

pkg/util/errors/errors.go
```diff
@@ -4,17 +4,40 @@ import (
...
+	if !strings.HasPrefix(err.Error(), "net/http: HTTP/1.x transport connection broken:") {
```
charlesakalugwu: what do you think about extracting that string to a variable `foo` so we can use `len(foo)` below instead of 47?

pkg/util/errors/errors.go
```diff
@@ -4,17 +4,40 @@ import (
...
+
 // IsMatchingSyscallError returns true when the error is one of the Errno's in match
 // it deals with the different ways of wrapping the syscalls.
 func IsMatchingSyscallError(err error, match ...syscall.Errno) bool {
 	for {
 		switch t := err.(type) {
+		case nil:
+			return false
 		case *azure.RequestError:
 			err = &t.DetailedError
```
charlesakalugwu: I really think you can save one iteration of the for loop in the case where this function is called with a `*azure.RequestError` by doing:

```suggestion
			err = t.DetailedError.Original
```

thekad:
this LGTM

pkg/util/errors/errors.go
```diff
@@ -4,17 +4,40 @@ import (
...
+	"network is unreachable":    syscall.ENETUNREACH,
+	"no such file or directory": syscall.ENOENT,
+}
+
+func unwrapNetHTTPError(err error) (bool, error) {
+	if !strings.HasPrefix(err.Error(), "net/http: HTTP/1.x transport connection broken:") {
+		return false, err
+	}
+	for sysMsg, errno := range str2Syscall {
+		if strings.Contains(err.Error()[47:], sysMsg) {
```
y-cote: Seems like
```
if strings.Contains(err.Error(), sysMsg) {
```
would be enough here since we assume the string will be constantly starting the same (HasPrefix) and doesn't conflict with the strings we compare with?

asalkeld: absolutely, I just wanted to save some searching time. if that is not worth it, i can remove it. it is in the error path so performance not really a priority

y-cote: that's right, clearer is better than faster here, but it's optional.

y-cote: /lgtm

asalkeld: i have cleaned it up a bit, i hope it's better to read now..

PR #1619, opened 2019-05-03
adding metric which is exactly 1 for every monitored cluster

Makdaam:
```release-note
NONE
```
This metric will be used for percentage calculation in monitoring/alerting. ALERT DeadMansSwitch is filtered somewhere after the metricsbridge and can't be used directly.


pkg/sync/v6/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -5,6 +5,8 @@ data:
     statsdSocket: /var/etw/mdm_statsd.socket
     queries:
     - query: 'ALERTS'
+    - query: 'ALERTS{alertname="DeadMansSwitch",alertstate="firing"}'
+      name: osacluster_monitored
```
charlesakalugwu: already discussed with @Makdaam regarding the name `osacluster_monitored`. Not sure it sounds right. What do the others think?

jim-minter:
Closing for now - think we need to better understand our Geneva performance characteristics

Makdaam:
Closing, as there might be a better explanation of missing DeadMansSwitch or a better base metric altogether.

PR #1621, opened 2019-05-09
groundwork for removing ci-operator

thekad:
```release-note
NONE
```
Trying to get rid of ci-operator, this is groundwork for removing ci-operator in the release repo

### DIFF
--- a/hack/tests/ci-prepare.sh
+++ b/hack/tests/ci-prepare.sh
@@ -30,6 +30,31 @@ stop_monitoring() {
     fi
 }
 
+generate_artifacts() {
+  if [[ -n "$ARTIFACTS" ]]; then
+      exec &>"$ARTIFACTS/cleanup"
+  fi
+
+  stop_monitoring
+
+  make artifacts
+}
+
+delete() {
+  realrp="$1"
+
+  if [[ -n "$NO_DELETE" ]]; then
+      return
+  fi
+
+  # only delete for fakerp
+  if [[ -z "$realrp" ]]; then
+    make delete
+  fi
+
+  az group delete -g "$RESOURCEGROUP" --yes --no-wait
+}
+
 if [[ ! -e /var/run/secrets/kubernetes.io ]]; then
     return
 fi
--- a/hack/vmimage.sh
+++ b/hack/vmimage.sh
@@ -3,21 +3,14 @@
 cleanup() {
   set +e
 
-  if [[ -n "$ARTIFACTS" ]]; then
-    exec &>"$ARTIFACTS/cleanup"
-  fi
+  generate_artifacts
 
-  make artifacts
-
-  if [[ -n "$NO_DELETE" ]]; then
-    return
-  fi
-  make delete
-  az group delete -g "$RESOURCEGROUP" --yes --no-wait
+  delete
 }
+
 trap cleanup EXIT
 
-. hack/tests/ci-operator-prepare.sh
+. hack/tests/ci-prepare.sh
 
 TAG=$(git describe --tags HEAD)
 if [[ $(git status --porcelain) = "" ]]; then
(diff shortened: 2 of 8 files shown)


hack/tests/ci-prepare.sh
```diff
@@ -30,6 +30,31 @@ stop_monitoring() {
...
 }
 
+cleanup() {
+    set +e
+
+    if [[ -n "$ARTIFACTS" ]]; then
+        exec &>"$ARTIFACTS/cleanup"
+    fi
+
+    if [[ -n "$MONITORING" ]]; then
```
asalkeld: do you need this check as stop_monitoring only does something if MON_PID is not empty

asalkeld:
also s/Azure 186/groundwork for removing ci-operator in the PR headline so we know what it is about without having to go to the jira

hack/tests/ci-prepare.sh
```diff
@@ -30,6 +30,31 @@ stop_monitoring() {
...
+      stop_monitoring
+    fi
+
+    make artifacts
+
+    if [[ -n "$T" ]]; then
```
jim-minter: the reason we hadn't condensed all the cleanup()s into one so far is that they're subtly different.  This PR doesn't yet handle things neatly enough to make me want to condense them.  Example: I don't want to rely on a variable called $T in a library cleanup().  Possibly renaming $T and making a wrapper function that cleans up $T then calls cleanup()?  Or still trapping the `rm -rf "$T"` in the other script?  Needs more experimentation.

hack/tests/e2e-create-prod.sh
```diff
@@ -1,23 +1,8 @@
...
-
-    make artifacts
-
-    if [[ -n "$NO_DELETE" ]]; then
-        return
-    fi
-    make delete
-    az group delete -g "$RESOURCEGROUP" --yes --no-wait
-}
 trap cleanup EXIT
```
jim-minter: can the trap be moved to ci-prepare.sh?

hack/vmimage.sh
```diff
@@ -17,7 +17,7 @@ cleanup() {
 }
 trap cleanup EXIT
```
jim-minter: which cleanup actually gets called?  This one or the one in ci-prepare.sh?

hack/tests/e2e-create-prod.sh
thekad: that would mean that `cleanup()` gets called at the end of every script, which may not be something you want to do, I rather keep it on each script to maintain some control

hack/tests/ci-prepare.sh
thekad: Taking a look at all the cleanup functions they are mostly the same except for some of them having stop_monitoring, which as Angus points out is already idempotent: won't fail if monitoring wasn't started.
I do agree about T, I was just trying to maintain the status quo but that does look off to me, let me see how to best tackle that

hack/tests/e2e-create-prod.sh
```diff
@@ -1,23 +1,8 @@
 #!/bin/bash -e
 
-cleanup() {
```
charlesakalugwu: it doesn't make sense to use the same cleanup() for the production e2e tests for reasons explained in https://github.com/openshift/openshift-azure/pull/1625

charlesakalugwu:
flake #1628 
/retest

hack/tests/e2e-create-prod.sh
thekad: better this way?

mjudeikis:
so       much       space       here        :)
/lgtm

charlesakalugwu:
/hold

Please can you improve the commit messages and/or squash this?

mjudeikis:
/lgtm
/hold cancel
@charlesakalugwu thanks for this :) @thekad is out so doing this so we can clean PR queue :)

PR #1627, opened 2019-05-09
Limit length of ci-generated resource groups for prod e2e tests

### LINKED ISSUE #1626: Limit length of ci-generated resource groups for prod e2e tests
There are several resource groups created during the creation of an ARO cluster. The customer specifies a cluster resource group `foo` and this is created with a hidden application resource group which contains the ARO managed application object. This application resource group is of the format `OS_resourcegroup_clustername_region`.  This application resource groups contains a further managed resource group whose name looks like a UUID. It is in this managed resource group that the customer's cluster components are created.

We currently generate cluster resource groups for e2e tests against the either the real or fake rp in CI. These names in the tests against the real rp are then duplicated to generate the application resource group `OS_foo_foo_region`. 

We need to limit the lengths of resource groups for cases where the e2e tests are to be targeted at the real rp.

Note that the maximum length of the application resource groups is 64

artifacts: 
https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-e2e-prod/16/build-log.txt
https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-e2e-prod/15/build-log.txt
https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-e2e-prod-vnet/15/build-log.txt
https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-e2e-prod-vnet/16/build-log.txt

charlesakalugwu:
Currently the resource group in ci is generated with the following expression
```
export RESOURCEGROUP="ci-$pullnumber$(basename "$0" .sh)-$(cat /dev/urandom | tr -dc 'a-z' | fold -w 6 | head -n 1)"
```

This `$(basename "$0" .sh)` component of the expression effectively extracts the base name of the prod e2e script to be used as part of the resultant RESOURCEGROUP.

By renaming the prod e2e script from `e2e-create-prod.sh` to `e2e-prod.sh` we effectively shave off 14 characters from the application resource group (format: `OS_rgname_rgname_region`) for a cluster.

closes #1626

/cc @mjudeikis @ehashman  @Makdaam 

```release-note
NONE
```


charlesakalugwu:
accompanying release repo PR: https://github.com/openshift/release/pull/3741

charlesakalugwu:
> /lgtm
> 
> This seems like a quick and easy fix and so I'm okay with merging this as is, but why are we repeating the resource group name in the cluster name twice? Is it possible to make that listed just once instead?

@ehashman as explained in https://github.com/openshift/openshift-azure/issues/1626 this is the nature of how prod clusters are created. Customer specifies a desire to create a cluster called `hive` in the resource group `foo` in region `bar` and the request is sent to the RP for the cluster to be created. An `application resource group` is created by the RP with a name that looks like `OS_foo_hive_bar`.
In our CI, we use the resource group to represent the cluster name. So we end up with prod `application resource groups` looking like `OS_foo_foo_bar` in CI. This is where the doubling comes from

Issue #1607, opened 2019-05-02
allow changing some more properties with an update


asalkeld:
For now enable changing the following properties by the user:

- VMSize
- AADIdentityProvider

Detect the change and make the required changes
- MasterServicePrincipalProfile
- WorkerServicePrincipalProfile

asalkeld:
I have changes, but waiting for my  integration test PR to go in first

asalkeld:
This still needs to support the following:

Detect the change and make the required changes

    MasterServicePrincipalProfile
    WorkerServicePrincipalProfile

/remove-lifecycle stale

PR #1629, opened 2019-05-10
Make some api properties updatable

asalkeld:
```release-note
Properties.AuthProfile.IdentityProviders[x].Provider and Properties.AgentPoolProfiles[x].VMSize can now be changed by the user using CreateOrUpdate.
```
partial #1607

### DIFF
--- a/pkg/api/validate/api.go
+++ b/pkg/api/validate/api.go
@@ -69,6 +69,7 @@ func (v *APIValidator) validateUpdateContainerService(cs, oldCs *api.OpenShiftMa
 			}
 		}
 	}
+	old.Properties.AuthProfile.IdentityProviders = cs.Properties.AuthProfile.IdentityProviders
 
 	if !reflect.DeepEqual(cs, old) {
 		// TODO: this is a hack because we're using deep.Equal.  To fix properly
--- a/pkg/api/validate/api.go
+++ b/pkg/api/validate/api.go
@@ -53,6 +53,12 @@ func (v *APIValidator) validateUpdateContainerService(cs, oldCs *api.OpenShiftMa
 	old.Properties.ProvisioningState = cs.Properties.ProvisioningState
 
 	for i, app := range old.Properties.AgentPoolProfiles {
+		for _, newApp := range cs.Properties.AgentPoolProfiles {
+			if newApp.Name == app.Name {
+				old.Properties.AgentPoolProfiles[i].VMSize = newApp.VMSize
+			}
+		}
+
 		if app.Role != api.AgentPoolProfileRoleCompute {
 			continue
 		}
--- a/pkg/plugin/integration_test.go
+++ b/pkg/plugin/integration_test.go
@@ -510,6 +510,81 @@ func TestHowAdminConfigChangesCausesRotations(t *testing.T) {
 	}
 }
 
+func TestHowUserConfigChangesCausesRotations(t *testing.T) {
+	tests := []struct {
+		name           string
+		change         func(cs *api.OpenShiftManagedCluster)
+		expectRotation map[rotationType]bool
+	}{
+		{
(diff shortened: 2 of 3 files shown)


pkg/plugin/integration_test.go
```diff
@@ -510,6 +510,88 @@ func TestHowAdminConfigChangesCausesRotations(t *testing.T) {
...
+			change: func(cs *api.OpenShiftManagedCluster) {
+				for i := range cs.Properties.AgentPoolProfiles {
+					if cs.Properties.AgentPoolProfiles[i].Role != api.AgentPoolProfileRoleCompute {
+						cs.Properties.AgentPoolProfiles[i].VMSize = "Standard_D16s_v3"
+					}
+				}
+			},
+		},
+		{
+			// Note: master is rotating here, is this expected?
```
charlesakalugwu: I don't think this is expected. @jim-minter any idea whats going on here?

pkg/api/validate/api.go
```diff
@@ -63,16 +69,12 @@ func (v *APIValidator) validateUpdateContainerService(cs, oldCs *api.OpenShiftMa
...
 
 	if !reflect.DeepEqual(cs, old) {
 		// TODO: this is a hack because we're using deep.Equal.  To fix properly
 		// we'd probably need to implement our own deep.Equal.
 		csCopy := cs.DeepCopy()
-		if csCopy.Properties.AuthProfile.IdentityProviders[0].Provider.(*api.AADIdentityProvider).Secret !=
-			old.Properties.AuthProfile.IdentityProviders[0].Provider.(*api.AADIdentityProvider).Secret {
-			csCopy.Properties.AuthProfile.IdentityProviders[0].Provider.(*api.AADIdentityProvider).Secret = "<hidden 1>"
-			old.Properties.AuthProfile.IdentityProviders[0].Provider.(*api.AADIdentityProvider).Secret = "<hidden 2>"
-		}
```
ehashman: I think removing these lines just removes the secret redaction in logging, are we sure we want to do this?

mjudeikis: Same question. What is the reasoning to remove it?

pkg/api/validate/api.go
```diff
@@ -63,16 +69,12 @@ func (v *APIValidator) validateUpdateContainerService(cs, oldCs *api.OpenShiftMa
 			}
 		}
 	}
+	old.Properties.AuthProfile.IdentityProviders = cs.Properties.AuthProfile.IdentityProviders
```
asalkeld: @ehashman and @mjudeikis here we are copying cs to old, so the 2 can't be different. So there is no point in having the redact - as far as i can see.

ehashman:
Just wanted to check in on the accuracy of the change note before approving, which says `Properties.AuthProfile.IdentityProviders[0]`. I think the PR affects all IdentityProviders, not just the first one?

asalkeld:
> Just wanted to check in on the accuracy of the change note before approving, which says `Properties.AuthProfile.IdentityProviders[0]`. I think the PR affects all IdentityProviders, not just the first one?

thanks, i'll fix the releasenote. (we do only use the [0] currently)

pkg/api/validate/api.go
mjudeikis: I think redact was for sync pod not to log secretes into the logs if those are changed. If those are not same, sync used to print diff values. I think this is not solved by this change?

asalkeld: @mjudeikis given that we are now coping cs to old, it is impossible for it to print out.
remember this was to print out properties that were invalid to change, since we are now saying that you can change this, this won't print.

asalkeld:
/retest
flake https://github.com/openshift/openshift-azure/issues/1438

PR #1640, opened 2019-05-13
[WIP] move build node to rhel8

mjudeikis:
```release-note
NONE
```
TODO:
* investigate vm to underlaying vm connectivity issue. 
* get rid of footers util


pkg/vmimage/resources.go
```diff
@@ -148,8 +148,8 @@ func vm(subscriptionID, resourceGroup, location string, sshPublicKey string) *co
 				ImageReference: &compute.ImageReference{
 					Publisher: to.StringPtr("RedHat"),
 					Offer:     to.StringPtr("RHEL"),
-					Sku:       to.StringPtr("7-RAW"),
-					Version:   to.StringPtr("latest"),
+					Sku:       to.StringPtr("8"),
+					Version:   to.StringPtr("8.0.2019050711"),
```
ehashman: Is this going to need to change on every update?

pkg/vmimage/data/script.sh
```diff
@@ -5,22 +5,18 @@ exec 2>&1
...
-baseurl=https://packages.microsoft.com/yumrepos/azure-cli
-enabled=1
-gpgcheck=1
-gpgkey=https://packages.microsoft.com/keys/microsoft.asc
-EOF
-yum -y install azure-cli
+yum -y install git golang libguestfs-tools-c libvirt-daemon-config-network virt-install libguestfs-xfs libguestfs-tools python36 python36-devel redhat-rpm-config
+alternatives --set python /usr/bin/python3
+
+pip3 install --upgrade --pre azure-cli --extra-index-url https://azurecliprod.blob.core.windows.net/edge --no-cache-dir
```
ehashman: We're switching from an rpm install of azure-cli to a `pip install` directly to the system, which is typically not recommended as it can irreparably break existing rpm Python installs that share dependencies. Are we doing this now because the rpm isn't available for RHEL8?

pkg/vmimage/data/script.sh
```diff
@@ -5,22 +5,18 @@ exec 2>&1
...
-cat >/etc/yum.repos.d/azure-cli.repo <<'EOF'
-[azure-cli]
-baseurl=https://packages.microsoft.com/yumrepos/azure-cli
-enabled=1
-gpgcheck=1
-gpgkey=https://packages.microsoft.com/keys/microsoft.asc
-EOF
-yum -y install azure-cli
+yum -y install git golang libguestfs-tools-c libvirt-daemon-config-network virt-install libguestfs-xfs libguestfs-tools python36 python36-devel redhat-rpm-config
+alternatives --set python /usr/bin/python3
```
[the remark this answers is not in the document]
mjudeikis: No :) what time to be alive! :D

pkg/vmimage/resources.go
mjudeikis: I think they will publish RAW image later on or "latest".

pkg/vmimage/data/script.sh
mjudeikis: Need to add comment. https://github.com/Azure/azure-cli/issues/8780 is the reason.

pkg/vmimage/data/script.sh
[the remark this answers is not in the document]
asalkeld: > No :) what time to be alive! :D

hilarious!

jim-minter:
I'm not seeing what the benefit of applying this change is right now?  Are we not just changing one set of hacks and workarounds for another?

mjudeikis:
> I'm not seeing what the benefit of applying this change is right now? Are we not just changing one set of hacks and workarounds for another?

the goal was to move to rhel8 so we could use new tooling available with the new kernel to be able to consume rhcos image in our build pipeline. At the moment all this still depends on next week.

PR #1646, opened 2019-05-15
release 20190514 Security fix

mjudeikis:
```release-note
Security release for Microarchitectural Data Sampling (MDS)
```

### DIFF
--- a/pkg/arm/arm.go
+++ b/pkg/arm/arm.go
@@ -29,9 +29,9 @@ func New(ctx context.Context, log *logrus.Entry, cs *api.OpenShiftManagedCluster
 	switch cs.Config.PluginVersion {
 	case "v3.2":
 		return v3.New(ctx, log, cs, testConfig), nil
-	case "v4.2", "v4.3":
+	case "v4.2", "v4.3", "v4.4":
 		return v4.New(ctx, log, cs, testConfig), nil
-	case "v5.0":
+	case "v5.1":
 		return v5.New(ctx, log, cs, testConfig), nil
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/arm/arm.go
```diff
@@ -29,9 +29,9 @@ func New(ctx context.Context, log *logrus.Entry, cs *api.OpenShiftManagedCluster
 	switch cs.Config.PluginVersion {
 	case "v3.2":
 		return v3.New(ctx, log, cs, testConfig), nil
-	case "v4.2", "v4.3":
+	case "v4.2", "v4.3", "v4.4":
 		return v4.New(ctx, log, cs, testConfig), nil
-	case "v5.0":
+	case "v5.1":
```
ehashman: Should this not be additive like the "4.2", "4.3", etc. case or is no one running "5.0" so it doesn't matter?

pkg/cluster/hash_test.go
```diff
@@ -83,6 +83,24 @@ func TestHashScaleSetStability(t *testing.T) {
...
+				role: api.AgentPoolProfileRoleInfra,
+				// this value should not change
+				expectedHash: "afa8a3e0cbe7264b811f231ca7577f213022ae7a8a7a3c92da715c5c813213c3",
+			},
+			{
+				role: api.AgentPoolProfileRoleCompute,
+				// this value should not change
+				expectedHash: "f6f60ca20f23036807e7c54686c304a653830028864830dd31b58de7358e3fee",
+			},
+
```
ehashman: I think this extra line here is what gofmt is complaining about in the verify suite.

ehashman:
Upgrade/create tests failing on

```
The platform image 'redhat:osa:osa_311:311.98.20190514' is not available. Verify that all fields in the storage profile are correct.
```

I think I just got an email about that going live so we should be good after a retest?

ehashman:
/hold

Wait for #1647 to be merged first.

pkg/arm/arm.go
mjudeikis: 5.0 is not yet out so we dont care

ehashman:
/retest

All three failures are `Entrypoint received interrupt: terminated`, hopefully this is load-related/temporary?

PR #1656, opened 2019-05-16
Properly clean up cluster NS records from osadev.cloud dns zone

### LINKED ISSUE #1655: NS recordsets for fakerp clusters are not cleaned up properly
There are several orphan `NS` dns recordsets within the `osadev.cloud` zone which have been accumulating as a result of the fact that our fakerp code wasn't doing anything to clean those up.

There are currently about 1014 records in the `osadev.cloud`, most of which belong to clusters which have long been deleted.

```
charles@titan:~/$ az network dns record-set list --resource-group=dns --zone-name=osadev.cloud | jq .[].fqdn --raw-output | wc -l
1014
```
/kind bug
/cc @mjudeikis @Makdaam

charlesakalugwu:
Adds a fix to properly clean up NS records and add more debug logs so we see better what is happening with respect to dns record set updates/deletes.
 
closes #1655 

/cc @Makdaam @mjudeikis 

```release-note
NONE
```

### DIFF
--- a/pkg/fakerp/customer_handlers.go
+++ b/pkg/fakerp/customer_handlers.go
@@ -36,7 +36,7 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	err = dm.deleteOCPDNS(req.Context(), cs)
+	err = dm.deleteDns(req.Context(), cs)
 	if err != nil {
 		s.badRequest(w, fmt.Sprintf("Failed to delete dns records: %v", err))
 		return
--- a/pkg/fakerp/dns.go
+++ b/pkg/fakerp/dns.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"strings"
 
-	azdns "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
+	azuredns "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 
@@ -14,10 +14,11 @@ import (
 )
 
 type dnsManager struct {
-	zc               dns.ZonesClient
-	rsc              dns.RecordSetsClient
+	zonesClient      dns.ZonesClient
+	recordSetsClient dns.RecordSetsClient
 	dnsResourceGroup string
 	dnsDomain        string
+	log              *logrus.Entry
 }
 
 func newDNSManager(ctx context.Context, log *logrus.Entry, subscriptionID, dnsResourceGroup, dnsDomain string) (*dnsManager, error) {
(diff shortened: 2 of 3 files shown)


pkg/fakerp/dns.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"strings"
 
-	azdns "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
+	azuredns "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2017-10-01/dns"
```
mjudeikis: nit: necessary change 
same with `zc` -> `zoneClient`

pkg/fakerp/customer_handlers.go
```diff
@@ -36,7 +36,7 @@ func (s *Server) handleDelete(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	err = dm.deleteOCPDNS(req.Context(), cs)
+	err = dm.deleteAroDns(req.Context(), cs)
```
mjudeikis: nit: it brings the necessary information. `dn.deleteDns` is fine too. I think there is no need to bake in names into functions names. I think it was a mistake in the first place to add `OCP`

pkg/fakerp/dns.go
```diff
@@ -27,23 +28,66 @@ func newDNSManager(ctx context.Context, log *logrus.Entry, subscriptionID, dnsRe
...
+		recordSetsClient: dns.NewRecordSetsClient(ctx, log, subscriptionID, authorizer),
 		dnsResourceGroup: dnsResourceGroup,
 		dnsDomain:        dnsDomain,
+		log:              log,
 	}, nil
 }
 
-func (dm *dnsManager) createOrUpdateZone(ctx context.Context, resourceGroup, zoneName, parentResourceGroup, parentZoneName string) error {
-	zone, err := dm.zc.CreateOrUpdate(ctx, resourceGroup, zoneName, azdns.Zone{
+func (dm *dnsManager) createOrUpdateAroDns(ctx context.Context, cs *api.OpenShiftManagedCluster) error {
```
mjudeikis: Does this have to be part of `DNSmanager`. Looks like product specific. What if we move it to outside `utils` package:
`https://github.com/openshift/openshift-azure/blob/c010f7798a1c33b2507fbe1acc4df33cbf4a827b/pkg/fakerp/fakerp.go#L172` and https://github.com/openshift/openshift-azure/blob/e7b4e323e1f7da47aaf770ece87c218fff584e76/pkg/fakerp/customer_handlers.go#L33 becomes one call to function in utils, which initializes DNS manager and does all the ARO/OCP logic. This way we keep DNS manager product agnostic? WDYT?

pkg/fakerp/dns.go
charlesakalugwu: `azdns` to `azuredns`, `dnsManager.zc` to `dnsManager.zonesClient` and `dnsManager.rsc` to `dnsManager.recordSetsClient` were all changed to enhance readability. I feel the code is slightly more readable now.

pkg/fakerp/customer_handlers.go
charlesakalugwu: sure I can change this. I just did a copy-replace however this is a good point.

pkg/fakerp/dns.go
charlesakalugwu: This project is about one product `ARO` and the way things were setup seems fine to me. I don't feel strongly against any further refactorings however that can be done in a follow up.

mjudeikis: it true. but thinking from code perspective - reusable block always. More code can be re-used without ties to the product - better. Like nothing would stop you to just pull out DNS manager to separate git repository as a separate standalone package. But not a priority now :) just flow of ideas :)

charlesakalugwu:
spoke with @kwoodson and he brought up the very good point that if someone deletes a cluster by deleting the cluster's resource group with something like `` then we still end up with those orphan NS records. 

Therefore he'll work on adding NS record set purging from the `osadev.cloud` zone into our cluster purging processes.

charlesakalugwu:
upgrade-v5.1 flake: #1438 
/retest

charlesakalugwu:
upgrade-v5.1 flake: #1438
/retest

PR #1659, opened 2019-05-16
[WIP] fixing context deadline issues with vnet test

Makdaam:
still trying to figure out how to pass this value from the test to the azure client

The issue was that PollingDuration is 15min by default if not set. And a blocking Delete() call on a resource group takes more time if there's a whole ARO cluster inside.

```release-note
NONE
```

### DIFF
--- a/pkg/util/azureclient/resources/groups.go
+++ b/pkg/util/azureclient/resources/groups.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"context"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
 	"github.com/Azure/go-autorest/autorest"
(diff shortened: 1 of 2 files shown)


Makdaam:
The actual timeout happens in https://github.com/Azure/go-autorest/blob/master/autorest/azure/async.go#L165 during the Delete() call.
So I needed to extend the timeout in AfterEach for the whole Delete() call, but also increase the PollingDuration so that the WaitForCompletion doesn't time out.

Makdaam:
flake #1438
/retest

mjudeikis:
@Makdaam can we discuss this in the scrum, please? Missing some background

Makdaam:
Sure.

Makdaam:
Turns out I was just lucky and the context timeout extension doesn't solve the problem.
The issue with the VNET test remains (delete times out).

PR #1666, opened 2019-05-20
WIP: Add support for istio (service mesh)

asalkeld:
There is a special openshift version that uses an operator to install it.
Note this is still in preview.

Jira-316

See:

  https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html/service_mesh_install/service-mesh-installation#installing-service-mesh

To turn it on, the user runs the following:
```
  oc create -n istio-system -f docs/istio-installation.yaml
```

View status of istio:
```
  oc get pods -n istio-system
  oc get controlplane/istio-installation -n istio-system
```

Note this support automatic injection of sidecars.

To run the upstream example:
```
  oc new-project myproject
  oc adm policy add-scc-to-user anyuid -z default -n myproject
  oc adm policy add-scc-to-user privileged -z default -n myproject
  oc apply -n myproject -f https://raw.githubusercontent.com/Maistra/bookinfo/master/bookinfo.yaml
  oc apply -n myproject -f https://raw.githubusercontent.com/Maistra/bookinfo/master/bookinfo-gateway.yaml
  oc get all -n myproject
  export GATEWAY_URL=$(oc get route -n istio-system istio-ingressgateway -o jsonpath='{.spec.host}')
  curl -o /dev/null -s -w "%{http_code}\n" http://$GATEWAY_URL/productpage
  xdg-open http://$GATEWAY_URL/productpage
```

```release-note
Add support for service mesh (istio)
```


docs/istio-installation.yaml
```diff
@@ -0,0 +1,47 @@
+  apiVersion: istio.openshift.com/v1alpha3
```
asalkeld: this is just an example, not sure if it should live here..

charlesakalugwu:
@asalkeld I am missing some context (or JIRA card) for this PR. Can you please state a few reasons why we are adding this? I think OpenShift 4.0 might already be coming with Istio integration. Are we planning to provide istio integration while we are on 3.11? How will this work? Will all clusters come out-of-the-box with istio? Or will the customer be required to specify it via their manifest/cli e.g. `az openshift create --enable-service-mesh`? If all we are not looking to support the entire istio deployment then only the changes made to the api server's config, by adding Validating- and MutatingAdmissionWebHooks, are required for Istio auth. Thanks!

asalkeld:
> @asalkeld I am missing some context (or JIRA card) for this PR. Can you please state a few reasons 

I did reference the Jira-316 (but it is lacking a lot of info)

why we are adding this? I think OpenShift 4.0 might already be coming with Istio integration. Are we planning to provide istio integration while we are on 3.11? How will this work? Do we plan to stand up all clusters with istio out-of-the-box? Or will the customer be required to specify it via their manifest/cli e.g. `az openshift create --enable-service-mesh`? If all we are not looking to support the entire istio deployment then only the changes you've made to the api server's config, by adding Validating- and MutatingAdmissionWebHooks, are required for Istio auth. Thanks!


Well we could wait for 4, esp. as this is in preview. It was partly for my understanding of istio.
This PR just adds the necessary operator which i think is not a big overhead. if the customer wants to enable it they deploy using
```
oc create -n istio-system -f docs/istio-installation.yaml
```
This starts all the necessary services that istio needs. (In the disabled state there is just an operator running)

I think this makes it easier for the customer to enable than doing everything from scratch.

mjudeikis:
I like the deployment model. Worth discussing in architecture call?

jim-minter:
Good to explore this, but I don't believe this should be in ARO/3.11.  Prefer to close this PR and target as a requirement for 4.

Issue #1628, opened 2019-05-09
Timed out waiting to ensure resource group is created


charlesakalugwu:
We found a test flake where a time out occurred while waiting to ensure that a resource group existed/was created.

```
time="2019-05-09T14:01:49Z" level=info msg="using region northeurope"
time="2019-05-09T14:01:49Z" level=info msg="ensuring resource group ci-1625-e2e-upgrade-xghagv"
time="2019-05-09T14:02:53Z" level=fatal msg="timed out waiting for the condition"
exit status 1
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
```

artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1625/pull-ci-azure-master-upgrade-v5.0/38/build-log.txt

/kind test-flake

charlesakalugwu:
I believe the issue outlined here is the cause of #1617. Furthermore it seems all the examples of this issue occur in the `northeurope` region

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1621/pull-ci-azure-master-upgrade-v5.0/42/build-log.txt

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1679/pull-ci-azure-master-upgrade-v5.1/245/build-log.txt

m1kola:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1782/pull-ci-azure-master-upgrade-v5.1/321/build-log.txt in #1782

m1kola:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1781/pull-ci-azure-master-upgrade-v5.2/140/build-log.txt in #1781

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1801/pull-ci-azure-master-upgrade-v5.1/420/build-log.txt in #1801 

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1801/pull-ci-azure-master-upgrade-v5.2/240/build-log.txt in #1801

m1kola:
[...]

PR #1668, opened 2019-05-21
Add ClusterRole for cluster-monitoring-view

### LINKED ISSUE #1667: clusterrole.rbac.authorization.k8s.io "cluster-monitoring-view" not found
In pkg/sync/v6/data/ClusterRoleBinding.rbac.authorization.k8s.io/cluster-monitoring-view.yaml

```
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-monitoring-view
```
But there is no ClusterRole called "cluster-monitoring-view"

found when looking at some logs for https://github.com/openshift/openshift-azure/issues/1438

asalkeld:
```release-note
NONE
```
So currently this ClusterRole is totally missing. But I don't have a clue what the intention for it is.
I have simply grabbed a read-only section out of cluster-monitoring-operator as a starting point.
If anyone can suggest what the contents of this ClusterRole should be, please let me know.

fixes #1667

### DIFF
--- a/pkg/sync/v6/bindata.go
+++ b/pkg/sync/v6/bindata.go
@@ -15,6 +15,7 @@
 // data/ClusterRole.rbac.authorization.k8s.io/customer-admin-controller.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/edit.yaml
+// data/ClusterRole.rbac.authorization.k8s.io/geneva-cluster-monitoring-view.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/namespace-viewer.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/registry-admin.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/registry-editor.yaml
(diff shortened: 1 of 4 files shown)


ehashman:
@asalkeld I think I figured it out.

Here's the upstream YAML from cluster-monitoring-operator, which has a bit more limited rights than what we're granting in this PR: https://github.com/openshift/cluster-monitoring-operator/blob/1f6529a333d36fbc4788ec9b52419cc0e19ecef5/assets/cluster-monitoring-operator/cluster-role.yaml

The [OpenShift Prometheus monitoring docs](https://docs.openshift.com/container-platform/3.11/install_config/prometheus_cluster_monitoring.html) say: "You need to use a role that has read access to all namespaces, such as the `cluster-monitoring-view` cluster role." It seems to me that is the sole purpose of this role.

If we're using cluster-monitoring-operator, why is this not already getting provisioned?

asalkeld:
> @asalkeld I think I figured it out.
> 
> Here's the upstream YAML from cluster-monitoring-operator, which has a bit more limited rights than what we're granting in this PR: https://github.com/openshift/cluster-monitoring-operator/blob/1f6529a333d36fbc4788ec9b52419cc0e19ecef5/assets/cluster-monitoring-operator/cluster-role.yaml
> 
> The [OpenShift Prometheus monitoring docs](https://docs.openshift.com/container-platform/3.11/install_config/prometheus_cluster_monitoring.html) say: "You need to use a role that has read access to all namespaces, such as the `cluster-monitoring-view` cluster role." It seems to me that is the sole purpose of this role.
> 
> If we're using cluster-monitoring-operator, why is this not already getting provisioned?

sync pod create rbac type objects before deployments, so the role will be missing as it gets created by the operator (within a deployment). 

/hold
If you agree, i'll close this, another option is we copy that role and call it something else and use that. - just so we don't get annoying errors.

charlesakalugwu:
> > @asalkeld I think I figured it out.
> > Here's the upstream YAML from cluster-monitoring-operator, which has a bit more limited rights than what we're granting in this PR: https://github.com/openshift/cluster-monitoring-operator/blob/1f6529a333d36fbc4788ec9b52419cc0e19ecef5/assets/cluster-monitoring-operator/cluster-role.yaml
> > The [OpenShift Prometheus monitoring docs](https://docs.openshift.com/container-platform/3.11/install_config/prometheus_cluster_monitoring.html) say: "You need to use a role that has read access to all namespaces, such as the `cluster-monitoring-view` cluster role." It seems to me that is the sole purpose of this role.
> > If we're using cluster-monitoring-operator, why is this not already getting provisioned?
> 
> sync pod create rbac type objects before deployments, so the role will be missing as it gets created by the operator (within a deployment).
> 
> /hold
> If you agree, i'll close this, another option is we copy that role and call it something else and use that. - just so we don't get annoying errors.

@asalkeld sounds good to me

ehashman:
> If you agree, i'll close this, another option is we copy that role and call it something else and use that. - just so we don't get annoying errors.

I think we probably shouldn't do that in case upstream changes something and we end up drifting away from their configs as a result.

mjudeikis:
Is this not the case what this role should be provisioned by operator? 
we need this role for granting `geneva` to read the data from Prometheus.
https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v6/data/ClusterRoleBinding.rbac.authorization.k8s.io/cluster-monitoring-view.yaml

Hence I think we need cut version of this role, and as suggested rename it a differnet name (potentially having geneva in the name so we know where it is being used.

pkg/sync/v6/sync.go
```diff
@@ -3,6 +3,8 @@ package sync
 //go:generate go get github.com/go-bindata/go-bindata/go-bindata
 //go:generate go-bindata -nometadata -pkg $GOPACKAGE -prefix data data/...
 //go:generate gofmt -s -l -w bindata.go
+//go:generate curl https://raw.githubusercontent.com/openshift/cluster-monitoring-operator/master/assets/cluster-monitoring-operator/cluster-role.yaml -o data/ClusterRole.rbac.authorization.k8s.io/geneva-cluster-monitoring-view.yaml
+//go:generate sed -i "s/cluster-monitoring-view/geneva-cluster-monitoring-view/" data/ClusterRole.rbac.authorization.k8s.io/geneva-cluster-monitoring-view.yaml
```
asalkeld: i'd prefer this in the content upgrade generator, i'll see how to get it in there... @ehashman just wanted to make this automatic so it doesn't deviate from upstream..

asalkeld:
flakes https://github.com/openshift/openshift-azure/issues/1628 https://github.com/openshift/openshift-azure/issues/1677
/retest

jim-minter:
I don't understand what the rationale is for this PR - please could you explain?

asalkeld:
> I don't understand what the rationale is for this PR - please could you explain?

Hi @jim-minter 
when going through logs, we have seen pages of the following log message:
```
clusterrole.rbac.authorization.k8s.io "cluster-monitoring-view" not found
```
it is because sync pod creates roles /bindings before deployments and the monitoring operator creates this role. So when sync pod creates the rolebinding the role does not exist, thus the error.
This PR is just trying to reduce the noise in the logs, be getting the role defined in the upstream definition
and adding it to what the sync pod knows about so we don't get the error (with a different name so there is no clash). I have added it to the "make content" so if it gets changed upstream we get the change.

charlesakalugwu:
seeing this error:

```
sync error: ClusterRoleBinding.rbac.authorization.k8s.io "cluster-monitoring-view" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"geneva-cluster-monitoring-view"}: cannot change roleRef
```

asalkeld:
> seeing this error:
> 
> ```
> sync error: ClusterRoleBinding.rbac.authorization.k8s.io "cluster-monitoring-view" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"geneva-cluster-monitoring-view"}: cannot change roleRef
> ```

Maybe I should just close this?

PR #1669, opened 2019-05-22
non-updateable objects are deleted and re-created

### LINKED ISSUE #1288: Update non-updatable objects - sync pod
Some object cant be updated. In example StorageClass:
```
parameters:
  fsType: xfs
  kind: managed
  location: '*** GENERATED ***'
  skuName: Premium_LRS
```
to
```
parameters:
  cachingmode: None
  fsType: xfs
  kind: managed
  location: '*** GENERATED ***'
  skuName: Premium_LRS
```

Error now:
```
INFO[2019-03-15T10:59:55Z] Update StorageClass.storage.k8s.io/azure-disk 
INFO[2019-03-15T10:59:55Z] - Object.map[parameters].map[cachingmode]: <does not have key> != None 
INFO[2019-03-15T10:59:55Z] Error while syncing: cannot sync cluster config: StorageClass.storage.k8s.io "azure-disk" is invalid: parameters: Forbidden: updates to parameters are forbidden. 
```

Sync pod should know about these and after making diff, should delete and create, instead of patch. 

/kind bug

asalkeld:
```release-note
NONE
```
/cc @mjudeikis 
fixes #1288

### DIFF
--- a/pkg/sync/v6/client.go
+++ b/pkg/sync/v6/client.go
@@ -200,6 +200,16 @@ func write(log *logrus.Entry, dyn dynamic.ClientPool, grs []*discovery.APIGroupR
 
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
+		if err != nil && strings.Contains(err.Error(), "updates to parameters are forbidden") {
+			log.Infof("object %s is not updateable, will delete and re-create", o.GetName())
+			err = dc.Resource(res, o.GetNamespace()).Delete(o.GetName(), &metav1.DeleteOptions{})
+			if err != nil {
+				return
+			}
+			o.SetResourceVersion("")
+			_, err = dc.Resource(res, o.GetNamespace()).Create(o)
+		}
+
 		return
 	})
 


ehashman:
/retest

flake upgrade-3.2 #552

pkg/sync/v6/client.go
```diff
@@ -200,6 +200,16 @@ func write(log *logrus.Entry, dyn dynamic.ClientPool, grs []*discovery.APIGroupR
...
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
+		if err != nil && strings.Contains(err.Error(), "updates to parameters are forbidden") {
+			log.Infof("object %s is not updateable, will delete and re-create", o.GetName())
+			err = dc.Resource(res, o.GetNamespace()).Delete(o.GetName(), &metav1.DeleteOptions{})
+			if err != nil {
+				return
+			}
+			o.SetResourceVersion("")
+			_, err = dc.Resource(res, o.GetNamespace()).Create(o)
```
kwoodson: Do we want to verify the delete was successful and the object no longer exists?
Do we want to check that the create was successful?

ehashman: I think this is handled by the consumer of `write`; in the return statement below we bubble up the error.

PR #1674, opened 2019-05-24
integration tests: add support for checking what files changed

asalkeld:
```release-note
NONE
```

This is to help prove what has changed based on config or admin changes.
/cc @jim-minter

### DIFF
--- a/pkg/plugin/integration_test.go
+++ b/pkg/plugin/integration_test.go
@@ -2,11 +2,15 @@ package plugin
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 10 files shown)


pkg/plugin/integration_test.go
```diff
@@ -744,3 +751,176 @@ func TestHowActionsCauseRotations(t *testing.T) {
...
+	}
+	for _, scaleset := range az.ComputeRP.State {
+		for _, vm := range scaleset.Vms {
+			prefix = fmt.Sprintf("%s:%s:", scaleset.Name, *vm.InstanceID)
+			parentDir = scaleset.VmsDir[*vm.InstanceID]
+			filepath.Walk(scaleset.VmsDir[*vm.InstanceID], beforeCallback)
+		}
+	}
+
+	return func() map[string]map[string][]string {
```
ehashman: I find the program flow here to be pretty confusing---we defined two local callbacks already, and here we return another function, to be invoked at some later time.

I think this would be easier for me to follow (and we'd gain a little more code reusability) if we wrote a single function that returned a map of hashes of all the files from a scaleset. Then, we could run that in our tests at specified before and after moments, letting the caller keep track of that state rather than trying to do that for them here. And rather than embedding the diffing logic here, we could split that out into its own helper.

asalkeld:
Thanks for the review @ehashman , just parking it for now given the pending v4 effort. I'll bring it back to life if needed.

PR #1682, opened 2019-05-29
Default timeouts in RoundTrippers' http transports

charlesakalugwu:
Moving these back to the default dial timeouts

```release-note
NONE
```

### DIFF
--- a/test/e2e/standard/cluster.go
+++ b/test/e2e/standard/cluster.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"net"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


charlesakalugwu:
upgrade-v4.4 flake #1684 
/retest

charlesakalugwu:
upgrade-v5.1 flake #1628

jim-minter:
Given the open question around IP reachability, is this PR something that you want to proceed with?

test/e2e/standard/cluster.go
```diff
@@ -302,7 +302,7 @@ func (sc *SanityChecker) checkCanAccessServices(ctx context.Context) error {
 			// except the dial timeout reduction from 30 to 10 seconds.
 			Proxy: http.ProxyFromEnvironment,
 			DialContext: (&net.Dialer{
-				Timeout:   10 * time.Second,
+				Timeout:   30 * time.Second,
```
ehashman: Given the comment above, should we just leave all these values as their defaults rather than overriding them explicitly?

charlesakalugwu:
> Given the open question around IP reachability, is this PR something that you want to proceed with?

I am just trying to get us back to the defaults. It is not exactly clear to me why we decided to reduce net.Dialer.Timeout to 10s in the first place.

test/e2e/standard/cluster.go
charlesakalugwu: yeah I would prefer we go back to defaults here

charlesakalugwu:
upgrade-5.1 flake #1696 
/retest

charlesakalugwu:
upgrade-5.1 flake #552 
/retest

ehashman:
I found a few other instances where we're overriding this default:

https://github.com/openshift/openshift-azure/blob/a7c580356482b67dfdcfed9bd01b30f0b9f863b6/pkg/cluster/kubeclient/roundtrip.go#L148-L154

https://github.com/openshift/openshift-azure/blob/a7c580356482b67dfdcfed9bd01b30f0b9f863b6/test/clients/openshift/client.go#L84-L90

https://github.com/openshift/openshift-azure/blob/a7c580356482b67dfdcfed9bd01b30f0b9f863b6/cmd/monitoring/monitoring.go#L190-L197

Do these also need to be updated?

charlesakalugwu:
Thanks @ehashman. I'll incorporate the changes into the cases you discovered.

charlesakalugwu:
e2e-create flake #1700 
/retest

thekad:
/lgtm
third time's the charm, i hope :grinning:

PR #1686, opened 2019-05-29
Embedded http server for azure-controllers

charlesakalugwu:
Unfortunately, controller-runtime, the component used to build
azure-controllers is at version 0.1.0 and is tightly bound to
certain versions of kubernetes. 0.1.0 has no metrics exposed and
later versions have metrics but depend on newer versions of
kubernetes which we cannot update to until our Microsoft colleagues
are ready.

Therefore this PR starts a standard HTTP server when azure-controllers is run
and is used to serve the health and metrics endpoints.

In addition the deployment spec was updated to take these changes into account and a service was created.

/cc @Makdaam @mjudeikis @y-cote @thekad 

```release-note
NONE
```


mjudeikis:
/lgtm
/retest
flake: https://github.com/openshift/openshift-azure/issues/1438

PR #1688, opened 2019-05-29
Decrease scraping interval in prom service monitors

charlesakalugwu:
All of these had their scrape interval set to 1 minute. I feel like etcd metrics, at least, need to be collected at a shorter interval (e.g. 30 seconds). Let's discuss. 

/cc @Makdaam @jim-minter @mjudeikis 

```release-note
NONE
```


charlesakalugwu:
upgrade-v3.2 error: could not resolve github.com
/retest

charlesakalugwu:
e2e-create flake #1628 
upgrade-v4.4 failure resource quota exceeded

ehashman:
Looks like the release note of "NONEF" has a typo :)

charlesakalugwu:
> Looks like the release note of "NONEF" has a typo :)

thanks :) fixed it

charlesakalugwu:
> /lgtm
> 
> 30s was the minimum scrape interval we used for Prom metrics at my last gig, I'm surprised we have stuff on a 60s interval (although looking through the Prometheus docs this might be the new default?)

I was also surprised to see `1m`. I somehow initially read that as `1s` haha. The control plane service monitors created by the cluster monitoring operator have `30s` scraping intervals

ehashman:
> IIRC any other interval resulted in Geneva incorrectly aggregating data points.

I'm curious why that would be the case; if Geneva is scraping metrics on a per-minute basis, shouldn't this change just result in more recent samples available? What data does it aggregate over?

mjudeikis:
Is this not a case of: 
```
1.00 - metrics -> sent
1.30 - metrics
2.00 - metrics -> sent
2.30 - metric
3.00 - metric -> sent 
```
Hence we lose each second metric? (I don't know, interpreting this here). Is it possible to change the main Prom config to configure all SM to be 1m?

charlesakalugwu:
verify failure: could not resolve host github.com
/retest

charlesakalugwu:
> Prometheus forwarding to Geneva is done on a per-minute basis. IIRC any other interval resulted in Geneva incorrectly aggregating data points.
> 
> Given the above, what does this PR achieve?

@jim-minter We forward to geneva on a per-minute basis, therefore (assuming timings align perfectly), we are sending only 1 sample per minute to geneva (for metrics from service monitors which are scraped every 1 minute). Changing the interval to 30 seconds will result in 2 samples sent to geneva per minute and hence better observability.

@ehashman the default scraping interval for service monitors is still `30s` as far as I can see [here](https://github.com/coreos/prometheus-operator/blob/5468fa74815cd82110f66d7ecd938dbdcd25ef3a/pkg/prometheus/promcfg.go#L150).

Here are the currently configured intervals for all service monitors on a fresh ARO cluster (missing values use the default of 30s)

```
[charles@akalugwu openshift-azure]$ kubectl get servicemonitors --all-namespaces -o jsonpath='{range .items[*]}{@.metadata.name}{" "}{@.spec.endpoints[0].interval}{"\n"}{end}'
alertmanager 30s
canary 1m
cluster-monitoring-operator 
etcd-metrics 1m
kube-apiserver 30s
kube-controllers 30s
kube-state-metrics 2m
kubelet 30s
node-exporter 30s
prometheus 30s
prometheus-operator 
router-stats 1m
sync 30s
```

I am slightly confused as to why the cluster monitoring operator team choose a `2m` interval for kube-state-metric's service monitor :man_shrugging: 

With regards to this PR, I feel `etcd metrics` should be more closely monitored with a 30s-interval monitor.

mjudeikis:
@charlesakalugwu can you give me an example where we might want to have 30s interval data over 1m data? I get that current default setting for out-of-the-box ServiceMonitors is 30s, but that not an argument to make all of ours to 30s. At the moment we drop each second operator managed SM metric so we send per minute worth data. My concert is an mdsd agent which buffers the metrics before sending. More data means a bigger memory footprint, which means potential OoO again.

charlesakalugwu:
> @charlesakalugwu can you give me an example where we might want to have 30s interval data over 1m data? I get that current default setting for out-of-the-box ServiceMonitors is 30s, but that not an argument to make all of ours to 30s. At the moment we drop each second operator managed SM metric so we send per minute worth data. My concert is an mdsd agent which buffers the metrics before sending. More data means a bigger memory footprint, which means potential OoO again.

@mjudeikis based on my previous experience on teams that used prometheus as well as seeing what others have been/are doing in the wild, I haven't seen critical components being monitored with a 1 minute scrape interval. It has always been 30s or less. I think etcd is the archetype of a critical component. 

Having said that, if you are worried about our metric forwarding agents not being able to deal with that then we should fix that instead of putting ourselves in a corner with less accurate monitoring. However this is not an issue in this case because the OOM-problematic mdsd is for logs and not metrics.

mjudeikis:
I dont know. This still looks like PR where we are making assumption without understanding udelaying problem (or lack of problem) :man_shrugging:

charlesakalugwu:
> I dont know. This still looks like PR where we are making assumption without understanding udelaying problem (or lack of problem)

I'll can close this for now but i'm pretty sure we'll come back to this (etcd-metrics) in the future. 
/close

PR #1691, opened 2019-05-31
individual artifact gathering on test failures

mjudeikis:
```release-note
NONE
```

Adds basic event and logs gathering for pods for individual tests. This should help to identify cluster state during test failure instead of looking to artefacts at the end of execution where failure mode is already gone


test/e2e/standard/artifacts.go
```diff
@@ -0,0 +1,120 @@
...
+
+		err = sc.getPodsLogs(namespace, dir)
+		if err != nil {
+			return err
+		}
+
+		err = sc.getPodsStatus(namespace, dir)
+		if err != nil {
+			return err
+		}
```
ehashman: Do we want to wait until we've attempted to collect all of the Events/Logs/PodStatus before erroring out? It might be that we get an error for one but not all three.

PR #1694, opened 2019-05-31
Add prometheus instrumentation to sync

charlesakalugwu:
This adds 5 new metrics for the sync process

- sync_info (gauge) with labels
  - plugin_version
  - image
  - period_seconds
- sync_errors_total (counter)
- sync_executions_inflight (gauge)
- sync_last_executed (gauge)
- sync_duration_seconds (summary)

/cc @jim-minter @kwoodson @Makdaam @ehashman 

```release-note
NONE
```

### DIFF
--- a/pkg/entrypoint/sync/sync.go
+++ b/pkg/entrypoint/sync/sync.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


charlesakalugwu:
e2e-create flake #1695 
/retest

pkg/entrypoint/sync/sync.go
```diff
@@ -71,6 +120,11 @@ func start(cfg *cmdConfig) error {
 		return err
 	}
 	log.Printf("running sync for plugin %s", cs.Config.PluginVersion)
+	syncInfoGauge.With(prometheus.Labels{
+		"plugin_version": cs.Config.PluginVersion,
+		"image":          cs.Config.Images.Sync,
+		"period_seconds": fmt.Sprintf("%d", int(cfg.interval.Seconds())),
```
thekad: Question: is `period_seconds` always going to be different? if so, i'd strongly advise against using it as a label (as per https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels ) and instead add another metric for it.

ehashman: This is an info gauge and the value is being read out of the config, so I read this as only changing if the sync pod is recreated with a new config (very infrequently).

charlesakalugwu: @thekad the `period_seconds` will change the least often from all labels in `syncInfoGauge`

PR #1697, opened 2019-06-05
send output from unit tests to app insights

Makdaam:
```release-note
NONE
```
A command line tool which sends unit test results and durations to app insights.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ LDFLAGS="-X main.gitCommit=$(GITCOMMIT)"
 
 AZURE_IMAGE ?= quay.io/openshift-on-azure/azure:$(GITCOMMIT)
 
-.PHONY: all artifacts azure-image azure-push clean create delete e2e generate monitoring monitoring-run monitoring-stop secrets sync-run test unit upgrade verify vmimage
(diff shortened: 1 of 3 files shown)


charlesakalugwu:
@Makdaam what's the need for this work? We already have a codecov integration for pushing, analysing and taking actions on unit test results. Something for e2e tests metrics instead would really be beneficial.

Makdaam:
@charlesakalugwu codecov gives us coverage info, but no pass/fail or performance data. This PR puts the pass/fail and timing data in the same format as the e2e insights.

thekad:
> @charlesakalugwu codecov gives us coverage info, but no pass/fail or performance data. This PR puts the pass/fail and timing data in the same format as the e2e insights.

So I get that, I am just curious what the usefulness would be?

Makdaam:
I wanted to put all the data regarding all tests in one place to include that in a test babysitter's dashboard. For unit tests the duration of passing tests would be the useful metric (fails would be caught by the PR author).

Then again after your question I'm wondering if durations measured in µs/ms per test are worth chasing.

ehashman:
Seems reasonable to me, if we don't need this we can revert it.

/lgtm

PR #1704, opened 2019-06-10
Openshift v4: more work

asalkeld:
```release-note
NONE
```

"dep ensure" is crashing for me 
/cc @mjudeikis


cmd/azure/azure.go
```diff
@@ -88,15 +113,93 @@ func run() error {
...
+	}
+
+	// doing this to prevent the stdin questions.
+	ic := &installconfig.InstallConfig{}
+	ic.Config = cfg
+	ic.File = &asset.File{
+		Filename: "install-config.yaml",
+		Data:     data,
+	}
+	waIc := []asset.WritableAsset{ic}
```
asalkeld: the above is effectively what InstallConfig.Generate() will give you (without the questions)

cmd/azure/azure.go
```diff
@@ -88,15 +113,93 @@ func run() error {
...
+	directory := path.Join("clusters", name)
+	assetStore, err := store.NewStore(directory)
+	if err != nil {
+		return errors.Wrap(err, "failed to create asset store")
+	}
+
+	targets := waIc
+	targets = append(targets, targetassets.IgnitionConfigs...)
+	targets = append(targets, targetassets.Manifests...)
+	targets = append(targets, targetassets.Cluster...)
```
asalkeld: note: somewhere in the above is the azure clientid and clientSecret, need to find those..

cmd/azure/azure.go
```diff
@@ -88,15 +113,93 @@ func run() error {
...
+	// doing this to prevent the stdin questions.
+	ic := &installconfig.InstallConfig{}
+	ic.Config = cfg
+	ic.File = &asset.File{
+		Filename: "install-config.yaml",
+		Data:     data,
+	}
+	waIc := []asset.WritableAsset{ic}
+
 	// TODO: Start consuming assets. Assets need to be file system agnostic
```
asalkeld: I am not sure what you mean above..

for the fakerp, i'd expect to use store.NewStore() and store like the current fakerp does and the v4 installer. however for the realrp, i'd expect we need to re-implement the Store API so that it stores this stuff in a DB like MSFT does currently.

cmd/azure/azure.go
```diff
@@ -20,40 +34,51 @@ var (
 	baseDomain = "osadev.cloud"
 )
 
+type EnvConfig struct {
```
asalkeld: still need to add clientId and clientSecret as we use them

cmd/azure/azure.go
mjudeikis: This is what I had in mind. Its was more metal note for myself.

mjudeikis:
/lgtm 
(close_eyes)

PR #1708, opened 2019-06-11
tide up v4 library usage

mjudeikis:
```release-note
NONE
```

Add TLS from master branch
Add SSH key generation


Makefile
```diff
@@ -14,3 +15,6 @@ verify:
 .PHONY: create
 create:
 	./hack/create.sh ${RESOURCEGROUP}
+
+azure:
+	go build -ldflags ${LDFLAGS} ./cmd/$@
```
ehashman: How does one use this make target? Can we add this to the README?

Makefile
```diff
@@ -1,6 +1,7 @@
 SHELL := /bin/bash
 GITCOMMIT=$(shell git describe --tags HEAD)$(shell [[ $$(git status --porcelain) = "" ]] || echo -dirty)
 LDFLAGS="-X main.gitCommit=$(GITCOMMIT)"
+OPENSHIFT_INSTALL_DATA = ./vendor/github.com/openshift/installer/data/data
```
ehashman: Where is this used? Also, can we use `:=` instead of `=`?

cmd/azure/azure.go
```diff
@@ -187,23 +79,47 @@ func run() error {
 		return err
 	}
 
-	ec, err := newEnvConfig()
+	//action, err := rootCmd.Flags().GetString("action")
+	//if err != nil {
+	//	return err
+	//}
```
ehashman: Since this is commented out, looks like we're not actually using the action flag anywhere yet?

cmd/azure/azure.go
```diff
@@ -256,10 +167,28 @@ func run() error {
...
-	//	err = waitForInstallComplete(ctx, config, rootOpts.dir)
 
+	// waiting routine
+	config, err := clientcmd.BuildConfigFromFlags("", filepath.Join(ec.Directory, "auth", "kubeconfig"))
+				if err != nil {
+					logrus.Fatal(errors.Wrap(err, "loading kubeconfig"))
+				}
+
+	// TODO: Implement context
+	ctx:= context.Background()
```
ehashman: I'm surprised gofmt isn't barfing on this... looks like CI maybe isn't running because this is being merged into the dev branch?

pkg/util/tls/tls_test.go
```diff
@@ -0,0 +1,299 @@
...
+		if err != nil {
+			t.Error(err)
+		}
+	}
+	goodCert, err := readCert(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, c := range []*x509.Certificate{cert, goodCert} {
```
ehashman: Just to make sure I understand this test case: we're zeroing out all the info below, then checking to see if the certs match?

pkg/util/tls/tls.go
```diff
@@ -0,0 +1,34 @@
+package tls
```
ehashman: I think we are copying this file and the tests from master, but this version is truncated... do we want to copy the whole thing over? I think as written the tests won't pass because e.g. `NewCA` isn't defined here but there's a corresponding test.

Makefile
mjudeikis: This is override for data directory. Installer expects it to be in `./data/data`. I think in the binary mode this will not be a problem, just when running with `go run stuff`

cmd/azure/azure.go
mjudeikis: yes, preparation for delete logic. But create logic broken due to https://coreos.slack.com/archives/CHFF1LCUR/p1560272128015500

cmd/azure/azure.go
mjudeikis: yes. need to sort this fast!

asalkeld:
@ehashman i'll merge, so i can work on it today and fix the issues you suggested..
/lgtm

PR #1709, opened 2019-06-11
enabling Sync pod test on EveryPR

Makdaam:
```release-note
NONE
```
After #1543 is fixed enables the sync pod test on Every PR.
This will detect changes which cause sync pod to update a resource on every run.


thekad:
opened flake #1710 
/retest

PR #1711, opened 2019-06-12
Openshift v4 move code to the plugin api and add "delete" command

asalkeld:
```release-note
NONE
```
Note: i had to revert the vendor commit (not included here).

/cc @mjudeikis


mjudeikis:
/lgtm
lets see where we can take it from here

PR #1712, opened 2019-06-12
Openshift v4 subnet fix

mjudeikis:
```release-note
NONE
```

Add subnet fix and linting 
/cc @asalkeld


hack/verify/validate-code-format.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -e
+
+FILES=$(gofmt -s -l pkg cmd)
+
+if [ -n "$FILES" ]; then
+    echo You have go format errors in the below files, please run "gofmt -s -w pkg cmd"
```
ehashman: Might wanna add single quotes around this whole line, as written those double quotes won't print.

hack/verify/validate-code-format.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo $FILES
+    exit 1
+fi
+
+FILES=$(goimports -e -l -local=github.com/openshift/openshift-azure pkg cmd)
+
+if [ -n "$FILES" ]; then
+    echo You have go import errors in the below files, please run "goimports -e -w -local=github.com/openshift/openshift-azure pkg cmd"
```
ehashman: Ditto here re quotes

PR #1714, opened 2019-06-12
Label sync and azure-controllers in conformance to CMO expectations

charlesakalugwu:
The Prometheus instances are laid down by prometheus operator which is in turn laid down by the cluster monitoring operator. The prometheus instances are created with a `serviceMonitorSelector` field which limits how service monitors are discovered and used

```
serviceMonitorSelector:
    matchExpressions:
    - key: k8s-app
      operator: Exists
```

This means that service monitors need to be labelled with `k8s-app` in order to be discovered.

```
$ kubectl get prometheus --namespace=openshift-monitoring k8s -o jsonpath='{.spec.serviceMonitorSelector.matchExpressions}'
[map[key:k8s-app operator:Exists]]
```

This PR updates labels on sync and azure-controllers to conform to the expectations of the cluster monitoring operator

```release-notes
NONE
```

/cc @mjudeikis @makdaam @kwoodson


charlesakalugwu:
upgrade-v4.4 flake #1715
/retest

ehashman:
I can't find where in our code we set up the `serviceMonitorSelector` to match `k8s-app` labels, would you be able to point me to a link?

charlesakalugwu:
> I can't find where in our code we set up the `serviceMonitorSelector` to match `k8s-app` labels, would you be able to point me to a link?

@ehashman we don't....that is all created via the cluster monitoring operator which creates the Prometheus operator which creates the Prometheus instances. The said field can be found on the Prometheus instances spec. Please refer to the description of this PR for how to inspect that instance spec

charlesakalugwu:
> I can't find where in our code we set up the `serviceMonitorSelector` to match `k8s-app` labels, would you be able to point me to a link?

@ehashman 
`kubectl get prometheus --namespace=openshift-monitoring k8s -o json` should show you the Prometheus yaml

charlesakalugwu:
upgrade-v4.4 flake #552 
/retest

charlesakalugwu:
upgrade-v5.1 flake #1628 
upgrade-v4.4 was due to excessive label updates. Modified the code to only update the service monitor's labels because thats really whats needed
/hold

kwoodson:
I'm not a huge fan of us having to modify our labels to conform with a hard coded upstream rule.  On the other hand, I don't see a better alternative that is easier to implement.  If all we need to do is add a label and the correct permissions then this is a :+1: from me.

/lgtm

charlesakalugwu:
upgrade-v4.4 flake #552

PR #1716, opened 2019-06-12
Add myself as an approver

ehashman:
```release-note
NONE
```

/assign @jim-minter


charlesakalugwu:
@ehashman @jim-minter how is it decided who becomes an approver? I want to become one as well :man_shrugging:

PR #1719, opened 2019-06-13
Add charles as an approver

charlesakalugwu:
Here we go

```release-note
NONE
```

/cc @jim-minter


jim-minter:
@charlesakalugwu, I'll reach out to discuss with you off GitHub.

Issue #1544, opened 2019-04-18
No such host when trying to dial api server


Makdaam:
```
time="2019-04-17T18:37:36Z" level=error msg="Post https://kapujsodowfxqrnvvtiy.eastus.cloudapp.azure.com/apis/project.openshift.io/v1/projectrequests: dial tcp: lookup kapujsodowfxqrnvvtiy.eastus.cloudapp.azure.com on 10.142.0.13:53: no such host" func="github.com/openshift/openshift-azure/test/e2e/standard.(*SanityChecker).ValidateCluster" file="/go/src/github.com/openshift/openshift-azure/test/e2e/standard/sanitychecker.go:190"
```
/kind test-flake

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1491/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/647

asalkeld:
Note there are lots of network errors in the above logs
```
Op:\"dial\", Net:\"tcp\", Source:net.Addr(nil), Addr:net.Addr(nil), 
                      Err:(*net.DNSError)(0xc0003dc680)
Op: "dial",     Net: "tcp",                      Source: nil,                      Addr: nil,
                      Err: {
                          Err: "server misbehaving",
                          Name: "openshift.e2e-upgrade-makdaam-1491-rlikxr.osadev.cloud",
                          Server: "10.142.0.13:53",
                          IsTimeout: false,
                          IsTemporary: true,
                      },
```

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1721/pull-ci-azure-master-upgrade-v4.4/163/build-log.txt

PR #1721, opened 2019-06-13
Add --all flag to git add in content-update script

charlesakalugwu:
This should fix the content update `make verify` flake which shows up
in the content update job from time to time.

The problem is that an older version of git is used in the ci-base image
(ci-base is used to run hack/periodic/content-update.sh) and the semantics
of `git add` is different in that version to what it is in version 2 with respect 
to deleted paths.

```
warning: You ran 'git add' with neither '-A (--all)' or '--ignore-removal',
whose behaviour will change in Git 2.0 with respect to paths you removed.
Paths like 'pkg/sync/v6/data/Template.template.openshift.io/openshift/eap71-amq-persistent-s2i.yaml' that are
removed from your working tree are ignored with this version of Git.

* 'git add --ignore-removal <pathspec>', which is the current default,
  ignores paths you removed from your working tree.

* 'git add --all <pathspec>' will let you also record the removals.
```

sample content update with `to be deleted` paths: https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-content-update/45/build-log.txt

```release-notes
NONE
```

/cc @mjudeikis @jim-minter @makdaam

```release-note
```

### DIFF
--- a/hack/periodic/content-update.sh
+++ b/hack/periodic/content-update.sh
@@ -6,7 +6,7 @@ go generate ./...
 
 . hack/tests/ci-prepare.sh
 
-git add pkg/sync/v$(ls -d pkg/sync/v* | sed -e 's/.*v//' | sort -n | tail -1)
+git add --all pkg/sync/v$(ls -d pkg/sync/v* | sed -e 's/.*v//' | sort -n | tail -1)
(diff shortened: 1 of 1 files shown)


charlesakalugwu:
upgrade-v5.1 flake #552 
/retest

charlesakalugwu:
upgrade-v4.4 flake #1544

Issue #1643, opened 2019-05-14
Timeout waiting to validate stateful test app in e2e runs


charlesakalugwu:
Found a flake with the following error signature:

```

• Failure [162.252 seconds]
Openshift on Azure end user e2e tests [EndUser][Fake][EveryPR]
/home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:12
  should create and validate test apps [EndUser][Fake][Apps] [It]
  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:13

  Expected
      <[]*standard.TestError | len:1, cap:1>: [
          {
              Bucket: "validateStatefulApp",
              Err: {
                  s: "timed out waiting for the condition",
              },
          },
      ]
  to be empty

  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/end_user.go:25
```

/kind test-flake

charlesakalugwu:
#1642 e2e-create

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1642/pull-ci-azure-master-e2e-create/192/build-log.txt

charlesakalugwu:
#1633 e2e-create
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1633/pull-ci-azure-master-e2e-create/193/build-log.txt

ehashman:
#1664 e2e-create https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1664/pull-ci-azure-master-e2e-create/230/build-log.txt

Makdaam:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1675/pull-ci-azure-master-e2e-create/270 in #1675

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1776/pull-ci-azure-master-e2e-create/504/build-log.txt

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1787/pull-ci-azure-master-upgrade-v5.2/199/build-log.txt in #1787

ehashman:
/reopen
/remove-lifecycle rotten

Seen in #2240.

PR #1722, opened 2019-06-13
Push sync metrics to geneva

charlesakalugwu:
```release-note
NONE
```

/cc @mjudeikis @jim-minter @makdaam


charlesakalugwu:
e2e-create flake #552 
/retest

charlesakalugwu:
upgrade-v5.1 flake #1628 
/retest

ehashman:
This LGTM, why is it on hold?

jim-minter:
looks like a start, but the metrics specifications don't exactly conform with the current ones (yes, we're missing documentation on standards here...)  @Makdaam are you able to check through this and help here?
Examples:
* shouldn't we forward sync_info to Geneva once and use preaggregates, rather than forward sync_info_cluster_versions, sync_info_container_images ?
* sync_error_rate should be sync_errors_total_rate; should the window be 1m?
* sync_average_duration - shouldn't we be calculating this Geneva-side?

charlesakalugwu:
> * shouldn't we forward sync_info to Geneva once and use preaggregates, rather than forward sync_info_cluster_versions, sync_info_container_images ?

I am ok with either path

> * sync_error_rate should be sync_errors_total_rate; should the window be 1m?

As far as I can see in the code, I have followed conventions here as much as possible...please see the `container_` family of metrics. The window should be over 3m (the current reconciliation period for the sync) otherwise we end up with graphs containing holes in them. 

> * sync_average_duration - shouldn't we be calculating this Geneva-side?

`sync_average_duration` is a summary metric which consists of 5 series (3 quantiles and 2 counters for _sum and _count). If we calculate this geneva-side we are sending the 5 series for each cluster. If this is calculated before shipping to geneva then we are sending only 1 series per cluster which is much more efficient.

pkg/sync/v6/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -186,6 +186,14 @@ data:
       name: grpc_server_handled_rate
     - query: 'etcd_disk_backend_commit_duration_seconds_count{namespace="openshift-azure-monitoring"} != 0'
     - query: 'etcd_disk_backend_commit_duration_seconds_sum{namespace="openshift-azure-monitoring"} != 0'
+    - query: 'sum without(image, period_seconds, pod, instance) (sync_info)'
+      name: sync_info_cluster_versions
+    - query: 'sum without(plugin_version, period_seconds, pod, instance) (sync_info)'
+      name: sync_info_container_images
+    - query: 'rate(sync_errors_total{service="sync"}[5m])'
+      name: sync_error_rate
```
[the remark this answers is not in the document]
Makdaam: My mistake. The pattern is to replace _total with _rate.

pkg/sync/v6/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -186,6 +186,14 @@ data:
       name: grpc_server_handled_rate
     - query: 'etcd_disk_backend_commit_duration_seconds_count{namespace="openshift-azure-monitoring"} != 0'
     - query: 'etcd_disk_backend_commit_duration_seconds_sum{namespace="openshift-azure-monitoring"} != 0'
+    - query: 'sum without(image, period_seconds, pod, instance) (sync_info)'
+      name: sync_info_cluster_versions
+    - query: 'sum without(plugin_version, period_seconds, pod, instance) (sync_info)'
+      name: sync_info_container_images
```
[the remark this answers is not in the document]
charlesakalugwu: alright I'll update this

charlesakalugwu:
> This LGTM, why is it on hold?

@ehashman this is on hold because I am working to remove the need to specify the `service=sync` instrumentation labels for sync metrics themselves and, by extension, other parts of the code that will be instrumented. Due to the fact that metrics are initialized at package level, our azure command structure (`cmd/azure/azure.go` imports all the `pkg/entrypoint/` packages which is where sync metrics are initialized) causes metrics for one subcommand to appear in the list of metrics for other subcommands. I am working on fixing this problem.

charlesakalugwu:
upgrade-v5.1 flake #1628 
/retest

charlesakalugwu:
e2e-create flake #1643 
/retest

pkg/sync/v6/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -186,6 +186,11 @@ data:
       name: grpc_server_handled_rate
     - query: 'etcd_disk_backend_commit_duration_seconds_count{namespace="openshift-azure-monitoring"} != 0'
     - query: 'etcd_disk_backend_commit_duration_seconds_sum{namespace="openshift-azure-monitoring"} != 0'
+    - query: 'sync_info'
+    - query: 'rate(sync_errors_total[5m])'
+      name: sync_errors_rate
+    - query: 'rate(sync_duration_seconds_sum[5m]) / rate(sync_duration_seconds_count[5m])'
```
ehashman: Isn't `sync_duration_seconds` a summary metric? Why not forward the whole timeseries (which will include p50, p90, p99 by default I think) rather than manually calculating the average?

Makdaam: We tried sending the summary metrics and there's no good way to present the data in Geneva. 

For alerting we wanted recent data so last run or an average of the last 2 runs in the worst case. Sending the metric with quantile=0.5 would show us the 50th percentile since the start of the sync pod which would grow much slower if sync runs suddenly took minutes instead of the typical 6s.

ehashman: Hm, the median will grow much slower but p90/p99 won't. Mostly I'm just wary of further aggregating this summary data into an average because averages tend to be very lossy.

charlesakalugwu: @ehashman summary quantiles should generally be avoided as you cannot do math such as averages on top of quantiles, which prevents one from aggregating client-side quantiles from across instances of a given service. 

That and the fact that it is better to send 1 time series instead of 5 series to geneva are the reasons why i chose not to forward 5 of them and do the precalculation.

charlesakalugwu:
upgrade-v5.1 flake #552 
/retest

pkg/sync/v6/data/ConfigMap/openshift-azure-monitoring/metrics-bridge.yaml
ehashman: > summary quantiles should generally be avoided as you cannot do math such as averages on top of quantiles, which prevents one from aggregating client-side quantiles from across instances of a given service.

This is also true of sending a precalculated average as we're doing here though... if this is a concern we should be storing this as a histogram (although that doesn't solve the concern of sending more time series to Geneva). I just generally don't think averages are useful metrics, they hide too much info.

PR #1723, opened 2019-06-14
Defining sane garbage collection thresholds

gburges:
```release-note
NONE
```


pkg/startup/v4/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,16 @@ kubeletArguments:
...
+  image-gc-high-threshold: 
+  - "80"
+  image-gc-low-threshold: 
+  - "60"
+  maximum-dead-containers:
+  - '50'
+  maximum-dead-containers-per-container:
+  - '2'
+  minimum-container-ttl-duration:
+  - 5m
```
mjudeikis: Don't change v4 and v5. Only `v6`. It's not si critical issue we would need to do v4 release for this.

gburges: fixing

jim-minter:
@gburges what is the rationale for this / where do these values come from?

gburges:
I was talking with @mjudeikis about having GC defined at a lower level than the default (OSD experience shows that busy clusters do run into trouble with those values), and we both agreed that the OSD settings work, so i copied them from that env

pkg/startup/v6/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,16 @@ kubeletArguments:
...
   - "50"
+  image-gc-high-threshold: 
+  - "80"
+  image-gc-low-threshold: 
+  - "60"
+  maximum-dead-containers:
+  - '50'
+  maximum-dead-containers-per-container:
+  - '2'
+  minimum-container-ttl-duration:
```
mjudeikis: Deprecated, need to remove.

pkg/startup/v6/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,16 @@ kubeletArguments:
   - RotateKubeletClientCertificate=true,RotateKubeletServerCertificate=true
   max-pods:
   - "50"
+  image-gc-high-threshold:
```
mjudeikis: Default is 85%. lets remove this.

pkg/startup/v6/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,16 @@ kubeletArguments:
   - RotateKubeletClientCertificate=true,RotateKubeletServerCertificate=true
   max-pods:
   - "50"
+  image-gc-high-threshold: 
+  - "80"
+  image-gc-low-threshold:
```
mjudeikis: This is the one im interested in. Default is 80. And I remember we always were setting this lower than that as 80 sometimes is too big when you are on 79% and somebody tries to deploy "big fat daemon".

pkg/startup/v6/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,16 @@ kubeletArguments:
   - RotateKubeletClientCertificate=true,RotateKubeletServerCertificate=true
   max-pods:
   - "50"
+  image-gc-high-threshold: 
+  - "80"
+  image-gc-low-threshold: 
+  - "60"
+  maximum-dead-containers:
```
mjudeikis: deprecated

mjudeikis: All below will be deprecated in favour of `kubelet eviction` . But they are still here, so worth keeping.

gburges: it is getting deprecated, but even in 1.15 k8s it's there, so it's not phased out yet.

ehashman: I think the image GC is also getting deprecated as well. Doc link: https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/#user-configuration

pkg/startup/v6/data/master/etc/origin/node/node-config.yaml
```diff
@@ -33,6 +33,14 @@ kubeletArguments:
...
   max-pods:
   - "50"
+  image-gc-low-threshold: 
+  - "60"
+  maximum-dead-containers:
+  - '50'
+  maximum-dead-containers-per-container:
+  - '2'
+  minimum-container-ttl-duration:
+  - 5m
```
ehashman: Upstream defaults ([doc link](https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/#user-configuration)):

- `image-gc-high-threshold` the percent of disk usage which triggers image garbage collection (85%) **<-- we're not modifying this but maybe we should?**
- `image-gc-low-threshold` the percent of disk usage to which image garbage collection attempts to free (80%)
- `minimum-container-ttl-duration` minimum age for a finished container before it is garbage collected. Default is 0 minute, which means every finished container will be garbage collected **<-- we seem to be increasing the minimum ttl here?**
- `maximum-dead-containers-per-container` maximum number of old instances to be retained per container. Default is 1. **<-- also increasing this?**
- `maximum-dead-containers` maximum number of old instances of containers to retain globally. Default is -1, which means there is no global limit. **<-- this one is the big change and I think worth implementing to override the defaults. In a past life I've seen production nodes with 3000+ dead containers.**

gburges: Originally i've set the high TH to 80, removed it on @mj's advice
minimum-container-ttl-duration - set to 5m, so there will be more trace for investigation purposes

Even the logs suggest to increase the default numbers in case of maximum-dead-containers*, and these setting are the same as OSD, for better product alignment.

OSD also has a maximum-dead-containers=2500 set, so I'll add that too

ehashman: We're currently setting `maximum-dead-containers` in this PR to 50, I suspect 200 or 250 might be a good number. (2500 is pathologically large.)

Also, while you're making changes, for some reason there's a mix of single and double quotes around the numbers, might wanna standardize :)

gburges: Yeah, i ctrl+c, ctrl+v 'd it from the osd files, not sure why it is not consistent there

ehashman:
/retest

Looks like a flake.

ehashman:
/retest

flake #1438

gburges:
https://github.com/openshift/openshift-azure/issues/1775

PR #1725, opened 2019-06-17
Bumping image version to 311.104.20190617

kwoodson:
Bumping image version to 311.104.20190617

```release-note
None
```

### DIFF
--- a/pkg/arm/arm.go
+++ b/pkg/arm/arm.go
@@ -29,7 +29,7 @@ func New(ctx context.Context, log *logrus.Entry, cs *api.OpenShiftManagedCluster
 	switch cs.Config.PluginVersion {
 	case "v4.2", "v4.3", "v4.4":
 		return v4.New(ctx, log, cs, testConfig), nil
-	case "v5.1":
+	case "v5.1", "v5.2":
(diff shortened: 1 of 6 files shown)


jim-minter:
This is not quite right yet:
* only v6.0 should be changed, and that should change the image and also all of the 3.11.98 references there should change to 3.11.104 (we happen to be bumping the ocp version as well as the image).
* you then need to add a v5.2 to the pluginconfig, copy the v5.1 but make the same patches as outlined above, as well as s/v5.1/v5.2/ in the newly added stanza.
* you then need to find all references to v5.1 in the actual codebase and add a reference to v5.2 side by side.

Then please prepare a second PR which you will merge to release-v5, which will add v5.2 there as well.  That one will be much the same as above, but no references to v6.0, and additionally update the pluginVersion: field in the pluginconfig from v5.1 to v5.2.

We can then merge both, build the new v5.2 images, and update the tests. @mjudeikis fyi

kwoodson:
@jim-minter,

Thanks for the instruction.  This makes much more sense.  I have also added:
https://github.com/openshift/openshift-azure/pull/1731

@mjudeikis Please review.

PR #1728, opened 2019-06-18
Openshift v4

mjudeikis:
```release-note
NONE
```
Add metrics logic for app insights
Hack in vendor for insights


cmd/azure/azure.go
```diff
@@ -90,7 +100,19 @@ func run(log *logrus.Entry) error {
 		if err != nil {
 			return errors.Wrap(err, "failed to enrich InstallConfig")
 		}
-		return p.Create(ctx, log, name, cfg)
+
+		err = p.Create(ctx, log, name, cfg)
+		if err != nil && metrics {
+			m.Stop(false)
+		}
+		m.Stop(true)
```
ehashman: In the case that `metrics == true`, we end up running `m.Stop(false)` and `m.Stop(true)` immediately after. Is that intended?

cmd/azure/azure.go
```diff
@@ -90,7 +100,19 @@ func run(log *logrus.Entry) error {
...
+		return err
+	}
+
+	err = p.Delete(ctx, log, name)
+	if err != nil && metrics {
+		m.Stop(false)
 	}
-	return p.Delete(ctx, log, name)
+	m.Stop(true)
```
ehashman: (same thing here per comment above)

cmd/azure/azure.go
mjudeikis: Yes. Maybe the name is wrong. Basically, `m.starts` records start of the process, and `m.Stop` stops and send data out. `metrics=true` acts as a check so we send metrics only if we managed to an iniciate reporter. Will redo this as now I see it can be reduced :) thanks

PR #1735, opened 2019-06-20
Use separate prometheus registry for sync metrics

charlesakalugwu:
This adds a separate Prometheus registry for the sync process and therefore prevents sync metrics from spilling over into metrics for other parts of the plugin cli. 

```release-note
NONE
```

This is a pre-requisite to unholding #1722.

In a follow-up we'll come up with a design which centralizes the registry, conditionally registers metrics from different parts of the cli into the registry and allows for testability of the instrumentation. 

/cc @Makdaam @mjudeikis


charlesakalugwu:
upgrade-v5.2 flake #1438
/retest

PR #1736, opened 2019-06-20
custom assets injection pattern

mjudeikis:
```release-note
NONE
```

Add pattern for injecting custom assets


asalkeld:
@mjudeikis doesn't seem to be passing the create - timing out...

pkg/assets/manifests/operators.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Name returns a human friendly name for the asset.
+func (*AroOperator) Name() string {
+	return "AroOperator Config"
+}
+
+// Dependencies returns all of the dependencies directly needed to generate
+// the asset.
+func (*AroOperator) Dependencies() []asset.Asset {
+	return []asset.Asset{}
```
asalkeld: shouldn't this have the other openshift manifests as a dependancy? other wise it might start this before the cluster?

PR #1740, opened 2019-06-20
Entrypoint for conformance test

thekad:
```release-note
NONE
```

### DIFF
--- a/hack/tests/origin-conformance.sh
+++ b/hack/tests/origin-conformance.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -ex
 
-PATH=`go env GOPATH`/bin::$PATH
+PATH="$(go env GOPATH)/bin":$PATH
 TEST_SUITE="${TEST_SUITE:-openshift/conformance/parallel/minimal}"
 
 cleanup() {
@@ -18,20 +18,17 @@ fetch_origin() {
   test -d /tmp/origin || git clone --depth=1 --branch=${branch} $remote /tmp/origin
 }
 
-install_binaries() {
-  # install ginkgo
-  go get -v github.com/onsi/ginkgo/ginkgo
-
+build_extended_test() {
   # build extended.tests (regular user can't write to /usr/local)
   # TODO: This will need updating once we move to 4.x
-  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/$(go env GOARCH)/* $(go env GOPATH)/bin/
 }
 
-function run-tests() {
+function run_tests() {
   # TODO: This will need updating once we move to 4.x
   ginkgo -v -noColor \
     -nodes="${TEST_NODES:-30}" \
-    `which extended.test` -- \
+    $(which extended.test) -- \
     -ginkgo.focus="${TEST_SUITE}" \
     -e2e-output-dir /tmp/artifacts \
     -report-dir /tmp/artifacts/junit \
@@ -53,6 +50,6 @@ make create
 
 fetch_origin "release-3.11"
 
-install_binaries
+build_extended_test
 
-run-tests
+run_tests
--- a/hack/tests/origin-conformance.sh
+++ b/hack/tests/origin-conformance.sh
@@ -1,0 +1,78 @@
+#!/bin/bash -ex
+
+REAL="`realpath $0`"
+HERE="`dirname ${REAL}`"
+
+PATH=`go env GOPATH`/bin::$PATH
+TEST_SUITE="${TEST_SUITE:-openshift/conformance/parallel/minimal}"
+
+cleanup() {
+  set +e
+
+  generate_artifacts
+
+  delete
+}
+
+fetch_origin() {
+  local branch="${1:-release-3.11}"
+  local remote="https://github.com/openshift/origin.git"
+
+  test -d /tmp/origin || git clone --depth=1 --branch=${branch} $remote /tmp/origin
+}
+
+install_binaries() {
+  # install ginkgo
+  go get -v github.com/onsi/ginkgo/ginkgo
+
+  # build extended.tests and/or openshift-tests (regular user can't write to /usr/local)
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
+
+  # install kubectl if not available
+  which kubectl || ( curl -s -L -o `go env GOPATH`/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/`go env GOARCH`/kubectl && chmod -v +x `go env GOPATH`/bin/kubectl )
+}
+
+function run-tests() {
+  if which openshift-tests && [[ -n "${TEST_SUITE}" ]]; then
+    openshift-tests run "${TEST_SUITE}" \
+      --provider "${TEST_PROVIDER-}" \
+      -o /tmp/artifacts/e2e.log \
+      --junit-dir /tmp/artifacts/junit
+    exit 0
+  fi
+  if ! which extended.test; then
+    echo "must provide TEST_SUITE variable"
+    exit 1
+  fi
+  # TODO: remove after this point once we move to 4.x
+  ginkgo -v -noColor \
+    -nodes="${TEST_NODES:-30}" \
+    `which extended.test` -- \
+    -ginkgo.focus="${TEST_SUITE}" \
+    -e2e-output-dir /tmp/artifacts \
+    -report-dir /tmp/artifacts/junit \
+    -test.timeout=2h \
+    -provider "${TEST_PROVIDER-}" \
+    ${PROVIDER_ARGS-} || rc=$?
+
+  exit ${rc:-0}
+}
+
+trap cleanup EXIT
+
+. ${HERE}/ci-prepare.sh
+
+start_monitoring
+set_build_images
+
+make create
+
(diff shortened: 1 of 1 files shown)


asalkeld:
can you squash the commits? you have 1 small file but 7 commits

charlesakalugwu:
> can you squash the commits? you have 1 small file but 7 commits

+1 and the commit messages could be better... e.g. `I shouldn't have commented this out` or `verbosity` is confusing/not helpful to have in our commit history

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,78 @@
...
+
+install_binaries() {
+  # install ginkgo
+  go get -v github.com/onsi/ginkgo/ginkgo
+
+  # build extended.tests and/or openshift-tests (regular user can't write to /usr/local)
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
+
+  # install kubectl if not available
+  which kubectl || ( curl -s -L -o `go env GOPATH`/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/`go env GOARCH`/kubectl && chmod -v +x `go env GOPATH`/bin/kubectl )
```
jim-minter: fix this at https://github.com/openshift/release/blob/master/projects/azure/base-images/ci-base.yaml#L79 : copy `kubectl` to /usr/local/bin as well as `oc`, and add a comment about why it's needed

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash -ex
+
+REAL="`realpath $0`"
+HERE="`dirname ${REAL}`"
```
jim-minter: all our tests currently hard code `hack/tests` instead of $HERE, keep consistency for now

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,78 @@
...
+trap cleanup EXIT
+
+. ${HERE}/ci-prepare.sh
+
+start_monitoring
+set_build_images
+
+make create
+
+if [ -n "$OS_GIT_MAJOR" -a -n "$OS_GIT_MINOR" ]; then
```
jim-minter: under what circumstances are these set?

hack/tests/origin-conformance.sh
thekad: it's already tucked in openshift/release#4125

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,78 @@
...
+
+  # build extended.tests and/or openshift-tests (regular user can't write to /usr/local)
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
+
+  # install kubectl if not available
+  which kubectl || ( curl -s -L -o `go env GOPATH`/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/`go env GOARCH`/kubectl && chmod -v +x `go env GOPATH`/bin/kubectl )
+}
+
+function run-tests() {
+  if which openshift-tests && [[ -n "${TEST_SUITE}" ]]; then
```
jim-minter: is openshift-tests a v4 only thing?  If so, remove references from here - solve the v3 problem, don't try to add code to solve a v4 problem here.

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,78 @@
...
+    openshift-tests run "${TEST_SUITE}" \
+      --provider "${TEST_PROVIDER-}" \
+      -o /tmp/artifacts/e2e.log \
+      --junit-dir /tmp/artifacts/junit
+    exit 0
+  fi
+  if ! which extended.test; then
```
jim-minter: can this stanza be removed? at least, the error message doesn't make sense to me given the test

hack/tests/origin-conformance.sh
thekad: Actually I put that in a diff PR here openshift/release#4191 instead

hack/tests/origin-conformance.sh
thekad: I simplified this and explicitly do 3.11, makes it much cleaner

hack/tests/origin-conformance.sh
thekad: Yes, openshift-tests is v4-only, removing all references to it to make it cleaner

thekad:
> > can you squash the commits? you have 1 small file but 7 commits
> 
> +1 and the commit messages could be better... e.g. `I shouldn't have commented this out` or `verbosity` is confusing/not helpful to have in our commit history

Which is why we usually squash before merge and leave a single, meaningful commit message :smile:

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,58 @@
...
+fetch_origin() {
+  local branch="$1"
+  local remote="https://github.com/openshift/origin.git"
+
+  test -d /tmp/origin || git clone --depth=1 --branch=${branch} $remote /tmp/origin
+}
+
+install_binaries() {
+  # install ginkgo
+  go get -v github.com/onsi/ginkgo/ginkgo
```
charlesakalugwu: can we get away with installing ginkgo in the `ci-base` image instead? https://github.com/openshift/release/blob/master/projects/azure/base-images/ci-base.yaml

charlesakalugwu: can we get away with installing ginkgo in the `ci-base` image instead? https://github.com/openshift/release/blob/master/projects/azure/base-images/ci-base.yaml

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,58 @@
...
+
+  # build extended.tests (regular user can't write to /usr/local)
+  # TODO: This will need updating once we move to 4.x
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
+}
+
+function run-tests() {
```
charlesakalugwu: can we rename this to use snake case? 
```suggestion
function run_tests() {
```

thekad:
This needs openshift/release#4191 to be merged first, and the prow job that uses this entrypoint is in openshift/release#4125 cc @charlesakalugwu

hack/tests/origin-conformance.sh
thekad: I thought about it, then thought what could happen if you need a specific version of ginkgo installed? granted, my argument is weak. I will add it to the PR.

hack/tests/origin-conformance.sh
thekad: this is bash though, we don't do snake case in bash afaics

charlesakalugwu: I was referring to how the other functions in the script are named. This is the only one using a hyphen. Others use an underscore. Just a cosmetic thing...

charlesakalugwu:
> This needs [openshift/release#4191](https://github.com/openshift/release/pull/4191) to be merged first, and the prow job that uses this entrypoint is in [openshift/release#4125](https://github.com/openshift/release/pull/4125) cc @charlesakalugwu

Thanks a lot @thekad :)

hack/tests/origin-conformance.sh
ehashman: > we don't do snake case in bash afaics

Every other subroutine in this file uses snake case :smile:

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash -ex
+
+PATH=`go env GOPATH`/bin::$PATH
```
ehashman: Please add quotes around this string.

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,58 @@
...
+  generate_artifacts
+
+  delete
+}
+
+fetch_origin() {
+  local branch="$1"
+  local remote="https://github.com/openshift/origin.git"
+
+  test -d /tmp/origin || git clone --depth=1 --branch=${branch} $remote /tmp/origin
```
ehashman: What if `/tmp/origin` exists but the repo is out of date? Or it contains something totally unexpected? I think it might make more sense to create a new temporary directory for the clone on each test run.

hack/tests/origin-conformance.sh
ehashman: We're not pinning here to a specific version of ginkgo so it seems safe to do it in the `ci-base` image. I think you've made that change already so this can be removed :)

hack/tests/origin-conformance.sh
```diff
@@ -0,0 +1,58 @@
...
+}
+
+install_binaries() {
+  # install ginkgo
+  go get -v github.com/onsi/ginkgo/ginkgo
+
+  # build extended.tests (regular user can't write to /usr/local)
+  # TODO: This will need updating once we move to 4.x
+  make -C /tmp/origin build-extended-test && mv -v /tmp/origin/_output/local/bin/linux/`go env GOARCH`/* `go env GOPATH`/bin/
```
ehashman: Small style nit, but the backticks syntax for command substitution is considered legacy. I think we use `$(...)` in the rest of our scripts, could we update this to be consistent?

hack/tests/origin-conformance.sh
thekad: This already runs in an ephemeral container in CI, so the checkout wouldn't exist in CI. I actually used this while testing because I could just use my local clone of origin mounted on /tmp/origin and avoid doing clones every single time.

ehashman: Eh, fair enough. I like things to be safe by default to reduce shooting oneself in the foot, but I suppose running this continuously locally should be approached with caution :)

ehashman:
Prow seems stuck...

/retest

PR #1742, opened 2019-06-21
Revert docker build strategy restriction

ehashman:
Reverts #1303.

Documented at https://docs.openshift.com/container-platform/3.11/admin_guidesecuring_builds.html#disabling-a-build-strategy-globally

```release-note
Add support for docker build strategy
```


ehashman:
This seems to be passing tests now so removing the WIP. This is a hard revert of #1303, not sure if we want to modify the tests rather than remove them but removing them was easy so I just did that.

asalkeld:
@ehashman looks good, happy to lgtm. Is everyone happy with the security concerns?

charlesakalugwu:
@ehashman I think it would be nicer to modify the test to confirm that docker builds are permitted. Besides that, this lgtm. 
@asalkeld are you referring to @grantomation's finding that the ose-docker-build pod needs to run with the "privileged" scc? I have reservations about this and look forward to hear what others think.

jim-minter:
/lgtm
this matches osd behaviour.
@ehashman @thekad please can you ensure our docs are in sync (will need to revert @thekad 's docs PR)

ehashman:
Yup, I'll follow up with a reversion of that PR. It's the last thing remaining in the JIRA.

PR #1738, opened 2019-06-20
Revert "Add default router profile to real rp sample manifest"

gburges:
This reverts commit 42a894b120140fcea4e10af8d4e2287676277e93.

```release-note
NONE
```

PR #1745, opened 2019-06-24
Re-add default router profile to real rp sample manifest

charlesakalugwu:
@ehashman's defaulting work got merged but hasn't made it to all earlier versions
of the plugin deployed on the RP. Therefore #1738 was hastily merged.

We can only take this out after all supported versions in production
contain the defaulting code.

sample error artifacts: https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-azure-e2e-prod-vnet/61/build-log.txt

/cc @ehashman @thekad @makdaam @gburges 

```release-note
NONE
```

### DIFF
--- a/test/manifests/realrp/create.yaml
+++ b/test/manifests/realrp/create.yaml
@@ -33,3 +33,5 @@ properties:
     vmSize: Standard_D4s_v3
     subnetCidr: 10.0.0.0/24
     osType: Linux
+  routerProfiles:
+  - name: default


ehashman:
Agreed. I guess this didn't show up in CI because we don't have prepush jobs that test against the RealRP? @charlesakalugwu do we have a dashboard of all the tests this change caused to fail?

charlesakalugwu:
> Agreed. I guess this didn't show up in CI because we don't have prepush jobs that test against the RealRP? @charlesakalugwu do we have a dashboard of all the tests this change caused to fail?

@ehashman yes currently all the `prod` e2e tests run as prow periodic jobs and unfortunately prow doesn't yet provide means to send notifications about job statuses of interest.

PR #1760, opened 2019-07-01
Metrics-bridge metrics

charlesakalugwu:
This adds 4 new metrics for metrics-bridge:

- `metrics_bridge_errors_total` (counter)
- `metrics_bridge_bytes_transferred_total` (counter)
- `metrics_bridge_metrics_transferred_total` (counter)
- `metrics_bridge_processing_duration_seconds` (summary)

*Sample Scrape*
```
# HELP metrics_bridge_bytes_transferred_total Total number of bytes transferred.
# TYPE metrics_bridge_bytes_transferred_total gauge
metrics_bridge_bytes_transferred_total 6.012035718e+09
# HELP metrics_bridge_errors_total Total number of errors.
# TYPE metrics_bridge_errors_total counter
metrics_bridge_errors_total 0
# HELP metrics_bridge_metrics_transferred_total Total number of metrics transferred.
# TYPE metrics_bridge_metrics_transferred_total gauge
metrics_bridge_metrics_transferred_total 8.955518e+06
# HELP metrics_bridge_processing_duration_seconds The duration of metrics bridge query processing
# TYPE metrics_bridge_processing_duration_seconds summary
metrics_bridge_processing_duration_seconds{quantile="0.5"} 2.325671467
metrics_bridge_processing_duration_seconds{quantile="0.9"} 2.809971859
metrics_bridge_processing_duration_seconds{quantile="0.99"} 2.809971859
metrics_bridge_processing_duration_seconds_sum 8102.594877420008
metrics_bridge_processing_duration_seconds_count 3329
```

/cc @Makdaam @mjudeikis @ehashman @thekad 

```release-note
NONE
```


Makdaam:
/lgtm

Metrics Bridge downtime has to be detected further down the pipeline via metrics rate, since prometheus metrics will not be forwarded if the metrics-bridge is down.

pkg/entrypoint/metricsbridge/metricsbridge.go
```diff
@@ -268,15 +273,18 @@ func (c *MetricsConfig) runOnce(ctx context.Context) error {
...
+			if err != nil {
 				return err
 			}
-
+			bytesCount += bytesSent
 			metricsCount++
 		}
 	}
+	metricsBridgeMetricsTransferredCounter.Add(float64(metricsCount))
+	metricsBridgeBytesTransferredCounter.Add(float64(bytesCount))
```
ehashman: Why wait until we've broken out of the loop to update this? Should we not do it live?

pkg/entrypoint/metricsbridge/metricsbridge.go
```diff
@@ -286,14 +294,31 @@ func start(cfg *cmdConfig) error {
...
 	log := logrus.NewEntry(logrus.StandardLogger())
 
+	log.Info("starting metrics endpoint")
+	l, err := net.Listen("tcp", fmt.Sprintf(":%d", cfg.httpPort))
+	if err != nil {
+		return err
+	}
+
+	mux := &http.ServeMux{}
+	mux.Handle("/healthz/ready", http.HandlerFunc(readyHandler))
```
ehashman: Why `/healthz/ready`? Isn't `/healthz` more standard?

pkg/sync/v6/data/Deployment.apps/openshift-azure-monitoring/metrics-bridge.yaml
```diff
@@ -44,6 +44,15 @@ spec:
         - /config/config.yaml
         image: '*** GENERATED ***'
         name: metricsbridge
+        ports:
+        - containerPort: 8080
+          name: http
+        readinessProbe:
+          httpGet:
+            path: healthz/ready
```
ehashman: Do we not need an initial `/` on the path?

pkg/entrypoint/metricsbridge/metricsbridge.go
```diff
@@ -286,14 +294,31 @@ func start(cfg *cmdConfig) error {
...
+	go http.Serve(l, mux)
+
 	log.Printf("metricsbridge starting")
 
 	if cfg.configDir == "" {
 		return fmt.Errorf("config value cant be empty")
 	}
 
 	if err := run(log, cfg.configDir); err != nil {
+		metricsBridgeErrorsCounter.Inc()
```
ehashman: By putting this increment in the start function, does that mean this will only increment once? Do we not want to include an error incrementer on line 277 above as well?

pkg/entrypoint/metricsbridge/metricsbridge.go
charlesakalugwu: It just felt cleaner to me.

pkg/entrypoint/metricsbridge/metricsbridge.go
charlesakalugwu: Yes `/healthz` is more standard but `/healthz/ready` is the convention we've used so far for the ready handlers for the `canary`, `sync` and `azure-controllers` pods.

pkg/sync/v6/data/Deployment.apps/openshift-azure-monitoring/metrics-bridge.yaml
charlesakalugwu: Not necessarily. This works without the initial slash. I'll add it however and we can update the other places in the codebase where its used in a follow-up.

pkg/entrypoint/metricsbridge/metricsbridge.go
charlesakalugwu: Good catch. Yes as this is right now, the counter will only ever increment once. I've moved this to a more appropriate place (inside the for loop in `run()`). Every error from `runOnce(context)` is returned to the `run()` caller. So we do not need an extra counter increment at 277.

pkg/entrypoint/metricsbridge/metricsbridge.go
ehashman: Eh, best to remain consistent then.

pkg/entrypoint/metricsbridge/metricsbridge.go
ehashman: My concern here is that the metrics won't be accurate if we end up failing out early due to an error---the bytes count won't be updated even if some bytes have been written already. What do you think?

pkg/entrypoint/metricsbridge/telemetry.go
```diff
@@ -0,0 +1,57 @@
...
+
+var (
+	metricsBridgeRegistry = prometheus.NewRegistry()
+
+	// TODO(charlesakalugwu): Add unit tests for the handling of these metrics once
+	//  the upstream library supports it
+	metricsBridgeErrorsCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "metrics_bridge_errors_total",
+			Help: "Total number of errors.",
```
thekad: non-blocking nitpick: make the ending period on all Help strings the same :smiley: (I'd vote for no ending period, not sure what's the "normal" here)

charlesakalugwu: Thanks @thekad :D The norm is actually for HELP strings to have a period at the end. I missed that on `metrics_bridge_processing_duration_seconds`. Thanks!

pkg/entrypoint/metricsbridge/metricsbridge.go
charlesakalugwu: Yes I agree. I'll move those lines into the loop :)

charlesakalugwu:
e2e-create flake #1778 
/retest

ehashman:
/retest

flake #1400 on upgrade-4.4

charlesakalugwu:
upgrade-v4.4 flake #1438 
upgrade-v5.1 flake #1628 
upgrade-v5.2 flake #1628
/retest

charlesakalugwu:
upgrade-v4.4 flake #1771

charlesakalugwu:
upgrade-v5.1 flake #1771 
/retest

PR #1765, opened 2019-07-01
adding alert definitions for the sync pod, router and api server

Makdaam:
```release-note
NONE
```


pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+      labels:
+        severity: warning
+  - name: aro-canary.rules
+    rules:
+    - alert: CanaryPodDown
+      annotations:
+        description: At least one Canary Pod is down for 10min
+        summary: Canary Pod is down
+      expr: |
+        absent(sum(up{service="canary"}) == 2)
```
ehashman: I don't think `absent` is what we want to use here, at least if I'm reading the description right. If `sum(up{service="canary"})` = 1, for example, this won't alert. Actually, even if the `up` vector for the canary service doesn't exist at all, because we're calling `sum`, this will always return an empty vector, at least according to the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#absent): 

```
absent(sum(nonexistent{job="myjob"}))
# => {}
```

Hence, this will never alert. I think the query we actually want is

```
sum(up{service="canary"}) < 2
```

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+      labels:
+        severity: warning
+  - name: aro-router.rules
+    rules:
+    - alert: RouterPodDown
+      annotations:
+        description: At least one Router Pod is down for 10min
+        summary: Router Pod is down
+      expr: |
+        absent(sum(up{service="router-stats"}) == 3)
```
ehashman: I think this query suffers from the same issue I identified above.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+      expr: |
+        rate(haproxy_server_http_responses_total{code="5xx",route=~"docker-registry|registry-console|branding|canary|console|alertmanager-main|grafana|prometheus-k8s"}[1m]) > 0
+      labels:
+        severity: warning
+    - alert: RouterDowntime
+      annotations:
+        description: Router noticed downtime of {{$labels.route}} pod {{$labels.pod}}
+        summary: Router noticed downtime
+      expr: |
+        rate(haproxy_server_downtime_seconds_total{route=~"docker-registry|registry-console|branding|canary|console|alertmanager-main|grafana|prometheus-k8s"}[1m]) > 0
```
ehashman: Is this something we want to alert on? This seems like it could be very noisy.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+      labels:
+        severity: warning
+  - name: aro-console.rules
+    rules:
+    - alert: APIBackendDown
+      annotations:
+        description: At least one API Backend is marked down in router
+        summary: API backend is down
+      expr: |
+        absent(sum(haproxy_backend_up{route="console"}) == 3)
```
ehashman: Has the `absent(sum(...` issue.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+      labels:
+        severity: warning
+    - alert: APINoMetrics
+      annotations:
+        description: Couldn't query metrics of at least one API server
+        summary: API not exposing metrics
+      expr: |
+        absent(sum(up{job="apiserver"})==3)
```
ehashman: Has the `absent(sum(...` issue.

charlesakalugwu:
upgrade-v4.4 flake #1438 
/retest

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
charlesakalugwu: @Makdaam I know that the out-of-the-box alerts that come with the installation use `absent` to detect services which are down. However I think its use together with `sum` is not correct as @ehashman points out. 

Personally, I would do the following 

```
avg without(instance)(up{job="canary"}) < 0.5
```
This will alert if more than half of the total number of canary instances are down. Such a query is resistant to changes in the actual topology of your service deployments i.e. it works if the canary job has 2 instances now or 100 instances in the future.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+    rules:
+    - alert: SyncPodDown
+      annotations:
+        description: Sync Pod is down
+        summary: Sync Pod is down
+      expr: |
+        absent(up{service="sync"} == 1)
+      for: 10m
+      labels:
+        severity: warning
```
charlesakalugwu: It might be more understandable if the following were used as severity values, `page` and `ticket`. However I understand this might not be straightforward due to the out-of-the-box rules use `warning` and `critical`. Whichever way is chosen, I would expect `ticket` or `warning` to create a SEV3/SEV4 ticket while `critical` and `page` end up as SEV1/SEV2 which triggers a page.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
```diff
@@ -0,0 +1,86 @@
...
+  name: prometheus-k8s-arorules
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: aro-sync.rules
+    rules:
+    - alert: SyncPodDown
+      annotations:
+        description: Sync Pod is down
+        summary: Sync Pod is down
```
charlesakalugwu: It would be really useful if we could add a `dashboard` entry (with link to a dashboard for the service) however I think this can be done on the geneva side albeit "manually" :man_shrugging:

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
Makdaam: @ehashman 
I tested the three scenarios with 3 alerts on a healthy cluster:
 - if the number of up metrics for a service is < expected number
 - if the number of up metrics for a service is > expected number
 - if there are no up metrics for a service

The expressions I used are:
absent(sum(up{service="canary"})  == 3)
absent(sum(up{service="canary"})  == 1)
absent(sum(up{service="heffalump"})  == 2)

All of them are active when there are 2 canary pods, and 0 heffalump pods/services.
The alert is active for every element in the returned vector https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/ 

The reason why I used absent is that sum(vector_expression) is an aggregation operator and not a function. If the vector_expression is an empty vector, sum will also be an empty vector, and not 0. Only when a vector_expression containing elements which sum up to 0 makes the sum 0. This means that if all pods are down and no up values are selected the alert doesn't trigger.

Makdaam: @charlesakalugwu
I wanted to detect whenever the number of pods is different than expected (so the expression has to know how many pods are supposed to be up in a healthy cluster). 

It might be a good idea to add a higher severity alert when all canary pods are down.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
Makdaam: I should document that, as this is the idea. Warnings should not wake people up (Sev3/4) while criticals should (Sev2).

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
Makdaam: I want to keep the abstraction layers clean, so no Geneva specific stuff in Prometheus.
Especially since the link will be different for staging and prod.

pkg/sync/v6/data/PrometheusRule.monitoring.coreos.com/openshift-monitoring/prometheus-k8s-arorules.yaml
Makdaam: It doesn't translate 1:1 to alerts/issues in Geneva. I wanted to send it all as part of the ALERTS metric and then select on what we actually alert.

If it turns out to be too noisy to use in Geneva alerting I'll remove it.

thekad:
/lgtm
thanks for explaining the `absent(sum())` logic in your queries.

Issue #1766, opened 2019-07-01
Group sync broken for Guest Accounts


mjudeikis:
Group sync does now work for guest accounts:
Login flow user metadata:
```
[root@master-000000 ~]# oc get user
NAME                                       UID                                    FULL NAME            IDENTITIES
live.com#mjudeiki@redhat.com               b9a96ebd-9c03-11e9-bf3e-000d3a4f7cf4   mjudeiki             Azure AD:bC37ZKYYOzbjtsKrK6Cgazvzxczxczxc
mjudeikis@rhcuppettgmail.onmicrosoft.com   30aead80-9c03-11e9-8117-000d3a4f705b   Mangirdas Judeikis   Azure AD:RoejElH-phVpBcRWgh-zxc
```
Sync groups:
```
[root@master-000000 ~]# oc get groups
NAME                  USERS
osa-customer-admins   b-majude@microsoft.com, jaquincy@microsoft.com, mjudeiki@redhat.com, testuserdisabled@rhcuppettgmail.onmicrosoft.com
```
Config:
```
 - login: true                                                                                                                                                                                                                                                                                                                                                    
    mappingMethod: claim                                                                                                                                                                                                                                                                                                                                                   
    name: "Azure AD"                                                                                                                                                                                                                                                                                                                                                       
    provider:                                                                                                                                                                                                                                                                                                                                                              
      apiVersion: v1                                                                                                                                                                                                                                                                                                                                                       
      claims:                                                                                                                                                                                                                                                                                                                                                              
        email:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
        - email                                                                                                                                                                                                                                                                                                                                                            
        id:                                                                                                                                                                                                                                                                                                                                                                
        - sub                                                                                                                                                                                                                                                                                                                                                              
        name:                                                                                                                                                                                                                                                                                                                                                              
        - name                                                                                                                                                                                                                                                                                                                                                             
        preferredUsername:                                                                                                                                                                                                                                                                                                                                                 
        - unique_name                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
      clientID: "1c700e16-9002-4184-b178-xxxx"                                                                                                                                                                                                                                                                                                                     
      clientSecret:                                                                                                                                                                                                                                                                                                                                                        
        file: client.secret                                                                                                                                                                                                                                                                                                                                                
[...]

PR #1768, opened 2019-07-02
group sync fix to include external users

mjudeikis:
```release-notes
fix group sync for guest accounts
```

Fix group sync for guest accounts

### DIFF
--- a/pkg/controllers/customeradmin/group_mapping.go
+++ b/pkg/controllers/customeradmin/group_mapping.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/go-autorest/autorest/azure"
 	v1 "github.com/openshift/api/user/v1"
+	userv1client "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
 	"github.com/sirupsen/logrus"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -18,7 +20,7 @@ import (
 
 // fromMSGraphGroup syncs the values from the given aad group into kubeGroup
 // it returns a boolean "changed" that tells the call if an update is required
-func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName string, msGroupMembers []graphrbac.User) (*v1.Group, bool) {
+func fromMSGraphGroup(log *logrus.Entry, userV1 userv1client.UserV1Interface, kubeGroup *v1.Group, kubeGroupName string, msGroupMembers []graphrbac.User) (*v1.Group, bool) {
 	var g *v1.Group
 	if kubeGroup == nil {
 		g = &v1.Group{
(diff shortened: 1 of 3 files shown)


mjudeikis:
This would fix guest account sync.
Anything else is too much intrusive and breaking changes.

We can't change claims, as it would require changing all `user` and `identity` objects on existing clusters.
Origin metada does not exist in AAD as per https://github.com/openshift/openshift-azure/issues/1766 debugging.

If we agree this is way forward, will add some unit tests to this. 
Or if you know a better way to fix this, let me know.

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +44,32 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
+// match it with already sign-ed in users and updates required metadata
+// to match external provider prefix.
+// Example: foo@bar.com external guest user in AAD after sign-in
+// would become live.com#foo@bar.com, where live.com# is the origin
+// Function would match user and convert foo@bar.com to live.com#foo@bar.com
+func reconcileGuestUsers(log *logrus.Entry, userV1 userv1client.UserV1Interface, user graphrbac.User) string {
+	// If AAD user is External type
+	// user.Mail - do not container ext reference
+	// user.MainNickmane - do have ext reference
+	if user.MailNickname != nil && strings.Contains(*user.MailNickname, "#EXT#") {
```
asalkeld: the above is always going to be true.

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +44,32 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
+		userList, err := userV1.Users().List(meta_v1.ListOptions{})
```
asalkeld: would be good to not list this so often

mjudeikis:
/hold
Need to optimise listing logic to be executed only once

asalkeld:
seems like an OK hack (minus the multiple user listing)

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +44,32 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
+			return *user.Mail
+		}
+		for _, usr := range userList.Items {
+			// if login name exist and contains email, we gonna use it as ref
+			if strings.Contains(usr.Name, *user.Mail) {
+				return usr.Name
+			}
+		}
+	}
+	return *user.Mail
```
asalkeld: return ""

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +44,32 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
+// Function would match user and convert foo@bar.com to live.com#foo@bar.com
+func reconcileGuestUsers(log *logrus.Entry, userV1 userv1client.UserV1Interface, user graphrbac.User) string {
+	// If AAD user is External type
+	// user.Mail - do not container ext reference
+	// user.MainNickmane - do have ext reference
+	if user.MailNickname != nil && strings.Contains(*user.MailNickname, "#EXT#") {
+		userList, err := userV1.Users().List(meta_v1.ListOptions{})
+		if err != nil {
+			log.Warnf("failed to list users, default to user.Main %s", err)
+			return *user.Mail
```
asalkeld: return ""

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -33,7 +35,7 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
 	g.Users = []string{}
 	for _, user := range msGroupMembers {
 		if user.UserType == graphrbac.Guest && user.Mail != nil {
-			g.Users = append(g.Users, *user.Mail) // doesn't include #EXT#
```
asalkeld: foo := reconcileGuestUsers(log, userV1, user)
if foo != "" {
g.Users = append(g.Users, foo)
}
This will prevent useless entries in the group

mjudeikis: I think the current logic is fine. we still want the record to be added if we reached that point of code. Its just question of "if it was a prefix or no". Or I am missing something?

ehashman: We don't validate within `reconcileGuestUsers` that `*AADUser.Mail` is non-empty so I think Angus has a decent point here. If you want to golf it even more I think you can write:

```go
if reconciled := reconcileGuestUsers(log, userV1, user); reconciled != "" {
    g.Users = append(g.Users, reconciled)
}
```

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +50,27 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
 
+// reconcileGuestUsers will take an external user reference from AAD
+// match it with already sign-ed in users and updates required metadata
+// to match external provider prefix.
+// Example: foo@bar.com external guest user in AAD after sign-in
+// would become live.com#foo@bar.com, where live.com# is the origin
+// Function would match user and convert foo@bar.com to live.com#foo@bar.com
+func reconcileGuestUsers(log *logrus.Entry, ocpUserList []v1.User, AADUser graphrbac.User) string {
+	// If AAD user is External type
+	// AADUser.Mail - do not container ext reference
```
ehashman:
```suggestion
	// AADUser.Mail - does not contain ext reference
```

pkg/controllers/customeradmin/group_mapping.go
```diff
@@ -42,6 +50,27 @@ func fromMSGraphGroup(log *logrus.Entry, kubeGroup *v1.Group, kubeGroupName stri
...
+	// usAADUserer.MainNickmane - do have ext reference
```
ehashman:
```suggestion
	// AADUser.MailNickname - does have ext reference
```

pkg/controllers/customeradmin/group_mapping.go
asalkeld: > I think the current logic is fine. we still want the record to be added if we reached that point of code. Its just question of "if it was a prefix or no". Or I am missing something?

user.Mail as a member of the group does nothing, correct? That's why we are fixing this bug...
I am just suggesting not to add it as it just clutters up the group list and will be confusing.
So if we have a group with lots of guests, but they haven't logged in we will have lots of these non-functional members (and they don't get cleaned up when the user does login).
Then only useful entry is the new one that you are adding "usr.Name" (which includes the foo.com#). Anyways, that is my understanding..

mjudeikis: We have a different type of guest users. in example, `@microsfot.com` email does not have any prefix. I assume MSFT trust their own accounts. I suspect similar behaviour will be with live.com and outlook.com accounts. 

And I would like to avoid this "I signed in but still can't see projects, and they appeared after 1 minute". So if user configured 10 users in Group we move all 10 into the group. And just modify the group if some of those users appear to be different (with prefix) post login.

mjudeikis: Otherwise all this delays user sync. And not sure if this is what we would like to do?

asalkeld: ok understood @mjudeikis I thought all had a prefix. lgtm..

mjudeikis: not all. And even more fun. Guest accounts from @microsoft.com or @live.com in your domain tree does not have this..

Issue #1695, opened 2019-05-31
Timeout in e2e sanity check - checkCanCreateLB


charlesakalugwu:
Caught a flake with the following signature

```
• Failure [1374.652 seconds]
Scale Up/Down E2E tests [ScaleUpDown][Fake][EveryPR][LongRunning]
/home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/scaleupdown.go:27
  should be possible to maintain a healthy cluster after scaling it out and in [It]
  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/scaleupdown.go:84

  Expected
      <[]*standard.TestError | len:1, cap:1>: [
          {
              Bucket: "checkCanCreateLB",
              Err: {
                  s: "timed out waiting for the condition",
              },
          },
      ]
  to be empty

  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/scaleupdown.go:81
```

artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1694/pull-ci-azure-master-e2e-create/305/build-log.txt

/kind test-flake

m1kola:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1781/pull-ci-azure-master-e2e-create/510/build-log.txt in #1781

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1806/pull-ci-azure-master-upgrade-v5.1/454/build-log.txt #1806

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1810/pull-ci-azure-master-upgrade-v5.1/453/build-log.txt #1810

PR #1781, opened 2019-07-05
Normalise vmimage archive

m1kola:
This will hopefully allow consistent archive generation across different operating systems (MacOS with GNU utils and Linux based distros).

```release-note
NONE
```


m1kola:
upgrade-v5.2 flake #1400
/retest

m1kola:
upgrade-v5.2 flake #552
upgrade-v4.4 flake #1438
/retest

m1kola:
Ready for review.

Note for reviewers: CI didn't complain, but I would appreciate if you checkout this branch and run `make generate` on your local machine and see if `pkg/vmimage/data/archive.tgz` (and `pkg/vmimage/bindata.go` as a result) gets changed. If you see changes - please let know (including information about your distro). Thanks!

hack/build-archive.sh
```diff
@@ -2,10 +2,24 @@
...
-# RHEL tar doesn't include --sort=name, so have to use find
+# Normalise the archive accross different OS. We need consistent results in CI and in . Known issues:
+#  * RHEL `tar` doesn't include `--sort=name`, so have to use `find`
+#  * MacOS sets GUID different to 0 even if you sepecify `--group=root` in `tar`,
+#    so we have to explicitly set GUID to 0
+#  * MacOS sets different file mode for symlinks, so we have to explicitly set permissions for them
 
-find -not -path ./archive.tgz -and -not -path ./script.sh -print0 \
+TMP_TAR=$(mktemp)
+TAR_COMMON_OPTIONS="--no-recursion --null --files-from=- --owner=root --group=root --owner=0 --group=0 --mtime=@1546300800"
```
thekad: wouldn't leaving `--owner=0 --group=0` work for all platforms?

hack/build-archive.sh
```diff
@@ -2,10 +2,24 @@
 
 cd data
 
-# RHEL tar doesn't include --sort=name, so have to use find
+# Normalise the archive accross different OS. We need consistent results in CI and in . Known issues:
+#  * RHEL `tar` doesn't include `--sort=name`, so have to use `find`
+#  * MacOS sets GUID different to 0 even if you sepecify `--group=root` in `tar`,
```
thekad: "sepecify"

hack/build-archive.sh
m1kola: Tar stores [`uid` and `gid` as well as `uname` and `gname`](https://www.gnu.org/software/tar/manual/html_node/Standard.html). Only having `--owner=0 --group=0` without `--owner=root --group=root` will not work as expected on macOS: I tried that and got `wheel` in the `gname` field.

Despite the fact that I'm using GNU tar on macOS.

m1kola:
@thekad thanks for the review. I updated the PR fixing the typo + answering your question about `--owner=0 --group=0`.

> I kinda feel like forcing people to use GNU Tar would be less cumbersome

The idea is that we force macOS users to use GNU tools. This archive is not the only issue. There are several instances where scripts using `find` are incompatible with BSD version of `find` that macOS ships by default, for example. There was also something with `base64`. So making scripts compatible with macOS versions of tools will be hard. So for now, I think, we should force macOS users to use GNU tools.

In this case, I had to make changes because even with GNU tar on macOS I'm getting a result different from Linux with GNU tar (different permissions for symlinks and different `gid` gets saved into the archive). I guess it's something on the syscall level, so we have to explicitly set all the fields (https://github.com/openshift/openshift-azure/pull/1781#discussion_r300984980).

m1kola:
e2e-create flake #1695
/retest

m1kola:
e2e-create flake #552
upgrade-v5.1 flake #1410
/retest

m1kola:
upgrade-v5.2 flake #1628

/retest

thekad:
/test conformance
(don't mind me, testing an optional test suite)

m1kola:
I guess https://github.com/openshift/release/pull/4347 should fix the conformance tests so I'm re-running.

/test conformance

charlesakalugwu:
@m1kola thanks a lot for this! I tried it out and it works on my Fedora without issues. However I am worried that such modifications could crop up in the future and I am wondering how far you got with the idea of running Fedora on the mac?

m1kola:
@charlesakalugwu thanks for testing this.

> However I am worried that such modifications could crop up in the future

I haven't experienced any any other issues so far that require modifications in the project's scripts to make it consistent with other environments. Everything else (like differences in `find` and other tools) can be fixed on the machine by installing GNU versions of these tools.

Even this change is not a blocker: it's just annoying because I have to run `git checkout -- pkg/vmimage/bindata.go pkg/vmimage/data/archive.tgz` before creating a commit and I can't run `make verify` locally.

> I am wondering how far you got with the idea of running Fedora on the mac?

I made a quick research and it looks like there is no support for some of the hardware installed in MacBook Pro models released after 2017. I guess I'll have to switch to a different laptop.

ehashman:
/hold

Let's hold off on making these changes for now---I'm not sure if we have sufficient test coverage to be confident in this PR.

m1kola:
@ehashman I changed the way we build the `tar` archive and the order how files appear in the archive, but content (including permissions) is the same. I'll try to give more details below on my changes and explain the ways we can check the archive to make sure that everything is ok.

**TL;DR**: Resulting archive is the same, but symlinks are now at the end of the archive.

## Archive build process

Previously we were creating the archive in one go explicitly specifying group and user names, but without explicitly specifying group id and user id. After this PR the process of creating a `tar` archive will be:

* Create an archive with all files *except symlinks*
  * On this step we explicitly set group and user names as well as group and user ids. Linux sets 0 and 0 in group & user id fields, but macOS sets some nonsense
* Append  symlinks into the archive from the previous step
  * Here we also explicitly set group & user names and group & user ids.
  * Plus we explicitly set `rwx` permissions for all symlinks. Linux already sets `rwx` permissions implicitly, but macOS implicitly sets 775 for some reason. Permissions on symlinks [do not matter](http://man7.org/linux/man-pages/man7/symlink.7.html) on Linux systems (on macOS actually too) so it's only to make them consistent in the archive across OSes.

So we (hopefully) will not be getting diff on archive just because of permissions and group & user ids.

## Archive structure

About the archive structure. Previously structure of the archive was:

* Normal files mixed with symlinks (everything ordered by name together)

Now it is:

* Normal files in the beginning of the archive (ordered by name)
* Symlinks at the end of the archive (ordered by name)

## Testing

To make sure that the content of the archive didn't change one can [download the archive from the commit that comes before my PR](https://github.com/openshift/openshift-azure/raw/7089ac7bca39bbeb9e0d4c7ed0888b8e4a806b62/pkg/vmimage/data/archive.tgz) and [download the archive from my commit](https://github.com/openshift/openshift-azure/raw/cb80acf6c6c592d45a91f67c8ba219722a25921b/pkg/vmimage/data/archive.tgz).

Then take a look at the list of files both archives contains and compare: `tar -tf archive.tgz | sort | less` (it will be the same). Or without `sort` to see the order of how the files appear in the archives (will be different - symlinks at the end).
To go deeper and examine content and the structure of the archives: `zcat archive.tgz| less`. It's not very readable, but shows the content of the files, their permissions, user & group names and ids, etc. This [document](https://www.gnu.org/software/tar/manual/html_node/Standard.html) describes tar format and has order and offsets of fields in the archive. `dd` can can help to read a specific value. For example, this is a `mode` field of the first file: `zcat archive.tgz | dd bs=1 skip=100 count=8 status=none`.

While looking at `zcat` of both archives we can verify that the content is the same, permissions are the same, etc... But symlinks will be at the end of the archive.

Example for a symlink:
```
./etc/docker/certs.d/docker-registry.default.svc-5000/node-client-ca.crt^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@0000777^@0000000^@0000000^@00000000000^@13412526600^@032120^@ 2/etc/origin/node/client-ca.crt^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@ustar  ^@root^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@root
```

Example for a regular file
```
./etc/NetworkManager/dispatcher.d/99-origin-dns.sh^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@0000775^@0000000^@0000000^@00000012602^@13412526600^@020631^@ 0^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@ustar  ^@root^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@root
```

charlesakalugwu:
@m1kola ^^

m1kola:
Rebased

charlesakalugwu:
I've tested and checked this again and it looks ok to me
/lgtm

m1kola:
@mjudeikis @charlesakalugwu @ehashman thanks for looking into this. This PR is currently on hold. Can the hold be lifted?

PR #1783, opened 2019-07-08
Add project-request, self-provisioner, managed-resource capabilities

mjudeikis:
```release-note
Add the ability to exclude certain objects from azure controller reconciliation loop.
Add project-request template into openshift namespace and enable customer-admin to change it
Add ability to modify self-provisioner clusterrolebinding for cluster-admin. 
Add ability to modify shared resources in openshift namespace
```

* Deploy `project-request` into the `openshift` namespace (default location to keep aligned with the product).
* Add label `"azure.penshift.io/reconcile-protect"` to enable "deploy-once" behaviour so the customer can edit the template and we don't touch it consequence runs.
* Ability to modify cluster role allows changing role-binding from `system:authenticated` to anything else hence enabled them to deny all users to create projects. 

Some designs considerations for this change:
* I didn't wanted to add any external API changes, so we could decouple these features from MSFT needing to do any work
* most of openshift default cluster roles are protected with `openshift.io/reconcile-protect` - had to remove some protection to enable editing.
* Instead of trying to invent something new tried to keep existing OCP behaviour as close to the on-prem product as possible. 
* We dont have a clean solution to "create objects and let them to be deleted after that but still have them in sync DB". This is why I introduce new annotation, where it allows object deployment but ignores all future updates. If at any point in time we will need to update these objects, we will have to write custom logic to keep customer logic and update ours. 

Desing: https://docs.google.com/document/d/1nFq61Hdw8qXwXanLWJwEpXWFLtUvsuw4reZI4r0JghM/edit 

TODO:
* documentation on MSFT side will be needed in additional PR when we go live.


pkg/sync/v6/sync.go
```diff
@@ -43,6 +43,7 @@ const (
 	ownedBySyncPodLabelKey            = "azure.openshift.io/owned-by-sync-pod"
 	syncPodWaitForReadinessLabelKey   = "azure.openshift.io/sync-pod-wait-for-readiness"
 	syncPodReadinessPathAnnotationKey = "azure.openshift.io/sync-pod-readiness-path"
+	deployOnceSyncPodLabelKey         = "azure.openshift.io/deploy-once"
```
jim-minter: it's an AnnotationKey

pkg/sync/v6/client.go
```diff
@@ -233,6 +233,12 @@ func needsUpdate(log *logrus.Entry, existing, o *unstructured.Unstructured) bool
 		return false
 	}
 
+	// check if object is marked deploy-only and deploy once
+	if o.GetAnnotations()[deployOnceSyncPodLabelKey] != "" {
```
jim-minter: can you have a think now about how this mechanism could be improved to in the future support objects that we might allow the end user to delete - e.g. templates in the openshift namespace?  If we can do that now, it might be useful.

pkg/sync/v6/data/ClusterRole.rbac.authorization.k8s.io/self-provisioner.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   annotations:
     authorization.openshift.io/system-only: "true"
     openshift.io/description: A user that can request projects.
+    openshift.io/reconcile-protect: "true"
```
jim-minter: why this change, out of interest?

pkg/sync/v6/sync.go
jim-minter: the deploy-once name is somewhat edge-triggered - is it possible to find an level-triggered name?

pkg/sync/v6/data/ClusterRole.rbac.authorization.k8s.io/self-provisioner.yaml
mjudeikis: Because internal OCP reconcile loop keeps defaulting them back :/ we need to remove it from the reconcide loop 
https://access.redhat.com/solutions/3304341

pkg/sync/v6/client.go
mjudeikis: Azure Arch document already has topic for this.

pkg/sync/v6/sync.go
mjudeikis: I might need an explanation what do you have in mind?

mjudeikis:
We still have a question about to enable the customer to manage their `openshift` namespace.

A suggestion would be:
* enable the customer to patch `openshift` namespace so they could add same annotation `azure.openshift.io/reconcile-protect: "true"`.
* if label is added we stop updating anything in that namespace.

mjudeikis:
/hold 
need some minor change

pkg/sync/v6/sync.go
```diff
@@ -329,6 +338,13 @@ func (s *sync) writeDB() error {
 	if err := s.applyResources(cfgFilter, keys); err != nil {
 		return err
 	}
+	// if SharedResources are managed - update/create
+	if s.config.managedSharedResource {
+		s.log.Debug("applying storageClass resources")
```
ehashman: This was copied from below I think, has nothing to do with storageClass though.

```suggestion
		s.log.Debug("applying managedShared resource")
```

pkg/sync/v6/client.go
```diff
@@ -254,3 +259,24 @@ func printDiff(log *logrus.Entry, existing, o *unstructured.Unstructured) bool {
 	}
 	return diffShown
 }
+
+// reconcileProtected check if object is marked with azure reconcile protection annotation
+// this will remove certain objects from being updated post cluster creation
+// access to modify these objects is limited by RBAC
+func (s *sync) reconcileProtected(existing, o *unstructured.Unstructured) bool {
+	if o.GetAnnotations()[reconcileProtectKeyAnnotationKey] != "" {
```
ehashman: If the value is not null but e.g. "false", this will still kick in.. do we want to explicitly check that it's true?

pkg/sync/v6/sync.go
```diff
@@ -481,6 +497,19 @@ type sync struct {
...
 	grs        []*discovery.APIGroupResources
+
+	config config
+}
+
+// config defines sync pod behaviour based on conditional logic in the cluster.
+// if set to false - components becomes unmanaged.
+// we need to make sure this for the nested dependencies (example: resource in the namespace)
+// we set these flags right time in the sync process
+type config struct {
```
ehashman: `config` seems like much too generic of a name to me, especially when we're including it in a struct that has a number of other different configs already. Can we give this a more targeted name?

test/e2e/specs/customer_admin.go
```diff
@@ -81,6 +81,50 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
+
+		crb.Subjects = subject
+
+		_, err = sanity.Checker.Client.CustomerAdmin.RbacV1.ClusterRoleBindings().Update(crb)
+		Expect(err).ToNot(HaveOccurred())
+
+		// HACK: openshift reconciler should not touch this role because of:
+		// openshift.io/reconcile-protect: "true"
+		// we need to confirm this
+		time.Sleep(5 * time.Second)
```
ehashman: Why is this sleep necessary? The comment doesn't make this any clearer to me.

test/e2e/specs/customer_admin.go
```diff
@@ -81,6 +81,50 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
+	})
+
+	It("should be able to edit self-provisioners clusterrolebinding", func() {
+		crbOriginal, err := sanity.Checker.Client.CustomerAdmin.RbacV1.ClusterRoleBindings().Get("self-provisioners", metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		crb := crbOriginal.DeepCopy()
+
+		// create ClusterRoleBinding which would allow us to test edit
+		// and would not break any other tests
+		subject := []rbacv1.Subject{
```
ehashman:
```suggestion
		subjects := []rbacv1.Subject{
```

test/e2e/specs/customer_admin.go
```diff
@@ -81,6 +81,50 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
+		Expect(err).ToNot(HaveOccurred())
+
+		// HACK: openshift reconciler should not touch this role because of:
+		// openshift.io/reconcile-protect: "true"
+		// we need to confirm this
+		time.Sleep(5 * time.Second)
+
+		crbUpdated, err := sanity.Checker.Client.CustomerAdmin.RbacV1.ClusterRoleBindings().Get("self-provisioners", metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(crbUpdated.Subjects)).To(Equal(len(crb.Subjects)))
```
ehashman: Do we want to check them one by one for equality? We didn't actually demonstrate these lengths weren't equal before, so it's possible this test is passing by coincidence.

pkg/sync/v6/sync.go
mjudeikis: reconcileConfig

test/e2e/specs/customer_admin.go
mjudeikis: This cole by default is managed by Openshift core rbac reconciler loop. we exclude it. This gives time for the controller to react and check if "it needs to change it". Our expectation is that we gonna manage it from now on. So we just checking if it does not touch it.

test/e2e/specs/customer_admin.go
mjudeikis: This makes sure this flow works:
	// Flow:
	// 1. Read rolebinding with default system:authenticated:oauth value (single value by default)
	// 2. Update it with 2 values (group + system:authenticated:oauth) and our reconcile annotation
	// 3. Commit to API
	// 4. Confirm what we committed is still in api/etcd

pkg/sync/v6/client.go
mjudeikis: done

mjudeikis:
Added some testing notes into https://docs.google.com/document/d/1nFq61Hdw8qXwXanLWJwEpXWFLtUvsuw4reZI4r0JghM/edit#

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
```
jim-minter: does this actually do anything?

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
```diff
@@ -0,0 +1,50 @@
...
+  verbs:
+  - update
+  - patch
+  - delete
+  - create
+- apiGroups:
+  - ""
+  - image.openshift.io
+  resources:
+  - imagestreams/layers
```
jim-minter: nit: can be rolled into stanza above?

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - create
+- apiGroups:
+  - ""
+  - project.openshift.io
+  resources:
+  - projects
+  - namespaces
+  verbs:
+  - get
+  - update
```
jim-minter: security: are you sure that this doesn't allow a customer admin to patch an arbitrary `openshift-*` project?  Probably an e2e to prove it would be a good plan.

pkg/sync/v6/client.go
```diff
@@ -254,3 +259,24 @@ func printDiff(log *logrus.Entry, existing, o *unstructured.Unstructured) bool {
...
+	if strings.ToLower(o.GetAnnotations()[reconcileProtectKeyAnnotationKey]) == "true" {
+		s.log.Infof("Sync reconcile protection detected for [%s/%s], ignore update", o.GetKind(), o.GetName())
+		return true
+	}
+	if strings.ToLower(existing.GetAnnotations()[reconcileProtectKeyAnnotationKey]) == "true" {
+		s.log.Infof("User reconcile protection detected for [%s/%s], ignore update", o.GetKind(), o.GetName())
+		// check sync pod managed/unmanaged dependencies
+		if (existing.GroupVersionKind().GroupKind() == schema.GroupKind{Kind: "Namespace"} &&
+			existing.GetName() == "openshift") {
+			s.reconcileConfig.managedSharedResource = false
```
jim-minter: a) the lifecycle of s.reconcileConfig.managedSharedResource is going to confuse someone in the future.  Can you just explicitly get the namespace and inspect it in writeDB() and then decide whether to call applyResources(sharedResourceFilter) as a result?  That saves the pseudo-global spaghetti.
b) I think I'd rather see 2 annotations - one which prevents reconciliation of a given object, and a second one which is applied to the openshift project which prevents reconciliation of all objects within.  I don't think it makes sense to condense these into one, but am open to being persuaded.

pkg/sync/v6/sync.go
```diff
@@ -43,6 +43,7 @@ const (
 	ownedBySyncPodLabelKey            = "azure.openshift.io/owned-by-sync-pod"
 	syncPodWaitForReadinessLabelKey   = "azure.openshift.io/sync-pod-wait-for-readiness"
 	syncPodReadinessPathAnnotationKey = "azure.openshift.io/sync-pod-readiness-path"
+	reconcileProtectKeyAnnotationKey  = "azure.openshift.io/reconcile-protect"
```
jim-minter: prefer name reconcileProtectAnnotationKey

pkg/sync/v6/sync.go
```diff
@@ -275,20 +276,25 @@ var (
 		return o.GroupVersionKind().GroupKind() == schema.GroupKind{Kind: "Secret"} ||
 			o.GroupVersionKind().GroupKind() == schema.GroupKind{Kind: "ConfigMap"}
 	}
+	sharedResourceFilter = func(o unstructured.Unstructured) bool {
+		// if Template OR ImageStream in openshift namespace
+		return o.GetNamespace() == "openshift" &&
+			(o.GroupVersionKind().GroupKind() == schema.GroupKind{Group: "image.openshift.io", Kind: "ImageStream"} ||
```
jim-minter: are you sure we don't need `schema.GroupKind{Kind: "ImageStream"} and schema.GroupKind{Kind: "Template"}` as well?

pkg/sync/v6/client.go
```diff
@@ -254,3 +259,24 @@ func printDiff(log *logrus.Entry, existing, o *unstructured.Unstructured) bool {
 	}
 	return diffShown
 }
+
+// reconcileProtected check if object is marked with azure reconcile protection annotation
+// this will remove certain objects from being updated post cluster creation
+// access to modify these objects is limited by RBAC
+func (s *sync) reconcileProtected(existing, o *unstructured.Unstructured) bool {
+	if strings.ToLower(o.GetAnnotations()[reconcileProtectKeyAnnotationKey]) == "true" {
```
jim-minter: isn't this a logic error?  Isn't the point of reconcileProtectKeyAnnotationKey that it's in the user's control?  We wouldn't ever want to see it in our yaml files?

pkg/sync/v6/data/ClusterRole.rbac.authorization.k8s.io/self-provisioner.yaml
jim-minter: still don't get it - is this happening on clusters today?  If not, what changes here to make it a problem?

test/e2e/specs/customer_admin.go
jim-minter: +1 with @ehashman - I thought this reconcile happened at apiserver start?  Do you see an issue without this sleep?

test/e2e/specs/customer_admin.go
```diff
@@ -81,6 +81,57 @@ var _ = Describe("Openshift on Azure customer-admin e2e tests [CustomerAdmin][Fa
...
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	// Flow:
+	// 1. Read rolebinding with default system:authenticated:oauth value
+	// 2. Update it with 2 values (group + system:authenticated:oauth)
+	// 3. Commit to API
+	// 4. Confirm what we committed is still in api/etcd
+	It("should be able to edit self-provisioners clusterrolebinding", func() {
+		crbOriginal, err := sanity.Checker.Client.CustomerAdmin.RbacV1.ClusterRoleBindings().Get("self-provisioners", metav1.GetOptions{})
```
jim-minter: I am trying to understand if there is a scenario in which customer admin can lock themselves out.  If customer admin removes everyone from self-provisioner, is it able to add anyone back in?

pkg/sync/v6/client.go
mjudeikis: It's an open question. We dont use it now, but I see potential use-cases in the future. But we can add them as we add the use-cases :)

pkg/sync/v6/client.go
mjudeikis: a) It makes logic little bit complex too, as now before applying `ImageStream` or `Template` I have to inspect each object to check which namespace it is in. So It can be done, but it will be case where we have to inspect all objects as as much as `sync` care its just `unstructured`. Im ok to change this. It looked cleaner this way in my head :)
b) I don't want to create 2 annotations as this one just does one thing. And in theory, it is the same thing. :man_shrugging:

pkg/sync/v6/data/ClusterRole.rbac.authorization.k8s.io/self-provisioner.yaml
mjudeikis: yes, it had to be on ClusterRoleBinding only. Not here. but it was 2 weeks when I was working with it. WIll retest and remove if this is a case.

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
mjudeikis: no

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
mjudeikis: Yes,  but its role and its assigned only to Openshift namespace. I think we would be testing RBAC here. But ok if we want to be sure our config is right.

pkg/sync/v6/sync.go
mjudeikis: all the current object complies with the current one. But need to confirm,

test/e2e/specs/customer_admin.go
mjudeikis: I see that sometimes our process runs faster than apiserver reacts to `watch` event and patches/not patches the object. In theroy we could read the object and see - "all good, this is what we submitted in" but in next half-second reconcile runs and changes it to default back. This  is my understanding of the process how it work (controller and reconcile loop). If this is not a case - I need to update my knowledge :)

pkg/sync/v6/data/Role.rbac.authorization.k8s.io/openshift/shared-resource-editor.yaml
mjudeikis: TODO: Write e2e test for config

mjudeikis:
TODO: move to `openshift.io/reconcile-protect: "true"`

mjudeikis:
TODO: Add sync pod to check annotation only on self-provisioner rolebinding and openshift namespace template imagestreams

pkg/sync/v6/sync.go
```diff
@@ -302,6 +310,20 @@ var (
...
+	// Reconcile protection logic is reconcileProtected func,
+	// but we check if openshift namespace has protection annotation
+	// to check if we need to apply sharedResourceFilter or not
+	// default - resources are managed
+	managedSharedResources := true
+	namespace, err := s.kc.CoreV1().Namespaces().Get("openshift", metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	if strings.ToLower(namespace.GetAnnotations()[reconcileProtectAnnotationKey]) == "true" {
```
jim-minter: are you missing a `namespace != nil &&` here?

mjudeikis: `openshift` namespace exists with the start of static pods. But we can add something like that just to be sure

ehashman:
/retest

quota flake

There's a bunch of dead jobs from this PR but I dunno which is which... perhaps we need to go purge them out a little early.

ehashman:
/retest

flake #552

Issue #1778, opened 2019-07-03
Sync e2e fails with unexpected log output


charlesakalugwu:
Caught a flake where the sync pod e2e test failed due to an expected entry not being found in the 
sync reconciliation logs.

artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1760/pull-ci-azure-master-e2e-create/479/build-log.txt

/kind test-flake

charlesakalugwu:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1838/pull-ci-azure-master-e2e-create/767 in #1838

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1836/pull-ci-azure-master-e2e-create/772/build-log.txt in #1836

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1863/pull-ci-azure-master-e2e-create/870/build-log.txt in #1863

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1871/pull-ci-azure-master-e2e-create/876/build-log.txt in #1871

asalkeld:
All the errored clusters have 
```
time=\"2019-07-03T08:29:45Z\" level=debug msg=\"Service catalog api is aggregated\"
time=\"2019-07-03T08:29:45Z\" level=debug msg=\"applying sc resources\"
time=\"2019-07-03T08:29:45Z\" level=info msg=\"sync error: couldn't find kind ClusterServiceBroker\"
time=\"2019-07-03T08:31:58Z\" level=info msg=\"starting sync
```
sync then runs again and completes

asalkeld:
So update on what is happening with this:

It always has updates in it, but the test only fails when there is an error as then there is one sync run.

This code below only works when the sync pod has done one run
https://github.com/openshift/openshift-azure/blob/master/test/e2e/specs/fakerp/sync.go#L27

If I change it to the below, it correctly fails all the time. (it can be seen by looking at the output that there are quite a lot of updates)
```go
		for _, run := range runs {
			Expect(run).To(Not(ContainSubstring("Update")))
		}
```

some examples of Updates

[...]

PR #1787, opened 2019-07-12
Rapid mitigation of CVEs with local yum updates

charlesakalugwu:
```release-note
Add support for cluster hot patches in response to CVEs
```


charlesakalugwu:
e2e-create flake #1778 
upgrade-v5.2 flake #1628 
/retest

pkg/arm/v6/data/master-startup.sh
```diff
@@ -1,5 +1,53 @@
...
+EOF
+cat > /etc/yum.repos.d/kickstart.repo <<'EOF'
+[rhel-7-server-rpms]
+name=Red Hat Enterprise Linux 7 Server (RPMs)
+baseurl=https://cdn.redhat.com/content/dist/rhel/server/7/7Server/$basearch/os
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslcacert=/etc/rhsm/ca/redhat-uep.pem
+sslclientcert=/var/lib/yum/client-cert.pem
+sslclientkey=/var/lib/yum/client-key.pem
+enabled=yes
```
mjudeikis: Should we keep them disabled by default and as part of the script enable? Just thinking...

Makefile
```diff
@@ -78,5 +78,8 @@ unit: generate testinsights
 e2e:
 	FOCUS="\[CustomerAdmin\]|\[EndUser\]\[Fake\]" TIMEOUT=60m ./hack/e2e.sh
 
+cvepatch:
```
charlesakalugwu: this will go away. it is here for now to allow me easily run the e2e tests for this particular piece of work. Triggering the job should be moved to the `release` repo

pkg/api/admin/config.go
```diff
@@ -8,6 +8,9 @@ import (
 
 // Config holds the cluster admin config structure
 type Config struct {
+	// SecurityPatchPackages defines a list of rpm packages for rapid cve mitigation
```
charlesakalugwu:
```suggestion
	// SecurityPatchPackages defines a list of rpm packages that fix security issues
```

pkg/api/admin/config.go
```diff
@@ -142,6 +145,9 @@ type CertificateConfig struct {
 	// geneva integration certificates
 	GenevaLogging *Certificate `json:"genevaLogging,omitempty"`
 	GenevaMetrics *Certificate `json:"genevaMetrics,omitempty"`
+
+	// package repository cdn client certificates
```
charlesakalugwu:
```suggestion
	// red hat cdn client certificates
```

pkg/fakerp/ssh.go
```diff
@@ -113,28 +112,27 @@ func (s *ssher) Dial(ctx context.Context, hostname string) (*ssh.Client, error)
 }
 
 func (s *ssher) RunRemoteCommandAndSaveToFile(client *ssh.Client, cmd, filename string) error {
-	f, err := os.Create(filename)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
+	result, err := s.RunRemoteCommand(client, cmd)
+	err = ioutil.WriteFile(filename, result, 0666)
```
charlesakalugwu:
```suggestion
	return ioutil.WriteFile(filename, result, 0666)
```

pkg/fakerp/ssh.go
```diff
@@ -113,28 +112,27 @@ func (s *ssher) Dial(ctx context.Context, hostname string) (*ssh.Client, error)
...
+	return err
```
charlesakalugwu:
```suggestion

```

pkg/vmimage/data/script.sh
```diff
@@ -174,7 +174,7 @@ EOF
 # not commited with a : so that Windows users can check out the repo
 mv /etc/docker/certs.d/docker-registry.default.svc-5000 /etc/docker/certs.d/docker-registry.default.svc:5000
 
-rm /etc/yum.repos.d/kickstart.repo /var/lib/yum/client-cert.pem /var/lib/yum/client-key.pem
+rm /var/lib/yum/client-cert.pem /var/lib/yum/client-key.pem
```
charlesakalugwu: we avoid deleting `kickstart.repo` because we want it baked into the image. We delete the accompanying tls keypair nevertheless.

pkg/vmimage/data/etc/yum.repos.d/kickstart.repo
```diff
@@ -7,6 +8,7 @@ sslclientkey=/var/lib/yum/client-key.pem
 enabled=yes
```
charlesakalugwu: it might be worth setting enabled=no and then enable them just in time in the scripts. Thanks @mjudeikis  :)

test/e2e/specs/fakerp/cvepatch.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+	"github.com/openshift/openshift-azure/test/sanity"
+	logger "github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = Describe("Apply CVE hot patch E2E tests [ApplyHotPatch][Fake][LongRunning]", func() {
+	var (
+		ctx       = context.Background()
+		patchRpms = []string{
```
charlesakalugwu: this list of rpms works for the checks in this test however I feel the test will break if this is run at a later date against a cluster with VMs which have been built with those rpms already in place.

test/e2e/specs/fakerp/cvepatch.go
```diff
@@ -0,0 +1,112 @@
...
+			"python-libs-2.7.5-80.el7_6.x86_64",
+			"python-2.7.5-80.el7_6.x86_64",
+			"vim-minimal-7.4.160-6.el7_6.x86_64",
+		}
+		log = logger.GetTestLogger()
+		// Do not use the internal config in other tests! This is necessary here in order to acquire
+		// ssh credentials which will be used to query the sample vm for rpm lists before and
+		// after cve hot patches are applied to a cluster. Usually fakerp tests should be written
+		// to use the admin config wherever possible.
+		internalConfig = func() (*api.OpenShiftManagedCluster, error) {
```
charlesakalugwu: haven't found a better way to grab the list of rpms before and after the hot patch as well as grabbing logs. This is why I resorted to this antipattern of using the internal config in an e2e test. Ideas for improvement welcome.

charlesakalugwu:
upgrade-v5.1 flake #1628
/retest

test/e2e/specs/fakerp/cvepatch.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/openshift/openshift-azure/pkg/api"
+	"github.com/openshift/openshift-azure/pkg/fakerp"
+	"github.com/openshift/openshift-azure/pkg/util/azureclient"
+	"github.com/openshift/openshift-azure/test/clients/azure"
+	"github.com/openshift/openshift-azure/test/sanity"
+	logger "github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = Describe("Apply CVE hot patch E2E tests [ApplyHotPatch][Fake][LongRunning]", func() {
```
charlesakalugwu:
```suggestion
var _ = Describe("Apply security patch E2E tests [ApplySecurityPatch][Fake][LongRunning]", func() {
```

pkg/api/validate/admin.go
```diff
@@ -78,6 +78,7 @@ func (v *AdminAPIValidator) validateUpdateContainerService(cs, oldCs *api.OpenSh
 	old.Config.ImageVersion = cs.Config.ImageVersion
 
 	old.Config.Images = cs.Config.Images
+	old.Config.SecurityPatchPackages = cs.Config.SecurityPatchPackages
```
jim-minter: will need to add suitable validation for SecurityPatchPackages

charlesakalugwu:
*Sample logs*
```
[cloud-user@master-000000 ~]$ sudo journalctl -t master-startup.sh -t node-startup.sh 
-- Logs begin at Tue 2019-07-09 11:40:41 UTC, end at Tue 2019-07-09 12:20:21 UTC. --
Jul 09 11:41:11 master-000000 master-startup.sh[8637]: installing red hat cdn configuration on master-000000
Jul 09 11:45:47 master-000000 master-startup.sh[11172]: installing python-libs-2.7.5-80.el7_6.x86_64 on master-000000
Jul 09 11:46:08 master-000000 master-startup.sh[11196]: installing python-2.7.5-80.el7_6.x86_64 on master-000000
Jul 09 11:46:09 master-000000 master-startup.sh[11206]: installing vim-minimal-7.4.160-6.el7_6.x86_64 on master-000000
Jul 09 11:46:16 master-000000 master-startup.sh[11217]: removing red hat cdn configuration on master-000000
```

pkg/arm/v6/data/master-startup.sh
```diff
@@ -1,5 +1,54 @@
...
+[rhel-7-server-ose-3.11-rpms]
+name=Red Hat OpenShift Container Platform 3.11 (RPMs)
+baseurl=https://cdn.redhat.com/content/dist/rhel/server/7/7Server/$basearch/ose/3.11/os
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslcacert=/etc/rhsm/ca/redhat-uep.pem
+sslclientcert=/var/lib/yum/client-cert.pem
+sslclientkey=/var/lib/yum/client-key.pem
+enabled=yes
+EOF
+yum makecache
```
jim-minter: why is yum makecache needed?

pkg/arm/v6/data/master-startup.sh
```diff
@@ -1,5 +1,54 @@
...
+
+{{ range .Config.SecurityPatchPackages }}
+logger -t master-startup.sh "installing {{.}} on $(hostname)"
+yum install -y {{.}}
```
jim-minter: need to confirm install/update @ehashman 
I think it's preferable to construct a single yum command line and run it once rather than range

pkg/arm/v6/data/master-startup.sh
```diff
@@ -1,5 +1,54 @@
...
+{{end}}
```
jim-minter: need to see the code that will handle reboots.  Also close on error handling/reporting.

pkg/arm/v6/data/master-startup.sh
```diff
@@ -1,5 +1,54 @@
 #!/bin/bash -ex
 
+{{ if gt (len .Config.SecurityPatchPackages) 0 }}
```
jim-minter: fwiw you can say `{{ if .Config.SecurityPatchPackages }}` - see https://golang.org/pkg/text/template/#hdr-Actions

test/e2e/specs/fakerp/cvepatch.go
jim-minter: Part of the problem may coincide with needing to improve the error handling/RPM version reporting side of this.

pkg/arm/v6/data/master-startup.sh
ehashman: I think we want to use `yum update`.

Also agree that it's preferable to specify this as a single yum command with one package rather than a range---less chance for failure and lets the yum dep resolver run just once.

pkg/arm/v6/data/master-startup.sh
charlesakalugwu: yep this is not really needed

pkg/arm/v6/data/master-startup.sh
charlesakalugwu: I think i'm going to use `yum install` here because:
- yum is smart enough that when you call install for a package which is already on the system, an update is done
```
[root@master-000001 cloud-user]# yum install -y python-libs-2.7.5-80.el7_6.x86_64
...
Updated:
  python-libs.x86_64 0:2.7.5-80.el7_6 

Dependency Updated: 
  python.x86_64 0:2.7.5-80.el7_6  

Complete!
```
- if you run `yum update` on a non-existent package, exit code 0 is returned. `yum install` returns the expected (imo) exit code 1 and this will make error handling more straightforward.
```
[root@master-000001 cloud-user]# yum update -y -q hell
No Match for argument: hell
[root@master-000001 cloud-user]# echo $?
0
[root@master-000001 cloud-user]# yum install -y -q hell
Error: Nothing to do
[root@master-000001 cloud-user]# echo $?
1
```
I'll join the packages and install them with a single `yum install`. Also adding a `--quiet` flag to the command.

charlesakalugwu:
upgrade-v5.1 flake #1628 
/retest

charlesakalugwu:
upgrade-v5.1 flake #1628
/retest

pkg/arm/v6/data/master-startup.sh
charlesakalugwu: added code to handle reboots and retries to the `yum install` command. Note that the reboot is handled in a trap. This is because if the reboot is done immediately when we determine it is needed then the rest of the waagent script will not be run and the cluster wont come up. Another option is to move the security updates code to the bottom of the script however I think updates should be applied to the system before bring up the control plane.

pkg/api/validate/admin.go
charlesakalugwu: I added an initial rudimentary check that the supplied rpm name doesn't end with `.rpm`. I also probably still need to check that the names match the format described here https://fedoraproject.org/wiki/Packaging:Naming?rd=Packaging:NamingGuidelines#Common_Character_Set_for_Package_Naming

pkg/arm/v6/data/master-startup.sh
ehashman: The risk of using `yum install` is that if a package is specified that isn't already installed on the system, it will go ahead and install that package, which is not a risk with `yum update`.

I don't think it's necessarily a problem that yum update returns exit code 0 on a nonexistent package, but would using `yum update-to` here instead fix that?

charlesakalugwu:
e2e-create flake #1787 
/retest

charlesakalugwu:
upgrade-v5.2 flake #1643 
/retest

charlesakalugwu:
upgrade-v5.2 flake #1438 
/retest

pkg/arm/v6/data/master-startup.sh
```diff
@@ -7,6 +7,62 @@
...
+baseurl=https://cdn.redhat.com/content/dist/rhel/server/7/7Server/$basearch/ose/3.11/os
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslcacert=/etc/rhsm/ca/redhat-uep.pem
+sslclientcert=/var/lib/yum/client-cert.pem
+sslclientkey=/var/lib/yum/client-key.pem
+enabled=yes
+EOF
+
+logger -t master-startup.sh "installing ARO security updates [{{ StringsJoin .Config.SecurityPatchPackages ", " }}] on $(hostname)"
+while ! yum install -y -q {{ StringsJoin .Config.SecurityPatchPackages " " }}; do
```
jim-minter: I don't think the infinite loop here is a good plan.  Suggest trying N (<= 5?) times and failing the script after that

pkg/arm/v6/data/master-startup.sh
```diff
@@ -7,6 +7,62 @@
...
+done
+
+logger -t master-startup.sh "removing red hat cdn configuration on $(hostname)"
+yum clean all
+rm -rf /var/lib/yum/client-cert.pem /var/lib/yum/client-key.pem
+
+if ! needs-restarting -r &>/dev/null; then
+  # reboot is done with a trap. waagent doesn't run on reboots therefore
+  # all instructions below will never get executed if the reboot is immediate
+  trap 'logger -t master-startup.sh "rebooting $(hostname) to complete ARO security updates"; reboot --reboot now' EXIT
```
jim-minter: 1) trap is a nice idea, but I think it would be better to reboot before `systemctl start atomic-openshift-node.service` to save the race.  Reorder the enables/starts at the end of the list, and then reboot if necessary?
2) I'm pretty tempted to remove the dependency on `needs-restarting` and always reboot `{{ if .Config.SecurityPatchPackages }}`

charlesakalugwu: 1. Yep this works too and saves the race. I'll do that.
2. Why exactly?

pkg/arm/v6/data/master-startup.sh
charlesakalugwu: i figured as long as red hat CDN isn't down for obscene amounts of time then it wasn't a problem. I'll limit retries to 5.

charlesakalugwu:
e2e-create flake #552 
/retest

ehashman:
(I don't think either of my comments are blocking fwiw.)

ehashman:
Hm, the master updates timed out after 2h.  Not sure if this is a flake or a genuine failed on the applysecurityupdates test.

test/e2e/specs/fakerp/securityupdates.go
```diff
@@ -0,0 +1,103 @@
...
+	logger "github.com/openshift/openshift-azure/test/util/log"
+)
+
+var _ = Describe("Apply security updates E2E tests [ApplySecurityUpdates][Fake][LongRunning]", func() {
+	var (
+		ctx                    = context.Background()
+		securityUpdatePackages = []string{
+			"python-libs-2.7.5-80.el7_6.x86_64",
+			"python-2.7.5-80.el7_6.x86_64",
+			"vim-minimal-7.4.160-6.el7_6.x86_64",
```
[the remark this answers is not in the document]
charlesakalugwu: yes I'll change these to 

```
"nano-2.3.1-10.el7.x86_64",
"wget-1.14-18.el7_6.1.x86_64",
```

pkg/api/validate/plugin.go
```diff
@@ -90,6 +95,17 @@ func (v *PluginAPIValidator) Validate(c *pluginapi.Config) (errs []error) {
 	return
 }
 
+func validateSecurityPatchPackage(name string) (errs []error) {
```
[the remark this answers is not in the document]
charlesakalugwu: sounds good...will get it in

charlesakalugwu:
> Hm, the master updates timed out after 2h. Not sure if this is a flake or a genuine failed on the applysecurityupdates test.

@ehashman the problem was that in my attempt to move the post-update restart to the end of the startup scripts, the statement to start `atomic-openshift-node` wasn't being called if `SecurityPatchPackages` was set therefore the control plane couldn't come up and the test got stuck waiting for `master-000000`. See [here](https://github.com/openshift/openshift-azure/blob/7f8d1f19ae5a7e993b751e9a55e4ad252e23fb8a/pkg/arm/v6/data/master-startup.sh). I've fixed the issue in the latest commits.

pkg/api/validate/plugin.go
```diff
@@ -90,6 +91,16 @@ func (v *PluginAPIValidator) Validate(c *pluginapi.Config) (errs []error) {
 	return
 }
 
+func validateSecurityPatchPackages(names []string) (errs []error) {
```
thekad: i hate named return values, but that's a personal pref only :man_shrugging:

pkg/api/admin/config.go
```diff
@@ -8,6 +8,9 @@ import (
 
 // Config holds the cluster admin config structure
 type Config struct {
+	// SecurityPatchPackages defines a list of rpm packages that fix security issues
+	SecurityPatchPackages *[]string `json:"securityPatchPackages,omitempty"`
```
thekad: I'd prefer CVEPackages, they can be installed with `yum update --cve` and etc. so the name kinda has existing history.

pkg/api/validate/plugin.go
charlesakalugwu: I just followed the function naming convention in that file. If others feel as strongly about this as you do then we could  discuss it and cook it up in a follow-up. What do you think?

charlesakalugwu:
e2e-create failure is due to quota issues.

ehashman:
/retest

quota

ehashman:
/retest

still quota :'(

Issue #206, opened 2018-08-15
remove dep https://github.com/go-test/deep


mjudeikis:
`k8s.io/apimachinery/pkg/equality.Semantic.DeepEqual` and `k8s.io/apimachinery/pkg/util/diff.ObjectGoPrintDiff` should be used instead of https://github.com/go-test/deep

kargakis:
More like cleanup

mjudeikis:
close in favor of AZURE-151

PR #1792, opened 2019-07-15
Replaces github.com/go-test/deep with github.com/google/go-cmp

m1kola:
The initial plan was to switch to [`diff.ObjectGoPrintDiff` from `k8s.io/apimachinery`](https://godoc.org/k8s.io/apimachinery/pkg/util/diff#ObjectGoPrintDiff), but this function was deprecated. Recommendation now is to use `github.com/google/go-cmp/cmp.Diff` which `diff.ObjectGoPrintDiff` itself uses in the latest versions of `k8s.io/apimachinery`.

```release-note
NONE
```


pkg/api/validate/admin.go
```diff
@@ -80,7 +79,7 @@ func (v *AdminAPIValidator) validateUpdateContainerService(cs, oldCs *api.OpenSh
 	old.Config.Images = cs.Config.Images
 
 	if !reflect.DeepEqual(cs, old) {
-		errs = append(errs, fmt.Errorf("invalid change %s", deep.Equal(cs, old)))
+		errs = append(errs, fmt.Errorf("invalid change %s", cmp.Diff(cs, old)))
```
m1kola: This and the following changes:

* https://github.com/openshift/openshift-azure/pull/1792/files#diff-9d1ebcff0c2c4390297fac3eb1aaebb1R82
* sync client v5: https://github.com/openshift/openshift-azure/pull/1792/files#diff-6b9eb417c2465f803e04c5286a756578R242
* same as above but in v6: https://github.com/openshift/openshift-azure/pull/1792/files#diff-323d45f40968c6b37fb8b3272de42e01R252

Change the logs output. But I didn't find any code/tools that depends on the output format. I assume it's safe to change. Please, correct me if I'm wrong.

pkg/api/validate/admin_test.go
```diff
@@ -43,8 +42,8 @@ func TestAdminAPIValidateUpdate(t *testing.T) {
 			f: func(oc *api.OpenShiftManagedCluster) {
 				oc.Config.PluginVersion = "latest" // the RP does this, but after validation: the user can't do this
 			},
-			expectedErrs: []error{
-				errors.New(`invalid change [Config.PluginVersion: latest != ]`),
+			expectedErrs: []*regexp.Regexp{
+				regexp.MustCompile(`PluginVersion:\s+?"latest"(?s).+?PluginVersion:\s+?""`),
```
m1kola: Output of `google/go-cmp` is unstable and the library [randomly adds non-breaking space and a regular space](https://github.com/google/go-cmp/blob/1b316004397f1f336546ca058ddb5b95c41a8772/cmp/report_text.go#L21-L49) into the output, so we can't just copy-paste the output and use it in tests: we have to use a regexp.

m1kola:
There are some "Could not resolve host: github.com" in logs so re-testing.

/retest

m1kola:
Prow says that jobs are still running after ~16 hours, so restarting.

/retest

asalkeld:
@m1kola why are we changing from go-test/deep - what's wrong with that? Can we have a better explanation please.

m1kola:
@asalkeld nothing wrong with it. The initial idea was to get rid of the extra dependency and use `k8s.io/apimachinery/pkg/util/diff.ObjectGoPrintDiff` instead (see jira AZURE-151). But `ObjectGoPrintDiff` has been deprecated and recommends to use `github.com/google/go-cmp` instead. So now it's more like for aligning dependencies with upstream Kubernetes.

asalkeld:
> @asalkeld nothing wrong with it. The initial idea was to get rid of the extra dependency and use `k8s.io/apimachinery/pkg/util/diff.ObjectGoPrintDiff` instead (see jira AZURE-151). But `ObjectGoPrintDiff` has been deprecated and recommends to use `github.com/google/go-cmp` instead. So now it's more like for aligning dependencies with upstream Kubernetes.

Thanks for the explanation @m1kola , maybe include the Jira reference next time?
old issue is #206

pkg/util/cmp/cmp_test.go
```diff
@@ -0,0 +1,52 @@
...
+			expect: false,
+		},
+		{
+			name:   "one nil: y",
+			x:      &x509.Certificate{},
+			y:      nil,
+			expect: false,
+		},
+		{
+			name:   "all non-nil and equal",
```
charlesakalugwu:
```suggestion
			name:   "both non-nil and equal",
```

pkg/util/cmp/cmp_test.go
```diff
@@ -0,0 +1,52 @@
...
+			x:      &x509.Certificate{Raw: []byte{1}},
+			y:      &x509.Certificate{Raw: []byte{1}},
+			expect: true,
+		},
+		{
+			name:   "all non-nil and not equal",
```
charlesakalugwu:
```suggestion
			name:   "both non-nil and not equal",
```

pkg/api/validate/admin_test.go
charlesakalugwu: these are really hard to read. I propose using a custom diff reporter for our go-cmp wrapper. The following is code which will produce such lists of differences in our tests:

```
admin_test.go:116: received errors:
admin_test.go:118: 	invalid change [{*api.OpenShiftManagedCluster}.Properties.AgentPoolProfiles[2].Count: 4 != 1]
admin_test.go:111: test case "invalid change"
admin_test.go:112: expected errors:
admin_test.go:114: 	invalid change [Name: new != openshift]
admin_test.go:116: received errors:
admin_test.go:118: 	invalid change [{*api.OpenShiftManagedCluster}.Name: new != openshift]
```

The code:

```
package cmp

import (
	"crypto/x509"
	"fmt"
	"strings"

	gocmp "github.com/google/go-cmp/cmp"
)

// simpleDiffReporter is a simple diff reporter that only records differences
// detected during comparison.
type simpleDiffReporter struct {
	path  gocmp.Path
	diffs []string
}

func (r *simpleDiffReporter) PushStep(ps gocmp.PathStep) {
	r.path = append(r.path, ps)
}

func (r *simpleDiffReporter) Report(rs gocmp.Result) {
	if !rs.Equal() {
		vx, vy := r.path.Last().Values()
		r.diffs = append(r.diffs, fmt.Sprintf("[%+v: %+v != %+v]", r.path.GoString(), vx, vy))
	}
}

func (r *simpleDiffReporter) PopStep() {
	r.path = r.path[:len(r.path)-1]
}

func (r *simpleDiffReporter) String() string {
	return strings.Join(r.diffs, "\n")
}

// Diff is a wrapper for github.com/google/go-cmp/cmp.Diff using the custom
// simple diff reporter
func Diff(x, y interface{}) string {
	// TODO(mikalai): Remove x509CertComparer after upgrading to a Go version
	//  that includes https://github.com/golang/go/issues/28743
	var r simpleDiffReporter
	if !gocmp.Equal(x, y, gocmp.Reporter(&r), gocmp.Comparer(x509CertComparer)) {
		return r.String()
	}
	return ""
}

func x509CertComparer(x, y *x509.Certificate) bool {
	if x == nil || y == nil {
		return x == y
	}
	return x.Equal(y)
}
```

You just need to change the diff strings in the tests and no need to use regexes

pkg/api/validate/admin_test.go
```diff
@@ -106,15 +105,13 @@ func TestAdminAPIValidateUpdate(t *testing.T) {
...
-		if !reflect.DeepEqual(errs, test.expectedErrs) {
-			t.Logf("test case %q", name)
-			t.Errorf("expected errors:")
-			for _, err := range test.expectedErrs {
-				t.Errorf("\t%v", err)
-			}
-			t.Error("received errors:")
-			for _, err := range errs {
-				t.Errorf("\t%v", err)
+		if len(test.expectedErrs) != len(errs) {
```
charlesakalugwu: revert this block of changes

pkg/api/validate/api_test.go
```diff
@@ -60,15 +59,13 @@ func TestAPIValidateUpdate(t *testing.T) {
...
-		if !reflect.DeepEqual(errs, test.expectedErrs) {
-			t.Logf("test case %q", name)
-			t.Errorf("expected errors:")
-			for _, err := range test.expectedErrs {
-				t.Errorf("\t%v", err)
-			}
-			t.Error("received errors:")
-			for _, err := range errs {
-				t.Errorf("\t%v", err)
+		if len(test.expectedErrs) != len(errs) {
```
charlesakalugwu: revert this block of changes

pkg/api/validate/admin_test.go
m1kola: @charlesakalugwu This is very similar (almost exactly the same) to what I had while working on this PR. I decided not to go with a custom reporter because:

* I didn't want to introduce an extra unit in our source code only to improve readability of tests. We will have to write tests and maintain this custom reporter.
* I wanted us to be able to just remove `pkg/util/cmp` after upgrading to a new version of Go and use `google/go-cmp` as is.

I agree that regexp is not very readable, but it's not too bad. Also test is not something one will be reading regularly. Someone will have to read unit tests when changing/breaking something in the code we test. In this case the person will know what they are doing and it will not be a problem to read the regexp.

Do you think improving test readability worth introducing new code?

We can simplify expressions a bit. We can check only for a line with a new value and ignore the old value. For example, we can replace ``regexp.MustCompile(`PluginVersion:\s+?"latest"(?s).+?PluginVersion:\s+?""`)`` with ``regexp.MustCompile(`PluginVersion:\s+?"latest"`)``

What do you think?

PR #1796, opened 2019-07-18
ETCD Backup before update

asalkeld:
```release-note
ETCD backups are now run prior to updating
```
Jira-263

@ehashman can you check the blobname validate. we currently create backup blobs with a "T" between the date and time.. I'd like to be consistent with how we do this

### DIFF
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -152,6 +152,15 @@ func isValidIPV4CIDR(cidr string) bool {
 
 func IsValidBlobName(c string) bool {
 	// https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata
+	if len(c) < 1 || len(c) > 1024 {
+		return false
+	}
+	if strings.HasSuffix(c, ".") || strings.HasSuffix(c, "/") {
+		return false
+	}
+	if strings.Contains(c, "./") || strings.Contains(c, "/.") {
+		return false
+	}
 	u, err := url.Parse(fmt.Sprintf("http://example.com/%s", c))
 	if err != nil {
 		return false
--- a/pkg/api/validate/types.go
+++ b/pkg/api/validate/types.go
@@ -1,7 +1,9 @@
 package validate
 
 import (
+	"fmt"
 	"net"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -34,8 +36,6 @@ var (
 
 	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
 	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
-
-	rxBlobContainerName = regexp.MustCompile(`^[a-z0-9-]{3,63}$`)
 
 	// This regexp is to check image version format
 	rxImageVersion = regexp.MustCompile(`^[0-9]{3}.[0-9]{1,4}.[0-9]{8}$`)
@@ -150,16 +150,15 @@ func isValidIPV4CIDR(cidr string) bool {
 	return true
 }
 
-func IsValidBlobContainerName(c string) bool {
+func IsValidBlobName(c string) bool {
 	// https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata
-	if !rxBlobContainerName.MatchString(c) {
+	u, err := url.Parse(fmt.Sprintf("http://example.com/%s", c))
+	if err != nil {
 		return false
 	}
-
-	if strings.Trim(c, "-") != c || strings.Contains(c, "--") {
+	if u.EscapedPath() != "/"+c {
 		return false
 	}
-
 	return true
 }
 
--- a/pkg/api/validate/types_test.go
+++ b/pkg/api/validate/types_test.go
@@ -1,8 +1,10 @@
 package validate
 
 import (
+	"fmt"
 	"net"
 	"testing"
+	"time"
 )
 
 func TestValidateImageVersion(t *testing.T) {
@@ -287,14 +289,11 @@ func TestIsValidAgentPoolHostname(t *testing.T) {
 	}
 }
 
-func TestIsValidBlobContainerName(t *testing.T) {
+func TestIsValidBlobName(t *testing.T) {
 	for _, tt := range []struct {
 		name  string
 		valid bool
 	}{
-		{
-			name: "12",
-		},
 		{
 			name:  "123",
 			valid: true,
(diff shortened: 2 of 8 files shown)


pkg/api/validate/types.go
```diff
@@ -35,7 +35,7 @@ var (
 	// This regexp is to guard against InvalidDomainNameLabel for hostname validation
 	rxCloudDomainLabel = regexp.MustCompile(`^[a-z][a-z0-9-]{1,61}[a-z0-9]\.`)
 
-	rxBlobContainerName = regexp.MustCompile(`^[a-z0-9-]{3,63}$`)
+	rxBlobContainerName = regexp.MustCompile(`^[a-zA-Z0-9-]{3,63}$`)
```
asalkeld: @ehashman this..

pkg/api/validate/types_test.go
```diff
@@ -329,6 +332,10 @@ func TestIsValidBlobContainerName(t *testing.T) {
 		{
 			name: "bad--bad",
 		},
+		{
+			name:  fmt.Sprintf("backup%s", time.Now().UTC().Format("2006-01-02T15-04-05")),
```
asalkeld: note the "T"

pkg/plugin/plugin.go
```diff
@@ -184,7 +184,7 @@ func (p *plugin) RecoverEtcdCluster(ctx context.Context, cs *api.OpenShiftManage
 	}
 
 	p.log.Info("running CreateOrUpdate")
-	if err := p.CreateOrUpdate(ctx, cs, true, deployFn); err != nil {
+	if err := p.createOrUpdateExt(ctx, cs, updateTypeNormal, deployFn); err != nil {
```
asalkeld: call the Ext version so we dont' backup before recovering from a backup

pkg/cluster/kubeclient/backup.go
```diff
@@ -22,7 +22,7 @@ func (u *Kubeclientset) BackupCluster(ctx context.Context, backupName string) er
 
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      backupName,
+			Name:      "etcd-backup",
```
ehashman: Why hardcode this constant? I see it's also used above on line 18... was this a bug?

pkg/plugin/integration_test.go
```diff
@@ -272,8 +274,49 @@ func setupNewCluster(ctx context.Context, log *logrus.Entry, cs *api.OpenShiftMa
 				},
 			},
 		}
+
 		return true, node, nil
 	})
+	cli.PrependReactor("get", "cronjobs", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		get := action.(k8stesting.GetAction)
+		cronjob := &batchv1b1.CronJob{
```
ehashman: In my experience, Kube cronjobs are notoriously unreliable on failures and most applications that they seemed right for we ended up using one-shot pods. Is this something we've considered? What's our error-handling story if this fails?

pkg/plugin/plugin.go
ehashman: Can we add a comment along those lines?

pkg/api/validate/types.go
ehashman: Seems fine to me... unless containers are case-insensitive?

pkg/plugin/integration_test.go
asalkeld: > In my experience, Kube cronjobs are notoriously unreliable on failures and most applications that they seemed right for we ended up using one-shot pods. Is this something we've considered? What's our error-handling story if this fails?

This PR is not introducing cronjobs, just adding the mocking for it.
But I don't believe when I did add it anyone else had a bad experience with them, else i would have done something different. They appear to be running just fine though... We were going to have some kind of monitoring to detect if the backups failed (or failed to happen - not sure if this is in place)..

pkg/plugin/plugin.go
asalkeld: sure

pkg/api/validate/types.go
asalkeld: OK, I see the problem. This is correct for the container, but too restrictive for the blob. We are using container name validation on a blob name. 
https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions
containers must be lowercase and blobnames can be any "Any URL characters". 

I'll make a new regex..

charlesakalugwu: According to the documentation, your implementation of `IsValidBlobName` doesn't take care of some requirements for valid blob names:
- A blob name must be at least one character long and cannot be more than 1,024 characters long, for blobs in Azure Storage.
- Avoid blob names that end with a dot (.), a forward slash (/), or a sequence or combination of the two.

```
=== RUN   TestIsValidBlobName
--- FAIL: TestIsValidBlobName (0.00s)
    types_test.go:347: //bad//: wanted valid false, got true
    types_test.go:347: : wanted valid false, got true
    types_test.go:347: xyypbtodfvqhbjggywtfuboirjpiytlqxpshwoucrwvpturicdagbwyqveglmpgckdjzzxjgyqtjzrqqlwkukdcifloehxffiwtpufdpgkrvyxbjmvipcmzdmprjoeiinsjkuyimtykptvghdwaaqaonopsmrtdfyhtiaqevxvdkdwtzdxefesbpjdrffopbmsinzcsvuxqkxykneqqrxavsywpisddgvzjswkqpwcnqpzogbcuxwvrcqikuzecurigjboopqaxsdzidxmlroblfybhhajwxffjhkykqgxagwcyyegesjxmubyyoaotkypnisujojmjwycshwhfyyefjgpraqgbqizwjbnpivaiutkptfyeynwsnpeybpllvqxjjcqwvfzdmmrhkvqzkafaakshgwbeewxbnytidqvcuuwnxxqnfmypsqjsengwiziwyjmdygaljavdozqpgauyimdpxtgfvjfakmzwxnysuffnvzlpwlhcxajinaasoxtsbejwbgcnyrugdakftvlriiugvlktgohmbhxdguellidywqtwnhvseirxffqnyfemafqdvuenlhrwzanmynwupsclybzhujdwlarydxyjanomttpnlasbphvqyvwtnlscjlgqrjknzqvlifqsjfqvdxelwmihbarzhzlrdevemufrmfkziuzhtvffucohekspkkkajzpwfdnsokcdpffbsdpykjvpcgmfbnaiiruwlbopveomhqzacokwthirhiehjdsufmdmesehygjboxbsgkblroeyztjxdjgnrqevuhvttpbbjxhylgnhczrmcrtekahnfruhgjmobwvnexipotpceqlaypptoolhinevldsagxytzufetiqueibuzivlholrmdhxmsseskfsxwxyiatgvapnlmthnrjbgicisafksdhbpnjqeiktxyskmdawoieuhamnyxnsnpzcbwhkouklhcfiocyhxurmgaboqdzvwmpvooadvrapkjjayt: wanted valid false, got true
    types_test.go:347: bad.: wanted valid false, got true
FAIL

```

pkg/cluster/kubeclient/backup.go
charlesakalugwu: Yeah I think this should be reverted as well.

pkg/plugin/integration_test.go
ehashman: This is fine, just wanted to mention it :) we were seeing something like 20% error rates on cronjobs failing (and they weren't retrying) at my last gig.

pkg/cluster/kubeclient/backup.go
ehashman: (waiting on an answer to this question, otherwise PR looks good to me)

asalkeld: So the validation on a kubernetes name is much more rigorous than a blob - to the extent that we can't use the timestamp (that is at the end of the backup name). Since the job is a transient entity, there is no real reason to have the job and blob named the same. Also since the blob lives on, i'd contend that it's name is more important.

pkg/plugin/integration_test.go
mjudeikis: Cronjobs are very sensitive to their config. took us a while to refactor our retry config in CI from purge jobs too. Agreed :)

pkg/cluster/kubeclient/backup.go
mjudeikis: I think its fine. It will create something like `etcd-backup-d8cc6`. And as stated blob-name is versioned.

mjudeikis:
@ehashman @charlesakalugwu merge?

ehashman:
/retest

RG creation timeout

PR #1807, opened 2019-07-19
Run validate on vmimage build

kwoodson:
Injecting vmimage validation into the vmimage build process.

```release-note
None
```


hack/vmimage.sh
```diff
@@ -35,6 +36,15 @@ make create
...
 make e2e
 
+# Run the image validation and capture the artifacts
+go generate ./... && go run -ldflags "-X main.gitCommit=$GITCOMMIT" ./cmd/vmimage -buildResourceGroup "${BUILD_RESOURCE_GROUP}" -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate -preserveBuildResourceGroup
+
+# Currently there are only 3 logs we want to capture
+mv /tmp/{yum_update_info,yum_check_update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
+
+# delete the image validation build_resource_group
+az group delete -g "${BUILD_RESOURCE_GROUP}" --yes --no-wait
```
charlesakalugwu: can you move this line into the cleanup function at the top to make sure `$BUILD_RESOURCE_GROUP` is cleaned up if an error occurs during `./cmd/vmimage`?

hack/vmimage.sh
```diff
@@ -35,6 +36,15 @@ make create
 
 make e2e
 
+# Run the image validation and capture the artifacts
+go generate ./... && go run -ldflags "-X main.gitCommit=$GITCOMMIT" ./cmd/vmimage -buildResourceGroup "${BUILD_RESOURCE_GROUP}" -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate -preserveBuildResourceGroup
```
charlesakalugwu: `go generate ./...` or `make generate`

hack/vmimage.sh
kwoodson: I think that should work.

kwoodson:
@charlesakalugwu 
> thanks @kwoodson! I was wondering if you plan to invoke vmimage-validate.sh from vmimage.sh? Currently the image validation functionality seems to be duplicated in both files.

This was my original attempt.  I spent time going through the entire process only to learn that the issue with calling `make vmimage-validate` was that the `cleanup()` executed because the `vmimage-validate.sh` script exited and there is a trap setup to catch `exit`.  This caused the `vmimage` script to exit early and never complete.

hack/vmimage-validate.sh
```diff
@@ -14,15 +12,17 @@ trap cleanup EXIT
...
 
 echo "Validating: ${IMAGE_RESOURCENAME}"
 
-go generate ./... && go run -ldflags "-X main.gitCommit=$(git rev-parse --short=10 HEAD)" ./cmd/vmimage -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate "true"
+# pass -preserveBuildResourceGroup so we can copy artifacts
+go generate ./... && go run -ldflags "-X main.gitCommit=$(git rev-parse --short=10 HEAD)" ./cmd/vmimage -buildResourceGroup "${BUILD_RESOURCE_GROUP}" -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate -preserveBuildResourceGroup
 
 # Currently there are only 3 logs we want to capture
-mv /tmp/{yum_update_info,yum_check_update,scap-report.html} ${ARTIFACTS:-/tmp}/
+mv /tmp/{yum_update_info,yum_check_update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
```
ehashman: When would this fail?

hack/vmimage-validate.sh
```diff
@@ -3,9 +3,7 @@
 cleanup() {
   set +e
 
-  generate_artifacts
-
-  delete
+  az group delete -g "${BUILD_RESOURCE_GROUP}" --yes --no-wait
```
ehashman: Does `delete` do something more than this? Trying to understand the need for this change.

pkg/vmimage/data/validate.sh
```diff
@@ -36,3 +36,5 @@ oscap xccdf eval \
   --cpe /usr/share/xml/scap/ssg/content/ssg-rhel7-cpe-dictionary.xml \
   /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml > /tmp/oscap.log || true
   
+python -c 'import sys; from lxml import html; h = html.fromstring(open("/tmp/scap-report.html").read()); sys.exit(1) if len(h.find_class("rule-result rule-result-fail")) > 0 else sys.exit(0)'
```
ehashman: :+1:

hack/vmimage-validate.sh
kwoodson: @ehashman @charlesakalugwu 
Delete calls the delete function [here](https://github.com/openshift/openshift-azure/blob/master/hack/tests/ci-prepare.sh#L67).  This calls `make delete` which executes https://github.com/openshift/openshift-azure/blob/master/hack/delete.sh.

This fires up the fakerp and attempts to call delete cluster.  At this point in time we do not have a cluster only a vm running our image.  The fakerp then fails to execute because it doesn't find the files in `_data` directory.  Failing then triggers the `cleanup` trap on exit.  This fails the validate script.  Hopefully this helps clarify the process.

mjudeikis: thanks for this :)

mjudeikis:
I think apart of `|| true` question, all looks good.

hack/vmimage-validate.sh
charlesakalugwu: I guess if one of those files `yum_update_info` or `yum_check_update` or `scap-report.html` isn't available in `/tmp` then mv will return a non-zero exit code. @kwoodson  I wonder if that is a possibility with the report generation...that one of those files might not get written for some reason?

charlesakalugwu:
This looks good to me. 
/lgtm

charlesakalugwu:
/lgtm cancel 

needs rebase

hack/vmimage-validate.sh
m1kola: I think, it happens when `ARTIFACTS` is not set or set to `/tmp` so the command will basically be moving files to the same location:

```
mv /tmp/{yum_update_info,yum_check_update,scap-report.html} /tmp/
mv: '/tmp/yum_update_info' and '/tmp/yum_update_info' are the same file
mv: '/tmp/yum_check_update' and '/tmp/yum_check_update' are the same file
mv: '/tmp/scap-report.html' and '/tmp/scap-report.html' are the same file

echo $?
1
```

kwoodson: @m1kola beat me to it!

@ehashman @charlesakalugwu @mjudeikis 
The major reason I placed the `|| true` here is that when running locally and the `$ARTIFACTS` variable is unset and defaults to tmp.  The `mv` essentially moves the file to the same location. `mv: '/tmp/scap-report.html' and '/tmp/scap-report.html' are the same file`.

I did not want to fail the entire process because the file gets moved to the same location.  If there is a preferred method to mitigate this then I'm open for suggestions.

hack/vmimage-validate.sh
m1kola: @kwoodson do we need to make similar changes to `hack/vmimage.sh`? I'm getting the following somewhere after `./hack/delete.sh mradchuk-tests`

> Resource group 'mradchuk-tests' could not be found.

kwoodson: @m1kola 
I have seen this message.  I think this must happen in the `make delete` and maybe somewhere here:
https://github.com/openshift/openshift-azure/blob/master/pkg/util/azureclient/openshiftmanagedcluster/2019-04-30/client.go#L208

I believe this to be a symptom of calling delete and then waiting for the delete to finish and then returning the message that it is actually gone. 
```
$ az group delete -g ktest-noexist
Are you sure you want to perform this operation? (y/n): y
Resource group 'ktest-noexist' could not be found.
```

I think hitting this in the trap `cleanup` just ensures that its missing.  I'm not sure exactly where that message comes from.

kwoodson:
Rebased and addressed questions.

PR #1812, opened 2019-07-23
Send logs to a customer's log analytics workspace

asalkeld:
This is just an adaptation of this: https://github.com/microsoft/OMS-docker/tree/master/Kubernetes
We have to do this as the container requires privileged mode.

```release-note
Send logs to a customer's log analytics workspace
```
Jira-124


pkg/fakerp/fakerp.go
```diff
@@ -201,6 +203,15 @@ func createOrUpdateWrapper(ctx context.Context, p api.Plugin, log *logrus.Entry,
 		return nil, err
 	}
 
+	if !isUpdate {
```
asalkeld: @amanohar the real RP would have to do something similar (create the workspace before creating the cluster)

charlesakalugwu:
@asalkeld Since we'll both need the OMS agent for our tasks what do you think about extracting the part which just sets up the OMS agent into a separate PR? Thanks

charlesakalugwu:
> @asalkeld Since we'll both need the OMS agent for our tasks what do you think about extracting the part which just sets up the OMS agent into a separate PR? Thanks

Actually most of the work for sending both customer metrics and logs involves just setting up the agent. So please disregard my previous suggestion.  :)

pkg/fakerp/fakerp.go
```diff
@@ -201,6 +203,15 @@ func createOrUpdateWrapper(ctx context.Context, p api.Plugin, log *logrus.Entry,
...
+		wID, wKey, err := createWorkspace(ctx, os.Getenv("AZURE_SUBSCRIPTION_ID"), cs.Location, os.Getenv("RESOURCEGROUP"))
```
asalkeld: Note the wID here is not the UUID that we need, it looks like this "/subscriptions/<sub>/resourcegroups/angus-dev/providers/microsoft.operationalinsights/workspaces/openshift".

asalkeld:
The tests are failing because I haven't made workspaces for each region yet..

pkg/api/admin/types.go
```diff
@@ -103,6 +106,11 @@ type NetworkProfile struct {
 	PeerVnetID *string `json:"peerVnetId,omitempty"`
 }
 
+// MonitorProfile configuration for Azure Monitor log analytics.
+type MonitorProfile struct {
+	WorkspaceID *string `json:"workspaceId,omitempty"`
```
asalkeld: admin api does not have the key as per design doc

pkg/api/validate/api.go
```diff
@@ -51,6 +51,9 @@ func (v *APIValidator) validateUpdateContainerService(cs, oldCs *api.OpenShiftMa
 	// validating ProvisioningState is the RP's responsibility
 	old.Properties.ProvisioningState = cs.Properties.ProvisioningState
 
+	old.Properties.MonitorProfile.WorkspaceID = cs.Properties.MonitorProfile.WorkspaceID
+	old.Properties.MonitorProfile.WorkspaceKey = cs.Properties.MonitorProfile.WorkspaceKey
```
asalkeld: MonitorProfile is updatable, so you can enable/disable Monitoring by doing an update.

pkg/sync/v7/data/DaemonSet.apps/openshift-azure-logging/omsagent.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - mountPath: /etc/kubernetes/host
+          name: azure-json-path
+        - mountPath: /etc/config
+          name: omsagent-config
+        - mountPath: /etc/config/settings
+          name: settings-vol-config
+          readOnly: true
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+        node-role.kubernetes.io/compute: "true"
```
asalkeld: only run on the compute node until we have figured out how to do audit logs (let's do that as a later work item).

pkg/sync/v7/translate.go
```diff
@@ -352,6 +352,41 @@ var translations = map[string][]struct {
 			Template: "{{ .Config.GenevaLoggingControlPlaneRegion }}",
 		},
 	},
+	"DaemonSet.apps/openshift-azure-logging/omsagent": {
+		{
```
asalkeld: by setting this label to true, we cause the sync pod to create the resource. when it is set to false, sync pod will delete it.

pkg/arm/v7/resources.go
```diff
@@ -365,6 +365,17 @@ func vmss(cs *api.OpenShiftManagedCluster, app *api.AgentPoolProfile, backupBlob
 		script = base64.StdEncoding.EncodeToString(b)
 	}
 
+	vmssTags := map[string]*string{
+		"clusterName": to.StringPtr(cs.Name),
+		"logAnalyticsWorkspaceResourceId": to.StringPtr(resourceid.ResourceID(
+			cs.Properties.AzProfile.SubscriptionID,
+			cs.Properties.AzProfile.ResourceGroup,
+			"Microsoft.OperationalInsights/workspaces",
+			cs.Properties.MonitorProfile.WorkspaceID)),
```
asalkeld: @charlesakalugwu this might be nice, but it is causing all nodes to be rotated if you enable/disable monitoring. Can we do this is a less intrusive way?

asalkeld: I have just pushed a commit to remove it, let me know what you think..

pkg/api/2019-04-30/types.go
```diff
@@ -65,6 +65,9 @@ type Properties struct {
 
 	// AuthProfile (in): configures OpenShift authentication
 	AuthProfile *AuthProfile `json:"authProfile,omitempty"`
+
+	// MonitorProfile configures which log analytics workspace to use.
+	MonitorProfile *MonitorProfile `json:"monitorProfile,omitempty"`
```
jim-minter: Is it agreed with Anubhuti that we are changing the existing external API?

pkg/api/2019-04-30/types.go
```diff
@@ -138,6 +141,12 @@ type AgentPoolProfile struct {
 	Role *AgentPoolProfileRole `json:"role,omitempty"`
 }
 
+// MonitorProfile configuration for Azure Monitor log analytics.
+type MonitorProfile struct {
+	WorkspaceID  *string `json:"workspaceId,omitempty"`
+	WorkspaceKey *string `json:"workspaceKey,omitempty"`
```
jim-minter: I thought WorkspaceKey wasn't going into the external API?

pkg/api/2019-04-30/types.go
jim-minter: (as opposed to creating a new external API version...)

ehashman: cc @amanohar

pkg/api/2019-04-30/types.go
ehashman: I think it's part of the external API, just not the admin API nor the az client.

pkg/sync/v7/data/ConfigMap/openshift-azure-logging/container-azm-ms-agentconfig.yaml
```diff
@@ -0,0 +1,292 @@
+apiVersion: v1
+data:
+  config-version: ver1
+  kube.conf: |
+    # Fluentd config file for OMS Docker - cluster components (kubeAPI)
```
ehashman: IMO it would be good to include a link somewhere (commit body or comment) to documentation on how these configs were generated.

pkg/api/2019-04-30/types.go
amanohar: `WorkspaceKey` shouldn't be in the public API model because RP will query it.

pkg/api/2019-04-30/types.go
amanohar: This will need to be a new API version. 

 (I also added it in the doc): Do we want `MonitorProfile` to be opinionated towards Log Analytics constructs? Or should it be more like authProfile, i.e. generic enough to support other Log providers? 

Note that its important to think of API model not just in context of `3.11` but also future ARO in general (and `4.x`) and options it will give to customers without big changes to the API model in the future.

mjudeikis:
Do we have a test (maybe manua) to confirm that existing clusters with no logging enabled, after the update will be able to make `put` and it?

ehashman:
/retest

flake #552 on upgrade-5.2

pkg/sync/v7/data/DaemonSet.apps/openshift-azure-logging/omsagent.yaml
```diff
@@ -63,16 +63,14 @@ spec:
         - mountPath: /hostfs
           name: host-root
           readOnly: true
-        - mountPath: /var/run/host
+        - mountPath: /var/run/host/docker.sock
           name: docker-sock
```
charlesakalugwu: maybe lets mount this readonly

pkg/sync/v7/data/Deployment.apps/openshift-azure-logging/omsagent-rs.yaml
```diff
@@ -0,0 +1,114 @@
+apiVersion: apps/v1
```
charlesakalugwu: i'm really not sure if this is  needed. In addition the daemonset version is better because that will continue to work as the size of the compute nodepool grows.

pkg/startup/v7/data/worker/etc/origin/node/node-config.yaml
```diff
@@ -45,6 +45,8 @@ kubeletArguments:
   - {{ print "node-role.kubernetes.io/" .Role "=true" | quote }}
   pod-manifest-path:
   - /etc/origin/node/pods
+  read-only-port:
```
charlesakalugwu: This enabled a local unauthenticated kubelet port from which `/metrics` can be scraped. We need to discuss and fully understand the risks of doing this.

pkg/util/enrich/enrich.go
```diff
@@ -82,3 +84,18 @@ func CertificatesFromVault(ctx context.Context, kvc keyvault.KeyVaultClient, cs
 
 	return nil
 }
+
+func MonitorIDAndKey(ctx context.Context, client operationalinsights.WorkspacesClient, cs *api.OpenShiftManagedCluster) error {
+	wID := ""
```
charlesakalugwu: `var wID, wKey string`

pkg/api/admin/config.go
```diff
@@ -103,6 +103,8 @@ type ImageConfig struct {
 	GenevaTDAgent *string `json:"genevaTDAgent,omitempty"`
 	GenevaStatsd  *string `json:"genevaStatsd,omitempty"`
 	MetricsBridge *string `json:"metricsBridge,omitempty"`
+
+	MonitorAgent *string `json:"monitorAgent,omitempty"`
```
charlesakalugwu: It might be easier to understand if this were renamed to `AzureMonitorAgent` or `LogAnalyticsAgent`

pkg/sync/v7/data/Deployment.apps/openshift-azure-logging/omsagent-rs.yaml
asalkeld: I think Jobe said we needed it, it would be nice to get some clarification on why.

asalkeld:
@charlesakalugwu l think it's good to go in, but @ehashman still needs to cut the release so we just need to decide if it goes in or not (I think the assumption is to wait until next release).

pkg/sync/v7/data/ConfigMap/openshift-azure-logging/container-azm-ms-agentconfig.yaml
charlesakalugwu: There's a ruby script inside the agent container which is responsible for generating default configs in the event that a wrong config or none was provided. This config was generated by running the agent with no config, letting the ruby script inside the agent image generate the config and then I copied that content.

pkg/sync/v7/data/Deployment.apps/openshift-azure-logging/omsagent-rs.yaml
charlesakalugwu: Ok its been clarified. The deployment is a log analytics cluster-scoped agent while the daemonset is a log analytics node-scoped agent. They are both needed on every cluster.

charlesakalugwu:
@asalkeld I think this will go in the next release
/lgtm

ehashman:
Please rebase this against v8 :)

charlesakalugwu:
upgrade-v5.2 flake #1400 
/retest

PR #1813, opened 2019-07-23
Improve release notes tool

charlesakalugwu:
This PR contains work which improves our release notes tool in various aspects. Some of these improvements include:

- Remove dependency on `libgit2`.
- Correctly lists the merged commits between two git references. The current release notes tool for example produces invalid results sometimes e.g. v5.2.1 to master lists changes that occurred before release of v5.2.1. 
- Release version is now passed as a parameter and written into the file together with the notes
- Split the reponame parameter (e.g. openshift/openshift-azure) to 2 parameters `github-org` and `github-repo`
- Split the commitrange parameter (e.g. v3..master) to 2 parameters `start-sha` and `end-sha`. This makes configuring the ranges more friendly to automation
- Updated the layout of the release notes document to make it much more pretty.
- First character of each release note is capitalized.

*Propsed considerations*:

In order to ensure that rendered changelogs are legible we, as a team, need to enforce a certain level of formatting of the content of `release-notes` tags.
If a new PR introduces multiple features, the multiline `release-notes` should be formatted properly e.g. like this:

```
Introduction to feature X:
  - foo is enabled
  - bar is disabled
Conclusion.
```

instead of like this:
```
Foo is enabled for feature X.
Bar is disabled for feature X.
``` 
How to run it
```
$ export GITHUB_TOKEN=<github_access_token>
$ go run cmd/releasenotes/releasenotes.go -start-sha=v5.2.1 -end-sha=master -release-version=v6.0 -output-file=release-v6-changelog.md

INFO[2019-07-23T11:17:24+02:00] parsing flags                                
INFO[2019-07-23T11:17:24+02:00] validating config                            
INFO[2019-07-23T11:17:24+02:00] creating release notes scraper               
INFO[2019-07-23T11:17:24+02:00] fetching release notes v5.2.1..master from github.com/openshift/openshift-azure 
INFO[2019-07-23T11:17:27+02:00] writing release notes to file release-v6-changelog.md   
```
/cc @y-cote @mjudeikis @ehashman 

```release-note
NONE
```


ehashman:
How do I test this? I ran `make releasenotes` and nothing happened.

charlesakalugwu:
@ehashman `make releasenotes` just builds the tool. Instructions for running the tool can be found in the associated README in this PR https://github.com/openshift/openshift-azure/blob/e8e6537009990a90c3864bf98a4da6960d9504e8/cmd/releasenotes/README.md#release-notes-sample-scrape-of-the-openshiftopenshift-azure-github-repo

charlesakalugwu:
upgrade-v5.1 flake #1481 
/retest

charlesakalugwu:
@y-cote please can you take a look at this?

mjudeikis:
My only concern here is that we now take a dependency into our vendor dir which will have to be vendored into azure RP. And it is big one (why it has GCP cloud API deps? those googles and their monorepos... :man_shrugging:  ) .
If we doing this, maybe it is worth moving this to azure-misc ? Or try to contribute these changes to  https://github.com/kubernetes/release/tree/master/cmd/release-notes as now we going way off the upstream.

charlesakalugwu:
> My only concern here is that we now take a dependency into our vendor dir which will have to be vendored into azure RP. And it is big one (why it has GCP cloud API deps? those googles and their monorepos...  ) .
> If we doing this, maybe it is worth moving this to azure-misc ? Or try to contribute these changes to https://github.com/kubernetes/release/tree/master/cmd/release-notes as now we going way off the upstream.

For what its worth, this PR (and all those vendor changes) causes a net addition of only 316 lines of code. To be honest I feel like most of those vendor changes aren't even connected to this PR but came up because I had to run `glide up -v` after removing `libgit2`. 

And I feel like the updated code in this PR is actually closer to the upstream https://github.com/kubernetes/release/tree/master/cmd/release-notes than what we had. Btw I still cannot run the current tool on my machine. This alone makes the changes in this PR very valuable. Everyone can finally run the release notes tool without resorting to ugly hacks of setting up libgit2 packages on their machine. I cannot even find a good `libgit2` on debian.

PR #1817, opened 2019-07-26
Moving geneva logging to the host + azsecpack

thekad:
```release-note
* Moved Geneva logging stack from openshift (running as a daemonset in the `openshift-azure-logging` namespace) to run in the hosts themselves as systemd services
* Added Azure Security Pack logging to all hosts
* Using fluentd from RHEL repos instead of td-agent
```
Rebased on top of v7 now, will probably need adjustments if we:
* need clamav


pkg/cluster/initialize.go
```diff
@@ -151,6 +159,7 @@ func (u *Upgrade) CreateOrUpdateConfigStorageAccount(ctx context.Context) error
 	return nil
 }
 
+// InitializeUpdateBlob sets some calculated values on the given blob
```
mjudeikis: its not `some` values. It `hashe values` used for updates.

pkg/startup/v7/data/master/etc/mdsd.d/mdsd.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="utf-8"?>
+<MonitoringManagement version="1.0" namespace="{{ .Config.GenevaLoggingNamespace | XMLEscape }}" eventVersion="1" timestamp="2017-08-01T00:00:00.000Z">
+    <Accounts>
+        <Account moniker="{{ .Config.GenevaLoggingAccount | XMLEscape }}" isDefault="true" autoKey="false"/>
```
mjudeikis: We usually do XML configs based on Derive pattern  https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v7/derive.go#L74

Gives us more control over the file. Is this possible here?

I think you might have double escape/`"`? 
`"{{ .Config.GenevaLoggingAccount | XMLEscape }}"` 
vs `{ .ResourceName | Escape }}` ?

you have `"` in some fields, but not other. What that about?

pkg/startup/v7/data/worker/etc/mdsd.d/mdsd.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mjudeikis: If you use the pattern above all files for masters and nodes could be re-used and not copy-paste. basically maintained once.

pkg/startup/v7/data/worker/etc/td-agent/td-agent.conf
```diff
@@ -0,0 +1,44 @@
+<source>
```
mjudeikis: same as xml files

pkg/startup/v7/startup.go
```diff
@@ -133,16 +135,23 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
-			"Config":           &s.cs.Config,
-			"Derived":          derived,
-			"Role":             role,
-			"Hostname":         hostname,
-			"DomainName":       domainname,
-		})
+		b, err := template.Template(filepath, tmpl,
+			map[string]interface{}{
+				"Deref": func(pi *int) int { return *pi },
+				"XMLEscape": func(s string) (string, error) {
```
mjudeikis: Duplicate  logic location with 
https://github.com/openshift/openshift-azure/blob/master/pkg/util/template/template.go#L15

pkg/startup/v7/startup.go
```diff
@@ -161,6 +170,7 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
 			perm = 0644
 		}
 
+		s.log.Debug(fmt.Sprintf("Writing file %s", filepath))
```
mjudeikis: Is this necessary? Would errors from `MkdirAll` and `WriteFile` would not be enough?

pkg/sync/v7/derive.go
```diff
@@ -70,59 +67,3 @@ func (derivedType) StatsdArgs(cs *api.OpenShiftManagedCluster) ([]interface{}, e
 		"-SourceRoleInstance", "OSA",
 	}, nil
 }
-
```
mjudeikis: Why we removing this over multiple files? same as above

thekad: This got moved from sync->startup

pkg/startup/v7/derive.go
```diff
@@ -27,17 +32,16 @@ func (derivedType) SystemReserved(cs *api.OpenShiftManagedCluster, role api.Agen
...
-				return "cpu=200m,memory=512Mi", nil
 			}
 
-		} else {
-			if role == api.AgentPoolProfileRoleMaster {
-				return "cpu=1000m,memory=1Gi", nil
-			} else {
-				return "cpu=500m,memory=512Mi", nil
-			}
+			return "cpu=200m,memory=512Mi", nil
```
thekad: changed the logic here because `go vet` was complaining it wasn't optimal, basically just removed the else statements.

pkg/startup/v7/derive.go
```diff
@@ -135,3 +139,216 @@ func (derivedType) CaBundle(cs *api.OpenShiftManagedCluster) ([]*x509.Certificat
 
 	return tls.UniqueCert(caBundle), nil
 }
+
+// TDAgentDotConf returns /etc/td-agent/td-agent.conf content
+func (derivedType) TDAgentDotConf() (string, error) {
+	// so far, this is a static config, may change later
+	return `<source>
```
ehashman: Do you have links to docs where you determined the format of these config files?

pkg/startup/v7/derive.go
```diff
@@ -135,3 +139,216 @@ func (derivedType) CaBundle(cs *api.OpenShiftManagedCluster) ([]*x509.Certificat
...
+    host 127.0.0.1
+    port 5001
+    weight 60
+  </server>
+</match>`, nil
+}
+
+// MDSDDotXML returns /etc/mdsd.d/mdsd.xml config contents
+func (derivedType) MDSDDotXML(cs *api.OpenShiftManagedCluster) (string, error) {
+	var tmpl = `<?xml version="1.0" encoding="utf-8"?>
```
ehashman: I am not a fan of this massive multi-line static config in the middle of a function... what's the idiomatic Golang way to handle this? I'd be inclined to read it from a file or at least put it at the top of this file.

pkg/startup/v7/startup.go
```diff
@@ -133,16 +134,18 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
-		}, map[string]interface{}{
-			"ContainerService": s.cs,
-			"Config":           &s.cs.Config,
-			"Derived":          derived,
-			"Role":             role,
-			"Hostname":         hostname,
-			"DomainName":       domainname,
-		})
+		b, err := template.Template(filepath, tmpl,
+			map[string]interface{}{
```
ehashman: Can we revert the whitespace change to reduce the size of this diff? I had a hard time figuring out what changed here.

pkg/vmimage/data/script.sh
```diff
@@ -164,7 +172,29 @@ yum -y install \
...
     WALinuxAgent \
-    yum-utils
+    yum-utils \
+    azure-mdsd \
+    azure-security \
+    azsec-monitor \
+    td-agent
+
+# td-agent needs access to the journal
+usermod -a -G systemd-journal td-agent
```
ehashman:
```suggestion
adduser td-agent systemd-journal
```

`usermod` is pretty low-level and the syntax is sort of arcane, I'd prefer using `adduser` unless there's a compelling reason not to?

pkg/vmimage/data/script.sh
```diff
@@ -164,7 +172,29 @@ yum -y install \
...
+
+# openshift-audit dir ends up with wrong permissions, preemptively fix that
```
ehashman: That sounds like a umask issue, should we just be setting that? Also, do we want the audit log dir to be world-readable? That doesn't seem right to me.

pkg/vmimage/data/script.sh
```diff
@@ -164,7 +172,29 @@ yum -y install \
...
+mkdir -pv /var/log/openshift-audit
+chmod -v 0755 /var/log/openshift-audit
+
+# we need to build gems
+yum -y groups install "Development Tools"
+
+# install the latest stable for td-agent3
+/opt/td-agent/embedded/bin/fluent-gem install fluent-plugin-systemd -v 1.0.1
```
ehashman: Why does fluent ship its own copy of rubygems!? It's also very frustrating that `fluent-plugin-systemd` doesn't ship binary gems because then this whole song and dance could be avoided.

The latest version is 1.0.2, does MS have a hard requirement on 1.0.1?

pkg/vmimage/data/script.sh
```diff
@@ -164,7 +172,29 @@ yum -y install \
...
+
+# td-agent needs access to the journal
+usermod -a -G systemd-journal td-agent
+
+# openshift-audit dir ends up with wrong permissions, preemptively fix that
+mkdir -pv /var/log/openshift-audit
+chmod -v 0755 /var/log/openshift-audit
+
+# we need to build gems
+yum -y groups install "Development Tools"
```
ehashman: Can we narrow this down a bit to just the build tools we need? This is a pretty large metapackage and I doubt that ruby needs e.g. doxygen.

pkg/vmimage/data/script.sh
```diff
@@ -174,7 +204,11 @@ EOF
 # not commited with a : so that Windows users can check out the repo
 mv /etc/docker/certs.d/docker-registry.default.svc-5000 /etc/docker/certs.d/docker-registry.default.svc:5000
 
-rm /var/lib/yum/client-cert.pem /var/lib/yum/client-key.pem
+rm -fv /etc/yum.repos.d/kickstart.repo \
```
ehashman: Previously we did not remove the kickstart repo, why are we doing that now?

pkg/startup/v7/startup.go
```diff
@@ -173,7 +176,10 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
 		if err != nil {
 			return err
 		}
+
+		s.log.Debug(fmt.Sprintf("Wrote file %s", filepath))
 	}
 
-	return nil
+	// write an empty file to signal files have all been written to disk
+	return w.WriteFile("/host/startup.finished", []byte{}, 0644)
```
ehashman: Where is this being used?

pkg/vmimage/data/etc/rc.d/rc.local
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -ex
 
 # https://github.com/openshift/openshift-azure/issues/1454
-for pos in /var/log/journald.pos/worker0/storage.json /var/log/openshift-audit.pos; do
+for pos in /var/run/td-agent/*.pos/*/storage.json /var/run/td-agent/*-cursor.json; do
```
ehashman: The upstream bug looks like it was fixed in May, do we even need this hack anymore?

pkg/sync/v7/translate.go
```diff
@@ -310,48 +304,6 @@ var translations = map[string][]struct {
 			Template: "{{ .Config.Images.ServiceCatalog }}",
 		},
 	},
-	"DaemonSet.apps/openshift-azure-logging/mdsd": {
```
jim-minter: How sure are you that an upgrade from v6 to this PR has a working logging stack at the end of it?

pkg/startup/v7/startup.go
jim-minter: If this can be reverted, do so :)

pkg/startup/v7/derive.go
jim-minter: discussed on Slack, a better direction would be to get the text under data/ ; use a common/ directory to limit duplication across master/ and worker/

pkg/startup/v7/derive.go
```diff
@@ -135,3 +139,216 @@ func (derivedType) CaBundle(cs *api.OpenShiftManagedCluster) ([]*x509.Certificat
...
+			var b bytes.Buffer
+			err := xml.EscapeText(&b, []byte(s))
+			return b.String(), err
+		},
+	}).Parse(tmpl))
+
+	b := &bytes.Buffer{}
+
+	err = t.Execute(b, map[string]string{
+		"Namespace":         cs.Config.GenevaLoggingNamespace,
```
jim-minter: I'd rather not encourage using intermediate variable names (I know this is a moved function, but this PR is then building on the pattern).  Best just to pass in cs, then refer to `{{ .ContainerService.Foo }}` like we do everywhere else.  Try to get to the point where template.Template() is being used and (reluctantly) add the XMLEscape function there.

pkg/vmimage/data/etc/systemd/system/mdsd.path
```diff
@@ -0,0 +1,8 @@
+[Path]
```
jim-minter: it may be less opaque to add systemctl enable(s) / systemctl start(s) to near the bottom of the master and worker startup scripts

pkg/vmimage/data/etc/systemd/system/td-agent.service.d/custom.conf
```diff
@@ -0,0 +1,4 @@
+[Unit]
```
jim-minter: what's this file do? does moving to systemctl enable(s) / systemctl start(s) to near the bottom of the master and worker startup scripts mean it can be avoided?

pkg/vmimage/data/script.sh
jim-minter: not a umask issue - https://github.com/openshift/origin/blob/a1e38f8c0d105eb1eed897de5ac46f2bbef832a9/vendor/gopkg.in/natefinch/lumberjack.v2/lumberjack.go#L191

0750 and a chgrp would probably be best

pkg/vmimage/data/script.sh
jim-minter: must not remove that - see https://github.com/openshift/openshift-azure/blob/93a30816ad15aa87636a608fe7b0233023c3af7b/pkg/arm/v7/data/master-startup.sh#L18

pkg/vmimage/data/script.sh
```diff
@@ -174,7 +204,11 @@ EOF
...
+  /etc/yum.repos.d/azurecore.repo \
```
jim-minter: makes sense not to remove these repo files in case there's a CVE in one of the RPMs and needs fixing?

pkg/vmimage/data/etc/yum.repos.d/td.repo
```diff
@@ -0,0 +1,5 @@
+[treasuredata]
```
jim-minter: We ship `fluentd` and `rubygem-fluent-plugin-systemd` RPMs in the Red Hat OpenShift Container Platform (v. 3.11 for x86_64) - is it feasible to use these instead of treasuredata and the build horror?

pkg/startup/v7/derive.go
thekad: I added a link to the docs to the top of the config file now

pkg/startup/v7/startup.go
thekad: a lot of the whitespace comes from gofmt, but you can use this: https://github.com/openshift/openshift-azure/pull/1817/files?w=1

pkg/vmimage/data/script.sh
thekad: that's ruby's "omnibus" pattern, which means ship all the things.

pkg/vmimage/data/etc/rc.d/rc.local
thekad: the fluentd version we'll use (from the yum repos) doesn't have that patch yet afaict

pkg/vmimage/data/script.sh
thekad: I added the azurecore repo to the kickstart repo file, makes no sense to use 2 diff files, wdyt?

pkg/vmimage/data/script.sh
thekad: installing from the yum repos means not having to do this now, let's see if that's possible.

pkg/vmimage/data/script.sh
thekad: btw, both openshift and fluentd run as root, so no chgrp needed, I did chmod 0750

pkg/startup/v7/data/common/etc/fluentd/fluent.conf
```diff
@@ -28,14 +28,14 @@
 </filter>
 <match **>
   @type forward
-  send_timeout 60s
-  recover_wait 10s
-  hard_timeout 60s
-
+  buffer_type memory
+  buffer_queue_full_action block
```
ehashman: What's the consequence here if the agent encounters a lot of backpressure? Is it possible for this to OOM the host or block indefinitely?

thekad: Hrm, good question, I had maintained the same settings @mjudeikis had in the previous version of the logging stack, but according to https://docs.fluentd.org/v/0.12/buffer#handling-queue-overflow I feel like either throwing an exception or dropping old chunks would be better... The arch is like this: journald -> fluentd -> memory buffer -> mdsd, what do you all think?

thekad:
I believe this is failing because it requires the image it is building at the moment, what's the best way to handle this @jim-minter ?

ehashman:
Unit tests are consistently failing so I'm not sure that's the (only) problem. Can you take a look at that?

pkg/arm/v7/data/node-startup.sh
```diff
@@ -100,6 +106,16 @@ mkdir -m 0750 -p /var/lib/origin/openshift.local.volumes
 systemctl disable rsyslog.service
 systemctl stop rsyslog.service
 
+# setting up geneva logging stack
+systemctl enable mdsd.service
```
mjudeikis: nit:
`systemctl enable mdsd.service azsecd.service ... ` should work too

pkg/startup/v7/startup.go
```diff
@@ -101,48 +102,70 @@ func (s *startup) Hash(role api.AgentPoolProfileRole) ([]byte, error) {
 func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, hostname, domainname string) error {
 	assetNames := AssetNames()
 	sort.Strings(assetNames)
+	var filesToWrite = map[string]string{}
```
jim-minter: @thekad golang does not guarantee the iteration order of a map.  By using a map here you invalidate the sort above.  You must use a slice.

thekad: fixed in 4719b2d

ehashman:
These test failures are real:

```
time="2019-08-08T19:26:50Z" level=debug msg="400 Bad Request: Failed to apply request: UpdateMasterAgentPoolStart: Code=\"VMExtensionProvisioningError\" Message=\"VM has reported a failure when processing extension 'cse'. Error message: \\\"Enable failed: failed to execute command: command terminated with exit status=1\\n[stdout]\\nmeta-data=/dev/disk/azure/resource-part1 isize=512    agcount=4, agsize=1048508 blks\\n         =                       sectsz=4096  attr=2, projid32bit=1\\n         =                       crc=1        finobt=0, sparse=0\\ndata     =                       bsize=4096   blocks=4194032, imaxpct=25\\n         =                       sunit=0      swidth=0 blks\\nnaming   =version 2              bsize=4096   ascii-ci=0 ftype=1\\nlog      =internal log           bsize=4096   blocks=2560, version=2\\n         =                       sectsz=4096  sunit=1 blks, lazy-count=1\\nrealtime =none                   extsz=4096   blocks=0, rtextents=0\\nTrying to pull repository quay.io/openshift-on-azure/ci-azure ... \\nv4.0-652-gebbd5b9: Pulling from quay.io/openshift-on-azure/ci-azure\\ncd9779fb83c9: Pulling fs layer\\n0bc0fb725b14: Pulling fs layer\\n663bf8f9a8a3: Pulling fs layer\\n0bc0fb725b14: Verifying Checksum\\n0bc0fb725b14: Download complete\\n663bf8f9a8a3: Verifying Checksum\\n663bf8f9a8a3: Download complete\\ncd9779fb83c9: Verifying Checksum\\ncd9779fb83c9: Download complete\\ncd9779fb83c9: Pull complete\\n0bc0fb725b14: Pull complete\\n663bf8f9a8a3: Pull complete\\nDigest: sha256:1462ce6a226c8f20c8e3e9537f1ea8bcbecc2e111f7fcc05d2cf02e4eb164a60\\nStatus: Downloaded newer image for quay.io/openshift-on-azure/ci-azure:v4.0-652-gebbd5b9\\n\\n[stderr]\\n+ grep /var/lib/docker /etc/fstab\\n+ systemctl stop docker-cleanup.timer\\n+ systemctl stop docker-cleanup.service\\n+ systemctl stop docker.service\\n+ mkfs.xfs -f /dev/disk/azure/resource-part1\\n+ echo '/dev/disk/azure/resource-part1  /var/lib/docker  xfs  grpquota  0 0'\\n+ mount /var/lib/docker\\n+ restorecon -R /var/lib/docker\\n+ cat\\n+ systemctl start docker.service\\n+ systemctl start docker-cleanup.timer\\n+ grep /var/lib/etcd /etc/fstab\\n+ docker pull registry.access.redhat.com/openshift3/ose-node:v3.11.117\\n+ mkfs.xfs /dev/disk/azure/scsi1/lun0\\nmkfs.xfs: /dev/disk/azure/scsi1/lun0 appears to contain an existing filesystem (xfs).\\nmkfs.xfs: Use the -f option to force overwrite.\\n+ true\\n+ echo '/dev/disk/azure/scsi1/lun0  /var/lib/etcd  xfs  defaults  0 0'\\n+ mount /var/lib/etcd\\n+ restorecon -R /var/lib/etcd\\n+ for dst in tcp,8444\\n+ proto=tcp\\n+ port=8444\\n+ iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 8444 -j ACCEPT -w\\n+ iptables-save\\n++ hostname -d\\n+ [[ dtcbb0usb5zupohn2otef03ddc.bx.internal.cloudapp.net == '' ]]\\n+ docker pull quay.io/openshift-on-azure/ci-azure:v4.0-652-gebbd5b9\\n+ set +x\\n+ docker run --privileged --rm --network host -v /:/host:z -e SASURI quay.io/openshift-on-azure/ci-azure:v4.0-652-gebbd5b9 startup\\ngitCommit v4.0-652-gebbd5b9\\ntime=\\\"2019-08-08T19:26:10Z\\\" level=info msg=\\\"startup pod starting\\\"\\ntime=\\\"2019-08-08T19:26:10Z\\\" level=info msg=\\\"reading config\\\"\\ntime=\\\"2019-08-08T19:26:11Z\\\" level=info msg=\\\"writing startup files\\\"\\ntime=\\\"2019-08-08T19:26:11Z\\\" level=info msg=\\\"creating clients\\\"\\ntime=\\\"2019-08-08T19:26:11Z\\\" level=info msg=\\\"enriching config\\\"\\ntime=\\\"2019-08-08T19:26:11Z\\\" level=info msg=\\\"all done successfully\\\"\\n+ unset SASURI\\n+ update-ca-trust\\n++ hostname\\n+ sed -i -re 's#( *server: ).*#\\\\1https://master-000000#' /etc/origin/master/openshift-master.kubeconfig\\n++ hostname\\n+ sed -i -re 's#( *server: ).*#\\\\1https://master-000000#' /etc/origin/node/node.kubeconfig\\n++ hostname\\n+ sed -i -re 's#( *server: ).*#\\\\1https://master-000000#' /etc/origin/node/sdn.kubeconfig\\n++ hostname\\n+ sed -i -re 's#( *server: ).*#\\\\1https://master-000000#' /root/.kube/config\\n+ tuned-adm profile openshift-control-plane\\n+ mkdir -m 0750 -p /var/lib/origin/openshift.local.volumes\\n+ systemctl disable rsyslog.service\\nRemoved symlink /etc/systemd/system/multi-user.target.wants/rsyslog.service.\\n+ systemctl stop rsyslog.service\\n+ systemctl enable mdsd.service\\nFailed to execute operation: No such file or directory\\n\\\".\"" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).badRequest" file="/home/prow/go/src/github.com/openshift/openshift-azure/pkg/fakerp/util.go:24"
```

Where is the mdsd.service file getting written? The master hosts aren't finding it.

thekad:
#1838 needs to merge before this for create to succeed

ehashman:
#1838 is merged, please rebase

ehashman:
/retest

flake #1438 on upgrade-5.2

pkg/sync/v7/translate.go
ehashman: I'm going to make sure this passes the upgrade-v6.0 tests (which are mysteriously missing from CI).

We know that we have a working logging stack on a 5.1/5.2 -> HEAD upgrade so I'm relatively confident...

PR #1821, opened 2019-07-30
Adds a check for pending security updates

m1kola:
Adds a check for pending security updates for vmimage

We currently perform the check in the following cases:
* As part of the vmimage build process (`make vmimage`). In this case we validate a freshly created image. It should always pass unless we get unlucky and a new security update gets released between `yum -y update` on the kickstart stage of the build process and  `yum updateinfo list security -q` on the validation stage.
* As part of the vmimage validation (`make vmimage-validate`) where we validate already existing image.

[AZURE-326](https://jira.coreos.com/browse/AZURE-326)

```release-note
NONE
```


hack/vmimage-validate.sh
```diff
@@ -25,4 +25,4 @@ echo "Validating: ${IMAGE_RESOURCENAME}"
 go generate ./... && go run -ldflags "-X main.gitCommit=$(git rev-parse --short=10 HEAD)" ./cmd/vmimage -buildResourceGroup "${BUILD_RESOURCE_GROUP}" -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate -preserveBuildResourceGroup
 
 # Currently there are only 3 logs we want to capture
-mv /tmp/{yum_update_info,yum_check_update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
+mv /tmp/{yum_updateinfo,yum_updateinfo_list_security,yum_check-update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
```
ehashman: For some reason we have renamed the old files and are mixing hyphens and underscores. Can we consistently use snake case here?

hack/vmimage.sh
```diff
@@ -43,7 +43,7 @@ make e2e
 go generate ./... && go run -ldflags "-X main.gitCommit=$GITCOMMIT" ./cmd/vmimage -buildResourceGroup "${BUILD_RESOURCE_GROUP}" -imageResourceGroup "${IMAGE_RESOURCEGROUP}" -image "${IMAGE_RESOURCENAME}" -imageStorageAccount "${IMAGE_STORAGEACCOUNT}" -validate -preserveBuildResourceGroup
 
 # Currently there are only 3 logs we want to capture
-mv /tmp/{yum_update_info,yum_check_update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
+mv /tmp/{yum_updateinfo,yum_updateinfo_list_security,yum_check-update,scap-report.html} ${ARTIFACTS:-/tmp}/ || true
```
ehashman: Same feedback as previous comment.

pkg/vmimage/data/validate.sh
```diff
@@ -37,3 +38,10 @@ oscap xccdf eval \
   /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml > /tmp/oscap.log || true
   
 python -c 'import sys; from lxml import html; h = html.fromstring(open("/tmp/scap-report.html").read()); sys.exit(1) if len(h.find_class("rule-result rule-result-fail")) > 0 else sys.exit(0)'
+
+SECURITY_LIST_SIZE=$(du /tmp/yum_updateinfo_list_security | cut -f1)
+if [ "$SECURITY_LIST_SIZE" -ne 0 ]; then
```
ehashman: This `du` check seems a bit convoluted, why not use the [`-s` built-in](http://www.tldp.org/LDP/abs/html/fto.html)?

```suggestion
if [ -s "/tmp/yum_updateinfo_list_security" ]; then
```

hack/vmimage-validate.sh
m1kola: We previously had `yum updateinfo` logging into `/tmp/yum_update_info` (why `_` in the middle of `updateinfo`?) and `yum check-update` logging into `/tmp/yum_check_update`.

I changed the names to match commands: `yum check-update` now writes to `/tmp/yum_check-update` and `yum check-update` to `/tmp/yum_check-update`. So basically replacing space with underscore and leaving rest of the command as is.

I don't mind replacing both hyphens and spaces with underscores, if that's what you mean.

pkg/vmimage/data/validate.sh
m1kola: Thanks! I had no idea about `-s`. Will try tomorrow

hack/vmimage-validate.sh
ehashman: Yes, let's use underscores everywhere.

m1kola:
Re-running to see if [this is](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1821/pull-ci-azure-master-vmimage/20/build-log.txt) a flake.

/test vmimage

m1kola:
@ehashman please take another look. I'm looking into what is going on with the `vmimage` job, but I'm pretty sure it's not related to my PR. Probably something to do with CI secrets rotation.

/test vmimage

pkg/vmimage/data/validate.sh
```diff
@@ -21,19 +21,26 @@ EOF
...
   --cpe /usr/share/xml/scap/ssg/content/ssg-rhel7-cpe-dictionary.xml \
   /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml > /tmp/oscap.log || true
   
-python -c 'import sys; from lxml import html; h = html.fromstring(open("/tmp/scap-report.html").read()); sys.exit(1) if len(h.find_class("rule-result rule-result-fail")) > 0 else sys.exit(0)'
+python -c 'import sys; from lxml import html; h = html.fromstring(open("/tmp/scap_report.html").read()); sys.exit(1) if len(h.find_class("rule-result rule-result-fail")) > 0 else sys.exit(0)'
+
+if [ -s "/tmp/yum_updateinfo_list_security" ]; then
+    >&2 echo "Found pending security updates:"
+    >&2 cat /tmp/yum_updateinfo_list_security
+    exit 1
```
ehashman: Hm, do we want this to fail CI? That might be overkill.

/cc @kwoodson

m1kola: In general, it works similarly to the scap report, but it shouldn't really fail when we run `make vmimage`, because we run `yum -y update` as part of the kickstart while building an image.

In case of `make vmimage-validate` on an older image, I would prefer it to fail as an indicator that some of the checks do not pass.

Why do you think we should not fail here? I feel like I'm missing something, so I would appreciate some input.

m1kola: We might need to move `python -c '...'` and this check to `vmimage.sh` and `vmimage-validate.sh`. So we do not perform any check in CSE script. Instead we will be grabbing logs from a VM and perform check on the higher level (in `vmimage.sh` and `vmimage-validate.sh`).

As a result we will have logs (scap report, npm opuputs, etc) in the build artifacts. It will make debugging easier.

/hold

m1kola:
Rebased on top of #1836 which needs to be merged first. So putting this on hold

/hold

m1kola:
Rebased

/test vmimage

m1kola:
Another rebase

/test vmimage

m1kola:
#1836 got merged, so rebasing again and removing the hold

@ehashman @kwoodson please take another look

/test vmimage
/hold cancel

kwoodson:
I have not tested this with the file name changes. :+1:  to normalizing names.

/lgtm

PR #1831, opened 2019-08-06
Make sure we can update Geneva and PackageRepository Secrets

### LINKED ISSUE #1828: can't update latched PackageRepository key/cert
Scenario: a v6 cluster is created.  During the process, pkg/api/plugin.ToInternal() is called to initialise the configured PackageRepository key/cert in the internal cluster model.  Subsequently that cert expires.  There's currently no way to update the key/cert in the internal cluster model.

I think that originally we had planned for a Geneva Action to handle this use case, along with rotation of Geneva Logging/Metrics client certificates.  At the moment that action does not exist and creating it is somewhat heavyweight.

A simpler mechanism would be to hook the already existing mechanism by which we administratively upgrade clusters, i.e. by PUTting `config.pluginVersion: latest` in the admin API, and cause that to re-read the keys/certs.  I think this boils down to adding `|| setVersionFields` to each of the if statements in https://github.com/openshift/openshift-azure/blob/83d6b27a6e8fa5fb8ae4760322169c2a88f4181c/pkg/api/plugin/convertertointernal.go#L95-L112 .

This trick should only be necessary for configuration items not exposed by the admin API, i.e. keys/certs and secrets.  As such it is also worth considering the if statements in https://github.com/openshift/openshift-azure/blob/83d6b27a6e8fa5fb8ae4760322169c2a88f4181c/pkg/api/plugin/convertertointernal.go#L140-L145 .

Comments in the code + unit test also needed.

As all of this is part of the internal admin specification, it can always be changed later.

cc @ehashman

asalkeld:
Closes #1828

```release-note
Bugfix: ensure that Geneva and PackageRepository certificates can be updated once set.
```

### DIFF
--- a/pkg/api/plugin/convertertointernal.go
+++ b/pkg/api/plugin/convertertointernal.go
@@ -92,22 +92,24 @@ func ToInternal(in *Config, old *api.Config, setVersionFields bool) (*api.Config
 		c.Images.TLSProxy = inVersion.Images.TLSProxy
 	}
 
-	if c.Certificates.GenevaLogging.Key == nil {
+	// use setVersionFields to override the secrets below otherwise
+	// they become un-updatable..
+	if c.Certificates.GenevaLogging.Key == nil || setVersionFields {
 		c.Certificates.GenevaLogging.Key = in.Certificates.GenevaLogging.Key
 	}
-	if c.Certificates.GenevaLogging.Cert == nil {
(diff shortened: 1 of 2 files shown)


jim-minter:
lgtm, check the upgrade test failure though

ehashman:
Looks like compute-1565053530-000000 never came up on upgrade-5.1... I think it's a flake but let's confirm.

/retest

PR #1833, opened 2019-08-07
Delay VM Reboot After Security Updates to Prevent Race with Kubelet

### LINKED ISSUE #1827: SecurityPatchPackages reboot should happen before starting atomic-openshift-node
In {master,node}-startup.sh, the needs-restarting/reboot logic should take place after `systemctl enable atomic-openshift-node.service` but before `systemctl start atomic-openshift-node.service &`

That way there is no race between kubelet startup and reboot taking place.

cc @ehashman

charlesakalugwu:
This prevents a possible race condition between the VM reboot and the VM Kubelet's startup by scheduling the reboot at some point in the future.

```release-note
Delay VM reboot after security updates to prevent possible race with Kubelet's startup.
```

This also changes when the reboots happen: they will always happen whenever security updates are installed. 

Scheduling the reboot 5 minutes in the future seems to be a good threshold, allowing the kubelet to bootstrap successfully. I ran several tests with lower thresholds but there were always issues achieving a completely healthy cluster.

/cc @jim-minter @kwoodson @ehashman 

/closes #1827


pkg/arm/v7/data/master-startup.sh
```diff
@@ -169,11 +169,10 @@ systemctl stop rsyslog.service
...
 # note: atomic-openshift-node crash loops until master is up
 systemctl enable atomic-openshift-node.service
-systemctl start atomic-openshift-node.service &
 {{ if .Config.SecurityPatchPackages }}
-needs-restarting --reboothint &>/dev/null || {
-  logger -t master-startup.sh "rebooting $(hostname) to complete ARO security updates"
-  shutdown --reboot now
-}
+logger -t master-startup.sh "scheduling $(hostname) reboot to complete ARO security updates"
+shutdown --reboot +5
```
kwoodson: Is there a way to shrink this `5` to `2` or is this the right answer?  Just want to make sure this remains efficient.

pkg/arm/v7/data/node-startup.sh
```diff
@@ -102,11 +102,10 @@ systemctl stop rsyslog.service
...
 # note: atomic-openshift-node crash loops until master is up
 systemctl enable atomic-openshift-node.service
-systemctl start atomic-openshift-node.service &
 {{ if .Config.SecurityPatchPackages }}
-needs-restarting --reboothint &>/dev/null || {
-  logger -t node-startup.sh "rebooting $(hostname) to complete ARO security updates"
-  shutdown --reboot now
-}
+logger -t node-startup.sh "scheduling $(hostname) reboot to complete ARO security updates"
+shutdown --reboot +5
```
kwoodson: same question as above.

pkg/arm/v7/data/master-startup.sh
```diff
@@ -169,11 +169,10 @@ systemctl stop rsyslog.service
...
 {{end}}
+systemctl start atomic-openshift-node.service &
```
kwoodson: I know this is an artifact but what is the intention of the `&` at the end here?  Why do we want to background the start of openshift-node?

ehashman: +1, systemctl will handle the daemonization of this process. Not sure why we need to fork `systemctl` into the background.

pkg/arm/v7/data/master-startup.sh
mjudeikis: Just to start discussion: 
```
systemctl enable atomic-openshift-node.service
{{ if .Config.SecurityPatchPackages }}
needs-restarting --reboothint
if [ $? -eq 0 ]; then
     logger -t master-startup.sh "scheduling $(hostname) reboot to complete ARO security updates"
     shutdown --reboot now
else
    systemctl start atomic-openshift-node.service
fi
{{else}}
    systemctl start atomic-openshift-node.service
{{end}}
```
This way we don't even start node on the first update and we go via full cycle after reboot. This would prevent nodes going `Ready, NotReady, Ready`.

But my concert is for a hypothetical situation where `needs-restarting --reboothint` always returns `true` ? Which should not be a case.

charlesakalugwu: I did a few local test runs yesterday with values from 2 minutes to 5 minutes and all values except 5 minutes were flaky.

pkg/arm/v7/data/master-startup.sh
jim-minter: it's a hack - if a node tries to start before the masters are up (likely), the kubelet will `exit 1` after some time.  systemd will keep trying to start the kubelet, but the initial kubelet `exit 1` causes `systemctl start` to `exit 1`, and hence the script to `exit 1`.

Could do `|| true` here instead if preferred.

kwoodson: I'm fine either way but adding a note might be helpful for future.  Thanks @jim-minter

ehashman: I think the intent is clearer if we use `|| true` so I'd prefer that.

charlesakalugwu:
upgrade-v7.0 flake `Could not resolve host: github.com`
/retest

charlesakalugwu:
upgrade-v5.1 flake #1628 
/retest

Issue #1356, opened 2019-03-26
Timeout waiting for master-000000 to be ready


charlesakalugwu:
Sometimes we see the following errors"

```
time="2019-03-26T13:41:22Z" level=info msg="draining master-000000" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:38"
time="2019-03-26T13:41:22Z" level=info msg="deallocating master-000000" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:44"
time="2019-03-26T13:44:37Z" level=info msg="updating master-000000" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:50"
time="2019-03-26T13:45:09Z" level=info msg="reimaging master-000000" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:58"
time="2019-03-26T13:45:40Z" level=info msg="starting master-000000" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:64"
time="2019-03-26T13:47:44Z" level=info msg="waiting for master-000000 to be ready" func="github.com/openshift/openshift-azure/pkg/cluster.(*simpleUpgrader).UpdateMasterAgentPool" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/update_master.go:70"
time="2019-03-26T13:47:54Z" level=warning msg="dial tcp 104.45.65.206:443: i/o timeout: retry 1" func="github.com/openshift/openshift-azure/pkg/cluster/kubeclient.(*retryingRoundTripper).RoundTrip" file="/go/src/github.com/openshift/openshift-azure/pkg/cluster/kubeclient/roundtrip.go:102"
[...]

PR #1834, opened 2019-08-07
remove tenant flag and os.exit(1) from arm scripts

mjudeikis:
```release-note
NONE
```
Remove tenant flag as it causing strange behaviour when tagging images. 
Remove `os.exit(1)` from arm script because none of the validation should produce `os.exit(1)` in the arm scripts. This is causing arm template to fail before results being returned. 
Validation will be added in the follow-up PR

https://coreos.slack.com/archives/CCV9YF9PD/p1565195172015700


ehashman:
/retest

flake #1356

kwoodson:
@ehashman  @mjudeikis  @m1kola 
The issue with removing this is that we are now passing `vmimage` build when it **_should_** be failing.  There are 2 outstanding issues currently.  The purpose of failing with the python is so that we would not pass unintentionally.

The solution would be to turn them off in the xml file,  document the exceptions by adding them to the rules sheet.  I've basically done that in #1835 

- Configure audispd Plugin To Send Logs To Remote Server
- Encrypt Audit Records Sent With audispd Plugin

mjudeikis:
@kwoodson we should fail CI, and not tag images. What this does currently it - it is failing ARM and we can't even get artefacts because we fail before. We need to move python code into CI and check `/tmp/artifacts` and not in the ARM code script.

PR #1838, opened 2019-08-08
Create new VMImage with mdsd/azsecd/azsecmond

thekad:
```release-note
Base VM image now has mdsd/azsecd/azsecmond installed (but disabled) by default
```
cc @ehashman @jim-minter


pkg/vmimage/data/script.sh
```diff
@@ -164,7 +164,21 @@ yum -y install \
...
+    azsec-monitor \
+    fluentd \
+    rubygem-fluent-plugin-systemd
+
+# openshift-audit dir ends up with wrong permissions, preemptively fix that
+mkdir -pv /var/log/openshift-audit
+chmod -v 0750 /var/log/openshift-audit
+
+# mdsd/azsecd/azmond are enabled by default (but misconfigured), disable them until configured
+find /etc/systemd -name azsecd\\* -or -name azsecmond\\* -or -name mdsd\\* -type l -delete
```
ehashman: I'd prefer that we disable the systemd units rather than delete these files, especially if the files are owned by the system packages.

Although, typically packages install their systemd service files into `/lib/systemd/` (and user overrides are installed in `/etc/systemd/`), so I'd be surprised if there was anything there; is that different on RHEL?

thekad: Services install the files to `/lib/systemd` and `systemctl enable` creates the symlink in `/etc/systemd/system/<level>`. Every time I have pushed a PR with a systemctl call in vmimage @jim-minter says I should just create the symlink instead, I have no opinion going either way, just erred on the side of history :man_shrugging:

jim-minter: systemctl disable/mask/whatever is fine here

thekad:
Two different kinds of timeout exceptions, seems odd, failure in the ci cluster? the error in e2e-create is also fishy, 
/retest

PR #1840, opened 2019-08-09
install golang 1.11 from source

mjudeikis:
```release-note
NONE
```

Install golang 1.11 from source, because rhel7 latest RAW ships with 1.9 by default.


mjudeikis:
bad rebase

PR #1844, opened 2019-08-11
Improve the docker pull retries in the startup script

asalkeld:
When quay.io is unavailable the user has to wait for the full cluster timeout
before knowing that it has failed as the pull will retry for ever.

Now it will retry every minute 5 times.

We now also log the reties.

```release-note
NONE
```


ehashman:
Please rebase against v8

asalkeld:
> Does it matter if the background `docker pull` of node image fails?
> 
> ```shell
> docker pull {{ .Config.Images.Node }} &>/dev/null &
> ```

I don't know why that is there... it is used in "systemctl enable atomic-openshift-node.service". That should fail if it can't get the image. at a guess, this is to just make startup faster, so we probably don't want to block on that.

m1kola:
Thanks for the explanation @asalkeld 

Changes look good to me, but there is a conflict in bindata

ehashman:
LGTM but needs rebase.

PR #1845, opened 2019-08-12
Allows make vmimage-validate to validate published marketplace vmimages

m1kola:
We can now run `IMAGE_VERSION=latest IMAGE_SKU=osa_311 make vmimage-validate` to validate the latest published marketplace vmimage or `IMAGE_VERSION=311.129.20190810 IMAGE_SKU=osa_311 make vmimage-validate` to validate a specific version.

Jira ref: AZURE-326

```release-note
NONE
```

### DIFF
--- a/pkg/vmimage/vmimage.go
+++ b/pkg/vmimage/vmimage.go
@@ -7,14 +7,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 	azresources "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/openshift-azure/pkg/util/arm"
 	"github.com/openshift/openshift-azure/pkg/util/azureclient/resources"
+	"github.com/openshift/openshift-azure/pkg/util/resourceid"
 	"github.com/openshift/openshift-azure/pkg/util/template"
 	"github.com/openshift/openshift-azure/pkg/util/tls"
 )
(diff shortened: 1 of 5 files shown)


m1kola:
Fixed import formatting in tests (the reason why the `verify` job failed).

/test vmimage

m1kola:
@kwoodson could you also please take a look at this? You recently worked with vmimage build/validation so hopefully it will be easy for you to review this.

/cc @kwoodson

pkg/vmimage/vmimage.go
```diff
@@ -114,8 +131,85 @@ func (builder *Builder) generateTemplate() (map[string]interface{}, error) {
 	return template, nil
 }
 
+func (builder *Builder) vmImageReference() *compute.ImageReference {
```
kwoodson: I like this function consolidating the options.

pkg/vmimage/vmimage.go
```diff
@@ -114,8 +131,85 @@ func (builder *Builder) generateTemplate() (map[string]interface{}, error) {
...
+			Offer:     to.StringPtr("RHEL"),
+			Sku:       to.StringPtr("7-RAW"),
+			Version:   to.StringPtr("latest"),
+		}
+	}
+
+	// Validating an existing marketplace image
+	if builder.hasMarketplaceVMImageRef() {
+		return &compute.ImageReference{
+			Publisher: to.StringPtr(builder.ImagePublisher),
```
kwoodson: suggestion/discussion: 
I know hard coding is bad but should we verify that the image reference is our exact one?

m1kola: I would like to avoid hardcoding SKU and version, because I guess we will have different SKU and versions for v4 openshift, but we can hardcode the publisher and offer. However I do not see why we would want to do this. Do you have any reasons in mind?

kwoodson: If the variables are tied to our publishing image then this prevents fat fingers and guarantees our publishing information is set. These will hopefully never change.  Just a suggestion.

I will not block this PR because of this.

m1kola: To guarantee that publisher and offer is set I added this info into `hack/vmimage-validate.sh` with ability to override, if needed:

```
IMAGE_PUBLISHER="${IMAGE_PUBLISHER:-redhat}"
IMAGE_OFFER="${IMAGE_OFFER:-osa}"
```

But taking into account that this is not going to change often (likely to never change), I think we can simplify my changes by removing these variables from the validation script, removing flags from the go command and just hardcode publisher and offer names here.

I'm happy to do this.

m1kola:
@kwoodson I updated the PR. Please take another look.

PR #1847, opened 2019-08-13
[wip] [proposal] arm versioning auto-load 

mjudeikis:
```release-note
NONE
```

This PR Externalises ARM Interface, adds it usage into the versioned packages and created "register" so versioned packages automatically register version they provide to the global registry. 

We discussed this on one of the scrums. Any ideas on this approach?
 
Ignore bindata changes for now, please :)


mjudeikis:
/close
I think it not worth the effort for now

Issue #1620, opened 2019-05-03
Follow-up after https://github.com/openshift/origin/pull/22742


jim-minter:
Follow-up after https://github.com/openshift/origin/pull/22742:

* When the DisableOutboundSNAT setting becomes available in OCP, start using it.  Check the upgrade behaviour.

* Investigate if `Microsoft.Storage/storageAccounts/read` can be removed from master and worker RBAC.

@mjudeikis @ingvagabund fyi

asalkeld:
@mjudeikis close this?

PR #1492, opened 2019-04-11
Enable cloud API rate limiting for nodes

ehashman:
```release-note
Enable cloud API rate limiting for nodes
```

Closes #1480.

PR #1854, opened 2019-08-14
add DisableOutboundSNAT for cloudconfig

mjudeikis:
```release-note
- Disable DisableOutboundSNAT for VMSS
```

Prepare for https://github.com/openshift/openshift-azure/issues/1620
context: https://github.com/kubernetes/kubernetes/pull/74490
https://bugzilla.redhat.com/show_bug.cgi?id=1703512

This is preparation PR. We don't have code yet released as errata but this will enable us to configure this option when we get it fast.

### DIFF
--- a/pkg/startup/v8/derive.go
+++ b/pkg/startup/v8/derive.go
@@ -62,11 +62,11 @@ func (derivedType) KubeReserved(cs *api.OpenShiftManagedCluster, role api.AgentP
 }
 
 func (derivedType) MasterCloudProviderConf(cs *api.OpenShiftManagedCluster) ([]byte, error) {
-	return derivedpkg.MasterCloudProviderConf(cs, true)
+	return derivedpkg.MasterCloudProviderConf(cs, false)
 }
 
 func (derivedType) WorkerCloudProviderConf(cs *api.OpenShiftManagedCluster) ([]byte, error) {
-	return derivedpkg.WorkerCloudProviderConf(cs, true)
+	return derivedpkg.WorkerCloudProviderConf(cs, false)
 }
 
 // MaxDataDisksPerVM is a stopgap until k8s 1.12.  It requires that a cluster
--- a/pkg/util/derived/derived.go
+++ b/pkg/util/derived/derived.go
@@ -35,6 +35,7 @@ func baseCloudProviderConf(cs *api.OpenShiftManagedCluster, disableOutboundSNAT 
 	return &cfg, nil
 }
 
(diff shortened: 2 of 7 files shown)


mjudeikis:
hmm. Any cloud provider changes will end up in hash changes for everything. Might need to consider moving them into versioned folders

ehashman:
There's a hacky way to avoid the hash changes: https://github.com/openshift/openshift-azure/pull/1492/files#diff-2018244c6f905eadd8b2581f21a978cbR10

mjudeikis:
chose more safe option instead of trying to bring all upstream into our codebase :/

pkg/util/derived/derived.go
```diff
@@ -30,18 +33,33 @@ func baseCloudProviderConf(cs *api.OpenShiftManagedCluster, cloudProviderBackoff
 		cfg.CloudProviderRateLimitQPS = 3.0
 		cfg.CloudProviderRateLimitBucket = 10
 	}
-	return &cfg
+	// match any version v8+
+	// TODO: make this default when all clusters are v8+
+	version, err := regexp.Compile("^v([8-9]|\\d{2,}).*")
```
ehashman: Hrm I am not a fan of doing regex semver matching, and this introduces a new exciting way to feature flag config changes between earlier versions... can you just add another bool like cloudProviderBackoff that we can purge once the code that uses it is gone?

pkg/startup/v8/derive.go
```diff
@@ -62,11 +62,11 @@ func (derivedType) KubeReserved(cs *api.OpenShiftManagedCluster, role api.AgentP
 }
 
 func (derivedType) MasterCloudProviderConf(cs *api.OpenShiftManagedCluster) ([]byte, error) {
-	return derivedpkg.MasterCloudProviderConf(cs, true)
+	return derivedpkg.MasterCloudProviderConf(cs, false)
```
ehashman: Is this plan to turn this on later in v8 once the setting is actually available?

mjudeikis: yes :) as soon as we get errata released and tested

jim-minter:
@mjudeikis why did you think that https://bugzilla.redhat.com/show_bug.cgi?id=1703512 would enable us to set disableOutboundSNAT?

mjudeikis:
@jim-minter this was follow-up for https://github.com/openshift/openshift-azure/issues/1620 as these options were backported(?). Or I misunderstood 1620?

PR #1850, opened 2019-08-13
gobindata remove pin

mjudeikis:
```release-note
NONE
```

Stop pinning go-bindata to commit

related to: https://github.com/openshift/release/pull/4727 
old topic: https://github.com/openshift/openshift-azure/pull/862 
fix: https://github.com/openshift/openshift-azure/issues/1639 
fix: https://github.com/openshift/openshift-azure/issues/1045

PR #1855, opened 2019-08-14
Update bindata for v8

m1kola:
There were a race condition between #1850 and #1852, so we need to update v8 bindata.

```release-note
NONE
```


m1kola:
@ehashman CC'ing you as we were discussing this issue in slack.

/cc @ehashman

m1kola:
Will be fixing it in #1856 as per https://github.com/openshift/openshift-azure/pull/1856#issuecomment-521303879

PR #1860, opened 2019-08-15
Removes unused code

m1kola:
```release-note
NONE
```


m1kola:
The above failures seem to happen because of hitting the subscription limits. Retesting.

/retest

PR #1867, opened 2019-08-20
Add log analytics rsyslog config file for deterministic syslog forwarding

charlesakalugwu:
This adds a rsyslog config file to the log analytics node agent in order to ensure that the out-of-the-box log forwarding experience is similar across all ARO clusters. This configuration option is documented [here](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/data-sources-syslog#configure-syslog-on-linux-agent) 

```release-note
NONE
```

/cc @asalkeld @kwoodson @thekad 

/hold


charlesakalugwu:
upgrade-v6.0 flake #1628 
upgrade-v7.0 flake #1628 
/retest

PR #1872, opened 2019-08-21
Modifying vmimage script to use the upstream

gburges:
```release-note
NONE
```


ehashman:
Why are so many v6/v7 bindata files being updated here? That doesn't seem right to me, but verify doesn't seem to be complaining.

gburges:
> Why are so many v6/v7 bindata files being updated here? That doesn't seem right to me, but verify doesn't seem to be complaining.

good question, I _think_ it's the go-bindata pin removal. I saw previously, some (un-)capitalization of words, and rewrites. I left the other PR (1822) as-is if we want to reference the failing verify test there

gburges:
/retest vmimage

quay.io timeout

mjudeikis:
Looks like something is off in the new VM image. need to run locally with this image and check why cluster is not starting.

gburges:
> Looks like something is off in the new VM image. need to run locally with this image and check why cluster is not starting.

Locally it ran successfully

y-cote:
> Why are so many v6/v7 bindata files being updated here? That doesn't seem right to me, but verify doesn't seem to be complaining.

I suspect the use of different go tools would do that.

y-cote:
Tried to verify your PR and hit a flake at the last command of the script it seems with : 

+ az image create -g images -n rhel7-3.11-201908252248 --source   ... 

FATA[2019-08-25T16:57:09-07:00][.../openshift/openshift-azure/cmd/vmimage/vmimage.go:140] main.main() Future#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded

PR #1875, opened 2019-08-22
Enabling boot diagnostics on all VMSS

thekad:
```release-note
Enabling boot diagnostics means serial console logs are available as
well as serial console access.
```

Using the `config` storage account to store the boot diagnostics
information, since this SA is already present by the time ARM is
deployed

### DIFF
--- a/pkg/arm/v8/resources.go
+++ b/pkg/arm/v8/resources.go
@@ -2,6 +2,7 @@ package arm
 
 import (
 	"encoding/base64"
+	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-07-01/network"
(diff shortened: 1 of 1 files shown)


thekad:
cc @jim-minter

asalkeld:
Nice, this looks like a good way to see errors in the startup script!

pkg/arm/v8/resources.go
```diff
@@ -441,6 +442,12 @@ func vmss(cs *api.OpenShiftManagedCluster, app *api.AgentPoolProfile, backupBlob
 						},
 					},
 				},
+				DiagnosticsProfile: &compute.DiagnosticsProfile{
```
ehashman: Relevant SDK docs:

- https://godoc.org/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute#VirtualMachineScaleSetVMProfile
- https://godoc.org/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute#BootDiagnostics

This LGTM, the only thing I'm not 100% sure of is whether this is the right storage account. @mjudeikis can you TAL?

Issue #1710, opened 2019-06-11
Unsolicited response received on idle HTTP channel starting with "HTTP/1.0 408 Request Time-out[...]"


thekad:
Getting the title error message when the e2e test is run some times, example:

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1709/pull-ci-azure-master-e2e-create/333/build-log.txt

/kind test-flake

thekad:
Found openshift/release#378 but that wasn't helpful

asalkeld:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1876/pull-ci-azure-master-e2e-create/884

asalkeld:
https://giedrius.blog/2018/10/25/gos-http-transport-and-408-response-code-whats-the-relationship/

asalkeld:
I see we don't use the RoundTripper (which disables keepalives) in the sync pod. I'll post a PR that enables it.

PR #1876, opened 2019-08-22
Make sure the allowed admin config update pass through

### LINKED ISSUE #1873: SSHSourceAddressPrefixes list does not get updated during admin upgrade
**issue description:**
Customer creates an ARO cluster and `SSHSourceAddressPrefixes` is used to create NSG to allow SSH access from certain IP ranges only. Then later a deployment of the service/RP changes/updates `SSHSourceAddressPrefixes` --> 
- All new ARO clusters should get new `SSHSourceAddressPrefixes` values configured in the NSG: This works as expected today.
- An admin initiated upgrade should also trigger the change to update `SSHSourceAddressPrefixes` in upgraded ARO clusters. However, today admin upgrade does update this on upgrade cluster.

Perusing the code shows that upgrade path might not be overriding this value (it should):
https://github.com/openshift/openshift-azure/blob/master/pkg/api/plugin/convertertointernal.go#L42

https://github.com/openshift/openshift-azure/blob/master/pkg/api/validate/admin.go#L68

From the RP side we have validated that plugin template with updated `SSHSourceAddressPrefixes ` is being passed to the plugin by the RP.

Tested using plugin version: v7.0

**Expected behavior:**
During upgrade plugin should override `SSHSourceAddressPrefixes ` with values provided to it by the RP (passed during plugin instantiation: `plugin.NewPlugin`).

asalkeld:
closes #1873

```release-note
allow the following to be set using the admin API:
- SSHSourceAddressPrefixes
- SecurityPatchPackages
- ComponentLogLevel
```

/cc @jim-minter @ehashman

### DIFF
--- a/pkg/api/plugin/convertertointernal.go
+++ b/pkg/api/plugin/convertertointernal.go
@@ -17,27 +17,16 @@ func ToInternal(in *Config, old *api.Config, setVersionFields bool) (*api.Config
 		c = old.DeepCopy()
 	}
 
-	if c.SecurityPatchPackages == nil {
-		c.SecurityPatchPackages = in.SecurityPatchPackages
-	}
-
 	// setting c.PluginVersion = in.PluginVersion is done up-front by the plugin
 	// code.  It could be done here as well (gated by setVersionFields) but
 	// would/should be a no-op.  To simplify the logic, we don't do it.
 
-	if c.ComponentLogLevel.APIServer == nil {
-		c.ComponentLogLevel.APIServer = &in.ComponentLogLevel.APIServer
-	}
-	if c.ComponentLogLevel.ControllerManager == nil {
-		c.ComponentLogLevel.ControllerManager = &in.ComponentLogLevel.ControllerManager
-	}
-	if c.ComponentLogLevel.Node == nil {
-		c.ComponentLogLevel.Node = &in.ComponentLogLevel.Node
-	}
+	c.ComponentLogLevel.APIServer = &in.ComponentLogLevel.APIServer
+	c.ComponentLogLevel.ControllerManager = &in.ComponentLogLevel.ControllerManager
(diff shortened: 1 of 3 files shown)


pkg/api/plugin/convertertointernal.go
```diff
@@ -17,27 +17,16 @@ func ToInternal(in *Config, old *api.Config, setVersionFields bool) (*api.Config
...
-	}
+	c.ComponentLogLevel.APIServer = &in.ComponentLogLevel.APIServer
+	c.ComponentLogLevel.ControllerManager = &in.ComponentLogLevel.ControllerManager
+	c.ComponentLogLevel.Node = &in.ComponentLogLevel.Node
 
-	if c.SSHSourceAddressPrefixes == nil {
-		c.SSHSourceAddressPrefixes = in.SSHSourceAddressPrefixes
-	}
+	c.SecurityPatchPackages = in.SecurityPatchPackages
+	c.SSHSourceAddressPrefixes = in.SSHSourceAddressPrefixes
```
ehashman: Why are some of these things being passed directly vs. some of them being passed by reference? Is it because these latter two settings are slices? Just curious.

ehashman:
If I'm reading this PR right, it also fixes a bug in which once the SecurityPatchPackages were set, they can't be updated? If so, can we update the changenotes to include that (and indicate that this is a bugfix)?

pkg/api/plugin/convertertointernal.go
asalkeld: Yeah, it's just 'cos they are slices. That they are not pointers is the real problem here because we don't know what has been provided new and what hasn't, BUT honestly I don't think we ever send partial updates (without some of the config) so I don't think that is an issue.

asalkeld:
/retest
flake #1710

PR #1881, opened 2019-08-23
Allow customer-admins to update the loganalytics agent config

charlesakalugwu:
This gives the `osa-customer-admins` group the ability to get, update or patch the data collection config `container-azm-ms-agentconfig` in the `openshift-azure-logging` namespace used by the log analytics agents.

AKS allows their customers to edit this configmap and the same ability is expected to be demanded by ARO customers when monitoring rolls out. 

```release-notes
Allow customer-admin to update the data collection config for the Log Analytics agent
```

/cc @amanohar @asalkeld


mjudeikis:
How dow we make sure sync pod will not convert config back to the original version?

pkg/sync/v9/client.go
```diff
@@ -294,6 +294,12 @@ func (s *sync) isReconcileProtected(existing, o *unstructured.Unstructured) bool
 			(o.GroupVersionKind().GroupKind() == schema.GroupKind{Group: "rbac.authorization.k8s.io", Kind: "ClusterRoleBinding"}) {
 			return true
 		}
+		// log analytics agent data collection configuration
```
thekad: I think it would make more sense to unset the "owned-by-sync-pod" annotation in the CM ?

charlesakalugwu: sounds good

charlesakalugwu:
> @thekad had a good suggestion, just set owned-by-sync-pod=false directly in the annotation

yep. its coming up

> @thekad had a good suggestion, just set owned-by-sync-pod=false directly in the annotation

@thekad @asalkeld @mjudeikis reconciliation of the configmap changes will be achieved (and documented) by setting the `azure.openshift.io/owned-by-sync-pod` label on the configmap to false.

charlesakalugwu:
On second thoughts, I felt like the `azure.openshift.io/owned-by-sync-pod` label is one which should only be set by the plugin/sync pod. The `openshift.io/reconcile-protect` annotation seemed more appropriate, in that case, for customer-facing disablement of sync pod reconciliation of resources. 

I also discussed this with the team and we collectively decided to go with `openshift.io/reconcile-protect` here. 

@thekad @asalkeld

pkg/sync/v9/client.go
charlesakalugwu: @thekad discussed this with the team and we decided to go with the `openshift.io/reconcile-protect` annotation since the `azure.openshift.io/owned-by-sync-pod` label should not be touched by the customer.

charlesakalugwu:
e2e-create flake #1932 
upgrade-v7.0 flake #1932 
/retest

PR #1882, opened 2019-08-23
Add check for checksum for downloaded golang binary

troy0820:
```release-note
NONE
```
* Add check to see if CHECKSUM for golang binary matches what we downloaded


pkg/vmimage/data/script.sh
```diff
@@ -11,8 +11,15 @@ xfs_growfs /dev/sda2
 yum -y update -x WALinuxAgent # updating WALinuxAgent kills this script
 yum -y install git libguestfs-tools-c libvirt-daemon-config-network virt-install
 
+CHECKSUM=14ec881815eb9e6618f95df5eb385d961283efc196d97912595ba6484a56180d
 # install golang-1.11 from source because rhel7 ships with golang 1.9 by default
-curl -s https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz | tar -C /usr/local -xz
+curl -s https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz --output /tmp/go.1.11.12.linux.amd64.tar.gz
```
ehashman: File here got renamed and then doesn't match the path below :)

```suggestion
curl -s https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz --output /tmp/go.1.11.12.linux-amd64.tar.gz
```

pkg/vmimage/data/script.sh
```diff
@@ -11,8 +11,15 @@ xfs_growfs /dev/sda2
...
+if [[ $(sha256sum /tmp/go.1.11.12.linux-amd64.tar.gz | awk '{print $1}') == $CHECKSUM ]]; then
```
ehashman: Might be simpler to use `sha256sum`'s `-c` flag here:

```suggestion
echo "14ec881815eb9e6618f95df5eb385d961283efc196d97912595ba6484a56180d  /tmp/go.1.11.12.linux-amd64.tar.gz" | sha256sum -c
```

Note that since we've `set -x`, if this errors out, then we will automatically exit (and don't need to include the if-else).

pkg/vmimage/data/script.sh
troy0820: I hate when that happens 😞

pkg/vmimage/data/script.sh
troy0820: NICE!! I like the -c option, so much easier.

troy0820: This makes sense, the `-x` will automatically exit.

pkg/vmimage/data/script.sh
```diff
@@ -12,7 +12,9 @@ yum -y update -x WALinuxAgent # updating WALinuxAgent kills this script
 yum -y install git libguestfs-tools-c libvirt-daemon-config-network virt-install
 
 # install golang-1.11 from source because rhel7 ships with golang 1.9 by default
-curl -s https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz | tar -C /usr/local -xz
+curl -s https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz --output /tmp/go1.11.12.linux-amd64.tar.gz
+echo "14ec881815eb9e6618f95df5eb385d961283efc196d97912595ba6484a56180d /tmp/go1.11.12.linux-amd64.tar.gz" | sha256sum -c
+cd /tmp && tar -xzf go1.11.12.linux-amd64.tar.gz -C /usr/local
```
ehashman: No need to cd here I think, can we just specify the full path of the tarball?

ehashman:
For the failing verify test, you may need to unpin and update your go-bindata. There was an email on how to do that, lmk if you need me to forward it :)

pkg/vmimage/data/script.sh
troy0820: Very true.  Weird directory issues 😬 I was facing.

troy0820:
Found the email and did the steps but when I do a `make verify` locally, I get some failures.  I generated the same bindata.go with my last PR that only touched docs.  🤔  @ehashman

ehashman:
Oh, `verify` is failing because you haven't checked in the bindata corresponding to the changes you made in this PR. Make sure you include that.

troy0820:
@ehashman That seemed to work 🤦‍♂ .  I didn't do a `make generate`.

PR #1883, opened 2019-08-23
Add ci-notify to vmimage build.

kwoodson:
Adding ci notifications to vmbuild.

```release-note:
None
```

### DIFF
--- a/hack/ci-notify/main.go
+++ b/hack/ci-notify/main.go
@@ -1,0 +1,184 @@
+package main
+
+/*
+
+A list of available environment variables can be found here:
+
+https://github.com/kubernetes/test-infra/blob/a9f967cb1235916fb5a4eca661fa083413211242/prow/jobs.md#job-environment-variables
+
+For periodics:
+- JOB_NAME
+- JOB_TYPE
+- JOB_SPEC
+- BUILD_ID
+- PROW_JOB_ID
+
+*/
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+var (
+	jobName   = flag.String("job-name", "", "Name of the CI job  e.g.  periodic-ci-azure-vmimage")
+	comment   = flag.String("comment", "", "The comment to add.")
+	owner     = flag.String("owner", "openshift", "Name of github organization.")
+	repo      = flag.String("repo", "openshift-azure", "Name of github repository")
+	creator   = flag.String("creator", "openshift-azure-robot", "Github user who will send the notifications.")
+	duration  = flag.Duration("since", 504*time.Hour, "How many hours in the past to be considered during the search since the last issue.")
+	succeeded = flag.Bool("success", false, "Did the ci job succeed?")
+)
+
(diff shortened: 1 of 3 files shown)


hack/vmimage.sh
```diff
@@ -18,25 +25,32 @@ else
   GITCOMMIT="$TAG-dirty"
 fi
 
+false
```
mjudeikis: what is this?

hack/ci-notify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Print("No issue found.  Creating one")
+		ir := &github.IssueRequest{
+			Title:     to.StringPtr(jobName),
+			Body:      to.StringPtr(fmt.Sprintf("The %s build has failed.  Please check the following [link](%s)<br/><br/>%s", jobName, getJobLogsURL(jobName), comment)),
+			Assignees: to.StringSlicePtr(users),
+		}
+		_, _, err = c.gh.Issues.Create(ctx, owner, repo, ir)
+		if err != nil {
+			return err
+		}
```
mjudeikis: return `nil` and drop `else` ?

hack/vmimage.sh
```diff
@@ -3,12 +3,19 @@
 cleanup() {
   set +e
 
+  if [[ -e /var/run/secrets/kubernetes.io ]]; then
+    if [ $PHASE == "build_complete" ]; then
+      ARGS="-success"
+    fi
+    go run ./hack/ci-notify/main.go -job-name periodic-ci-azure-vmimage -comment "Phase: $PHASE" $ARGS
```
m1kola: Do we need to hard code `periodic-ci-azure-vmimage` here? Can we replace it with `$JOB_NAME`?

hack/vmimage.sh
```diff
@@ -3,12 +3,19 @@
 cleanup() {
   set +e
 
+  if [[ -e /var/run/secrets/kubernetes.io ]]; then
```
m1kola: Notification is not part of a cleanup. Can we move this check into a separate function-trap?

hack/ci-notify/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+	issues, _, err := c.gh.Issues.ListByRepo(ctx, owner, repo, opt)
+	if err != nil {
+		return nil, err
+	}
+
+	return issues, nil
+}
+
+func (c *client) getVMImageBuildIssue(ctx context.Context, jobName, owner, repo string, since time.Duration) (*github.Issue, error) {
```
m1kola: The name is very specific. I think we will be using this script not only for `vmimage` builds (`vmimage-validate`, for example). Maybe`getIssueByName` or something like that?

hack/vmimage.sh
```diff
@@ -3,12 +3,20 @@
 cleanup() {
   set +e
 
+  if [[ -e /var/run/secrets/kubernetes.io ]] && [ "${JOB_TYPE} == "periodic" ]; then
+    if [ $PHASE == "build_complete" ]; then
```
m1kola: In traps we can check exit code with with `$?` ;)

m1kola: Ah, right! We can make it work somehow like this:

```bash
    exit_code=$?
    set +e
```

But I'm not sure if the efforts worth it.

Issue #1617, opened 2019-05-03
fake rp server not starting


asalkeld:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1583/pull-ci-azure-master-upgrade-v5.0/14

```
./hack/create.sh ci-1583-e2e-upgrade-umtbdc
time="2019-05-02T08:25:01Z" level=info msg="using region northeurope"
time="2019-05-02T08:25:01Z" level=info msg="ensuring resource group ci-1583-e2e-upgrade-umtbdc"
time="2019-05-02T08:26:04Z" level=fatal msg="timed out waiting for the condition"
exit status 1
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
time="2019-05-02T08:26:10Z" level=info msg="starting the fake resource provider" func=main.main file="/tmp/tmp.6kMNo70TBC/src/github.com/openshift/openshift-azure/cmd/fakerp/main.go:20"
time="2019-05-02T08:26:10Z" level=info msg="validating external plugin api data models" func="github.com/openshift/openshift-azure/pkg/plugin.(*plugin).ValidatePluginTemplate" file="/tmp/tmp.6kMNo70TBC/src/github.com/openshift/openshift-azure/pkg/plugin/plugin.go:86"
time="2019-05-02T08:26:10Z" level=info msg="starting server on localhost:8080" func="github.com/openshift/openshift-azure/pkg/fakerp.(*Server).Run" file="/tmp/tmp.6kMNo70TBC/src/github.com/openshift/openshift-azure/pkg/fakerp/server.go:69"
{"component":"entrypoint","level":"error","msg":"Process did not finish before 4h0m0s timeout","time":"2019-05-02T12:20:39Z"}
{"component":"entrypoint","level":"error","msg":"Process did not exit before 15s grace period","time":"2019-05-02T12:20:54Z"}
```

/kind test-flake

Makdaam:
same in
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1471/pull-ci-azure-master-e2e-create/139
and
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1471/pull-ci-azure-master-upgrade-v5.0/16

ehashman:
[...]

PR #1884, opened 2019-08-26
Fix fakerp "trap" and used azureclient.resources.Group

asalkeld:
```release-note
NONE
```

relates to #1617


asalkeld:
> I understand most of this PR, but what's the benefit to mocking the groups client and using that here?

Hi @ehashman , it's so that we can start using https://github.com/openshift/openshift-azure/blob/master/pkg/util/azureclient/retry.go rather than inventing a new mechanism.

PR #1890, opened 2019-08-27
fix sync e2e test to ignore the first run

### LINKED ISSUE #1778: Sync e2e fails with unexpected log output

asalkeld:
1. There *are* lots of Updates in the first sync run
2. The sync e2e test was incorrectly passing except when it thought there was only one sync run (there was only one run, or there was an error causing a missing "sync done" log message.
3. There were continuous Updates to SecurityContextConstraints.security.openshift.io/privileged by https://github.com/openshift/openshift-azure/blob/master/pkg/cluster/kubeclient/syncpod.go#L30-L48

```
 unstructured.Unstructured{
+   "allowedUnsafeSysctls":     []interface{}{string("*")},
    "users": []interface{}{
-    string("system:serviceaccount:kube-system:sync"),
    },
```

What this PR does is to fix the e2e test:
1. make sure we have more than one sync run
2.  make sure there are no error logs in the sync run
3.  ignore the first sync run
4. check all remaining sync runs for Updates

Then the continuous Updates to SecurityContextConstraints.security.openshift.io/privileged are fixed by:
1. add the missing system:serviceaccount:kube-system:sync to the sync data
2. the allowedUnsafeSysctls gets removed by syncpod.go because our openshift api is too old and does not have this field, so it get removed during this update. I have updated openshift/api and the client to version release-3.11.

```release-note
NONE
```
@fixes #1778 
/cc @jim-minter


test/e2e/specs/fakerp/sync.go
```diff
@@ -18,12 +20,28 @@ var _ = Describe("sync pod tests [EveryPR]", func() {
...
-		rx := regexp.MustCompile(`(?ms)^[^\n]* msg="starting sync"[^\n]*$.*?^[^\n]* msg="sync done"[^\n]*$`)
-		runs := rx.FindAllString(string(b), -1)
-		// check for constantly updated objects
-		Expect(runs).To(ContainElement(Not(ContainSubstring("Update"))))
+			rx := regexp.MustCompile(`(?ms)^[^\n]* msg="starting sync"[^\n]*$.*?^[^\n]* msg="sync done"[^\n]*$`)
+			runs := rx.FindAllString(string(b), -1)
+			if len(runs) > 1 {
+				for i, run := range runs {
+					Expect(run).To(Not(ContainSubstring("level=error")))
+					if i < 3 {
```
asalkeld: oops, was testing. should be 1

test/clients/openshift/template.go
```diff
@@ -23,7 +23,7 @@ func init() {
 	// additional OpenShift types must be registered manually.  AddToScheme
 	// registers the API group types (e.g. route.openshift.io/v1, Route) only.
 
-	if err := templatev1.AddToSchemeInCoreGroup(scheme.Scheme); err != nil {
+	if err := templatev1.Install(scheme.Scheme); err != nil {
```
asalkeld: this is causing the test fail, busy investigating..

jim-minter:
Ideally we really ought to be bumping all the k8s libraries to 1.11 if we're bumping the openshift libraries to 3.11.

Either way, before merging this we need to ensure that MSFT are prepared and willing to revendor the equivalent libraries when they revendor us (and the effort to do this will no doubt slow them down because this kind of thing is always fiddly).

jim-minter:
(for all the libraries in question, best would be to use the same ones which are vendored in openshift/origin release-3.11 if this is possible)

vendor/github.com/openshift/api/security/v1/types.go
```diff
@@ -96,6 +104,26 @@ type SecurityContextConstraints struct {
...
+	// AllowedUnsafeSysctls is a list of explicitly allowed unsafe sysctls, defaults to none.
+	// Each entry is either a plain sysctl name or ends in "*" in which case it is considered
+	// as a prefix of allowed sysctls. Single * means all unsafe sysctls are allowed.
+	// Kubelet has to whitelist all allowed unsafe sysctls explicitly to avoid rejection.
+	//
+	// Examples:
+	// e.g. "foo/*" allows "foo/bar", "foo/baz", etc.
+	// e.g. "foo.*" allows "foo.bar", "foo.baz", etc.
+	// +optional
+	AllowedUnsafeSysctls []string `json:"allowedUnsafeSysctls,omitempty" protobuf:"bytes,24,rep,name=allowedUnsafeSysctls"`
```
asalkeld: this is why i am updating the package.

asalkeld:
> (for all the libraries in question, best would be to use the same ones which are vendored in openshift/origin release-3.11 if this is possible)

I'll post a new PR with all those vendor changes and we can see what people want to do.

test/clients/openshift/template.go
ehashman: Any follow-up on this?

test/e2e/specs/fakerp/sync.go
```diff
@@ -18,12 +20,28 @@ var _ = Describe("sync pod tests [EveryPR]", func() {
...
-		Expect(err).ToNot(HaveOccurred())
+		for {
+			b, err := sanity.Checker.Client.Admin.CoreV1.Pods("kube-system").GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw()
+			Expect(err).ToNot(HaveOccurred())
 
-		rx := regexp.MustCompile(`(?ms)^[^\n]* msg="starting sync"[^\n]*$.*?^[^\n]* msg="sync done"[^\n]*$`)
-		runs := rx.FindAllString(string(b), -1)
-		// check for constantly updated objects
-		Expect(runs).To(ContainElement(Not(ContainSubstring("Update"))))
+			rx := regexp.MustCompile(`(?ms)^[^\n]* msg="starting sync"[^\n]*$.*?^[^\n]* msg="sync done"[^\n]*$`)
```
ehashman: Small nit: we're compiling this regex every loop... should we pull this out?

asalkeld: sure, i assumed it was e2e testing so performance was not  an issue. but i can pull it out

test/clients/openshift/template.go
asalkeld: I fixed it here: https://github.com/openshift/openshift-azure/pull/1890/files#diff-aaf8e1ec8557614866fe4f35fa93a036

PR #1891, opened 2019-08-27
fixing small syntax error in hack/delete.sh

Makdaam:
```release-note
NONE
```
just a small syntax error fix


m1kola:
Nice catch!

PR #1894, opened 2019-08-27
Update client binary download links

ehashman:
```release-note
NONE
```

These now match the image versions in the pluginconfig: https://github.com/openshift/openshift-azure/blob/b7ada76b8555adfa9562dfa35ba5413e16a8c6b0/pluginconfig/pluginconfig-311.yaml#L101-L112

I also added this as a step in the release instructions.


asalkeld:
@ehashman make generate

asalkeld:
> I wonder if this can be turned into a template? if the versions are already present in the plugin config yaml file, would it make sense to have them automatically interpolated here?

Yeah, @ehashman I hope you don't mind I'll give that a go and post a PR

asalkeld:
@ehashman dynamic way of doing it: https://github.com/openshift/openshift-azure/pull/1896

PR #1895, opened 2019-08-27
Make log paths relative to make them more readable

asalkeld:
This PR aims to make log messages more readable by converting absolute file names into relative ones. An example of the difference is:

It was like this
```
INFO[2019-08-28T06:54:03+10:00]/home/angus/code/go/src/github.com/openshift/openshift-azure/pkg/fakerp/client/config.go:48 github.com/openshift/openshift-azure/pkg/fakerp/client.NewConfig() using region eastus
```
now will be
```
INFO[2019-08-28T08:19:26+10:00]pkg/fakerp/client/config.go:48 pkg/fakerp/client.NewConfig() using region eastus 
```

```release-note
Microsoft: when using logrus.SetReportCaller() you can now use the following to convert absolute file names in logs to relative ones.
	logrus.SetFormatter(&logrus.TextFormatter{
		FullTimestamp:    true,
		CallerPrettyfier: utillog.RelativeFilePathPrettier,
	})

```


asalkeld:
> @asalkeld I'd definitely check with MS about upgrading the version, I agree that it's low-risk and potentially of a lot of benefit here reducing log noise.

Anubhuti said she is fine with it.

PR #1897, opened 2019-08-28
[WIP] PrivateLink plugin side implementation 

mjudeikis:
```release-note
Adds management subnet
Adds an internal load balancer
Extends API to configure vnet configuration to prepare for expressRoute and vnet
```

TODO: 
- [ ] Add callback function to `createOrUpdate` to provision the PLS/PE
- [ ] Change internal type/config to have placeholders for PE DNS/IP
- [ ] Make plugin kubeconfig avare of the above and use round tripper to call API via External Loadbalanancer OR PE
- [ ] Unit tests for new functions

### DIFF
--- a/pkg/api/2019-09-30-preview/default.go
+++ b/pkg/api/2019-09-30-preview/default.go
@@ -45,4 +45,18 @@ func setDefaults(oc *OpenShiftManagedCluster) {
 			},
 		}
 	}
+
+	if oc.Properties.NetworkProfile == nil {
+		oc.Properties.NetworkProfile = &NetworkProfile{}
+	}
+
+	if oc.Properties.NetworkProfile.VnetCIDR == nil {
+		oc.Properties.NetworkProfile.VnetCIDR = to.StringPtr("10.0.0.0/8")
+	}
+	if oc.Properties.NetworkProfile.DefaultCIDR == nil {
+		oc.Properties.NetworkProfile.DefaultCIDR = to.StringPtr("10.0.0.0/24")
+	}
+	if oc.Properties.NetworkProfile.ManagementCIDR == nil {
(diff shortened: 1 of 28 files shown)


pkg/api/2019-09-30-preview/default.go
```diff
@@ -45,4 +45,18 @@ func setDefaults(oc *OpenShiftManagedCluster) {
 			},
 		}
 	}
+
+	if oc.Properties.NetworkProfile == nil {
+		oc.Properties.NetworkProfile = &NetworkProfile{}
+	}
+
```
asalkeld: any reason the other networkProfile members are not already copied here?

pkg/api/admin/default.go
```diff
@@ -0,0 +1,26 @@
+package admin
+
+import (
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+func setDefaults(oc *OpenShiftManagedCluster) {
```
asalkeld: you could be even more explicit here setDefaultsForV7()

mjudeikis: It should not matter. Moved it to work with the internal model as I think doing this for external might cause issues. I think we dont care if it is done on v6 to v8+ or v7 to V9.

pkg/util/cmp/cmp.go
```diff
@@ -24,3 +25,15 @@ func x509CertComparer(x, y *x509.Certificate) bool {
 
 	return x.Equal(y)
 }
+
+func rsaPrivateKeyComparer(x, y *rsa.PrivateKey) bool {
```
mjudeikis: Anybody knows a better way to do this?

PR #1901, opened 2019-08-29
vendor updates to make us more consistent with origin

asalkeld:
```release-note
NONE
```
this is a folllow up to #1890 . It updates kubernetes and all it's deps to be consistent with origin 3.11


asalkeld:
/hold
maybe merge in next cycle?

glide.yaml
```diff
@@ -1,19 +1,26 @@
 package: github.com/openshift/openshift-azure
 import:
 - package: github.com/openshift/api
-  version: release-3.10
```
mjudeikis: This was an issue with MSFT side of vendor. Problems with ServiceCatalog dep clash. Need to confirm with  MSFT before merging. Or maybe move to separate PR and send it to MSFT to check.

m1kola:
@jhendrixMSFT I think, the summary of the issue is: there is currently no *official* way to build an [ARM template](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates) as code using Azure SDK and a workaround gets broken after upgrading SDK version.

The workaround is to use types from the SDK that were designed to work with REST API, but not with ARM. As a result there are few issues:

* Types get changed based on swagger definitions without taking ARM into account. The `Type` field is one of the examples: the older versions of SDK were marshaling it into JSON, but newer version do not. Not marshaling this field makes sense for REST API, because type can be received from the endpoint URL, but not for ARM where type needs to be explicitly specified.
* Some fields are not present in REST types: for example, ARM expects to receive `apiVersion` field in JSON, but in REST API it's [a URL parameter](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/virtualnetworks/createorupdate#uri-parameters), so it's not present in the type definition generated from the swagger spec. Another example is `dependsOn`: it's not a thing in the REST API, but it's an important field in the ARM template.

>  Is it just the `Type` field that's causing problems? The fact it's `READ-ONLY`? Something else?

It's `Name` and `Type` on most of the types that represent resources in Azure SDK: these fields are present in the structs and [they get unmarshaled from JSON](https://github.com/Azure/azure-sdk-for-go/blob/v33.2.0/services/network/mgmt/2018-01-01/network/models.go#L14308-L14325), but becasue they are read only [they don't get marshaled into JSON](https://github.com/Azure/azure-sdk-for-go/blob/v33.2.0/services/network/mgmt/2018-01-01/network/models.go#L14252-L14270). In some other types apart from `Name` and `Type` similar issue can be found with other fields: for example, [storage `Account` doesn't marshal `Sku`, `Kind` into JSON](https://github.com/Azure/azure-sdk-for-go/blob/v33.2.0/services/storage/mgmt/2019-04-01/storage/models.go#L581-L597), but ARM expects these fields.

Also [ARM API supports `apiVersion`, `dependsOn`, `condition`, `resources` and some other fields](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates#resources) in the resource section. Existing REST types simply do not have these fields.

> Generally speaking, any codegen enhancements that introduce breaking changes are pretty much out of the question so please keep that in mind when coming up with ideas.

It's hard to suggest something specific, because I didn't manage to find any code or API specs for the ARM API publicly available. Maybe having separate types for the ARM API and for the REST API would help us to achieve what we need without breaking backward compatibility? OR wrap REST types in a structure with extra fields and custom marshaling? ARM API types to include `apiVersion`, `dependsOn`, `resources` (to accept child resources), `condition`, etc in the JSON.

mjudeikis:
```
pkg/sync/v71/sync.go:497:13: undefined: dynamic.ClientPool
pkg/sync/v71/sync.go:498:16: undefined: discovery.APIGroupResources
```

Needs rebase. I would like this to merge after we do v8/v9 release. As I see this potentially blocking vendoring of new plugin code on MSFT side, hence creating delays in the metrics/logging release. 
/cc @ehashman @jim-minter

asalkeld:
just rebased to include the new v10 directories

ehashman:
flake #1410 

/retest

mjudeikis:
/hold cancel
Im ok to merge it now

jim-minter:
@asalkeld I'm good with this but I'd like you to have agreed with MSFT that they will be doing the same on their side before this merges.  Have you done that?

ehashman:
/hold

I'm good to merge this but we need confirmation from MS first.

ehashman:
Closing as stale.

PR #1902, opened 2019-08-30
Adds CI notifications into vmimage-validate

m1kola:
So we can run it as a periodic job and get feedback from CIin form of a GitHub issue.

```release-note
NONE
```

/test vmimage

### DIFF
--- a/hack/tests/ci-prepare.sh
+++ b/hack/tests/ci-prepare.sh
@@ -62,6 +62,17 @@ delete() {
   make delete
 }
 
+ci_notify() {
+  # We create notifications in CI and only for periodic jobs. This will open an issue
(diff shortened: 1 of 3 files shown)


m1kola:
@kwoodson could you please take a look?

/cc @kwoodson

Also see the related PR in the release repo that adds the periodic job to validate the latest marketplace image: https://github.com/openshift/release/pull/4901

hack/tests/ci-prepare.sh
```diff
@@ -62,6 +62,17 @@ delete() {
   make delete
 }
 
+ci_notify() {
```
kwoodson: I like this change.

PR #1904, opened 2019-08-30
Avoid VM image build timeout

ehashman:
```release-note
NONE
```

I got a successful build out this, but it failed the VM validation due to a token expiry.

/test vmimage


ehashman:
This succeeded locally on Friday.

PR #1905, opened 2019-09-03
Attempt to convert vmimage build to ssh instead of cse.

kwoodson:
An attempt to update the vmimage build to move away from CSE and towards ssh.

This is a naive approach.

```release-note:
None
```


pkg/vmimage/vmimage.go
```diff
@@ -252,6 +253,13 @@ func (builder *Builder) Run(ctx context.Context) error {
 		return err
 	}
 
+	go builder.sshInstall()
+	// wait for install
+	err = builder.waitForInstall(ctx)
```
thekad: code suggests this is to wait until the install _begins_, is that correct? I am thinking you should need a /var/log/install.done or similar to check when the install is finished, touch it at the end of the install script ?

kwoodson: When the `install.sh` finishes it writes here: https://github.com/openshift/openshift-azure/pull/1905/files#diff-037e2465ca7d49e6da64f82bd06ca8e1R349

This was the intention.  There might be a better way

kwoodson:
@ehashman @jim-minter @mjudeikis PTAL

jim-minter:
I feel like this approach adds technical debt.  How much do we need to use ssh instead of cse at this point?

kwoodson:
@jim-minter 
> I feel like this approach adds technical debt. How much do we need to use ssh instead of cse at this point?

As per our discussion in slack the recommendation was to remove the dependency on the CSE timeout.  This moves the timeout to ssh vs CSE.  Now that we have a new 7.7 image (#1925) and builds are completing again in a reasonable time(<90 minutes) then I think we can probably close this.

PR #1909, opened 2019-09-05
add ACR images

mjudeikis:
```release-note
NONE
```

Update release v8.0 images to point to ACR


mjudeikis:
/hold
until https://github.com/openshift/openshift-azure/pull/1911 merges
+ we need to add pull secret for ACR. So suggestion is to use quay.io for v8 for now and make v9 to have new secret

PR #1910, opened 2019-09-05
[master] fix groups sync rbac

mjudeikis:
```release-note
bugfix: Regression fix in Group sync for guest users
```

Group sync fails for guest users with RBAC error:
```
time="2019-09-05T10:13:34Z" level=debug msg="AAD Group Reconciler (poll).."
time="2019-09-05T10:13:34Z" level=warning msg="failed to list users users.user.openshift.io is forbidden: User \"system:serviceaccount:openshift-infra:customer-admin-controller\" cannot list users.user.openshift.io at the cluster scope: no RBAC policy matched"
```


Makdaam:
/lgtm
The controller should be able to manipulate users in the cluster.

PR #1911, opened 2019-09-05
[v8] fix group sync

mjudeikis:
```release-note
Fix group sync rbac
```


Makdaam:
/lgtm
The controller should be able to manipulate users in the cluster.
Same as #1910

PR #1915, opened 2019-09-06
Update container versions for v8 release

ehashman:
```release-note
NONE
```

OpenShift container versions need to match the VM image. We determined we don't need to do a VM image update this sprint, and since the newest one doesn't match the container images we tested against, reverting back to 129 for now.


ehashman:
@mjudeikis I'm going to update tag v8.0 to point to c967905ba689107b300ec1112b6198760c3975a3, please ack.

mjudeikis:
@ehashman did you pushed new tag already?

ehashman:
Not yet, pushing now.

PR #1916, opened 2019-09-06
Update image versions in plugin config for v8/v9

ehashman:
```release-note
NONE
```

This is a cherry-pick of cbfb63eb524adf70c4f7cb57dfa8ed85432595b1 plus an update to the v9 VM image version.

### DIFF
--- a/pluginconfig/pluginconfig-311.yaml
+++ b/pluginconfig/pluginconfig-311.yaml
@@ -62,28 +62,28 @@ versions:
     imageSku: osa_311
     imageVersion: 311.129.20190810
     images:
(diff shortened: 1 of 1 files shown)


pluginconfig/pluginconfig-311.yaml
```diff
@@ -101,7 +101,7 @@ versions:
     imageOffer: osa
     imagePublisher: redhat
     imageSku: osa_311
-    imageVersion: 311.129.20190810
+    imageVersion: 311.141.20190904
```
mjudeikis: We bumping VM to 141? And images to 129?

mjudeikis: Ignore

PR #1917, opened 2019-09-06
[master] logging fix

mjudeikis:
```release-note
- Fix logging
```

Fixes missing logs issue


pkg/startup/v7/startup.go
```diff
@@ -119,7 +119,7 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
 			filepath = strings.TrimPrefix(filepath, "common")
 
 			filesToWrite[filepath] = string(b)
-
+			fileKeys = append(fileKeys, filepath)
```
asalkeld: are you sure we should be changing this? we never create old versions (except in CI), they are there for hash comparisons really so you are now changing the old hashes..

pkg/cluster/hash_test.go
```diff
@@ -36,34 +36,34 @@ func TestHashScaleSetStability(t *testing.T) {
 			{
 				role: api.AgentPoolProfileRoleMaster,
 				// this value should not change
```
asalkeld: this ^
I don't think we should "fix" the old versions

mjudeikis: I think this will be aligned with other PR's, no?

asalkeld: i am ok with it..

mjudeikis:
manually tagging

PR #1918, opened 2019-09-06
[v8] Logging fix

mjudeikis:
```release-note
- Fix logging
- Fix group sync rbac
```


pkg/arm/v7/data/node-startup.sh
```diff
@@ -107,9 +107,9 @@ systemctl disable rsyslog.service
 systemctl stop rsyslog.service
 
 # setting up geneva logging stack
-systemctl unmask mdsd.service azsecd.service azsecmond.service
-systemctl enable mdsd.service azsecd.service azsecmond.service
-systemctl start mdsd.service azsecd.service azsecmond.service
+systemctl unmask mdsd.service azsecd.service azsecmond.service fluentd.service
+systemctl enable mdsd.service azsecd.service azsecmond.service fluentd.service
+systemctl start mdsd.service azsecd.service azsecmond.service fluentd.service
```
ehashman: Since this is for v8 release, I don't think we need to touch any of the v7 files? Can we avoid these changes?

mjudeikis: I dont know. This is following the practice to keep all in sync. Like v7 will have a different version of v7 files than v8 branch v7 files :man_shrugging:

ehashman: I think that's okay, that's the point of the hashes. We don't ever deploy v7 off the v8 release branch, right?

mjudeikis: yes, we just upgrade v7 with v8 branch. Hashes so we would not produce "VM rollout, where we dont need it"

mjudeikis:
/hold
need https://github.com/openshift/openshift-azure/pull/1923 to merge and rebase

PR #1919, opened 2019-09-06
[v7]Logging and rbac fixes

mjudeikis:
```release-note
- Fix logging
- Fix group sync rbac
```


ehashman:
```
fatal: unable to access 'https://github.com/openshift/openshift-azure.git/': Could not resolve host: github.com 
```

flake...

ehashman:
@mjudeikis I'm concerned this is a real bug---we had the same failure on upgrade and scaleupdown on both this PR and #1918 at the same test point, and this failure references a data race. Can you PTAL?

/cc @jim-minter

pkg/cluster/kubeclient/backup.go
```diff
@@ -29,10 +30,18 @@ func (u *Kubeclientset) BackupCluster(ctx context.Context, backupName string) er
...
 	job.Spec.BackoffLimit = to.Int32Ptr(0)
-	job.Spec.Template.Spec.Containers[0].Args = []string{
-		"etcdbackup",
-		fmt.Sprintf("--blobName=%s", backupName),
-		"--action=save",
+
+	// Special code to make etcd backup work for v4+ branches
+	// if image is etcdbackup - old cluster
+	// else - new
+	if strings.Contains(cronjob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Image, "etcdbackup") {
```
jim-minter: sorry, what does this change do and why do we need it in 7.1?

pkg/plugin/plugin.go
```diff
@@ -55,7 +55,7 @@ func (p *plugin) Validate(ctx context.Context, new, old *api.OpenShiftManagedClu
 
 	// if this is an update and not an upgrade, check if we can service it, and
 	// if not, fail early
-	if old != nil && new.Config.PluginVersion != "latest" {
+	if old != nil && new.Properties.ClusterVersion != "latest" {
```
jim-minter: I think this makes sense but we need to check it through with MSFT

pkg/util/cmp/cmp.go
```diff
@@ -1,20 +1,20 @@
...
-		// FIXME: Remove x509CertComparer after upgrading to a Go version that includes https://github.com/golang/go/issues/28743
-		gocmp.Comparer(x509CertComparer),
-	)
+	// FIXME: Remove x509CertComparer after upgrading to a Go version that includes https://github.com/golang/go/issues/28743
+	opts = append(opts, gocmp.Comparer(x509CertComparer))
 
-	return gocmp.Diff(x, y, newOpts...)
+	opts = append(opts, gocmp.Comparer(rsaPrivateKeyComparer))
+
+	return gocmp.Diff(x, y, opts...)
```
jim-minter: why is this file being changed in 7.1?

pkg/plugin/plugin.go
jim-minter: probably needs an extra line of comment here...

pluginconfig/pluginconfig-311.yaml
```diff
@@ -177,6 +177,46 @@ versions:
...
+      serviceCatalog: registry.access.redhat.com/openshift3/ose-service-catalog:v3.11.129
+      templateServiceBroker: registry.access.redhat.com/openshift3/ose-template-service-broker:v3.11.129
+      webConsole: registry.access.redhat.com/openshift3/ose-web-console:v3.11.129
+      format: registry.access.redhat.com/openshift3/ose-${component}:v3.11.129
+      httpd: registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-86
+      masterEtcd: registry.access.redhat.com/rhel7/etcd:3.2.22-28
+      genevaLogging: osarpint.azurecr.io/acs/mdsd:master.20190228.1
+      genevaStatsd: osarpint.azurecr.io/acs/mdm:git-a909a2e76
+      genevaTDAgent: osarpint.azurecr.io/acs/td-agent:master.20190228.1
+      azureControllers: quay.io/openshift-on-azure/azure:v7.1
```
ehashman: We need to make sure this image is tagged and exists.

ehashman:
/retest

failing due to quota in australiasoutheast

pkg/plugin/plugin.go
mjudeikis: +1

pkg/util/cmp/cmp.go
mjudeikis: gocmp panics when you run it on full structure because of ssh keys. It was an issue when I moved logic around. I think it is no-op now. Its a bug in current our Diff logic in master too (see my PLS PR).

Will remove.

mjudeikis:
we should dump Australia quota. i would it to become a proper 3rd region. Make sense for future VPN to be geo-distributed. `s/northeurope/southcentralaustralia/` in our ci

pkg/cluster/kubeclient/backup.go
mjudeikis: v7 was never designed to run updated on v4. with image consolidation v7 branch way of initializing etcbackup does not work on v4 image.

pkg/plugin/plugin.go
asalkeld: so we have this code in fakerp:

```golang
	switch {
	case cs.Properties.ClusterVersion == "":
	case isAdmin && cs.Properties.ClusterVersion == "latest":
		oldMajor, _, err := parsePluginVersion(cs.Config.PluginVersion)
		if err != nil {
			return nil, err
		}
		if oldMajor < 3 {
			return nil, fmt.Errorf("tried to upgrade a cluster that is too old")
		}
		cs.Properties.ClusterVersion = ""
		cs.Config.PluginVersion = "latest"
	}
```
So we are assuming that MSFT does it in exactly the same order as us
1. Validate()
2. cs.Properties.ClusterVersion = ""
3. cs.Config.PluginVersion = "latest"
4. CreateOrUpdate()

pkg/cluster/kubeclient/backup.go
jim-minter: ah, I see - and because we run a backup on the cluster before we upgrade it, we need this to work. sgtm.

jim-minter:
This lgtm.  Not sure whether we want to discuss with Anubhuti before or after committing it.

jim-minter:
(for anyone else who is reviewing: `for i in pkg/*/v71; do diff -uNr `sed -e 's/1$//' <<<$i` $i; done |less` ftw)

mjudeikis:
Tested:
* v4.3 Upgrade to v7.1
*v7.0 Upgrade to v7.1
*v7.0 scaleupdown with v7.1

Master needs to be synced with these changes too. I would do this as a follow-up.

jim-minter:
@mjudeikis @ehashman I have added two commits backporting @Makdaam 's work at #1926 and #1930

ehashman:
#1932 on upgrade-7.0 and scaleupdown-7.0

ehashman:
e2e-create looks like a real failure:

```
panic: template: /etc/mdsd.d/mdsd.xml:10:67: executing \\\\\\\"/etc/mdsd.d/mdsd.xml\\\\\\\" at \\u003c.Derived.CustomerRes...\\u003e: error calling CustomerResourceGroup: parsing failed for . Invalid resource Id format
```

jim-minter:
@mjudeikis please check the last commit - based on a call with MSFT.

mjudeikis:
@jim-minter i think its fine. WIll run on more test and if fine, move tags and LGTM, rebuild images

ehashman:
Completed testing on `ehashman-test` cluster:

- Upgraded a 4.3 cluster to 110e7bd805369f76329e94cd5a78a936d8ea8361
- Confirmed logs are available for upgraded `ehashman-test` cluster in Azure

PR #1924, opened 2019-09-09
Allow upgrades from 4.4 and 4.3

asalkeld:
```release-note
Allow upgrades from v4
```
/cc @mjudeikis 
/cc @ehashman


pkg/plugin/plugin.go
```diff
@@ -212,6 +212,11 @@ func (p *plugin) createOrUpdateExt(ctx context.Context, cs *api.OpenShiftManaged
 		isUpdate = false
 	}
 
+	switch cs.Config.PluginVersion {
+	case "v4.4", "v4.3":
+		backupEtcd = false
```
asalkeld: the backup didn't seem to work, so disabling..

mjudeikis:
This is not enough. Something is off here as after upgrade it still runs with v4.4 images

jim-minter:
I am pretty convinced we must not do this.  Closing.  If I'm absolutely wrong it can be re-opened.

PR #1927, opened 2019-09-09
fix fakerp logic error

mjudeikis:
```release-note
NONE
```

Fix fakerp logic bug. We are doing validation too early in the process


mjudeikis:
/hold
still wrong

jim-minter:
@mjudeikis is this PR dead?

mjudeikis:
/close
I think it will be more master alignment post v7.1 merges/lgtm'ed

PR #1928, opened 2019-09-09
remove v8 release, it was not to be

jim-minter:
```release-note
NONE
```


jim-minter:
cf https://github.com/openshift/release/pull/4969

PR #1935, opened 2019-09-11
align master to release-v7

mjudeikis:
```release-note
NONE
```

Align master  with v7 branch


mjudeikis:
let's not do this again!

pkg/plugin/plugin.go
```diff
@@ -215,6 +224,12 @@ func (p *plugin) createOrUpdateExt(ctx context.Context, cs *api.OpenShiftManaged
 		isUpdate = false
 	}
 
+	switch cs.Config.PluginVersion {
+	case "v4.4", "v4.3":
+		backupEtcd = false
```
asalkeld: I don't think this hack needs to come back to master

PR #1942, opened 2019-09-13
WIP: azure-sdk-for-go upgrade assessment

m1kola:
This is not for code review, at the moment. Just for me to do some testing.

```release-note
NONE
```


pkg/arm/v9/resources.go
```diff
@@ -51,8 +52,8 @@ const (
 	vmssAdminUsername                             = "cloud-user"
 )
 
-func (g *simpleGenerator) vnet() *network.VirtualNetwork {
```
mjudeikis: Im not sure I am big fan of putting all objects under one package.
I think these could go under `https://github.com/openshift/openshift-azure/tree/master/pkg/util/azureclient` individual subpackages, replicating SDK structure.

m1kola: @mjudeikis this is not ready for review: it's a very dirty PR :) I'm looking into what we need to change in our code base to be able to upgrade to the latest version of SDK and this PR is more like notes for me.

m1kola: I mean, I'll be happy to rearrange it later, if we decide to go this way at the end.

ehashman:
Closing as stale.

PR #1931, opened 2019-09-10
Use Iptables to restrict access to Worker Kubelet read-only port

charlesakalugwu:
```release-note
Use iptables rules to restrict access to the Kubelet's unauthenticated read-only port
```

/cc @ehashman @jim-minter @asalkeld

PR #1943, opened 2019-09-15
Log Analytics workspace key should be retrieved and set by fake RP

charlesakalugwu:
Currently, the ARO plugin is fetching the workspace key. This will not work in the RP environment because the plugin has a service principal with access to the publisher tenant, but the Log analytics workspace is not in that tenant. As a result, the plugin will get a 404 or auth error.  

The workspace key retrieval logic should not have gone into the plugin as per the design doc for this feature. Workspace key retrieval should be handled by the RP (fake RP in our case).

```release-note
Log analytics workspace key should not be retrieved via plugin
```

depends on #1931 

/cc @asalkeld @mjudeikis


charlesakalugwu:
e2e-create flake #1778 
artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1943/pull-ci-azure-master-e2e-create/1083/build-log.txt

Noticed that the sync pod constantly updates the log analytics agent deployment and daemonset. 
*log-analytics-node-agent (daemonset)*
- sync wants to set spec.containers[0].securityContext.runAsGroup = 0
- sync wants to apiVersion = v1 on the `valueFrom.fieldRef` in the env variables section of the yaml

*log-analytics-cluster-agent (deployment)*
- sync wants to set spec.replicas = 1
- sync wants to set spec.containers[0].securityContext.runAsGroup = 0
- sync wants to apiVersion = v1 on the `valueFrom.fieldRef` in the env variables section of the yaml

These issues are fixed by updating those resource yamls to set those values to make them invariants which prevents constant reconciliation of those resources by the sync pod

pkg/plugin/plugin.go
```diff
@@ -238,21 +235,6 @@ func (p *plugin) createOrUpdateExt(ctx context.Context, cs *api.OpenShiftManaged
...
-		if err != nil {
-			return &api.PluginError{Err: err, Step: api.PluginStepClientCreation}
-		}
-		wc := operationalinsights.NewWorkspacesClient(ctx, p.log, cs.Properties.AzProfile.SubscriptionID, authorizer)
-		if err != nil {
-			return &api.PluginError{Err: err, Step: api.PluginStepClientCreation}
-		}
-		err = enrich.MonitorIDAndKey(ctx, wc, cs)
-		if err != nil {
-			return &api.PluginError{Err: err, Step: api.PluginStepEnrichMonitorProfile}
```
asalkeld: This "api.PluginStepEnrichMonitorProfile" is not used anywhere else, we should remove it.

charlesakalugwu: Yes. However that doesn't block this PR.

mjudeikis: I think they should be removed here too as steps removal is happening here too

JackQuincy:
LGTM

PR #1944, opened 2019-09-15
Add unit test file for passwd in envars.

y-cote:
1- The test detects environment variable populated with specific values, but will only log any hits and NOT make the test fail at the moment. This could be changed in the future.
2- The white list is used to deal with false positives OR yaml configs we prefer not to change and continue using environment variables.
3- The test is extensible by adding new routines starting with "checkX()".
 
```release-note
NONE
```


mjudeikis:
rebase is needed.
What is the background for this PR?

thekad:
I am guessing it is related to AZURE-324

pkg/sync/v9/envpass_test.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+	return false
+}
+
+// We are looking for a structure like this:
+// env:
+// - name: SomeVarName
+//   value: '*** GENERATED ***'
+func checkValueGenerated(path string, o unstructured.Unstructured, t *testing.T) bool {
```
asalkeld: t is not used (maybe some debugging?)

pkg/sync/v9/envpass_test.go
```diff
@@ -0,0 +1,125 @@
...
+		if r := checkValueGenerated(path, u, t); r == true {
+			t.Logf("%s: contains '*** GENERATED ***' in env[*].value which probably means that secrets are passed in environment variables. This is not allowed for security reasons.", path)
+		}
+
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(paths) > 0 {
```
asalkeld: paths is never set

pkg/sync/v9/envpass_test.go
```diff
@@ -0,0 +1,125 @@
...
+		u, err := unmarshal(b1)
+		if err != nil {
+			return err
+		}
+
+		if r := checkSecretKeyRef(path, u); r == true {
+			t.Logf("%s: contains 'password' in env[*].valueFrom.secretKeyRef.key which probably means that secrets are passed in environment variables. This is not allowed for security reasons.", path)
+		}
+		if r := checkValueGenerated(path, u, t); r == true {
+			t.Logf("%s: contains '*** GENERATED ***' in env[*].value which probably means that secrets are passed in environment variables. This is not allowed for security reasons.", path)
```
asalkeld: these never result in the test failing
were you going to do something like this below?
```
paths = append(paths, path)
```

pkg/sync/v9/envpass_test.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(paths) > 0 {
+		t.Errorf("invalid files (rerun with REGENERATE environment variable set):\n  %s", strings.Join(paths, "\n  "))
```
asalkeld: this error message is misleading. i don't think that running REGENERATE will fix the password being passed in.

asalkeld:
I have rebased and fixed (kept @y-cote 's commit). Moved to https://github.com/openshift/openshift-azure/pull/1954

PR #1946, opened 2019-09-16
Not allowing 2 infra nodes anymore. Only accepting 3.

Makdaam:
```release-note
MSFT: don't allow cluster creation with 2 infra nodes (azure-cli<=2.0.63 will stop working with openshift)
```
Verified that the azure-cli version recommended in documentation (v2.0.65)
https://docs.microsoft.com/en-us/azure/openshift/howto-setup-environment

already tries to create an openshift cluster with 3 infra nodes.
https://github.com/Azure/azure-cli/blob/azure-cli-2.0.64/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/custom.py#L2523

### DIFF
--- a/pkg/api/validate/validators.go
+++ b/pkg/api/validate/validators.go
@@ -247,8 +247,7 @@ func validateAgentPoolProfile(app *api.AgentPoolProfile, vnet *net.IPNet) (errs 
 		if app.Name != string(app.Role) {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q]", app.Name))
 		}
-		// TODO: remove count "2" check when CLI is adjusted
-		if app.Count < 2 || app.Count > 3 {
+		if app.Count != 3 {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].count %d", app.Name, app.Count))
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/api/validate/validators.go
```diff
@@ -247,8 +247,7 @@ func validateAgentPoolProfile(app *api.AgentPoolProfile, vnet *net.IPNet) (errs
 		if app.Name != string(app.Role) {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q]", app.Name))
 		}
-		// TODO: remove count "2" check when CLI is adjusted
-		if app.Count < 2 || app.Count > 3 {
+		if app.Count != 3 {
```
kwoodson: Is there a chance large clusters would want more?

Makdaam: I guess there is a breaking point for the routers or Azure VMs somewhere. I think that's pretty far away for now, but worth considering as a feature. The size of the infra scale set would have to be customer controlled for that to work at scale, so it's not a small feature.

ehashman: Given that our max supported cluster size is 20 compute nodes right now, I think we don't need to worry about this yet.

PR #1947, opened 2019-09-16
Fix multiple aad urls single-cluster case

charlesakalugwu:
This is a follow-up to #1937. The patch api doesn't accept an empty slice for `identifiersUri` and therefore blows up in the case that a single cluster is linked to an AAD app, the cluster is deleted and the cluster is attempted to be unlinked from the AAD app.  This PR fixes that problem.

```release-note
NONE
```

/cc @mjudeikis @kwoodson

### DIFF
--- a/cmd/createorupdate/createorupdate.go
+++ b/cmd/createorupdate/createorupdate.go
@@ -38,6 +38,10 @@ var (
 	restoreFromBlob = flag.String("restore-from-blob", "", "If set, request a restore of the cluster from the provided blob name.")
 )
 
+const (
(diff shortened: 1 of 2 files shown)


mjudeikis:
would be good to have unti tested this, but lets merge as it is causing errors now. 
Can we have the issue raised with "good first issue" label?

/lgmt

mjudeikis:
/lgtm 
test in separate PR please

charlesakalugwu:
upgrade-v7.1 flake #1628 
/retest

PR #1948, opened 2019-09-16
VMImage improvements

thekad:
Updated scap rules for vmimage, it complained about ssmtp, possibly pulled as a dependency on recently-added clamav
Also added small tweak to `hack/vmimage.sh` so we can easily add more tags if needed. Tags will be added at the same time it is labelled "valid" 

```release-note
NONE
```


hack/vmimage.sh
```diff
@@ -44,7 +44,7 @@ make vmimage-validate
 
 PHASE=image_tag
 TAGS=$(az image show -g "$IMAGE_RESOURCEGROUP" -n "$IMAGE_RESOURCENAME" --query tags | python -c 'import sys; import json; tags = json.load(sys.stdin); print " ".join(["%s=%s" % (k, v) for (k, v) in tags.items()])')
-az resource tag -g "$IMAGE_RESOURCEGROUP" -n $IMAGE_RESOURCENAME --resource-type Microsoft.Compute/images --tags $TAGS valid=true
+az resource tag -g "$IMAGE_RESOURCEGROUP" -n $IMAGE_RESOURCENAME --resource-type Microsoft.Compute/images --tags $TAGS valid=true $EXTRA_TAGS
```
kwoodson: What is the format of `$EXTRA_TAGS`?

thekad: `key=value` pairs, I am thinking of a use case where you tag an image as rc=true and the azure-purge bin doesn't delete it

kwoodson: Thanks, :+1:

kwoodson:
This looks good to me, please move to ready.

hack/vmimage.sh
mjudeikis: why we adding this now if we dont use it?

thekad: yeah you're right, reverting this part.

Issue #1839, opened 2019-08-09
sync pod test "should not continuously update objects" fails


thekad:
Hadn't seen this error before, not sure if transient in #1838 

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1838/pull-ci-azure-master-e2e-create/767

m1kola:
Seen it in #1836 

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1836/pull-ci-azure-master-e2e-create/772/build-log.txt

charlesakalugwu:
this is a duplicate of #1778 
/close

PR #1950, opened 2019-09-18
Azure Monitor end-to-end tests

charlesakalugwu:
In order to enable/disable the Azure Monitor agent, the customer would currently need to either entirely remove the `MonitorProfile` in the config or set `MonitorProfile.WorkspaceResourceId=""`

However during the design phase we discussed and agreed to use a boolean field `MonitorProfile.Enabled` for this instead. This would allow a customer to temporarily disable forwarding of telemetry data into the workspace without having to entirely delete the information about the workspace.

Furthermore, ARO SREs currently have no way to determine whether a given cluster has Azure Monitor integration enabled or not since we neither add the workspace key (admin api does not contain secrets) nor the workspace id (the workspace id does not add more information than the already existing workspace resource id). An enabled flag would solve this problem for SREs. 

```release-note
Enable/disable Azure Monitor integration with Enabled flag in MonitorProfile
```

/cc @amanohar @asalkeld @JackQuincy @ehashman


asalkeld:
@charlesakalugwu it was my understanding that during the design phase the MonitorProfile.Enabled was discussed and throw out. It really is redundant.

ehashman:
Agreed that this is redundant, I thought we said we were not including this? See unresolved comment on https://docs.google.com/document/d/1KCKI4_rjP5c7_qP0jYZMQhJrpdg1j3ZSy-fTqzhhriU/edit?ts=5d82d4c5 /cc @jim-minter

JackQuincy:
Every other profile/field leaving it off the request means no change. For this if it means remove it is an inconsistency in our api. If we want consistency with the rest of our api we need the enabled field to allow the customer to turn off the logging.

amanohar:
Its not redundant. It follows the API model pattern of today which is if a customer excludes a field/property, service doesn't delete it. E.g. if customer omits `masterProfile` masters are not deleted. Mixing design patterns can be confusing to the customer and lead to bugs.

Also, regarding:
```
it was my understanding that during the design phase the MonitorProfile.Enabled was discussed and throw out.
```
In the joint design discussions this was the conclusion was `Enabled` field would be used to indicate intent.

Also, we shared all questions/issues in the email thread.

asalkeld:
OK, for clarity sake I am fine with this.

jim-minter:
@ehashman @charlesakalugwu will open a JIRA to cover automated testing for this feature.
@mgahagan73 and @charlesakalugwu will manually test, and if good this can merge.

ehashman:
Okay, waiting on test results before we LGTM it.

/approve

ehashman:
`verify` is suggesting you need to run `goimports -e -w -local=github.com/openshift/openshift-azure pkg cmd test`

Once that's fixed, waiting on @mgahagan73's signoff.

charlesakalugwu:
@jim-minter @ehashman @mgahagan73 end-to-end tests have been added for this feature.

charlesakalugwu:
upgrade-v7.0 flake #1628 
e2e-create flake #1839

pkg/fakerp/fakerp.go
```diff
@@ -200,7 +200,12 @@ func createOrUpdateWrapper(ctx context.Context, p api.Plugin, log *logrus.Entry,
 		return nil, err
 	}
 
+	if !isUpdate {
+		cs.Properties.MonitorProfile.Enabled = true
```
asalkeld: I thought we were not going to turn this on by default, isn't quite expensive?

pkg/fakerp/client/config.go
```diff
@@ -46,5 +46,11 @@ func NewConfig(log *logrus.Entry) (*Config, error) {
 		return nil, fmt.Errorf("must set AZURE_REGIONS to a comma separated list")
 	}
 	log.Infof("using region %s", c.Region)
+	if c.WorkspaceResourceID == "" {
```
asalkeld: IMHO this should be turned on in the environment so not all dev clusters have it on.

pkg/sync/v9/translate.go
```diff
@@ -326,7 +326,7 @@ var translations = map[string][]struct {
 		{
 			Path: jsonpath.MustCompile("$.metadata.labels['azure.openshift.io/sync-pod-optionally-apply']"),
 			F: func(cs *api.OpenShiftManagedCluster, o interface{}) (interface{}, error) {
-				if cs.Properties.MonitorProfile.WorkspaceID != "" && cs.Properties.MonitorProfile.WorkspaceKey != "" {
```
asalkeld: do we have some validation that says "if enabled and other workspace fields are not set, error"?

test/e2e/specs/azure_monitor.go
```diff
@@ -0,0 +1,142 @@
...
+			Tags: map[string]*string{
+				"now": to.StringPtr(fmt.Sprintf("%d", time.Now().Unix())),
+				"ttl": to.StringPtr("3h"),
+			},
+		})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(workspaceGroup).NotTo(BeNil())
+		defer azure.RPClient.Groups.Delete(ctx, rgName)
+
+		By(fmt.Sprintf("Creating workspace %s in resource group %s", rgName, rgName))
```
asalkeld: I think we should not be creating workspaces on each e2e run, have pre-made workspaces and just use them..
Anubhuti said these were quite heavy weight resources and we should share them across clusters.

pkg/fakerp/client/config.go
charlesakalugwu: I think all dev clusters should have monitoring enabled by default because:
- It gives the SREs confidence that the monitoring integration is working similarly for the customers
- Clusters created with a service such as ours almost always have native monitoring enabled out of the box (AKS, EKS, GKE etc)

pkg/fakerp/fakerp.go
charlesakalugwu: I think its a good idea to have this always on in development so we are confident that the monitoring integration works (eating our own dog food). There are only a few dev clusters being created each of which lasts a maximum of 3 days. Our workspaces currently have a retention period of 31 days which is included in our pricing plan.

test/e2e/specs/azure_monitor.go
charlesakalugwu: I am not sure that using pre-made workspaces would be cheaper. These workspaces created in the e2e live for less than 5 minutes and are deleted at the end. I think it is important to create these workspaces here so we can test the workspace update functionality as well as the container insights solution deployment process for workspaces

pkg/sync/v9/translate.go
charlesakalugwu: We don't and it sounds like a good idea. I'll add it. :+1:

charlesakalugwu:
e2e-create flake #1778 
upgrade-v7.1 flake #1481

pkg/fakerp/fakerp.go
jim-minter: this should not be here - update the create.yaml manifest.

pkg/fakerp/fakerp.go
```diff
@@ -311,5 +316,8 @@ func enrichCs(cs *api.OpenShiftManagedCluster, conf *client.Config) error {
 		}
 	}
 
+	if cs.Properties.MonitorProfile.WorkspaceResourceID == "" {
```
jim-minter: this should not be here either

test/manifests/fakerp/admin-update.yaml
```diff
@@ -1,3 +1,6 @@
 # ADMIN_MANIFEST=test/manifests/fakerp/admin-update.yaml hack/upgrade.sh
 properties:
   clusterVersion: latest
+  monitorProfile:
```
jim-minter: I do not believe this file should be being changed.  What is the rationale for changing it?

test/manifests/fakerp/create.yaml
```diff
@@ -29,4 +29,5 @@ properties:
     vmSize: {{ if eq .RunningUnderTest "true" }}Standard_D2s_v3{{ else }}Standard_D4s_v3{{ end }}
     subnetCidr: 10.0.0.0/24
   monitorProfile:
+    enabled: true
```
jim-minter: good.  I think the line below should be something like:
`    workspaceResourceId: {{ print "/subscriptions/" .SubscriptionID "/resourcegroups/defaultresourcegroup-" .Region "/providers/microsoft.operationalinsights/workspaces/DefaultWorkspace-" .Region | quote }}`

hack/tests/e2e-upgrade.sh
```diff
@@ -35,6 +35,7 @@ cp -a "$T/src/github.com/openshift/openshift-azure/_data" .
 
 set_build_images
 
+export AZURE_REGIONS=$(awk '/^location:/ { print $2 }' <_data/containerservice.yaml)
```
jim-minter: I do not think this change should be here.  Why is it here?

test/e2e/specs/azure_monitor.go
```diff
@@ -0,0 +1,142 @@
...
+
+		By("Waiting for the log analytics agents to get ready")
+		err = wait.PollImmediate(2*time.Second, 5*time.Minute, ready.CheckDaemonSetIsReady(sanity.Checker.Client.Admin.AppsV1.DaemonSets(azureMonitorProject), nodeAgentName))
+		Expect(err).NotTo(HaveOccurred())
+		err = wait.PollImmediate(2*time.Second, 5*time.Minute, ready.CheckDeploymentIsReady(sanity.Checker.Client.Admin.AppsV1.Deployments(azureMonitorProject), clusterAgentName))
+		Expect(err).NotTo(HaveOccurred())
+	})
+})
+
+// withAzureMonitorIt wraps It blocks to make sure the tests are run on plugin versions supporting Azure Monitor (>= v9.0).
```
jim-minter: e2e tests already implicitly require a very latest cluster - don't try to solve that problem in this PR.  Remove withAzureMonitorIt and revert the changes associated with test/manifests/config.go.

test/e2e/specs/fakerp/securityupdates.go
```diff
@@ -3,20 +3,19 @@ package fakerp
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
```
jim-minter: all the changes in this file should be reverted

test/manifests/config.go
```diff
@@ -0,0 +1,19 @@
+package manifests
```
jim-minter: all the changes in this file should be reverted

test/e2e/specs/azure_monitor.go
jim-minter: I think I'm OK with this.

test/e2e/specs/azure_monitor.go
```diff
@@ -0,0 +1,142 @@
...
+
+		By("Waiting for the log analytics agents to get ready")
+		err = wait.PollImmediate(2*time.Second, 5*time.Minute, ready.CheckDaemonSetIsReady(sanity.Checker.Client.Admin.AppsV1.DaemonSets(azureMonitorProject), nodeAgentName))
+		Expect(err).NotTo(HaveOccurred())
+		err = wait.PollImmediate(2*time.Second, 5*time.Minute, ready.CheckDeploymentIsReady(sanity.Checker.Client.Admin.AppsV1.Deployments(azureMonitorProject), clusterAgentName))
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Update cluster config to disable azure monitor and update the workspace resource id")
+		*clusterConfig.Properties.MonitorProfile.Enabled = false
+		newWorkSpaceId := fmt.Sprintf("/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourcegroups/"+
```
jim-minter: How is this new workspace being used, if it is only set in conjunction with enabled being false?

test/e2e/specs/azure_monitor.go
```diff
@@ -0,0 +1,142 @@
...
+		rgName, err := random.LowerCaseAlphanumericString(5)
+		Expect(err).ToNot(HaveOccurred())
+		rgName = "e2e-workspace-" + rgName
+
+		By("Reading the cluster config before the azure monitor checks")
+		clusterConfig, err := azure.RPClient.OpenShiftManagedClusters.Get(ctx, os.Getenv("RESOURCEGROUP"), os.Getenv("RESOURCEGROUP"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(clusterConfig).NotTo(BeNil())
+		Expect(*clusterConfig.Properties.MonitorProfile.Enabled).To(Equal(true))
+		originalWorkapceId := *clusterConfig.Properties.MonitorProfile.WorkspaceResourceID
```
jim-minter: typo in variable name

hack/tests/e2e-upgrade.sh
charlesakalugwu: This change is here because there is currently a bug in `hack/tests/e2e-upgrade.sh` where a cluster could be created in one region and the upgrade is done with another region. See [these logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1950/pull-ci-azure-master-upgrade-v7.0/361/build-log.txt) for an example of this bug in action. The 7.0 cluster is created in westeurope and then the upgrade to the latest plugin version is attempted in australiasoutheast :man_shrugging:

pkg/api/validate/admin.go
```diff
@@ -68,6 +68,7 @@ func (v *AdminAPIValidator) validateUpdateContainerService(cs, oldCs *api.OpenSh
 	// validating ProvisioningState and ClusterVersion is the RP's responsibility
 	old.Properties.ProvisioningState = cs.Properties.ProvisioningState
 	old.Properties.ClusterVersion = cs.Properties.ClusterVersion
+	old.Properties.MonitorProfile = cs.Properties.MonitorProfile
```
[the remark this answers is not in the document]
charlesakalugwu: alright :+1:

ehashman:
/approve cancel

-- this PR has changed a bunch since I initially looked at it

ehashman:
Closing as stale.

Issue #1903, opened 2019-08-30
Connection reset by peer on scale down


m1kola:
```
time="2019-08-30T12:37:07Z" level=debug msg="400 Bad Request: Failed to apply request: UpdateWorkerAgentPoolDeleteVM: compute.VirtualMachineScaleSetVMsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: Delete https://management.azure.com/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/ci-1902-e2e-upgrade-seztls/providers/Microsoft.Compute/virtualMachineScaleSets/ss-infra-1567165546/virtualmachines/2?api-version=2018-10-01: read tcp 172.16.122.151:53208->13.71.185.242:443: read: connection reset by peer" func="pkg/fakerp.(*Server).badRequest()" file="pkg/fakerp/util.go:24"
time="2019-08-30T12:37:07Z" level=debug msg="ending:   PUT /admin/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/ci-1902-e2e-upgrade-seztls/providers/Microsoft.ContainerService/openShiftManagedClusters/ci-1902-e2e-upgrade-seztls" func="pkg/fakerp.(*Server).logger.func1()" file="pkg/fakerp/middleware.go:17"
time="2019-08-30T12:37:07Z" level=fatal msg="400: 400 Bad Request: Failed to apply request: UpdateWorkerAgentPoolDeleteVM: compute.VirtualMachineScaleSetVMsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: Delete https://management.azure.com/subscriptions/225e02bc-43d0-43d1-a01a-17e584a4ef69/resourceGroups/ci-1902-e2e-upgrade-seztls/providers/Microsoft.Compute/virtualMachineScaleSets/ss-infra-1567165546/virtualmachines/2?api-version=2018-10-01: read tcp 172.16.122.151:53208->13.71.185.242:443: read: connection reset by peer\n"
exit status 1
make: *** [upgrade] Error 1
```

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1902/pull-ci-azure-master-upgrade-v6.0/185/build-log.txt in #1902

/kind test-flake

thekad:
[...]

PR #1951, opened 2019-09-18
MDSD + azsec-clamav

thekad:
```release-note
Sending clamav scan events to geneva
```
This uses the new image built with azsec-clamav and clamav
Also removing the kickstart blocks now that v7 got pushed with kickstart.repo enabled cc @charlesakalugwu ptal


thekad:
/hold
this needs the new image published which @ehashman already triggered

ehashman:
@thekad image is published and ready for this PR

thekad:
Looking into failures, the docs say the instruction should succeed but that seems to not be the case, digging.

thekad:
Flakes #1903 and #1438 

/retest

thekad:
/hold cancel
cc @openshift/sig-azure

PR #1955, opened 2019-09-19
Fix iptables rule to restrict kubelet read-only port access to lo interface

### LINKED ISSUE #1952: Use iptables to restrict access to worker kubelet read-only port
(follow up from #1931)

AFAICS, end-user compute workload is able to read from the kubelet r/o port of the node it is running on.  This should be prevented.

```
[root@master-000000 ~]# oc new-app centos/ruby-25-centos7~https://github.com/sclorg/ruby-ex.git
warning: Cannot find git. Ensure that it is installed and in your path. Git is required to work with git repositories.
--> Found Docker image 4de2768 (3 hours old) from Docker Hub for "centos/ruby-25-centos7"

    Ruby 2.5 
    -------- 
    Ruby 2.5 available as container is a base platform for building and running various Ruby 2.5 applications and frameworks. Ruby is the interpreted scripting language for quick and easy object-oriented programming. It has many features to process text files and to do system management tasks (as in Perl). It is simple, straight-forward, and extensible.

    Tags: builder, ruby, ruby25, rh-ruby25

    * An image stream tag will be created as "ruby-25-centos7:latest" that will track the source image
    * A source build using source code from https://github.com/sclorg/ruby-ex.git will be created
      * The resulting image will be pushed to image stream tag "ruby-ex:latest"
      * Every time "ruby-25-centos7:latest" changes a new build will be triggered
    * This image will be deployed in deployment config "ruby-ex"
    * Port 8080/tcp will be load balanced by service "ruby-ex"
      * Other containers can access this service through the hostname "ruby-ex"

--> Creating resources ...
    imagestream.image.openshift.io "ruby-25-centos7" created
    imagestream.image.openshift.io "ruby-ex" created
    buildconfig.build.openshift.io "ruby-ex" created
    deploymentconfig.apps.openshift.io "ruby-ex" created
    service "ruby-ex" created
--> Success
    Build scheduled, use 'oc logs -f bc/ruby-ex' to track its progress.
    Application is not exposed. You can expose services to the outside world by executing one or more of the commands below:
     'oc expose svc/ruby-ex' 
    Run 'oc status' to view your app.
[root@master-000000 ~]# oc get pods
NAME              READY     STATUS      RESTARTS   AGE
ruby-ex-1-build   0/1       Completed   0          2m
ruby-ex-1-spwdw   1/1       Running     0          28s
[root@master-000000 ~]# oc get nodes -o wide|grep compute
compute-1568806316-000000   Ready     compute   7h        v1.11.0+d4cacc0   10.0.0.10     <none>        Red Hat Enterprise Linux Server 7.7 (Maipo)   3.10.0-1062.1.1.el7.x86_64   docker://1.13.1
[root@master-000000 ~]# oc rsh ruby-ex-1-spwdw
sh-4.2$ curl http://10.0.0.10:10255/
404 page not found
```

^ I would expect to see curl report `Connection refused` or to hang here.

Please additionally implement an e2e test to ensure this functionality is correct going forwards.

charlesakalugwu:
This is a follow-up to #1952 by additionally requests to only pass when coming through the loopback interface. Also added an e2e test which verifies the fix. 

```release-note
NONE
```

closes #1952 

/cc @jim-minter @mjudeikis @ehashman @kwoodson @Makdaam


test/e2e/specs/fakerp/kubeletsecurity.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/openshift/openshift-azure/pkg/util/random"
+	"github.com/openshift/openshift-azure/pkg/util/ready"
+	"github.com/openshift/openshift-azure/test/clients/openshift"
+	"github.com/openshift/openshift-azure/test/sanity"
+)
+
+const nginxTemplate = "nginx-example"
+
+func executeCommandInPod(client *openshift.Client, namespace, pod, container, command string) (string, string, error) {
```
kwoodson: consider moving this to `test/util` as it seems very useful.  Will not block on that.

test/e2e/specs/fakerp/kubeletsecurity.go
```diff
@@ -0,0 +1,118 @@
...
+	var stdout, stderr bytes.Buffer
+	req := client.CoreV1.RESTClient().
+		Post().
+		Namespace(namespace).
+		Resource("pods").
+		Name(pod).
+		SubResource("exec").
+		VersionedParams(&corev1.PodExecOptions{
+			Command:   []string{"/bin/sh", "-c", command},
```
kwoodson: are there any shell escaping libraries for `command`?  Just curious how the format of this string affects the outcome.

charlesakalugwu:
I got help from @Makdaam investigating why the rules only work momentarily (directly after applying them). The issue is because the rules added by this PR get periodically pushed down by rules added by the kubelet when it starts up. Particularly, pod traffic jumps to the `OPENSHIFT-FIREWALL-ALLOW` chain where it ends up always allowed. Our modifications to the iptables end up lower in the chain and are never applied to the pod traffic.

```
[root@compute-1568977772-000000 cloud-user]# iptables -S
-P INPUT ACCEPT
-P FORWARD DROP
-P OUTPUT ACCEPT
-N DOCKER
-N DOCKER-ISOLATION
-N KUBE-EXTERNAL-SERVICES
-N KUBE-FIREWALL
-N KUBE-FORWARD
-N KUBE-NODEPORT-NON-LOCAL
-N KUBE-SERVICES
-N OPENSHIFT-ADMIN-OUTPUT-RULES
-N OPENSHIFT-FIREWALL-ALLOW
-N OPENSHIFT-FIREWALL-FORWARD
-N OS_FIREWALL_ALLOW
-A INPUT -m conntrack --ctstate NEW -m comment --comment "kubernetes externally-visible service portals" -j KUBE-EXTERNAL-SERVICES
-A INPUT -m comment --comment "Ensure that non-local NodePort traffic can flow" -j KUBE-NODEPORT-NON-LOCAL
-A INPUT -m comment --comment "firewall overrides" -j OPENSHIFT-FIREWALL-ALLOW
-A INPUT -j KUBE-FIREWALL
-A INPUT -i lo -p tcp -m tcp --dport 10255 -j ACCEPT
-A INPUT -p tcp -m tcp --dport 10255 -j DROP
-A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
-A INPUT -p icmp -j ACCEPT
-A INPUT -i lo -j ACCEPT
-A INPUT -p tcp -m state --state NEW -m tcp --dport 22 -j ACCEPT
```

We need to investigate further.

jim-minter:
@charlesakalugwu and @kwoodson will work together to get to the bottom of what the iptables configuration needs to be here.

@charlesakalugwu this is failing verify tests.

kwoodson:
@charlesakalugwu @Makdaam and myself had a chat directly after our meeting this morning.  I'll attempt to summarize the ideas we discussed:
- The iptables rules get deprioritized when the kubelet starts.  The current method becomes unreliable when attempting to preemptively insert denial rules for the port in question.  The kubelet injects its own rules at the top of the chain thus never hitting the reject rules.
- We agreed that the proper fix lies in fixing the service that talks to the unauthenticated port.  If there is a place where we can contribute this fix then we could take a look.
- The fix that lives in OSD is a similar iptables rule that gets reapplied often.  It is in no means robust but slightly better than what is applied here because it is config managed and reapplied often.
- We had a brief chat about alternative solutions but felt uncomfortable applying them this late in the sprint cycle.
- @charlesakalugwu and @mgahagan73 were starting to implement test cases or manual tests. 

After a chat with @ehashman, we discussed that the iptables rules were probably insufficient and could probably be reverted since they fail to achieve the desired outcome.

ehashman:
Thanks @kwoodson for the detailed and helpful writeup. I'm going to revert the iptables change for the v9 release and we'll need to brainstorm about other possible solutions for this issue.

Closing in favour of #1962.

PR #1956, opened 2019-09-19
CI test driver script for Verification tests

mgahagan73:
script for starting verification-tests  (Bushslicer) in CI, needed dependences have been added to the ci base image by a previous commit.

```release-note
NONE
```

### DIFF
--- a/hack/tests/verification-tests.sh
+++ b/hack/tests/verification-tests.sh
@@ -16,7 +16,7 @@ function get_az_hosts() {
 BRANCH="${BRANCH:-v3}"
 VERIFICATION_TESTS_GIT="${VERIFICATION_TESTS_GIT:-https://github.com/openshift/verification-tests.git}"
 # Env vars for setting up test
-ARO_PUBLIC_HOSTNAME=$(awk '/^  publicHostname:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)
+ARO_PUBLIC_HOSTNAME="${ARO_PUBLIC_HOSTNAME:-$(awk '/^  publicHostname:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)}"
 RESOURCEGROUP="${RESOURCEGROUP:-$(awk '/^name:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)}"
 export OPENSHIFT_ENV_OSE_WEB_CONSOLE_URL="${OPENSHIFT_ENV_OSE_WEB_CONSOLE_URL:-https://${ARO_PUBLIC_HOSTNAME}/}"
 export BUSHSLICER_CONFIG='{"global": {"browser": "chrome"}}'
@@ -33,8 +33,6 @@ env
 echo =========================
 
 git clone --branch $BRANCH $VERIFICATION_TESTS_GIT
-#scl enable rh-ror50 verification-tests/tools/install_os_deps.sh
-#scl enable rh-ror50 verification-tests/tools/hack_bundle.rb
 pushd verification-tests
 tools/openshift-ci/verification_tests_ci_entrypoint.sh $TESTS_URL
 # do what we need to do to publish logs under junit-report/
--- a/hack/tests/verification-tests.sh
+++ b/hack/tests/verification-tests.sh
@@ -1,0 +1,40 @@
+#!/bin/bash -e
+# Get verification tests source code prepare it and run the verification tests (Bushslicer)
+
+set -o pipefail
+
+. hack/tests/ci-prepare.sh
+
+function get_az_hosts() {
+  local IFS=$'\n'
+  for ip in $(az vmss list-instance-public-ips -g $RESOURCEGROUP -n ss-master --query "[].ipAddress" -o tsv) ; do
+    i+=${ip}':master:node,'
+  done
+  echo ${i%?}
+}
+
+BRANCH="${BRANCH:-v3}"
+VERIFICATION_TESTS_GIT="${VERIFICATION_TESTS_GIT:-https://github.com/openshift/verification-tests.git}"
+# Env vars for setting up test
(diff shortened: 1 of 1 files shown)


hack/tests/verification-tests.sh
```diff
@@ -0,0 +1,38 @@
...
+export OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_SSH_PRIVATE_KEY="${OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_SSH_PRIVATE_KEY:-${HOME}/go/src/github.com/openshift/openshift-azure/_data/_out/id_rsa}"
+export OPENSHIFT_ENV_OSE_USER_MANAGER_USERS="${OPENSHIFT_ENV_OSE_USER_MANAGER_USERS:-:MYLOGINTOKEN}"
+export OPENSHIFT_ENV_OSE_HOSTS="${OPENSHIFT_ENV_OSE_HOSTS:-$(get_az_hosts)}"
+# must be URL to new line separated case IDs in format OCP-xxxx
+TESTS_URL="${TESTS_URL:-http://aosqe-test-lists.s3-website-us-east-1.amazonaws.com/lists/3.11_ded_azure.lst}"
+echo =========================
+env
+echo =========================
+
+git clone --branch $BRANCH $VERIFICATION_TESTS_GIT
```
thekad: make a shallow clone? i.e. `--depth 1`

hack/tests/verification-tests.sh
```diff
@@ -0,0 +1,38 @@
...
+# Env vars for setting up test
+ARO_PUBLIC_HOSTNAME="${ARO_PUBLIC_HOSTNAME:-$(awk '/^  publicHostname:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)}"
+RESOURCEGROUP="${RESOURCEGROUP:-$(awk '/^name:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)}"
+export OPENSHIFT_ENV_OSE_WEB_CONSOLE_URL="${OPENSHIFT_ENV_OSE_WEB_CONSOLE_URL:-https://${ARO_PUBLIC_HOSTNAME}/}"
+export BUSHSLICER_CONFIG='{"global": {"browser": "chrome"}}'
+export BUSHSLICER_DEFAULT_ENVIRONMENT="${BUSHSLICER_DEFAULT_ENVIRONMENT:-ose}"
+export OPENSHIFT_ENV_OSE_API_PORT="${OPENSHIFT_ENV_OSE_API_PORT:-443}"
+export OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_USER="${OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_USER:-cloud-user}"
+export OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_SSH_PRIVATE_KEY="${OPENSHIFT_ENV_SERVICES_AZURE_HOST_CONNECT_OPS_SSH_PRIVATE_KEY:-${HOME}/go/src/github.com/openshift/openshift-azure/_data/_out/id_rsa}"
+export OPENSHIFT_ENV_OSE_USER_MANAGER_USERS="${OPENSHIFT_ENV_OSE_USER_MANAGER_USERS:-:MYLOGINTOKEN}"
```
thekad: where does `MYLOGINTOKEN` come from?

hack/tests/verification-tests.sh
```diff
@@ -0,0 +1,38 @@
...
+  for ip in $(az vmss list-instance-public-ips -g $RESOURCEGROUP -n ss-master --query "[].ipAddress" -o tsv) ; do
+    i+=${ip}':master:node,'
+  done
+  echo ${i%?}
+}
+
+BRANCH="${BRANCH:-v3}"
+VERIFICATION_TESTS_GIT="${VERIFICATION_TESTS_GIT:-https://github.com/openshift/verification-tests.git}"
+# Env vars for setting up test
+ARO_PUBLIC_HOSTNAME="${ARO_PUBLIC_HOSTNAME:-$(awk '/^  publicHostname:/{ print $2}' ${HOME}/go/src/github.com/openshift/openshift-azure/_data/containerservice.yaml)}"
```
thekad: these calls that wrap awk in bash param expansion are perhaps too convoluted, can you move them to more clear `if` statements? e.g. `if [[ -z "$ARO_PUBLIC_HOSTNAME" ]]; then #do the awk magic here`

hack/tests/verification-tests.sh
mgahagan73: Thanks for the suggestion, less load on upstream is always good. I'll fix it.

hack/tests/verification-tests.sh
mgahagan73: At present exactly how we get the login token is still being worked out. The previous job for Bushslicer pulled a username and password from the containerservice file. That user account is no longer available. I intend to have the caller of the script provide the token as one of the environment variables as is the case for most of the other config variables. MYLOGINTOKEN is just a placeholder to hopefully make it easier to see why nothing ran in the event that we never set a login token.

hack/tests/verification-tests.sh
mgahagan73: I can refactor those if it will make it more readable. Other option is to make these come from the environment variables set by the caller of the script and simply exit if they are not provided. I'll also see if there is a less convoluted way to parse yaml in the shell.

ehashman: I'm not sure this will be more readable with a bunch of if statements, it's going to be a mess either way :)

I think that matching the exact spacing for the `publicHostname` here might be fragile though, can we match some arbitrary amount of whitespace there?

jim-minter:
@mgahagan73 are you planning to make further changes on this?

mgahagan73:
@jim-minter  Yes I have another revision on the way.

asalkeld:
happy to merge this if it is going to help you progress/iterate.. In you view is this ready to merge now?

PR #1959, opened 2019-09-19
Removing assignee from issue creation.

kwoodson:
Removing sig-azure as assignee.  Will leave uncommented until we determine what we want to do in the meanwhile.  

Options:
- look up primary in the on call
- default list of assignees

```release-note:
None
```

### DIFF
--- a/hack/ci-notify/main.go
+++ b/hack/ci-notify/main.go
@@ -98,12 +98,10 @@ func (c *client) createOrUpdateJobIssue(ctx context.Context, jobName, comment, o
 	}
 
 	if issue == nil {
-		users := []string{"openshift/sig-azure"}
 		fmt.Print("No issue found.  Creating one")
 		ir := &github.IssueRequest{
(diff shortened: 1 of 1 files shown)


m1kola:
Do I understand correctly that the issue now is that we will not get notified about new tickets without having asignee?

Should we modify the issue/comment message to metion the `@openshift/sig-azure` team so everyone on team will get a notification about a comment?

kwoodson:
> Do I understand correctly that the issue now is that we will not get notified about new tickets without having asignee?

The assignee is what is causing this to not work.

> Should we modify the issue/comment message to metion the @openshift/sig-azure team so everyone on team will get a notification about a comment?

That sounds fine to me.

thekad:
It doesn't really matter who the assignee is, anyone who lgtm or approves or etc. will be assigned the issue. Adding the team mention in the body of the issue should suffice imo.

PR #1965, opened 2019-09-23
Adding workspaceID to admin api

gburges:
```release-note
NONE
```


gburges:
@charlesakalugwu The doesn't fail with this, but not sure if it's a false negative

ehashman:
What's the context around this PR? Is there a JIRA? I'm not understanding what this is needed for :)

PR #1967, opened 2019-09-24
Improves make vmimage-validate locally

m1kola:
Currently scap report and yum log files are not being cleaned up between `make vmimage-validate` runs. It's not an issue in CI, because we run jobs in clean environments, but can cause confusion when run locally because the validation will fail even if the image is valid. It happens because the command checks if files contain data or not.

Example scenario:

* You run validation locally and it fails due to pending security updates
* You fix the issue/update vm image
* You run validation locally again: it fails becasue `yum_updateinfo_list_security` log contains data from previous run

```release-note
NONE
```

### DIFF
--- a/pkg/vmimage/ssh.go
+++ b/pkg/vmimage/ssh.go
@@ -49,7 +49,7 @@ func (builder *Builder) scp(files []string) error {
 		}
 
 		builder.Log.Infof("download %s", file)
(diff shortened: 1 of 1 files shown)


kwoodson:
@m1kola sgtm

/lgtm

m1kola:
Forgot to start the vmimage job.

/test vmimage

PR #1972, opened 2019-09-24
start v10.0

jim-minter:
```release-note
NONE
```


jim-minter:
@asalkeld note second commit fixes logic error introduced in 8f881e236b39d821508f6a6fd10e3895b7ed7e49

PR #1973, opened 2019-09-24
Make the way we get the latest dev version more reliable and run "make content"

asalkeld:
esp. since we are going into double digits, "ls" is not enough.
numeric listing for our dev directories currently produces "v10  v6  v7  v71  v9" instead of
"v10  v9 v71 v7 v6"

Also run the content update as it doesn't seem to have run for a while.

```release-note
NONE
```


Makefile
```diff
@@ -3,7 +3,7 @@ GITCOMMIT=$(shell git describe --tags HEAD)$(shell [[ $$(git status --porcelain)
 LDFLAGS="-X main.gitCommit=$(GITCOMMIT)"
 
 AZURE_IMAGE ?= quay.io/openshift-on-azure/azure:$(GITCOMMIT)
-LATEST_PLUGIN_VERSION=v$(shell ls -d pkg/sync/v* | sed -e 's/.*v//' | tail -1)
+LATEST_PLUGIN_VERSION=$(shell go run hack/dev-version/dev-version.go)
```
thekad: adding a `sort` would've sufficed in here.

hack/dev-version/dev-version.go
```diff
@@ -0,0 +1,35 @@
...
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
+
+	pluginapi "github.com/openshift/openshift-azure/pkg/api/plugin"
+)
+
+func pluginToDevVersion(pluginVersion string) string {
+	var major, minor int
+	fmt.Sscanf(pluginVersion, "v%d.%d", &major, &minor)
```
thekad: This is wrong. This would cause e.g. `v2.2` to equal `v22`

hack/dev-version/dev-version_test.go
```diff
@@ -0,0 +1,34 @@
...
+			want:          "v92",
+		},
+		{
+			pluginVersion: "v11.0",
+			want:          "v11",
+		},
+		{
+			pluginVersion: "v12.4",
+			want:          "v124",
+		},
```
thekad: you need test cases for e.g. `v1.24 != v12.4`

hack/dev-version/dev-version.go
asalkeld: actually, that is exactly what we have currently. see: https://github.com/openshift/openshift-azure/tree/master/pkg/sync
v7.1 becomes the directory v71
v9.0 becomes v9
v10.0 becomes v10
no sort is going to figure that out

hack/dev-version/dev-version_test.go
asalkeld: i am not sure what you mean here as the directory never has a "." in it. we are probably safe from the following collision ("v11.0" -> "v11" and "v1.1" -> v11) as we only keep 2-3 versions in tree.

thekad:
/lgtm
thanks for clarifying that for me :)

ehashman:
Re: run `make content` -- what happened to our content update bot? @mjudeikis ?

PR #1981, opened 2019-10-02
Refactor fakeRP to support APIVersion

mjudeikis:
```release-note
NONE
```

This is very rough refactoring to support multiple APIVersions. I know it can be improved. Like everything and everywhere:) 

See last commit only


Makdaam:
One suggestion, to add 
```
if [[ -n "$API_VERSION" ]]; then
    API_VERSION="-api-version=$API_VERSION"
fi
```
and change the go run in hack/ scripts which run cmd/createorupdate to include ${API_VERSION:-}

Right now there's no way to interface with the new code without executing go run manually.

PR #1989, opened 2019-10-04
Updates imports to use specific versions

m1kola:
Let's use explicit versions in our imports instead of `latest` alias. It will make SDK upgrades more predictable.

```release-note
NONE
```


m1kola:
@mjudeikis cc'ing, because it changes the code you are originally wrote. Please take a look when you have a moment.

/cc @mjudeikis

PR #1991, opened 2019-10-07
auto-generated docs

mjudeikis:
```release-note
NONE
```

AutoGenerated test commands docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,7 @@ releasenotes:
 
 content:
 	go test -timeout=300s -tags=content -run=TestContent ./pkg/sync/$(LATEST_PLUGIN_VERSION)
+	go run ./hack/generate-test-commands >docs/commands.md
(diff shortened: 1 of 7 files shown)


test/commands_test.go
```diff
@@ -0,0 +1,68 @@
...
+	err = yaml.Unmarshal(b, &prow)
+	if err != nil {
+		t.Error(err)
+	}
+
+	for _, c := range prow.Presubmits.OpenshiftOpenshiftAzure {
+		commands = append(commands, c.RerunCommand)
+	}
+
+	f, err := os.OpenFile("../docs/commands.md",
```
jim-minter: use `os.Create()`?

ehashman: Oh, yeah, this is it... opened with APPEND mode below is the source of the bug.

jim-minter:
fixed
/lgtm

PR #1999, opened 2019-10-15
Pass the privateEndpoint to the customer using FQDN and PublicHostname

asalkeld:
```release-note
When using private cluster, the API server's IP address can be obtained from the FQDN and PublicHostname fields.
```


asalkeld:
this is wrong (privateEndpoint is what the RP uses)

PR #2001, opened 2019-10-16
Private cluster work items

asalkeld:
```release-note
Add PrivateAPIServer boolean to the new public API
```
/cc @jim-minter 

Current state: **e2e tests working locally**

Private Cluster TODO list: could all be follow ups?

1. ssh: implement hack/ssh.sh using proxy (I still need to copy id_rsa's about the place) and fix test/util/diagnostics ssh (optional)


pkg/api/2019-10-27-preview/converterfrominternal.go
```diff
@@ -34,6 +34,7 @@ func FromInternal(cs *api.OpenShiftManagedCluster) *OpenShiftManagedCluster {
 		ClusterVersion:    &cs.Properties.ClusterVersion,
 		PublicHostname:    &cs.Properties.PublicHostname,
 		FQDN:              &cs.Properties.FQDN,
+		PrivateAPIServer:  &cs.Properties.PrivateAPIServer,
```
amanohar: Was the final decision to expose this at the top level? IMO this shouldn't be a top level property.

asalkeld: I am happy to move it (to MasterAgentPool?)

pkg/config/v11/generate.go
```diff
@@ -26,6 +40,15 @@ func (g *simpleGenerator) Generate(template *pluginapi.Config, setVersionFields
 	g.cs.Config = *config
 	c := &g.cs.Config
 
+	if g.cs.Properties.PrivateAPIServer {
+		managementIP, err := getLastUsableIP(*g.cs.Properties.NetworkProfile.ManagementSubnetCIDR)
```
asalkeld: this obviously should be g.cs.Properties.AgentPoolProfiles[0].SubnetCIDR

pkg/arm/v11/resources.go
```diff
@@ -60,21 +60,24 @@ func (g *simpleGenerator) vnet() *network.VirtualNetwork {
...
-				DomainNameLabel: to.StringPtr(derived.MasterLBCNamePrefix(g.cs)),
-			},
-			IdleTimeoutInMinutes: to.Int32Ptr(15),
+			IdleTimeoutInMinutes:     to.Int32Ptr(15),
 		},
 		Name:     to.StringPtr(armconst.IPAPIServerName),
 		Type:     to.StringPtr("Microsoft.Network/publicIPAddresses"),
 		Location: to.StringPtr(g.cs.Location),
 	}
+	if !g.cs.Properties.PrivateAPIServer {
```
asalkeld: note: derived.MasterLBCNamePrefix() uses FQDN and with private cluster that is an IP

pkg/api/2019-10-27-preview/converterfrominternal.go
asalkeld: I have now moved it to the MasterPoolProfile

asalkeld:
@jim-minter I am still getting the following error:

```
DEBU[2019-10-18T15:40:55+10:00]pkg/util/wait/wait.go:71 pkg/util/wait.ForHTTPStatusOk.func1() ForHTTPStatusOk: will retry on the following error Get https://10.0.0.254/healthz: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers) 
DEBU[2019-10-18T15:41:03+10:00]pkg/fakerp/util.go:24 pkg/fakerp.(*Server).badRequest() 400 Bad Request: Failed to apply request: WaitForOpenShiftAPI: Get https://10.0.0.254/healthz: x509: certificate is valid for 172.30.0.1, not 10.0.0.254 
```
Could it be because I am using australiasoutheast region?

pkg/arm/v11/resources.go
```diff
@@ -265,6 +306,13 @@ func (g *simpleGenerator) ilbAPIServer() *network.LoadBalancer {
 		Location: to.StringPtr(g.cs.Location),
 	}
 
+	if g.cs.Properties.PrivateAPIServer {
```
jim-minter: we only call ilbAPIServer() if PrivateAPIServer is set, fwiw, so these might go more clearly rolled in above.

pkg/arm/v11/resources.go
```diff
@@ -558,11 +606,14 @@ func vmss(cs *api.OpenShiftManagedCluster, app *api.AgentPoolProfile, backupBlob
...
 			},
 		}
-		(*(*vmss.VirtualMachineProfile.NetworkProfile.NetworkInterfaceConfigurations)[0].VirtualMachineScaleSetNetworkConfigurationProperties.IPConfigurations)[0].PublicIPAddressConfiguration = &compute.VirtualMachineScaleSetPublicIPAddressConfiguration{
-			Name: to.StringPtr(armconst.VmssNicPublicIPConfigurationName),
-			VirtualMachineScaleSetPublicIPAddressConfigurationProperties: &compute.VirtualMachineScaleSetPublicIPAddressConfigurationProperties{
-				IdleTimeoutInMinutes: to.Int32Ptr(15),
-			},
+
+		if !cs.Properties.PrivateAPIServer {
+			(*(*vmss.VirtualMachineProfile.NetworkProfile.NetworkInterfaceConfigurations)[0].VirtualMachineScaleSetNetworkConfigurationProperties.IPConfigurations)[0].PublicIPAddressConfiguration = &compute.VirtualMachineScaleSetPublicIPAddressConfiguration{
```
jim-minter: this, and the removal of the public LB rules, will both affect the monitoring script - it will hang waiting for these IPs to be populated - will need an update.

pkg/fakerp/proxy.go
```diff
@@ -23,37 +24,26 @@ func (c *conn) Read(b []byte) (int, error) {
...
-			return err
-		}
-		ca, err := ioutil.ReadFile("secrets/proxy-ca.pem")
-		if err != nil {
-			return err
-		}
-		if ok := roots.AppendCertsFromPEM(ca); !ok {
-			return fmt.Errorf("error configuring proxy")
-		}
+	secretsPath := "secrets"
```
jim-minter: thanks!

jim-minter:
> @jim-minter I am still getting the following error


I think you're not baking the ILB IP address into the serving cert properly around here:

https://github.com/openshift/openshift-azure/blob/98cd1a1d56299a8bf4f8cd9fd4d68cf439d39bff/pkg/config/v11/generate.go#L164-L173

In the external case, FQDN is a hostname and goes in DNSNames.  In the internal case, it's an IP and needs to go in IPAddresses.

This issue would almost certainly be enough to prevent non-master nodes registering too.

pkg/arm/v11/resources.go
asalkeld: will do

pkg/arm/v11/resources.go
asalkeld: I know, will do. i'll post a todo list in the description

asalkeld:
@jim-minter ptal, imho we *could* merge and do other work as followup? It's all protected with PrivateAPIServer..

asalkeld:
/test e2e-create-20190930 
for the above: quay.io pull error

The upgrade-v10.0 has "PrivateAPIServer:  true", I think this is because the v10 fakerp assumed that if the managementsubnet was non-nil it was a private cluster

ehashman:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2001/pull-ci-azure-master-e2e-create-20191027-private/7 failure looks like #2034 to me?

pkg/arm/v11/resources.go
```diff
@@ -243,15 +284,41 @@ func (g *simpleGenerator) ilbAPIServer() *network.LoadBalancer {
...
+						EnableFloatingIP:     to.BoolPtr(false),
+					},
+					Name: to.StringPtr(armconst.LbSSHLoadBalancingRuleName),
+				},
 			},
 			Probes: &[]network.Probe{
 				{
 					ProbePropertiesFormat: &network.ProbePropertiesFormat{
-						Protocol:          network.ProbeProtocolHTTPS,
+						Protocol:          network.ProbeProtocolTCP,
```
jim-minter: does that want to be reverted?

jim-minter: (and the /healthz deletion below too)

test/e2e/standard/cluster.go
```diff
@@ -333,6 +333,7 @@ func (sc *SanityChecker) checkCanAccessService(ctx context.Context, url string,
 	}
 
 	if sc.cs.Properties.NetworkProfile.PrivateEndpoint != nil && usePE {
+		inner.TLSClientConfig.InsecureSkipVerify = true
```
jim-minter: Please don't do this even in test - override DialTLS - copy roundtrippers.HealthCheck

test/sanity/sanity.go
```diff
@@ -28,8 +31,25 @@ func init() {
...
+		conf, err := client.NewServerConfig(log, cs)
+		if err != nil {
+			panic(err)
+		}
+		authorizer, err := azureclient.NewAuthorizerFromEnvironment("")
+		if err != nil {
+			panic(err)
+		}
+		ctx = context.WithValue(ctx, api.ContextKeyClientAuthorizer, authorizer)
+		cs.Properties.NetworkProfile.PrivateEndpoint, err = fakerp.GetPrivateEndpointIP(ctx, log, cs.Properties.AzProfile.SubscriptionID, conf.ManagementResourceGroup, cs.Properties.AzProfile.ResourceGroup)
```
jim-minter: a comment about what's being done here and why would be useful

cmd/monitoring/monitoring.go
```diff
@@ -267,14 +186,41 @@ func main() {
...
-	if err := m.init(ctx, log); err != nil {
-		log.Fatalf("Cannot initialize monitor: %v", err)
+	oc, err := loadOCConfig()
+	if err != nil {
+		log.Fatalf("Cannot load clusterConfiguration from %s: %v", *configfile, err)
 	}
 
-	if err := m.run(ctx); err != nil {
+	var mAPI monitor
+	privateAPIServer := (os.Getenv("PRIVATE_API_SERVER") == "true")
```
jim-minter: this is shady - can't you get at the value via the oc variable fetched above?

jim-minter:
I haven't reviewed the monitoring changes closely, but overall this is looking pretty sound.  A couple of minor nits, check the test and merge.

pkg/arm/v11/resources.go
asalkeld: yes, but at the time of pushing i still needed it. the fix should be rolled out by now, i'll test

pkg/plugin/plugin_test.go
```diff
@@ -409,3 +409,53 @@ func TestValidateUpdateBlock(t *testing.T) {
 		t.Error("expected current cluster to pass update validation")
 	}
 }
+
+func TestGetPrivateAPIServerIPAddress(t *testing.T) {
```
ehashman: This test serves as decent documentation but we might also want to write down how this works somewhere for posterity's sake.

pkg/arm/v11/resources.go
asalkeld: I chatted to Julien this morning in the scrum and he said the azure fix is not totally rolled out yet
but when it is, we need to revert it (2 liner). I'd suggest merge as-is and after a couple of days revert it.

test/util/diagnostics/ssh.go
```diff
@@ -91,7 +91,7 @@ func (s *ssher) Dial(ctx context.Context, hostname string) (*ssh.Client, error)
 			return nil, err
 		}
 
-		if int(instance) > len(s.masterIPs) {
+		if int(instance)+1 > len(s.masterIPs) {
```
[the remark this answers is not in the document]
asalkeld: yes

pkg/plugin/plugin_test.go
asalkeld: i'll follow up with docs on the interface

asalkeld:
flake #1628 
/test upgrade-v7.1

Issue #1775, opened 2019-07-03
scale up/down timed out waiting for the condition


gburges:
Looks like a new unhandled error

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1723/pull-ci-azure-master-e2e-create/465

/kind test-flake

asalkeld:
scaling ss-infra-1569551297 capacity up from 2 to 3

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1977/pull-ci-azure-master-upgrade-v7.1/124

Makdaam:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2128/pull-ci-azure-master-e2e-create-20191027-private/218

PR #2004, opened 2019-10-16
Add PrivateAPIServer to the public API

asalkeld:
```release-note
Add PrivateAPIServer field to MasterPoolProfile 
```

### DIFF
--- a/pkg/api/2019-10-27-preview/convertertointernal.go
+++ b/pkg/api/2019-10-27-preview/convertertointernal.go
@@ -90,6 +90,9 @@ func mergeProperties(oc *OpenShiftManagedCluster, cs *api.OpenShiftManagedCluste
 	if oc.Properties.FQDN != nil {
 		cs.Properties.FQDN = *oc.Properties.FQDN
 	}
+	if oc.Properties.MasterPoolProfile != nil && oc.Properties.MasterPoolProfile.PrivateAPIServer != nil {
+		cs.Properties.PrivateAPIServer = *oc.Properties.MasterPoolProfile.PrivateAPIServer
+	}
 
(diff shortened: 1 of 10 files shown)


pkg/api/2019-10-27-preview/convertertointernal.go
```diff
@@ -90,6 +90,9 @@ func mergeProperties(oc *OpenShiftManagedCluster, cs *api.OpenShiftManagedCluste
 	if oc.Properties.FQDN != nil {
 		cs.Properties.FQDN = *oc.Properties.FQDN
 	}
+	if oc.Properties.MasterPoolProfile != nil && oc.Properties.MasterPoolProfile.PrivateAPIServer != nil {
+		cs.Properties.PrivateAPIServer = *oc.Properties.MasterPoolProfile.PrivateAPIServer
+	}
```
ehashman: Is this something that should be updateable? What would the user expect to happen if this was switched from "true" to "false" or vice versa?

pkg/cluster/hash_test.go
```diff
@@ -155,6 +155,7 @@ func TestHashScaleSetStability(t *testing.T) {
 			cs.Config.Images.ImagePullSecret = []byte{}
 			cs.Config.Images.GenevaImagePullSecret = []byte{}
 		default:
+			cs.Properties.NetworkProfile.ManagementSubnetCIDR = nil
```
ehashman: Why do we need to set this in this PR?

asalkeld: removed

asalkeld:
flake https://github.com/openshift/openshift-azure/issues/1775

PR #2014, opened 2019-10-17
Fix the latest dev-version for minor releases that don't have a direcory

asalkeld:
```release-note
NONE
```
/cc @ehashman 

This situation is as follows:
we make a new minor release say 10.1, but don't have any sync/arm/config changes.
When we try and get the the latest versioned directory it breaks.


ehashman:
flake for v9 upgrade is a quay timeout issue
new flake #2020 for v10 upgrade

/retest

Issue #1932, opened 2019-09-10
Directory_ResourceSizeExceeded the size of the object has exceeded its limit


kwoodson:
Error while attempting to run `test e2e_create` on #1925 
```
time="2019-09-10T15:23:30Z" level=fatal msg="cannot update aad app secret: graphrbac.ApplicationsClient#Patch: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code=\"Unknown\" Message=\"Unknown service error\" Details=[{\"odata.error\":{\"code\":\"Directory_ResourceSizeExceeded\",\"date\":\"2019-09-10T15:23:30\",\"message\":{\"lang\":\"en\",\"value\":\"The size of the object has exceeded its limit. Please reduce the number of values and retry your request.\"},\"requestId\":\"3484ebe7-c4ea-4a0f-a17d-07f920659062\"}}]"
```

Makdaam:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1926/pull-ci-azure-master-e2e-create/1029 on #1926

Makdaam:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1926/pull-ci-azure-master-upgrade-v7.0/230 on #1926 in australiasoutheast

Makdaam:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1926/pull-ci-azure-master-e2e-create/1035 and https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1926/pull-ci-azure-master-upgrade-v7.0/236

kwoodson:
We need to look at this:
```
cannot update aad app secret: graphrbac.ApplicationsClient#Patch: Failure responding to request: StatusCode=403
```

jim-minter:
Looking at the AAD audit logs.  This suddenly started failing from 10:23:30 AM (central) today.  It's mostly failing but occasionally succeeding.  I can't see anything wrong with the object in question, 
aro-ci-aad-shared.  Going to open a support case with MSFT.

jim-minter:
[...]

PR #2021, opened 2019-10-18
Allow validation of an existing built image

ehashman:
```release-note
NONE
```

This lets you export `IMAGE_RESOURCENAME` so you don't need to build a new image if you've done something silly that messed up the test cluster create or e2e, or if your e2e flakes.

Note: you may also need to set `AZURE_REGIONS` to ensure validation happens in the same region the test image was built.

### DIFF
--- a/hack/vmimage.sh
+++ b/hack/vmimage.sh
@@ -21,12 +21,15 @@ else
 fi
 
 export IMAGE_RESOURCEGROUP="${IMAGE_RESOURCEGROUP:-images}"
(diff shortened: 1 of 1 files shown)


ehashman:
/test upgrade-v10.0

network flake

ehashman:
So `make vmimage-validate` runs validation steps but doesn't run e2e tests or tag an image as valid. Hence why I made this change. One should be able to run the vmimage e2e stuff on an existing image IMO. I don't know if that means we should change the `vmimage` target to be `vmimage-e2e` or something.

jim-minter:
I think this is fine/
/lgtm
/hold cancel

ehashman:
/hold

Waiting until we fix CI

ehashman:
flake #1932 

/retest

Issue #1870, opened 2019-08-20
quay.io timeout from southcentral


jim-minter:
```
docker pull quay.io/openshift-on-azure/azure:v6.0
Get https://quay.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

@charlesakalugwu @Makdaam @ehashman @shrutir25 fyi

jim-minter:
issues from at least 11th august
[stderr.txt](https://github.com/openshift/openshift-azure/files/3522529/stderr.txt)

jim-minter:
interesting that on this occasion, eventually the pull succeeded, but presumably the ARM template had timed out by that point.
Also notice that a previous `docker pull` succeeded so the VM must have had internet access at least in the beginning

asalkeld:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2036/pull-ci-azure-master-e2e-create-20190430/3
(eastus region)

asalkeld:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2030/pull-ci-azure-master-upgrade-v9.0/190

PR #2030, opened 2019-10-23
Validate that VM and container versions of OCP match in pluginconfig (golang)

asalkeld:
```release-note
NONE
```
/cc @ehashman
(easier than working around different python versions?)


asalkeld:
flake upgrade-v9.0: https://github.com/openshift/openshift-azure/issues/1870

PR #2038, opened 2019-10-24
Support upgrading from v10.1

asalkeld:
```release-note
NONE
```
/cc @ehashman


asalkeld:
This is not working as v10.1.1 is an invalid version according to https://github.com/openshift/openshift-azure/blob/master/pkg/api/validate/plugin.go#L38-L44

Issue #661, opened 2018-10-25
Unknown network allocation error.


kargakis:
```
time="2018-10-25T14:19:57Z" level=debug msg="400 Bad Request: Failed to apply requestCode=\"NetworkingInternalOperationError\" Message=\"Unknown network allocation error.\"" resourceGroup=ci-op-lb5vf50g-226a1
```
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/660/pull-ci-openshift-openshift-azure-master-e2e-azure-upgrade/5/build-log.txt

/kind test-flake
@amanohar @sozercan  any idea what's going on here?

amanohar:
@kargakis I looked into and the error happening during VMSS and NRP communication. I am following up with the VMSS team (tracked via an incident) to understand the root cause.

thekad:
@amanohar any updates on this?

jim-minter:
haven't seen this recently, closing

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1802/pull-ci-azure-master-upgrade-v5.2/244/build-log.txt in #1802

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1853/pull-ci-azure-master-upgrade-v5.2/439/build-log.txt #1853

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1853/pull-ci-azure-master-upgrade-v7.0/1/build-log.txt also #1853

m1kola:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1860/pull-ci-azure-master-upgrade-v7.0/49/build-log.txt in #1860

asalkeld:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2044/pull-ci-azure-master-upgrade-v7.1/313

PR #2044, opened 2019-10-29
Remove PeerVnetID from 2019-10-27-preview API

asalkeld:
```release-note
Remove  PeerVnetID from 2019-10-27-preview API
```

Note microsoft will store the api that a cluster is created with and not allow a cluster to be updated
with an older version.


asalkeld:
/retest
flake https://github.com/openshift/openshift-azure/issues/661

PR #2045, opened 2019-10-29
Make the dns server more dynamic (for private cluster case)

asalkeld:
```release-note
NONE
```
This is required by private cluster when the customer provides their own DNS server.
We just want to populate /etc/origin/node/resolv.conf with what the dnsmasq has in
/etc/dnsmasq.d/origin-upstream-dns.conf
/cc @jim-minter


pkg/startup/v11/derive.go
```diff
@@ -71,6 +74,38 @@ func (derivedType) WorkerCloudProviderConf(cs *api.OpenShiftManagedCluster) ([]b
...
+		return nil, fmt.Errorf("no servers found in origin-upstream-dns.conf")
+	}
+	return servers, nil
+}
+
+func (d *derivedType) NameServers() ([]string, error) {
+	dnsConfFile := "/etc/dnsmasq.d/origin-upstream-dns.conf"
+	var content string
+	if _, err := os.Stat(dnsConfFile); os.IsNotExist(err) {
+		// only in unittests
```
jim-minter: can you be sure? if there's a startup ordering issue this could hide it.  I'd rather not risk that.

pkg/startup/v11/derive.go
```diff
@@ -71,6 +74,38 @@ func (derivedType) WorkerCloudProviderConf(cs *api.OpenShiftManagedCluster) ([]b
 	return derivedpkg.WorkerCloudProviderConf(cs, false)
 }
 
+func getServerFromDNSConf(content string) ([]string, error) {
+	//  cat /etc/dnsmasq.d/origin-upstream-dns.conf
+	//  server=168.63.129.16
+	servers := []string{}
+	for _, line := range strings.Split(content, "\n") {
+		if !strings.Contains(line, "server=") {
```
jim-minter: HasPrefix?

jim-minter:
a bit grubby but works - minor nits then please merge

pkg/startup/v11/derive.go
asalkeld: I know, trying to work around integration tests :-(
I'll try and improve it

asalkeld:
/hold
I'll clean this up tomorrow morning

jim-minter:
@asalkeld the test errors are genuine, ptal.  Using https://github.com/spf13/afero would be a neater approach test-wise but can accept what's here for now.

asalkeld:
@jim-minter this is horrible, i'll merge my 2 related prs and hopefully have something better

PR #2049, opened 2019-10-29
fix typo

jim-minter:
```release-note
NONE
```
CI secrets have been updated

### DIFF
--- a/pkg/fakerp/config.go
+++ b/pkg/fakerp/config.go
@@ -67,7 +67,7 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/fakerp/config.go
```diff
@@ -67,7 +67,7 @@ func GetPluginTemplate() (*pluginapi.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	genevaPullSecret, err := ioutil.ReadFile("secrets/acr-docker-pull-secert")
+	genevaPullSecret, err := ioutil.ReadFile("secrets/acr-docker-pull-secret")
```
ehashman: Why did this not break anything?

jim-minter: file was misnamed in the secret itself - have duplicated the file to a new name, once this merges and goes back all the branches eventually we'll get rid of the old named file

Issue #1515, opened 2019-04-15
checkCanDeployRedhatIoImages e2e test fails


asalkeld:
https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1514/pull-ci-openshift-openshift-azure-master-upgrade-v4.2/148/

/kind flake-test

jim-minter:
thanks @asalkeld, please can you triage to see what went wrong?

jim-minter:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1511/pull-ci-openshift-openshift-azure-master-upgrade-v3.2/593

asalkeld:
I'll close, can always reopen if we get more.
/close

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1997/pull-ci-azure-release-v10-e2e-scaleupdown-v9.0/1/build-log.txt #1997

asalkeld:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2055/pull-ci-azure-release-v11-e2e-scaleupdown-v9.0/1

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2055/pull-ci-azure-release-v11-e2e-scaleupdown-v7.1/1

PR #2057, opened 2019-11-06
Add ImagePullSecret back into older worker cs

asalkeld:
This was removed in v10, but older clusters still need it. This part of the code
is heavily depended on by versioned code, but has not been versioned. This has
lead to a bug when scaling older clusters. https://github.com/openshift/openshift-azure/issues/1515

```release-note
NONE
```
/cc @jim-minter


jim-minter:
thanks @asalkeld 
/lgtm

PR #2060, opened 2019-11-07
Reduce journal rotation and restart fluentd once a day

jim-minter:
This aims to resolve the issue of fluentd exhausting its open file descriptors
on boxes where logs rotate a lot.

Fluentd's official recommendation is to increase the file descriptor limit
(https://docs.fluentd.org/installation/before-install#increase-max-of-file-descriptors)
but a tactical restart seems saner.

```release-note
NONE
```


jim-minter:
@grantomation @pnasrat this doesn't please me but ptal

PR #2063, opened 2019-11-09
fix and extend e-mail matching in customer-admin-controller

jim-minter:
```release-note
Resolve customer-admin group memberships issue when case of AAD mail domain differs from case of AAD login
```

### DIFF
--- a/pkg/controllers/customeradmin/group_mapping_test.go
+++ b/pkg/controllers/customeradmin/group_mapping_test.go
@@ -309,6 +309,38 @@ func TestFromMSGraphGroup(t *testing.T) {
 				Users: []string{"user@home.com"},
 			},
 		},
+		{
+			name:          "domain case can match",
+			kubeGroupName: osaCustomerAdmins,
+			want1:         true,
+			msGroupMembers: []graphrbac.User{
+				{
+					Mail: to.StringPtr("user@Home.com"),
+				},
+			},
+			want: &v1.Group{
+				ObjectMeta: meta_v1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


jim-minter:
/hold
to see if this fixes customer issue

@mjudeikis ptal
@thekad @Makdaam fyi

Makdaam:
I missed the AADUser.Mail return. :(
lgtm

pkg/controllers/customeradmin/group_mapping_test.go
```diff
@@ -309,6 +309,38 @@ func TestFromMSGraphGroup(t *testing.T) {
...
+			},
+			want: &v1.Group{
+				ObjectMeta: meta_v1.ObjectMeta{
+					Name: osaCustomerAdmins,
+				},
+				Users: []string{"user@home.com"},
+			},
+		},
+		{
+			name:          "user case can't match",
```
ehashman: I don't really understand this test case, how is it different than e.g. the one on L234?

mjudeikis:
from the context, I assume this is @Makdaam prod incident 
/lgtm

pkg/controllers/customeradmin/group_mapping_test.go
jim-minter: previously in some cases we were incorrectly doing case insensitive match on the whole e-mail address.  In fact the domain part is case insensitive and the username part is case sensitive.  The first additional test case checks that we correctly coalesce Home.com -> home.com; the second checks that we do not incorrectly coalesce User@ to user@.

PR #2064, opened 2019-11-11
[WIP] adding restart vm geneva action

gburges:
```release-note
NONE
```


ehashman:
I think you need to update your `go-bindata` @gburges, a bunch of stuff is getting updated here that shouldn't be.

gburges:
opening a new one with the removed bindata pin

PR #2071, opened 2019-11-13
Updating image on v10.1 to 3.11.153.

kwoodson:
```release-note
Bumping image to 311.153.20191113
```


jim-minter:
@kwoodson going to need to create a 10.2 here for this new image, so this PR is going to be more involved - please take another look, can chat about it in the morning if needed.

PR #2072, opened 2019-11-13
Bumping image version to 311.153.20191113

kwoodson:
```release-note:
Bumping image version to 311.153.20191113
```


jim-minter:
@kwoodson this is going to need to be a 12.1 on the release-v12 branch, then you'll need a 3rd PR for the master branch which fixes v13, and also adds v12.1 and v10.2 on master.

PR #1687, opened 2019-05-29
Add race detection to e2e tests

charlesakalugwu:
This adds race detection to our unit and e2e tests

/cc @jim-minter @grantomation @Makdaam 

```release-note
NONE
```

PR #2081, opened 2019-11-14
revert f66053121e8fd74acb6b40c0837471c3de5083f3 - it is a case study of premature optimisation

jim-minter:
```release-note
NONE
```

cf https://github.com/openshift/openshift-azure/pull/1687


asalkeld:
happy with change, just note it was done because unit tests were starting to flake with timeouts - they are slow

PR #2085, opened 2019-11-18
Fixes vendor

m1kola:
* `github.com/apparentlymart/go-cidr` was not reflected in `glide.lock`
* `golang.org/x/tools` needed to be pined because of missing packages

```release-note
NONE
```

/test vmimage


asalkeld:
I am OK with this, have you got someone from msft to test it?

m1kola:
We are going to upgrade Azure SDK which they are going to test. The PR for upgrade will include this PR, so it probably makes sense to put this on hold and give MSFT to test everything once the upgrade PR is ready.

/hold

jim-minter:
@m1kola actually don't think any changes are needed to glide.yaml now that Leszek's work merged?  But would accept a PR with the results of having run `glide up -v`.

m1kola:
Right, Leszek fixed both issues. Closing this PR.

PR #2100, opened 2019-11-25
update twistlock regex

jim-minter:
```release-note
NONE
```

(also using this PR as a test in case there is a lurking flake in the new privileged container e2e)


jim-minter:
ha, no flake.

PR #2101, opened 2019-11-25
Upgrade Azure Go SDK

m1kola:
~~Includes #2102 which needs to be merged first.~~
~~Includes #2113 which needs to be merged first.~~

```release-note
NONE
```


m1kola:
Will have update the PR to include v14.

m1kola:
Investigating upgrade job failures. My wild guess for now is that `github.com/Azure/go-autorest` has introduced some default deadline since `v11.2.7`, but changelog only has [this](https://github.com/Azure/go-autorest/blob/master/CHANGELOG.md#v1132):

> In `Future.WaitForCompletionRef()` if the provided context has a deadline don't add the default deadline.

m1kola:
I think I figured out the upgrade issue. I was initially blaming changes in Azure SDK, but it's not relevant it seems.

I made changes to ARM templates that affect all versions. As a result I had to change `TestHashScaleSetStability`. The problem is that: we [spin up a cluster using code from an older tag](https://github.com/openshift/openshift-azure/blob/8c6e8392680369e6be8abf3b67ac5d86e48f0f77/hack/tests/e2e-upgrade.sh#L31). And then run  [a E2E test](https://github.com/openshift/openshift-azure/blob/8c6e8392680369e6be8abf3b67ac5d86e48f0f77/hack/tests/e2e-upgrade.sh#L39) using the new code (where hash changed). Plugin compares hashes from a blob (generaded by old code) and hash generated by the current version of the plugin and they do not match. It triggers rotation and the test times out.

m1kola:
Rebased on top of the master branch and updated commit dates

hack/proxy-infrastructure/resources.go
```diff
@@ -40,10 +52,14 @@ func vnet(conf *Config) *network.VirtualNetwork {
 		Type:     to.StringPtr("Microsoft.Network/virtualNetworks"),
 		Location: to.StringPtr(conf.region),
 	}
+	return &arm.Resource{
```
jim-minter: globally, please can you change the pattern from
```
foo := thing{...}
return &arm.Resource{Resource: thing}
```
to
```
return &arm.Resource{Resource: &thing{...}}
```

notes:
* `git diff -w` will make this acceptable
* please standardise on .Resource being a pointer to a thing, not the thing itself

hack/proxy-infrastructure/setup.go
```diff
@@ -228,23 +228,19 @@ func generate(ctx context.Context, conf *Config) (map[string]interface{}, error)
...
 	}
 
-	resources := []interface{}{
-		vnet(conf),
-		ip(conf),
-		nsg(conf),
-		nic(conf),
-		vm(conf),
-		cs,
+	resources := []arm.Resource{
```
jim-minter: `[]*arm.Resource` ? (occurs multiple times in PR).  Let's avoid the shallow copying implied by all this dereferencing.

pkg/arm/v10/resources.go
```diff
@@ -385,24 +419,42 @@ func (g *simpleGenerator) nsgMaster() *network.SecurityGroup {
...
 func (g *simpleGenerator) Vmss(app *api.AgentPoolProfile, backupBlob, suffix string) (*compute.VirtualMachineScaleSet, error) {
-	return vmss(g.cs, app, backupBlob, suffix, g.testConfig)
+	r, err := vmss(g.cs, app, backupBlob, suffix, g.testConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	rr, ok := r.Resource.(compute.VirtualMachineScaleSet)
+	if !ok {
+		return nil, errors.New("can not convert value into VirtualMachineScaleSet")
```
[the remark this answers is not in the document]
jim-minter: makes sense, thanks

pkg/arm/v10/resources.go
```diff
@@ -628,5 +680,24 @@ func vmss(cs *api.OpenShiftManagedCluster, app *api.AgentPoolProfile, backupBlob
 		}
 	}
 
-	return vmss, nil
+	resource := &arm.Resource{
+		Resource:   vmss,
+		APIVersion: apiVersions["compute"],
+	}
+
+	if app.Role == api.AgentPoolProfileRoleMaster {
```
jim-minter: I'm doubting whether it makes sense to remove FixupDepends (at the very least, as part of this PR).  Removing it gives more opportunity for manual error.

pkg/cluster/hash_test.go
```diff
@@ -36,153 +36,153 @@ func TestHashScaleSetStability(t *testing.T) {
 			{
 				role: api.AgentPoolProfileRoleMaster,
 				// this value should not change
-				expectedHash: "51df71b5d8b4586dbdb5736f54f65d201b31ee5d7facd0378a4c59507aaa2e61",
+				expectedHash: "5bbd5e596e21a53eb6593ff64dcb034035dd0263dafe5db28283226b270faf83",
```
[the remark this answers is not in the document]
jim-minter: don't want to merge until this is better understood

pkg/util/arm/arm.go
```diff
@@ -1,128 +1,223 @@
...
+	Variables      map[string]interface{} `json:"variables,omitempty"`
+	Parameters     map[string]Parameter   `json:"parameters,omitempty"`
+	Functions      []interface{}          `json:"functions,omitempty"`
+	Resources      []Resource             `json:"resources,omitempty"`
+	Outputs        map[string]interface{} `json:"outputs,omitempty"`
 }
 
-// FixupAPIVersions inserts an apiVersion field into the ARM template for each
-// resource (the field is missing from the internal Azure type).
-func FixupAPIVersions(template map[string]interface{}, versions map[string]string) error {
```
jim-minter: I am unconvinced that we should remove FixupAPIVersions or FixupDepends, at the very least in this PR - let's discuss.

pkg/util/arm/arm_test.go
```diff
@@ -2,139 +2,182 @@ package arm
...
-
+func TestResource(t *testing.T) {
 	tests := []struct {
-		name      string
-		resources []interface{}
-		expect    []string
-		ignore    map[string]struct{}
+		name     string
+		input    *Resource
+		expected string
```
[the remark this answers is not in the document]
jim-minter: Yes, the same comments as there apply here.

pkg/util/azureclient/openshiftmanagedcluster/2018-09-30-preview/client.go
```diff
@@ -73,7 +73,7 @@ type OpenShiftManagedClustersUpdateTagsFuture struct {
 // If the operation has not completed it will return an error.
 func (future *OpenShiftManagedClustersUpdateTagsFuture) Result(client OpenShiftManagedClustersClient) (osmc v20180930preview.OpenShiftManagedCluster, err error) {
 	var done bool
-	done, err = future.Done(client)
+	done, err = future.DoneWithContext(context.Background(), client)
```
jim-minter: why these changes?  What do they add?

m1kola: @jim-minter `github.com/Azure/go-autorest` [deprecated `Done` in v11.2.0](https://github.com/Azure/go-autorest/blob/3492b2aff5036c67228ab3c7dba3577c871db200/CHANGELOG.md#v1120) and [removed in v12.0.0](https://github.com/Azure/go-autorest/blob/3492b2aff5036c67228ab3c7dba3577c871db200/CHANGELOG.md#v1200):

```go
future.DoneWithContext(context.Background(), client)
```
is an equvivalent of

```go
future.Done(client)
```

from the older SDK versions

m1kola: See implementation: https://github.com/Azure/go-autorest/blob/v11.2.0/autorest/azure/async.go#L89-L93

jim-minter: Oh I see, it's that upstream is marking it as deprecated.  Whatever ;-)

jim-minter: In fact, no, these changes don't make sense.  Pulling a context out of thin air to avoid calling a deprecated function is not better than calling a deprecated function.  We're not going to put the effort into passing contexts all around here; leave the calls to Done() as they were.

m1kola: @jim-minter `Done()` no longer exists in `go-autorest`

jim-minter: got it, sorry.  Then please use context.TODO() here.

hack/proxy-infrastructure/resources.go
```diff
@@ -5,16 +5,26 @@ import (
 	"compress/gzip"
 	"os"
 
-	"github.com/Azure/azure-sdk-for-go/profiles/latest/network/mgmt/network"
```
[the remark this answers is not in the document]
m1kola: Agree, but let's do it separately, becasue this PR is already too big and it's not blocking. I created [this Jira task](https://jira.coreos.com/browse/AZURE-497) for it.

pkg/arm/v10/resources.go
m1kola: I hate `FixupDepends` and `FixupApiVersions`, but I agree: it makes maintanance of our ARM templates harder.

hack/proxy-infrastructure/setup.go
m1kola: Makes sense. I think, it will require to change `arm.Template` definition, but it should not be a problem.

hack/proxy-infrastructure/resources.go
m1kola: Updated.

We can't do `return &arm.Resource{Resource: &thing{...}}` in 100% of cases, because there are some pieces of code where we set some values conditionally (see `vmss` functions in ARM, for example).

I mean, we can do type assertions and dereferencing (something like below), but I don't see the benifit:

```go
r := &arm.Resource{Resource: &thing{...}}

if r.Resource is a right type {
    // set values conditionally here 
}

return r
```

If you have any ideas about these cases - let me know.

pkg/arm/v10/resources.go
m1kola: I reverted `FixupDepends` and `FixupApiVersions` changes

m1kola:
@jim-minter I updated the PR. I expect the `TestHashScaleSetStability` to fail, because it now contains the old hashes (I reverted to `master`) and I didn't have a chance to fix it today. I'll look into it tomorrow.

pkg/arm/v10/resources.go
```diff
@@ -22,12 +24,12 @@ var (
...
 		"Microsoft.Compute": "2018-10-01",
-		"Microsoft.Network": "2018-07-01",
+		"Microsoft.Network": "2019-04-01",
 		"Microsoft.Storage": "2018-02-01",
 	}
 )
 
-func (g *simpleGenerator) vnet() *network.VirtualNetwork {
+func (g *simpleGenerator) vnet() *arm.Resource {
 	vn := &network.VirtualNetwork{
```
m1kola: Missed this one: need to do `return &arm.Resource{Resource: &...}` here too

m1kola: Ah, no. It's one with condition.

m1kola:
Retesting to see if it's a flake

m1kola:
/hold

I think, it's ready for review. Putting it on hold, because MSFT needs to test it as well.

pkg/arm/v10/json_hack.go
```diff
@@ -0,0 +1,32 @@
...
+	*arm.Resource
+}
+
+func (r *fixupArmResource) MarshalJSON() ([]byte, error) {
+	b, err := json.Marshal(r.Resource)
+	if err != nil {
+		return nil, err
+	}
+
+	var fixupMap map[string]*json.RawMessage
```
jim-minter: nit - json.RawMessage is actually a `[]byte` under the covers so the extra pointer is extraneous

pkg/util/arm/arm.go
```diff
@@ -1,22 +1,264 @@
...
-	ContentVersion string        `json:"contentVersion,omitempty"`
-	Parameters     struct{}      `json:"parameters,omitempty"`
-	Variables      struct{}      `json:"variables,omitempty"`
-	Resources      []interface{} `json:"resources,omitempty"`
-	Outputs        struct{}      `json:"outputs,omitempty"`
+	Schema         string                 `json:"$schema,omitempty"`
+	APIProfile     string                 `json:"apiProfile,omitempty"`
+	ContentVersion string                 `json:"contentVersion,omitempty"`
+	Variables      map[string]interface{} `json:"variables,omitempty"`
+	Parameters     map[string]Parameter   `json:"parameters,omitempty"`
```
jim-minter: note: in v4 I'm changing this to `map[string]*Parameter`

pkg/util/arm/arm.go
```diff
@@ -1,22 +1,264 @@
...
+// named types; to get around this we go weakly typed
+func shadowCopy(i interface{}) interface{} {
+	return _shadowCopy(reflect.ValueOf(i)).Interface()
+}
+
+func _shadowCopy(v reflect.Value) reflect.Value {
+	switch v.Kind() {
+	case reflect.Array:
+		a := reflect.New(reflect.ArrayOf(v.Len(), emptyInterfaceType)).Elem()
+		for i := 0; i < v.Len(); i++ {
```
jim-minter: v4 needed a fix for the uuid.UUID - that would probably be good to backport as a separate PR to keep in sync

jim-minter:
waiting for MSFT to confirm before releasing hold.
/lgtm

pkg/arm/v10/json_hack.go
m1kola: Yeah, right! Extra pointer doesn't hurt in this case, but I'll update it in the PR with UUID & `map[string]*Parameter` changes.

pkg/arm/v10/arm.go
```diff
@@ -85,12 +85,6 @@ func (g *simpleGenerator) Generate(ctx context.Context, backupBlob string, isUpd
 
 	arm.FixupDepends(g.cs.Properties.AzProfile.SubscriptionID, g.cs.Properties.AzProfile.ResourceGroup, azuretemplate, nil)
 
-	// HACK: Current SDK version is v24. Private link support comes into the version v28+.
```
mjudeikis: :clap:

m1kola:
@mjudeikis @jim-minter https://github.com/openshift/openshift-azure/pull/2123 got merged and I had to rebase the PR. Could you please re-approve this but still keep it on hold?

mjudeikis:
@julienstroheker @amanohar can you please check if you happy with this going in?

kwoodson:
@m1kola What is the status of this?

m1kola:
@kwoodson checking with MSFT.

kwoodson:
@m1kola Thanks.  It appears the priority on this has decreased as work on v4 has increased.  Awaiting a response from @amanohar

ehashman:
Can this be closed for now @m1kola ?

m1kola:
Yes, I think we can close it.

PR #2112, opened 2019-11-28
don't allow upper-case domain names

jim-minter:
```release-note
MSFT: update plugin validator to reject upper-case domain names for public hostname and router.
```


ehashman:
Does this fix #1571?

Issue #2119, opened 2019-12-05
Move log-analytics deployment from infra to compute


asalkeld:
https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v14/data/Deployment.apps/openshift-azure-logging/log-analytics-cluster-agent.yaml#L90

/assign @olga-mir 

This is so that the node agent can monitor the cluster-agent

olga-mir:
I've passed on a cluster for validation, once confirmed that it achieves the goal I'll put the PR up.

olga-mir:
PR has been merged
/close

PR #2122, opened 2019-12-05
Move Log Analytics cluster agent to compute node

olga-mir:
```release-note
Move Log Analytics cluster agent to compute node
```
This will enable monitoring of the cluster level agent by the node agents daemonset pods (they are not scheduled on master/infra nodes)
https://github.com/openshift/openshift-azure/issues/2119


olga-mir:
test timeouts, happening in other places as well: https://github.com/openshift/openshift-azure/pull/2116#issuecomment-561893250

`Timeout: request did not complete within allowed duration`

Issue #2126, opened 2019-12-06
"Timeout: request did not complete within allowed duration" flake


ehashman:
```
 Openshift on Azure customer-admin e2e tests [CustomerAdmin][EveryPR] 
  should not be able to create a non-whitelisted privileged container
  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/customer_admin.go:344
• Failure [44.468 seconds]
Openshift on Azure customer-admin e2e tests [CustomerAdmin][EveryPR]
/home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/customer_admin.go:27
  should not be able to create a non-whitelisted privileged container [It]
  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/customer_admin.go:344
  Unexpected error:
      <*errors.StatusError | 0xc0003b0b40>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
              Status: "Failure",
              Message: "Timeout: request did not complete within allowed duration",
              Reason: "Timeout",
              Details: {Name: "", Group: "", Kind: "", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 504,
          },
      }
      Timeout: request did not complete within allowed duration
  occurred 
```

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2123/pull-ci-azure-master-e2e-create-20191027-private/179/build-log.txt #2123 

/kind test-flake

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2122/pull-ci-azure-master-e2e-create-20191027-private/177/build-log.txt #2122

ehashman:
#2121:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2121/pull-ci-azure-master-e2e-create-20191027-private/184/build-log.txt
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2121/pull-ci-azure-master-e2e-create-20191027-private/183/build-log.txt
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2121/pull-ci-azure-master-e2e-create-20191027-private/181/build-log.txt

[...]

PR #2129, opened 2019-12-10
Enable enduser to manage Kafka clusters

olga-mir:
```release-note
Enable enduser to manage Kafka clusters
```
Regular users should be able to manage Kafka clusters, not only customer admins. 

/assign asalkeld


olga-mir:
/retest
flakes: 
https://github.com/openshift/openshift-azure/issues/2126 
https://github.com/openshift/openshift-azure/issues/1778

mjudeikis:
/retest
looks like timeout again

jim-minter:
/hold

I don't think this is right - isn't this giving every user admin on *all* cluster strimzi CRDs?  I don't actually think anyone should have that.

I think we want users who have the view clusterrole bound to their namespace able to view the strimzi CRDs in that namespace only; equivalents for edit and admin.

I think I'm expecting to see 3 new cluster roles, each making use of `rbac.authorization.k8s.io/aggregate-to-{view,edit,admin}: "true"` labels.

Customer admin should automatically get access to all the end users strimzi CRDs by virtue of being given admin on all end user projects, not via an additional RBAC rule (so I think I expect to see the existing customer-admin-strimzi ClusterRoleBinding removed as well).

mjudeikis:
ping @olga-mir

olga-mir:
@mjudeikis this is not forgotten, just getting back to this now.

olga-mir:
thank you @jim-minter for suggesting this approach. 

I think that the initial confusion that led to this task is when a customer (maybe even a user from customer-admin) tried to create a cluster via Web Console. This fails because template controller SA in openshift-infra namespace lacks permissions to manipulate kafka resources. 

Screenshot from 2020-01-10 14-23-56

With the approach that you suggested this issue resolves automatically, without manually modifying the template controller (that's the point of the aggregation, neat):

Screenshot from 2020-01-10 14-34-23

olga-mir:
/retest 
flakes https://github.com/openshift/openshift-azure/issues/2126

olga-mir:
the tests fail on sync pod `should not continuously update objects `
which is not a flake in this PR. 

When I run `make sync-run` (while the cluster's sync pod deployment is scaled to 0) I do get updates on every run, for a few aggregated roles (admin, view, edit, cluster-reader):

```
NFO[2020-01-13T11:34:26+11:00] Update ClusterRole.rbac.authorization.k8s.io/admin                                                                                                                                                                                                                                                                         
INFO[2020-01-13T11:34:26+11:00]   unstructured.Unstructured{                                                                                                                                                                                                                                                                                               
                ... // 2 identical entries                                                                                                                                                                                                                                                                                                                 
                "kind":     string("ClusterRole"),                                                                                                                                                                                                                                                                                                         
                "metadata": map[string]interface{}{"annotations": map[string]interface{}{"openshift.io/description": string("A user that has edit rights within the project and can change the project's membership."), "rbac.authorization.kubernetes.io/autoupdate": string("false")}, "labels": map[string]interface{}{"azure.openshift.io/owned-by-sync
-pod": string("true"), "kubernetes.io/bootstrapping": string("rbac-defaults")}, "name": string("admin")},                                                                                                                                                                                                                                                  
                "rules": []interface{}{                                                                                                                                                                                                                                                                                                                    
                        map[string]interface{}{"apiGroups": []interface{}{string("automationbroker.io")}, "resources": []interface{}{string("access")}, "verbs": []interface{}{string("create")}},                                                                                                                                                         
-                       map[string]interface{}{                                                                                                                                                                                                                                                                                                            
-                               "apiGroups": []interface{}{string("kafka.strimzi.io")},                                                                                                                                                                                                                                                                    
-                               "resources": []interface{}{                                                                                                                                                                                                                                                                                                
-                                       string("kafkas"),                                                                                                                                                                                                                                                                                                  
-                                       string("kafkaconnects"),                                                                                                                                                                                                                                                                                           
-                                       string("kafkaconnects2is"),                                                                                                                                                                                                                                                                                        
-                                       string("kafkamirrormakers"),                                                                                                                                                                                                                                                                                       
-                                       string("kafkausers"),                                                                                                                                                                                                                                                                                              
-                                       string("kafkatopics"),                                                                                                                                                                                                                                                                                             
-                                       string("kafkabridges"),                                                                                                                                                                                                                                                                                            
-                               },                                                                                                                                                                                                                                                                                                                         
-                               "verbs": []interface{}{                                                                                                                                                                                                                                                                                                    
-                                       string("get"),                                                                                                                                                                                                                                                                                                     
-                                       string("list"),                                                                                                                                                                                                                                                                                                    
-                                       string("watch"),                                                                                                                                                                                                                                                                                                   
-                                       string("create"),                                                                                                                                                                                                                                                                                                  
-                                       string("delete"),                                                                                                                                                                                                                                                                                                  
-                                       string("patch"),                                                                                                                                                                                                                                                                                                   
-                                       string("update"),                                                                                                                                                                                                                                                                                                  
-                               },                                                                                                                                                                                                                                                                                                                         
-                       },   
```

Currently there are only few very fundamental roles that make use of aggregation, e.g `aggregate-to-view` label is present only on these roles:
```
ClusterRole.rbac.authorization.k8s.io/cluster-reader.yaml
ClusterRole.rbac.authorization.k8s.io/system-aggregate-to-view.yaml
ClusterRole.rbac.authorization.k8s.io/system-openshift-aggregate-to-view.yaml
ClusterRole.rbac.authorization.k8s.io/system-service-catalog-aggregate-to-view.yaml
ClusterRole.rbac.authorization.k8s.io/view.yaml
```

Which make me think that aggregation is not used often. Maybe sync pod implementation doesn't take this flow into account? Another option could be that I am missing some labels or I have some redundant labels (something like autoupdate, reconcile protect or similar)

olga-mir:
the way it is done currently for other aggregated roles, is by copying the rules to the parent view, edit and admin roles: 
https://github.com/openshift/openshift-azure/blob/ef64521c1b5684ffc31edc1d56765432e23ac4d1/pkg/sync/v10/data/ClusterRole.rbac.authorization.k8s.io/view.yaml#L248-L258
is from https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v10/data/ClusterRole.rbac.authorization.k8s.io/system-service-catalog-aggregate-to-view.yaml
Added the commit that follows this pattern: https://github.com/openshift/openshift-azure/pull/2129/commits/9f4876d8694b43203df617c095760d490291634e
It makes the aggregated roles redundant, is there any value in keeping them?

olga-mir:
/retest

seems some problem in CI

olga-mir:
@mjudeikis Can you please take a look at this PR?

(current failing failing is a flake, but there could be not much point in battling the CI if there are changes I need to make to the PR)

pkg/sync/v15/data/ClusterRole.rbac.authorization.k8s.io/edit.yaml
```diff
@@ -19,6 +19,24 @@ rules:
   - access
   verbs:
   - create
+- apiGroups:
```
mjudeikis: Do we need these role definitions here if we use aggregations? I think this duplicates the effort

pkg/sync/v15/data/ClusterRole.rbac.authorization.k8s.io/admin.yaml
```diff
@@ -19,6 +19,24 @@ rules:
   - access
   verbs:
   - create
+- apiGroups:
```
mjudeikis: same. Do we need this if we use aggregates?

mjudeikis:
might be flake, but looks particularly interesting sync failure. lets try again
/retest

pkg/sync/v15/data/ClusterRole.rbac.authorization.k8s.io/edit.yaml
olga-mir: this is exactly my question. 
without this sync pod is continuously updating, because there are ALWAYS changes from sync pod perspective.  
https://github.com/openshift/openshift-azure/pull/2129#issuecomment-573478271
https://github.com/openshift/openshift-azure/pull/2129#issuecomment-573901132

pkg/sync/v15/data/ClusterRole.rbac.authorization.k8s.io/admin.yaml
olga-mir: same :) 
as above
this is exactly my question. 
without this sync pod is continuously updating, because there are ALWAYS changes from sync pod perspective.  
https://github.com/openshift/openshift-azure/pull/2129#issuecomment-573478271
https://github.com/openshift/openshift-azure/pull/2129#issuecomment-573901132
@mjudeikis

mjudeikis:
Got you! Didnt understood the issue before.
/lgtm

olga-mir:
/hold

v7 image has been deleted, this test is going to fail

olga-mir:
/hold cancel 
 upgrade v71 test has been removed, ok to test again

olga-mir:
/hold

wait to fix `verify`: https://github.com/openshift/openshift-azure/pull/2189

jim-minter:
lgtm as well - in an ideal world the sync pod would understand aggregate-to-*, but it doesn't, and I don't think we're going to implement it now.

Issue #1442, opened 2019-04-05
ssh rejected - administratively prohibited


charlesakalugwu:
Saw a flake with the following entries in the logs

```
time="2019-04-05T13:34:41Z" level=warning msg="Dial failed: ssh: rejected: administratively prohibited (open failed)" func=github.com/openshift/openshift-azure/pkg/fakerp.debugDeployerError file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/fakerp.go:69"
time="2019-04-05T13:34:42Z" level=warning msg="Dial failed: ssh: rejected: administratively prohibited (open failed)" func=github.com/openshift/openshift-azure/pkg/fakerp.debugDeployerError file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/fakerp.go:69"
time="2019-04-05T13:34:42Z" level=warning msg="Dial failed: ssh: rejected: administratively prohibited (open failed)" func=github.com/openshift/openshift-azure/pkg/fakerp.debugDeployerError file="/go/src/github.com/openshift/openshift-azure/pkg/fakerp/fakerp.go:69"
```
artifacts: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1430/pull-ci-openshift-openshift-azure-master-e2e-azure/3407/build-log.txt

/kind test-flake

charlesakalugwu:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/1520/pull-ci-openshift-openshift-azure-master-upgrade-v4.2/162/build-log.txt

jim-minter:
The fixes were made to master but not backported, thus this error can still occur on v3/v4 upgrade flows.  Will consider the priority of fixing this.

ehashman:
/reopen

```
 time="2019-10-10T12:10:58Z" level=warning msg="Dial failed: ssh: rejected: administratively prohibited (open failed)" func="test/util/diagnostics.(*scalesetDebugger).GatherHostLogs()" file="test/util/diagnostics/scaleset.go:51"
time="2019-10-10T12:10:58Z" level=debug msg="400 Bad Request: Failed to apply request: UpdateWorkerAgentPoolUpdateScaleSet: Future#WaitForCompletion: context has been cancelled: StatusCode=200 -- Original Error: context deadline exceeded" func="pkg/fakerp.(*Server).badRequest()" file="pkg/fakerp/util.go:24"
[...]

PR #1440, opened 2019-04-05
re-add nil check which I should not have removed

jim-minter:
```release-note
NONE
```

fixes #1435 

@charlesakalugwu

PR #2133, opened 2019-12-11
Rename Log Analytics to omsagent

olga-mir:
```release-note
Rename Log Analytics to omsagent
Upgrade omsagent to 12042019
```


pkg/sync/v13/data/DaemonSet.apps/openshift-azure-logging/log-analytics-node-agent.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     metadata:
       annotations:
         agentVersion: 1.10.0.1
-        dockerProviderVersion: 7.0.0-6
+        dockerProviderVersion: 8.0.0-1
```
mjudeikis: Why do we change old release this? This will invalidate old hash and trigger rotation?

pkg/sync/v14/data/ConfigMap/openshift-azure-logging/omsagent-rs-config.yaml
```diff
@@ -0,0 +1,207 @@
...
+     </source>
+     <filter mdm.kubepodinventory** mdm.kubenodeinventory**>
+      type filter_inventory2mdm
+      custom_metrics_azure_regions eastus,southcentralus,westcentralus,westus2,southeastasia,northeurope,westeurope,southafricanorth,centralus,northcentralus,eastus2,koreacentral,eastasia,centralindia,uksouth,canadacentral,francecentral,japaneast,australiaeast
+      log_level info
+     </filter>
+     #custom_metrics_mdm filter plugin for perf data from windows nodes
+     <filter mdm.cadvisorperf**>
+      type filter_cadvisor2mdm
+      custom_metrics_azure_regions eastus,southcentralus,westcentralus,westus2,southeastasia,northeurope,westeurope,southafricanorth,centralus,northcentralus,eastus2,koreacentral,eastasia,centralindia,uksouth,canadacentral,francecentral,japaneast,australiaeast
```
mjudeikis: Dont have any background, but is this hardcoding needed?

pkg/sync/v13/data/DaemonSet.apps/openshift-azure-logging/log-analytics-node-agent.yaml
mjudeikis: I think this is causing you failing tests as hash changed and our mocked cluster upgrade test fails. They are not very verbose in the test output

olga-mir: thank you for catching this. v13 was indeed mistake. I will fix it

pkg/sync/v14/data/ConfigMap/openshift-azure-logging/omsagent-rs-config.yaml
olga-mir: yes I believe it is needed as is. This hardcoding is part of OMS config: a/pkg/sync/v13/data/Deployment.apps/openshift-azure-logging/log-analytics-cluster-agent.yaml
And we already have it in our code base: https://github.com/openshift/openshift-azure/blob/master/pkg/sync/v14/data/ConfigMap/openshift-azure-logging/log-analytics-cluster-agent-config.yaml#L67
it is longer in the new OMS agent version: https://github.com/microsoft/OMS-docker/compare/61729c6e3250910592d2e2b8391353f4c579d65a...6928f685f2c61f4ce3759a485c5ebfd4ddaf905d#diff-f4081688aa41659e0d49872cfbf54e1bR122

jim-minter:
All this lgtm, not sure why the unit test is failing though

olga-mir:
@jim-minter unit was failing because of the test whitelist: https://github.com/openshift/openshift-azure/pull/2133/commits/638797032e6c7995e860d7d5012bcbed97114b2c 
I wasn't aware of it, and mj has pointed this out. Unit test is passing now

pkg/sync/v14/data/Deployment.apps/openshift-azure-logging/omsagent-rs.yaml
```diff
@@ -54,11 +54,11 @@ spec:
           name: in-rs-tcp
         resources:
           limits:
-            cpu: 150m
-            memory: 600Mi
+            cpu: 1
+            memory: 740Mi
           requests:
-            cpu: 75m
```
asalkeld: just check when you run multiple sync's that it doesn't keep changing this. when I did this, i  remember there was..

olga-mir: I believe it is not an issue this time, because it is not updating the same deployment, but completely replacing it. At least after `make sync-run` this is what I can see:
```
[ openshift-azure (omsagent) ] k get deploy omsagent-rs -n openshift-azure-logging -o yaml | grep -A 6 resources 
        resources:
          limits:
            cpu: "1"
            memory: 740Mi
          requests:
            cpu: 150m
            memory: 250Mi
```

pkg/sync/v14/data/SecurityContextConstraints.security.openshift.io/privileged.yaml
```diff
@@ -41,6 +41,7 @@ users:
 - system:serviceaccount:openshift-infra:build-controller
 - system:serviceaccount:openshift-etcd:etcd-backup
 - system:serviceaccount:openshift-azure-logging:log-analytics-agent
+- system:serviceaccount:openshift-azure-logging:omsagent
```
olga-mir: @Makdaam can you please have a look at this? I can't find a way to replace `log-analytics` with `omsagent` (not even edit this as cluster-admin). Can we add `omsagent` for now and remove `log-analytics` later or is there a better way to do it?

Makdaam: If you'd like to replace it for testing on your dev cluster you can
`oc annotate scc privileged openshift.io/reconcile-protect=true`
as a cluster-admin.
Otherwise the sync pod will reset it to what you have compiled in the azure image pkg/sync/vX/data.

If this is about replacing the accounts on prod, then I'd do it like you described: add a new account, migrate, remove old account. Or if it can be synced with our release cycle just flip the switch at once.

olga-mir: @Makdaam The problem is not the sync-pod. (in my setup it is scaled down to 0 and I only run `make sync-run`) The problem that if I replace `log-analytics` with `omsagent` in SCC in pkg/sync/vX/data and run the sync-pod I get: 

```
INFO[2019-12-13T09:46:37+11:00] Update SecurityContextConstraints.security.openshift.io/privileged 
INFO[2019-12-13T09:46:37+11:00]   unstructured.Unstructured{
                ... // 20 identical entries
                "seccompProfiles":    []interface{}{string("*")},
                "supplementalGroups": map[string]interface{}{"type": string("RunAsAny")},
                "users": []interface{}{
                        string("system:admin"),
                        string("system:serviceaccount:openshift-infra:build-controller"),
                        string("system:serviceaccount:openshift-etcd:etcd-backup"),
-                       string("system:serviceaccount:openshift-azure-logging:log-analytics-agent"),
                        string("system:serviceaccount:openshift-azure-logging:omsagent"),
-                       string("system:serviceaccount:openshift-azure-logging:newsagent"),
                        string("system:serviceaccount:kube-system:sync"),
                },
                "volumes": []interface{}{string("*")},
        },
  } 
INFO[2019-12-13T09:46:37+11:00] sync error: admission webhook "sccs.aro-admission-controller.aro.openshift.io" denied the request: users: Required value: must include user system:serviceaccount:openshift-azure-logging:log-analytics-agent 
```

So if there is no issue with adding the "omsagent" first and then in future remove "log-analytics" I think it would be the safest approach.

pkg/sync/v14/data/Deployment.apps/openshift-azure-logging/omsagent-rs.yaml
olga-mir: @asalkeld I found an issue with limit 1 - it does seem to try to update it every run due to types: 
```
                                                                "limits": map[string]interface{}{                                                                                                                                                                                                                                                          
-                                                                       "cpu":    string("1"),                                                                                                                                                                                                                                                             
+                                                                       "cpu":    int64(1),                                                                                                                                                                                                                                                                
                                                                        "memory": string("740Mi"),                                                                                                                                                                                                                                                         
                                                                }, 
```
I have changed that limit to 950 instead there. 
But as for this line that you've commented - I don't see any problem. On the second run (without any changes), sync-pod did not try to perform any changes

Also noticed 740Mi, I'll fix this typo

olga-mir:
Hi  @ganga1980, I've updated the names as you requested: https://github.com/openshift/openshift-azure/pull/2133/commits/27f342ffb0e058e18d3c5ec941be11775da97c23

olga-mir:
/retest

flakes: 

pull-ci-azure-master-upgrade-v10.1.1: https://github.com/openshift/openshift-azure/issues/1442 https://github.com/openshift/openshift-azure/issues/1440
pull-ci-azure-master-e2e-create-20191027-private: https://github.com/openshift/openshift-azure/issues/2126
pull-ci-azure-master-upgrade-v7.1, pull-ci-azure-master-upgrade-v12.2:  https://github.com/openshift/openshift-azure/issues/2168 (new)

PR #2134, opened 2019-12-11
Allowing annotation openshift.io/reconcile-protect=true instead of a label

Makdaam:
```release-note
Fixes bug which prevented adding reconcile-protect annotation to SCCs
```


jim-minter:
@Makdaam good catch, thanks.
Please work with @ehashman to make sure this gets backported and sent to MSFT, they are still in the process of vendoring it

/lgtm

Makdaam:
azure keyvault creation timed out flake this time

PR #2139, opened 2019-12-12
[release-v13] Prep for release 13.1

ehashman:
```release-note
NONE
```

### DIFF
--- a/pkg/cluster/hash_test.go
+++ b/pkg/cluster/hash_test.go
@@ -185,6 +185,23 @@ func TestHashScaleSetStability(t *testing.T) {
 				expectedHash: "959b10157d2864150339914d0e5ddee5dc7d7e1e6d2318f0ef9e5325cc6b09df",
 			},
 		},
+		"v13.0": {
+			{
+				role: api.AgentPoolProfileRoleMaster,
+				// this value should not change
+				expectedHash: "752cc90913885e480132837ef71e915933a80fc9933767b7594867fb3848be9a",
+			},
+			{
+				role: api.AgentPoolProfileRoleInfra,
+				// this value should not change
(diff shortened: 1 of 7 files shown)


pkg/cluster/hash_test.go
```diff
@@ -185,6 +185,23 @@ func TestHashScaleSetStability(t *testing.T) {
 				expectedHash: "959b10157d2864150339914d0e5ddee5dc7d7e1e6d2318f0ef9e5325cc6b09df",
 			},
 		},
+		"v13.0": {
```
pnasrat: Confused by this should this not be adding a 13.1 here looking at the file there are hashes for eg 12.2. Also do you need to add the 13.1 case in TestHashScaleSetStability earlier in this file

pkg/cluster/hash_test.go
```diff
@@ -314,6 +331,10 @@ func TestHashSyncPodStability(t *testing.T) {
 			// this value should not change
 			expectedHash: "e6e1637db32ca54384e7b975f9ce652cdee06d3e494ec28a8b663e4f70e84af8",
 		},
+		"v13.0": {
```
pnasrat: Ditto

pkg/cluster/hash_test.go
ehashman: This test ensures that we don't accidentally modify the contents of 13.0 now that it's been released; it's okay for us to modify the contents of 13.1 until we officially release and tag it. (I lifted the hashes from master, where 13.0 was already frozen.)

ehashman: The unit tests require us to have hashes for every version release except for the current `pluginVersion` in the pluginconfig: 

https://github.com/openshift/openshift-azure/blob/ffed473e54f737db539a0bfa9e1a4939d542f45f/pkg/cluster/hash_test.go#L226-L230

PR #2153, opened 2019-12-17
Backend scripts for geneva actions

thekad:
Backend commands for new RunCommand geneva action options.

```release-note
NONE
```


pkg/vmimage/data/usr/local/bin/oc-get-all
```diff
@@ -0,0 +1,15 @@
...
+# check where we're running
+if [[ $HOST != master-* ]]
+then
+  echo "This geneva action can only be executed from one of the masters"
+  exit
+fi
+
+for target in group user deploy service pod route
+do
+  oc get --all-namespaces $target
```
kwoodson: Does it make sense to return these in json or yaml by adding the output flag e.g `-o json`

thekad: I thought about it, but since the output of the geneva action reencodes the output of the command in another json object it makes it kind of messy, wdyt?

pkg/vmimage/data/usr/local/bin/get-live-net-status
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
mjudeikis: Please use things from https://github.com/openshift/openshift-azure/blob/master/pkg/vmimage/data/usr/local/bin/get-node-logs#L19-L32 and clear that script

pkg/vmimage/data/usr/local/bin/oc-get-all
mjudeikis: What is the current output example of this?

mjudeikis: Any particual reason we dont use golang kubeclient for this?

thekad: for speed, we wanted to leverage the existing RunCommand action and not write up a whole new geneva action for this (but we will retrofit the get control plane pods into a more flexible geneva action)

pkg/vmimage/data/usr/local/bin/get-live-net-status
thekad: these are 2 distinct scripts for 2 different purposes, I think they are fine without sharing code between them

mjudeikis: get-node-logs are not being used and should be deleted and refactored.Networking information in `get-node-logs` is more granual

PR #2155, opened 2019-12-18
Fix RefreshCluster validation for non-admin and fix openshift resolv.conf

### LINKED ISSUE #2144: Potential DNS mismatch between agent nodes
After a DNS settings update on the vnet. The vms need to be restarted to take the update. A user can scale the cluster with out using refresh cluster. The new vm will have the new dns settings while the rest of the cluster will have the old settings. Creating potential for inconsistent behavior in apps depending on which node they land on. If the dns settings have changed. All scale operations should be blocked until a refresh cluster has gone through.

asalkeld:
```release-note
Correct the behaviour of RefreshCluster when updating as non-admin
Correct the content of OpenShift resolve.conf
```
/cc @ehashman 

fixes #2144


pkg/plugin/plugin.go
```diff
@@ -78,7 +78,7 @@ func (p *plugin) Validate(ctx context.Context, new, old *api.OpenShiftManagedClu
 		// don't call checkIfClusterWillRefresh() if we already have errors.
 		return
 	}
-	if !externalOnly && old != nil {
```
asalkeld: this was preventing the validation below as these 2 conditions never happen

pkg/startup/v14/data/master/etc/origin/node/resolv.conf
```diff
@@ -1,3 +1,3 @@
-{{ range .Properties.NetworkProfile.Nameservers }}
+{{ range .ContainerService.Properties.NetworkProfile.Nameservers }}
```
asalkeld: serious oops :-(

pkg/plugin/integration_test.go
```diff
@@ -699,6 +699,30 @@ func TestHowActionsCauseRotations(t *testing.T) {
 		expectNodes    map[rotationType]int
 		expectCalls    []string
 	}{
+		{
+			name:           "scale up, with dns change",
+			expectRotation: map[rotationType]bool{rotationMaster: true, rotationInfra: true, rotationSync: false, rotationCompute: true},
```
asalkeld: this test confirm that all nodes rotate

pkg/plugin/plugin.go
asalkeld: this needs some discussion as in the real rp clients don't exist in the first validate.

ehashman: Where did we land with this?

pkg/startup/v14/startup_test.go
```diff
@@ -55,6 +57,34 @@ func TestFilePermissions(t *testing.T) {
...
+		Properties: api.Properties{
+			NetworkProfile: api.NetworkProfile{
+				Nameservers: []string{"1.2.3.4", "5.6.7.8"},
+			},
+		},
+	}
+	tmpl := `
+{{ range .ContainerService.Properties.NetworkProfile.Nameservers }}
+nameserver {{ . }}
+{{ end }}`
```
ehashman: Is there any way to keep this in sync with the contents of `pkg/startup/v14/data/master/etc/origin/node/resolv.conf`? I'm worried that since this test is sort of unrelated they might drift. Can we read in the template from that file (or that of the current plugin) directly for the purposes of testing this?

pkg/plugin/plugin.go
asalkeld: i am changing it to (pending confirmation from Jack)
```
if !externalOnly && new.Properties.ProvisioningState != api.Creating {
```

pkg/startup/v14/startup_test.go
asalkeld: i'll just delete the test, it was just me confirming it was working

asalkeld:
/retest
checkEnforcesEmptyDirQuotas failed

PR #2156, opened 2019-12-18
Adding imagestream access to customer admin project

kwoodson:
```release-note
Adding imagestream access to the customer-admin-project cluster role
```


pkg/sync/v14/data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
```diff
@@ -41,3 +41,12 @@ rules:
   - delete
   - patch
   - update
+- apiGroups:
+  - ""
```
jim-minter: add the line ` - image.openshift.io` here

kwoodson: Will update

pkg/sync/v14/data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
```diff
@@ -41,3 +41,12 @@ rules:
   - delete
   - patch
   - update
+- apiGroups:
+  - image.openshift.io
```
ehashman: Quick question for you @kwoodson ... we currently have some imagestream permissions granted in the customer-admin-cluster YAML: https://github.com/openshift/openshift-azure/blob/b2aa52db592e401df36c12a10b1cb6c6c06beb4f/pkg/sync/v14/data/ClusterRole.rbac.authorization.k8s.io/customer-admin-cluster.yaml#L133-L142

Do you or @jim-minter know the difference between the two ClusterRoles and why we would put these permissions on one vs. the other? From the looks of running through git blame, those capabilities have existed on customer-admin-cluster since before the addition of the admissioncontroller.

ehashman: (From what I can see in https://github.com/openshift/openshift-azure/blob/b2aa52db592e401df36c12a10b1cb6c6c06beb4f/pkg/controllers/customeradmin/shared.go#L14-L48

both roles get bound to the `osa-cluster-admins` group so I'm not actually sure there's much difference, I'm just trying to grok the historic reasons why the roles are separate)

ehashman:
504 timeout on e2e

pkg/sync/v14/data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
kwoodson: @ehashman This is a good question.  The original intention was to divide the customer abilities into project level and cluster level access.  We did this by creating two separate cluster roles.  The cluster abilities were creating service brokers where as the project abilities included project management (crud at the project level).  

I think over time the customer-admin abilities have been augmented and now the distinction might have been lost.  I'm fine with placing the changes in either location, `customer-admin-cluster` or `customer-admin-project`.

Would you like for me to move these changes into `customer-admin-cluster` and consolidate them with the other options that were pre-existing?

ehashman: I think this is fine as is, just wanted to get the context.

pkg/sync/v14/data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
```diff
@@ -41,3 +41,12 @@ rules:
   - delete
   - patch
   - update
+- apiGroups:
```
ehashman: Don't we still need the apiGroup `- ""`?

kwoodson: I'm not sure what the difference is.  What does supplying an empty "" do vs adding the specific api group `image.openshift.io`.  I'll happily add both but unsure as to why.

ehashman: I'm not actually sure what the difference is, trawling through docs, but I noticed this didn't match the other file.

ehashman:
/retest
more 504 timeouts...

PR #2158, opened 2019-12-19
Fix Template() usage so bugs are not hidden

asalkeld:
```release-note
NONE
```
This is because the template function will return interface{} on a non-existing
value instead of an error. With struct we get the error correctly.
This has hidden an error in the startup assets.


ehashman:
Needs rebase after #2155 lands.

PR #2166, opened 2020-01-06
Adds privateApiServer field into the admin API

m1kola:
Addming the `privateApiServer` field into the admin API makes debugging easier for an SRE: it is more obvious that the cluster has a private api enabled.

```release-note
NONE
```

/cc @mjudeikis

### DIFF
--- a/pkg/api/admin/convertertointernal.go
+++ b/pkg/api/admin/convertertointernal.go
@@ -90,6 +90,7 @@ func mergePropertiesAdmin(oc *OpenShiftManagedCluster, cs *api.OpenShiftManagedC
 	if oc.Properties.FQDN != nil {
 		cs.Properties.FQDN = *oc.Properties.FQDN
 	}
+
 	// always set this at it's a per/request action, not a configuration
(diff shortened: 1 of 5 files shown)


pkg/api/admin/convertertointernal.go
```diff
@@ -90,6 +90,10 @@ func mergePropertiesAdmin(oc *OpenShiftManagedCluster, cs *api.OpenShiftManagedC
 	if oc.Properties.FQDN != nil {
 		cs.Properties.FQDN = *oc.Properties.FQDN
 	}
+	if oc.Properties.PrivateAPIServer != nil {
```
mjudeikis: I think we should drop this. This would allow admin to set cluster to the public, and potentially break the cluster.
This should be `read-only`

m1kola: @mjudeikis  Makes sense. I removed this from conversion to internal.

PTAL. Also note `TestRoundTrip`: had to add an exception.

PR #2173, opened 2020-01-09
Prepare v15 plugin for development

ehashman:
```release-note
NONE
```

I had to revert the changes that happened on master to the v14 directories since I cut the v14 branch yesterday morning. They've all been copied into v15. I started with the hashes from the v14 branch (see #2172) and then worked backwards to make sure everything matched and the unit tests passed.


ehashman:
Hrm, unit tests are passing locally and I don't see any files that would be causing a stink, I'm not sure what's wrong.

PR #2182, opened 2020-01-15
Increase supported compute cluster size to 100

pnasrat:
```release-note
Intended as a soft launch support larger cluster creation based on previous performance tests now that the timeout issues for cluster creation of this size have been mitigated

Feature launch requires more extensive test plan and documentation updates
```

### DIFF
--- a/pkg/api/validate/validators.go
+++ b/pkg/api/validate/validators.go
@@ -305,7 +305,7 @@ func validateAgentPoolProfile(app *api.AgentPoolProfile, np *api.NetworkProfile)
 		if !rxAgentPoolProfileName.MatchString(app.Name) {
 			errs = append(errs, fmt.Errorf("invalid properties.agentPoolProfiles[%q].name %q", app.Name, app.Name))
 		}
-		if app.Count < 1 || app.Count > 30 {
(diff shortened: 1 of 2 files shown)


pnasrat:
Thanks for the quick review @nilsanderselde holding as this requires discussion per release note (and the tests haven't even run yet!)

pnasrat:
Tests passing ready for review.

mjudeikis:
@pnasrat im missing some background. Was this agreed with MSFT and is there a story for this?

mjudeikis:
as per comments on the https://issues.redhat.com/browse/ARO-397, if scaling created consistent failure with outbound rules, we should think about how to change this. Potentially adding more IPs to outbound rules or make adding dynamic. 

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview#scale

mjudeikis:
I want more tests on this to know how snap behaves when scaling up and down. And potentially suggest way to work with it

pnasrat:
As I am no longer on the team I'm closing my PRs. People can pull from my branches and take forward if they want

PR #2187, opened 2020-01-18
add client id for use in unit tests which disables roleLister

jim-minter:
```release-note
Add client id for use in unit tests which disables roleLister
```


olga-mir:
/hold

upgrade v71 needs to be removed from this PR (it has been removed from `release`) need to find a way to drop it from existing PRs
verify fails for all PRs with `Potential DoS vulnerability via decompression bomb`

olga-mir:
/hold cancel 
`verify` test fixed

PR #2188, opened 2020-01-18
add client id for use in unit tests which disables roleLister

jim-minter:
```release-note
add client id for use in unit tests which disables roleLister
```


mjudeikis:
failing due to https://github.com/openshift/openshift-azure/pull/2190

PR #2191, opened 2020-01-22
make config struct invisible for update

mjudeikis:
```release-note
hide config from update return errors
```


mjudeikis:
@jim-minter this is important. PTAL.

PR #2210, opened 2020-02-04
Backend scripts for geneva actions

thekad:
Also removing unused get-node-logs script

```release-note
NONE
```

Re-opening to get rid of the stale checks in the previous PR.


thekad:
Much better now, hit a flake tho
/test e2e-create-20191027-public

PR #2212, opened 2020-02-07
added a customer-view role, as a subset of customer-admin-cluster

Makdaam:
Added a customer-view role, which is assignable by Customer Admins to Service Accounts and other users. It allows viewing (get,list,watch) all resources that Cusotmer Admins can view + issuing Subject Access Reviews and Resource Access Reviews.

This role would enable customers to run their own inventory software with a cluster-wide read-only access. The "view" cluster role is wider than "customer-admin-cluster" and not assignable by a customer.

```release-note
- Added a customer-view role, which is assignable by Customer Admins to Service Accounts and other users. It allows viewing (get,list,watch) all resources that Cusotmer Admins can view + issuing Subject Access Reviews and Resource Access Reviews.
```

### DIFF
--- a/pkg/sync/v15/bindata.go
+++ b/pkg/sync/v15/bindata.go
@@ -15,6 +15,7 @@
 // data/ClusterRole.rbac.authorization.k8s.io/customer-admin-cluster.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/customer-admin-controller.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/customer-admin-project.yaml
+// data/ClusterRole.rbac.authorization.k8s.io/customer-view.yaml
 // data/ClusterRole.rbac.authorization.k8s.io/edit.yaml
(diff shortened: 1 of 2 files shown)


kwoodson:
@Makdaam For history sake, what is interesting about this PR is that we initially introduced this customer view role back when we added the customer admin.  We decided to remove it in favor of granting customer admin instead for simplifying our code paths and for maintenance sake. I'm not saying we should or shouldn't, just wanted to add this POV.

ehashman:
This seems fine to me, but can you point me at the incident where this came up as a requested feature? Want to better understand the rationale and customer use case.

thekad:
LGTM but I also wanna hear the background on this

ehashman:
I'm gonna hold this for now; waiting for a specific customer case to justify the feature.

/hold

ehashman:
/close

let's reopen this one if we need it, gonna pause for now.

PR #2213, opened 2020-02-10
Adding init-network to startup container.

kwoodson:
I was able to test this in my private cluster setup by overwriting the pluginconfig with my custom image.  The cluster came up out of the box with no custom dns and then I ran a `refreshCluster: true` and this was able to work as well.

The only issue currently is that we need to ensure that 168.63.129.16 is in our `/etc/dnsmasq.d/origin-upstream-dns.conf` or we need to place it into `/etc/dnsmasq.conf` as a `server=`.  I'd prefer to just include it in the dnsmasqd file.

Thoughts?

```release-note:
Adding init-network to start up container
```


pkg/entrypoint/startup/cmd.go
```diff
@@ -8,6 +8,8 @@ import (
 
 type cmdConfig struct {
 	config.Common
+	initNetwork bool
+	role        string
```
mjudeikis: We dont use role anymore?

pkg/util/azureclient/network/networkinterface.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-06-01/network"
+
```
mjudeikis: linting should complain?

pkg/util/azureclient/network/networkinterface.go
```diff
@@ -13,6 +14,8 @@ import (
 // InterfacesClient is a minimal interface for azure NewInterfacesClient
 type InterfacesClient interface {
 	Get(ctx context.Context, resourceGroupName string, networkInterfaceName string, expand string) (network.Interface, error)
+	GetVirtualMachineScaleSetNetworkInterface(ctx context.Context, resourceGroupName string, virtualMachineScaleSetName string, virtualmachineIndex string, networkInterfaceName string, expand string) (result network.Interface, err error)
+	ListVirtualMachineScaleSetNetworkInterfaces(ctx context.Context, resourceGroupName string, virtualMachineScaleSetName string) (result network.InterfaceListResultPage, err error)
```
mjudeikis: I think List is not used

kwoodson:
I think this is ready now.  I have tested it and with the latest iteration we do not need the customer to supply the custom Azure dns string by writing it to dnsmasq.conf.

@asalkeld @mjudeikis @jim-minter  
PTAL

kwoodson:
Appears to have timed out running e2e:
```
time="2020-02-11T21:57:56Z" level=debug msg="validating that pod disruption budgets are immutable" func="github.com/openshift/openshift-azure/test/e2e/standard.(*SanityChecker).ValidateCluster" file="/home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/standard/sanitychecker.go:165"
panic: test timed out after 1h0m0s
```

pkg/arm/v15/data/master-startup.sh
```diff
@@ -91,6 +87,21 @@ done
...
+
+# run the startup --init to bootstrap DNS setup
+docker run --privileged --rm --network host -v /:/host:z -e SASURI {{ .Config.Images.Startup }} startup --init-network
+
+# restart network manager to pick up new host settings in the dhclient
+/bin/systemctl restart NetworkManager
+# set the /etc/resolv.conf
+/etc/NetworkManager/dispatcher.d/99-origin-dns.sh
+# restart dnsmasq to get the new settings in /etc/dnsmasq.conf
+/bin/systemctl restart dnsmasq.service
```
jim-minter: I thought that a NetworkManager restart would run 99-origin-dns.sh, and maybe even that 99-origin-dns.sh would restart dnsmasq.  Are both of these lines definitely necessary?  Why?

pkg/startup/v15/startup.go
```diff
@@ -251,3 +254,55 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
+
+	s.log.Info("creating clients")
+	authorizer, err := azureclient.NewAuthorizer(spp.ClientID, spp.Secret, s.cs.Properties.AzProfile.TenantID, "")
+	if err != nil {
+		return err
+	}
+
+	ncli := network.NewInterfacesClient(ctx, log, s.cs.Properties.AzProfile.SubscriptionID, authorizer)
+	var dnsString string
+	nic, err := ncli.GetVirtualMachineScaleSetNetworkInterface(ctx, s.cs.Properties.AzProfile.ResourceGroup, "ss-master", "0", "nic", "")
```
jim-minter: nit: s/ss-master/names.MasterScalesetName/

pkg/startup/v15/startup.go
```diff
@@ -251,3 +254,55 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
+	if err != nil {
+		return err
+	}
+	if nic.DNSSettings != nil {
```
jim-minter: in what circumstances is this nil? why is leaving dnsString empty the correct behaviour in this case? can you just use `*nic.DNSSettings.InternalDomainNameSuffix` ?

pkg/startup/v15/startup.go
```diff
@@ -251,3 +254,55 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
+		dnsString = to.String(nic.DNSSettings.InternalDomainNameSuffix)
```
jim-minter: `dnsString = *nic.DNSSettings.InternalDomainNameSuffix`

pkg/startup/v15/startup.go
```diff
@@ -251,3 +254,55 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
+	err = writeContentsToFile("/host/etc/dhcp/dhclient-eth0.conf", domainSettings)
+	if err != nil {
+		return err
+	}
+
+	err = writeContentsToFile("/host/etc/dnsmasq.conf", []string{fmt.Sprintf("server=/%s/%s\n", dnsString, constants.AzureNameserver)})
+	return err
+}
+
+func writeContentsToFile(filename string, contents []string) error {
```
jim-minter: use ioutil.WriteFile

pkg/startup/v12/startup.go
```diff
@@ -244,3 +244,9 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
 
 	return w.Close()
 }
+
+// WriteSearchDomain queries for the search domain and writes to
+// /etc/dhcp/dhclient-eth0.conf.  This is for private api clusters only
+func (s *startup) WriteSearchDomain(ctx context.Context, log *logrus.Entry) error {
+	return fmt.Errorf("not supported in this release")
```
asalkeld: might be better to return nil here incase we need to patch a prod cluster with a newer startup image (for an unrelated bug). in this case we would want this to be a no-op, not fail

pkg/startup/v15/startup.go
```diff
@@ -251,3 +254,55 @@ func (s *startup) writeFiles(role api.AgentPoolProfileRole, w writers.Writer, ho
...
+	s.log.Info("writing custom dns")
+
+	domainSettings := []string{
+		"interface \"eth0\" {\n",
+		fmt.Sprintf("    supersede domain-name \"%s\";\n", dnsString),
+		fmt.Sprintf("    supersede domain-search \"%s\";\n", dnsString),
+		"}\n",
+	}
+
+	// write to dnsmasq.conf and see if that is sufficient
```
asalkeld: is this a TODO or an old comment, somewhat confusing

pkg/arm/v15/data/master-startup.sh
kwoodson: What I found during testing(from memory) was the following. 

1. The restart to network manager was required to pick up the dhclient-eth0.conf changes.
1. On restart the network manager would place the dhclient-eth0.conf into /var/lib/NetworkManager/dhclient-eth0.conf which would cause a refresh of the lease file.
1. This would cause NetworkManager to update /etc/resolv.conf with the new search domain and the dns servers.  I'm not 100% sure if the 99-origin-dns.sh would rerun.  I had found where the settings for the local dnsmasq's 10. address were not in the resolv.conf.  A rerun of the 99-origin-dns.sh would ensure that the 10. address was in resolv.conf and that the search domains were ordered correctly, cluster.local first, followed by azure domain.
1. Then finally the dnsmasq.service restart that causes dnsmasq to pick up the newly appended search string `server=/<Azure domain>/168.63.129.16`.  In testing, I think that the NetworkManager restart did not pick up the latest settings from dnsmasq.conf.

pkg/startup/v15/startup.go
kwoodson: From looking at the description of `ioutil.WriteFile`, I did not see an append mode.  For the dnsmasq.conf I needed to append the `server=` string.  Is this possible with `ioutil.WriteFile`?  The many examples that I looked at showed using the `os.OpenFile` method with the `os.O_APPEND|os.O_CREATE|os.O_WRONLY`.

pkg/startup/v12/startup.go
kwoodson: sgtm

pkg/startup/v15/startup.go
kwoodson: removed

ehashman:
can't spell..

/test upgrade-private-v14.1

pkg/arm/v15/data/master-startup.sh
```diff
@@ -74,10 +74,6 @@ done
 
 iptables-save >/etc/sysconfig/iptables
 
-# when starting node waagent and network utilities goes into race condition.
-# if waagent runs before dns is known to the node we end up with empty string
-while [[ $(hostname -d) == "" ]]; do sleep 1; done
```
ehashman: Why do we move this after image pulls?

pkg/startup/v15/startup.go
ehashman: What's wrong with opening the file using `os.O_APPEND`? We only use this helper twice, could just call `Write` twice against the open file?

pkg/arm/v15/data/master-startup.sh
kwoodson: When setting custom dns on the private link vnet the internal Azure naming is no longer given to the cluster.  This means that we will never pass the above `hostname -d` check.  How we solve this is to query for the Azure dns domain inside of the startup container and write the settings to the correct files.  Then check for the hostname afterwards.

pkg/startup/v15/startup.go
kwoodson: On one of the files we want to create and on one of the files we want to append.  Having this code two times in the same function seemed to violate DRY so I created a helper function on the side that accommodates both calls.

jim-minter: Ooh, sorry, didn't see the O_APPEND, ignore me

jim-minter:
/lgtm
/hold

holding just in case @kwoodson or @ehashman don't want this to merge yet.  Unhold if happy.  LGTM.

ehashman:
/hold cancel

Assuming tests pass I think this is good to go, once it merges I'll cut v15.

kwoodson:
```
Ran 23 of 40 Specs in 1203.798 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 17 Skipped
--- PASS: TestE2E (1203.80s)
PASS
ok  	github.com/openshift/openshift-azure/test/e2e	1211.386s
```

ehashman:
Test passed but CI still isn't reporting it and I don't see any reason after nearly 20 hours that we hold this up any longer...

/override upgrade-private-v14.1

ehashman:
Tide has this queued behind #2221, I'm just going to manually merge as I'd like to cut the release.

PR #2215, opened 2020-02-11
disabling DenyEscalatingExec

Makdaam:
```release-note
Allow binary builds.
```

PR #2170, opened 2020-01-07
E2E test for working binary builds 

pnasrat:
Create a basic e2e test for binary builds

https://docs.openshift.com/container-platform/3.11/dev_guide/dev_tutorials/binary_builds.html

```release-note
NONE
```

PR #2216, opened 2020-02-12
[release-v14] Prep v14.2, disable DenyEscalatingExec

ehashman:
```release-note
Bugfix: fixed oc binary builds
```

Backports #2170 to release-v14.

Opening this in favour of #2215 since there's a bunch of extra prep work that needs to be done for the v14.2 release, as we're touching files that are laid down on disk. Even though we only make changes in one of the four directories, I think we should be making copies of all of them?

I also made sure to pull in the e2e test in this PR. I checked to ensure that it failed on a 14.1 cluster (huzzah for integration test driven development). It should now pass with the fix.

@m1kola for more context around why this is the fix we opted for, see commit message of b882857.


ehashman:
https://github.com/openshift/release/pull/7129 adds 14.1 upgrade tests.

asalkeld:
looking good to me, waiting for tests

PR #2217, opened 2020-02-12
update image version

mjudeikis:
```release-note
Update images to 14.2
```


ehashman:
/hold

Do we actually have any clusters in prod running a vendored 14.2 on the RP? I thought we just hotfixed?

(i.e. I think this is all moot because I can probably remove all the 14.2 references as it was never deployed in prod, yes?)

ehashman:
/override verify

missing the gosec fix on this branch

/override upgrade-v12.2
/override upgrade-v10.1.1

those VM images are gone now I think

PR #2219, opened 2020-02-13
Would be nice to know when the etcdbackup finishes

thekad:
```release-note
NONE
```

### DIFF
--- a/pkg/entrypoint/etcdbackup/etcdbackup.go
+++ b/pkg/entrypoint/etcdbackup/etcdbackup.go
@@ -94,5 +94,8 @@ func start(cfg *cmdConfig) error {
 	default:
 		flag.Usage()
 	}
+
+	log.Print("etcdbackup finished")
+
 	return err
 }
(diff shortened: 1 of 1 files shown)


pkg/entrypoint/etcdbackup/etcdbackup.go
```diff
@@ -94,5 +94,8 @@ func start(cfg *cmdConfig) error {
 	default:
 		flag.Usage()
 	}
+
+	log.Print("etcdbackup finished")
```
olga-mir: just out of curiosity is there any rule where we use `log.Info`[f] and where `log.Print`, I can see we have both, but "Print" not as much as "info"

kwoodson: I think I'd prefer `.Info` as I do know that logging generally has a switch to turn on logging levels and `Info` falls into a specific level whereas I don't know what `Print` does.

thekad: Yeah I only copy/pasted the line from a few lines above where the "starting" message appears. I do agree it is confusing, I'll change it to Info.

thekad:
Couple of timeout errors
/retest

Issue #2118, opened 2019-12-04
Privileged container test flaking between versions


ehashman:
Upgrade 7.1 test failing with:

```
• Failure [8.194 seconds]
Openshift on Azure customer-admin e2e tests [CustomerAdmin][EveryPR]
/home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/customer_admin.go:27
  should not be able to create a non-whitelisted privileged container [It]
  /home/prow/go/src/github.com/openshift/openshift-azure/test/e2e/specs/customer_admin.go:344

  Expected
      <string>: pods "test" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
  to equal
      <string>: admission webhook "aro-admission-controller.aro.openshift.io" denied the request: spec.containers[0]: Forbidden: requires privileges but image is not whitelisted on platform 
```

Not sure if this is specific to an old version upgrade? I've only seen one instance of the failure.

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2116/pull-ci-azure-master-upgrade-v7.1/537/build-log.txt #2116 

/kind test-flake
/assign @Makdaam

ehashman:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2122/pull-ci-azure-master-upgrade-v10.1.1/272/build-log.txt #2122

ehashman:
Just saw this on https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-azure/2292/pull-ci-azure-master-upgrade-v15.0/132 for #2292

PR #2222, opened 2020-02-13
added an alternative error

Makdaam:
The error  can be received before admission controller initializes.
Fixes flake from #2118 
```release-note
NONE
```


olga-mir:
/lgtm 
but can you please explain when the second error can occur? Is this sort of a race condition?

jim-minter:
@Makdaam do you understand how the race happens?  Please can you explain it here before we merge this?

ehashman:
/close

closing this as low priority

PR #2223, opened 2020-02-13
Adding cronjobs/jobs/daemonsets to the output

thekad:
Might be useful for cases when e.g. etcd-backup is stuck

```release-note
NONE
```


thekad:
flake #1778 

/retest

pkg/vmimage/data/usr/local/bin/oc-get-all
```diff
@@ -7,7 +7,7 @@ then
   exit
 fi
 
-for target in group user deploy service pod route
+for target in node group user daemonset service route deploy cronjob job pod
```
ehashman: Are cronjobs different than jobs?

thekad: yes, a job named "foo" will run once and you have to re-create again if you want it to run again. A cronjob named "bar" will spawn jobs named "bar-${timestamp}" on whatever schedule the parent cronjob has defined.

PR #2226, opened 2020-02-14
Fix for custom dns on non-master

kwoodson:
After more testing and demonstrating, I noticed the new startup container was not applying the fix to the non-master nodes.  This is unfortunate due to timing of the v15 release.  I was able to determine that the `cs.Properties.PrivateAPIServer` attribute is not available on non-master nodes during CSE boot time.

The proposed fix is to run it on boot regardless of private API server or not.  This shouldn't have any effect on clusters where Azure dns is being used and will be a basic no-op.

```release-note:
Fix custom dns on non-master
```


ehashman:
@kwoodson can you rebase this PR against the v15 branch and backport to v16?

PR #2227, opened 2020-02-14
Ensure that all nodes run the init-network

kwoodson:
```release-note:
Applying init-network to all nodes
```


pkg/entrypoint/startup/startup.go
```diff
@@ -56,11 +56,9 @@ func runInitNetwork(ctx context.Context, log *logrus.Entry) error {
 		return err
 	}
 
-	if cs.Properties.PrivateAPIServer {
```
asalkeld: I think this is because this function https://github.com/openshift/openshift-azure/blob/master/pkg/startup/v15/startup.go#L45 is stripping it out. An alternative solution is to add privateApiServer to that function.

pkg/arm/v15/data/node-startup.sh
```diff
@@ -72,8 +72,6 @@ docker run --privileged --rm --network host -v /:/host:z -e SASURI {{ .Config.Im
 
 # restart network manager to pick up new host settings in the dhclient
 /bin/systemctl restart NetworkManager
-# set the /etc/resolv.conf
-/etc/NetworkManager/dispatcher.d/99-origin-dns.sh
```
asalkeld: this is not needed?

mjudeikis: It is run as part of NM restart

pkg/entrypoint/startup/startup.go
mjudeikis: Does this makes this to be applied to all clusters?

kwoodson: Testing this: 
https://github.com/openshift/openshift-azure/pull/2229

ehashman:
Closing for now in favour of the approach in #2229 

/close

PR #2229, opened 2020-02-17
Include private api server inside of worker properties.

kwoodson:
```release-note
Include private api server property to the worker nodes
```

PR #2230, opened 2020-02-17
Include private api server inside of worker properties.

kwoodson:
```release-note
Including private api server inside of worker properties
```

### DIFF
--- a/pkg/startup/v15/startup.go
+++ b/pkg/startup/v15/startup.go
@@ -47,6 +47,7 @@ func (s *startup) GetWorkerCs() *api.OpenShiftManagedCluster {
 		ID:   s.cs.ID,
 		Name: s.cs.Name,
 		Properties: api.Properties{
+			PrivateAPIServer: s.cs.Properties.PrivateAPIServer,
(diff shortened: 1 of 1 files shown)


kwoodson:
v15 backport of #2229

kwoodson:
Failed the upgrade already:
```
time="2020-02-17T20:29:47Z" level=fatal msg="400: 400 Bad Request: Failed to apply request: UpdateMasterAgentPoolReimage: Code=\"CannotAllocateRemainingVMsInAvailabilitySet\" Message=\"Delete/Deallocate operation on VM 'ss-master_0' failed because the remaining VMs in the Availability Set '_ss-master_a34b811327774622a9acd4af07b53c13' cannot be allocated together. Changes in Availability Set allocation need to be executed atomically. Please deallocate or delete some or all of these VMs before retrying the current operation. \\r\\nPlease note that this VM is not allocated and won't accrue any charges.\\r\\nDetails: The VM allocation failed due to an internal error. Please retry later or try deploying to a different location.\"\n"
```

PR #2261, opened 2020-03-19
Update customer-admin role with create clusterrole ability

gburges:
```release-note
NONE
```


ehashman:
/hold

I don't recall reporting this -- is there an associated JIRA or incident?

gburges:
> /hold
> 
> I don't recall reporting this -- is there an associated JIRA or incident?

https://issues.redhat.com/browse/ARO-509

m1kola:
This can allow customers to escalate their privileges. Are we sure that we want to do this?

m1kola:
Also I do not see where the documentation implies the ability to create clusterroles: doc in [the link from jira says](https://docs.openshift.com/aro/admin_guide/index.html) says that customer admins can describe (get) roles. But I do not see anything about creating roles.

ehashman:
3.11 is in feature freeze while we work towards 4.x GA. Unprioritized JIRAs like this one should not be considered valid for implementation right now.

PR #2270, opened 2020-04-08
Allow omsagent daemonset recreation in sync pod

olga-mir:
```release-note
Allow to recreate omsagent daemonset in sync pod
```
I've tested this manually on master branch:
1. Create dev-cluster `make create` (make sure `AZURE_WORKSPACE_ID` is set to valid resourceid)
2. Check sync pod logs.
3. Remove sync pod by scaling it to 0: `oc scale deployment --replicas=0 --namespace=kube-system sync`
4. Apply this patch to simulate the upgrade problem that we have now: https://github.com/openshift/openshift-azure/commit/83d79e744258d1af223c69d0a5290f2491415bd0
5. `make run-sync` and observe it crashes ==> this proves that problem is reproduced.
6. Apply this fix: https://github.com/openshift/openshift-azure/commit/ba4a620287707b4837d774988f22d3a7572887e9
7. Restore the cluster and reproduce the problem again - this time sync pod runs with no problem.

* There are no changes on selectors on omsagent `deployment`, therefore we have this problem only on daemonset. 
* I suggest in addition to checking the error message we apply this fix for omsagent only, to minimize the change. 
* when daemonset is recreated, it takes in my dev cluster about 10 minutes for the new pod to come up. There is 10 minutes blackout in some (most?) metrics, e.g. active pods count is not affected.

### DIFF
--- a/pkg/sync/v16/client.go
+++ b/pkg/sync/v16/client.go
@@ -204,14 +204,17 @@ func (s *sync) write(o *unstructured.Unstructured) error {
 
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
-		if err != nil && strings.Contains(err.Error(), "updates to parameters are forbidden") {
-			s.log.Infof("object %s is not updateable, will delete and re-create", o.GetName())
-			err = dc.Resource(res, o.GetNamespace()).Delete(o.GetName(), &metav1.DeleteOptions{})
-			if err != nil {
-				return
+		if err != nil {
+			if strings.Contains(err.Error(), "updates to parameters are forbidden") ||
+				(strings.Contains(err.Error(), "field is immutable") && strings.Contains(o.GetName(), "omsagent")) {
+				s.log.Infof("object %s is not updateable, will delete and re-create", o.GetName())
+				err = dc.Resource(res, o.GetNamespace()).Delete(o.GetName(), &metav1.DeleteOptions{})
+				if err != nil {
+					return
+				}
+				o.SetResourceVersion("")
+				_, err = dc.Resource(res, o.GetNamespace()).Create(o)
 			}
-			o.SetResourceVersion("")
-			_, err = dc.Resource(res, o.GetNamespace()).Create(o)
 		}
 
 		return
--- a/pkg/sync/v17/client.go
+++ b/pkg/sync/v17/client.go
@@ -204,14 +204,17 @@ func (s *sync) write(o *unstructured.Unstructured) error {
 
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
-		if err != nil && strings.Contains(err.Error(), "updates to parameters are forbidden") {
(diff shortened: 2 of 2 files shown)


olga-mir:
Actually I think we don't need this fix in v17

pkg/sync/v16/client.go
```diff
@@ -204,14 +204,17 @@ func (s *sync) write(o *unstructured.Unstructured) error {
...
 		o.SetResourceVersion(rv)
 		_, err = dc.Resource(res, o.GetNamespace()).Update(o)
-		if err != nil && strings.Contains(err.Error(), "updates to parameters are forbidden") {
-			s.log.Infof("object %s is not updateable, will delete and re-create", o.GetName())
-			err = dc.Resource(res, o.GetNamespace()).Delete(o.GetName(), &metav1.DeleteOptions{})
-			if err != nil {
-				return
+		if err != nil {
+			if strings.Contains(err.Error(), "updates to parameters are forbidden") ||
+				(strings.Contains(err.Error(), "field is immutable") && strings.Contains(o.GetName(), "omsagent")) {
```
thekad, line 209: while I agree this would minimize the potential impact, I see no harm in dropping the ``o.GetName() == "omsagent"`` conditional, the original conditional was a doing a blanket check based on the error message.

ehashman: I think I agree with this -- @jim-minter is there any risk in always deleting and recreating sync-managed objects that have immutable field changes?

olga-mir: I am not very fussed about v3 and I am okay to remove the extra check for `omsagent`.
`It is generally discouraged to make label selector updates and it is suggested to plan your selectors up front.` -  https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates
In my opinion, going forward if we do need to change spec.selectors we need to be aware that this is a bigger change than for example adding a label or changing resources. 
I agree that broken upgrade cycle is high price to pay, but I think this should be solved by adding tests, not by making exceptions wider. 
(we have an upgrade test, why didn't it fail?)

jim-minter: `strings.Contains(err.Error(), "field is immutable") && strings.Contains(o.GetName(), "omsagent")` feels a fraction tight but presumably works for now to get unblocked.

`strings.Contains(err.Error(), "field is immutable")` possibly feels a bit loose.

I think I'd ship as is.  Failing that I think I'd either try to fix on a longer error message and/or kind DaemonSet.

jim-minter:
lgtm, but letting @ehashman drive.

pkg/sync/v16/client.go
olga-mir: Answering my own question from above - the upgrade test did not fail because they don't run with azure monitoring enabled.

olga-mir:
/retest

(ci aad has been updated now)

olga-mir:
/retest

 flake `"timed out waiting for the condition"`

ehashman:
c'mon bot

/lgtm

PR #2279, opened 2020-04-22
Content update

ehashman:
```release-note
NONE
```

Let's see if this passes CI reliably vs. #2275.


ehashman:

timeout flake

ehashman:
I think we should go ahead and merge this, it's not running into the same failures we were seeing in the bot content update.

PR #2284, opened 2020-05-04
Disable gosec checks

ehashman:
Upstream changes in gosec keep breaking validation on builds. Since this
repo is in maintenance mode, I'd like to disable the checks.

```release-note
NONE
```


thekad:
I am OK with this, but I think someone else should weigh in.

PR #2289, opened 2020-05-19
resolve "http2: no cached connection was available" error in admission controller

### LINKED ISSUE #2287: aro-admission-controller fails when running for a long time.
The following error messages show up in the log of aro-admission-controller (api server above get's Internal Server Error from aro-admission-controller). The issue occurs after the process is running for 3-4 weeks on a relatively busy cluster.

"error fetching namespace X required to preallocate values for restricted: Get https://clusterurl.region.cloudapp.azure.com/api/v1/namespaces/X: http2: no cached connection was available"

At first glance it seems to be a problem with the outbound connections not being released properly back into the pool.

This issue is to track progress on the investigation.

jim-minter:
fixes #2287

```release-note
resolve "http2: no cached connection was available" error in admission controller
```

This is basically a rebase to get https://github.com/golang/go/issues/22091 .


glide.yaml
```diff
@@ -81,6 +81,8 @@ import:
   version: v3
 - package: google.golang.org/grpc
   version: v1.16.0
+- package: golang.org/x/net
+  version: ab555f366c4508dbe0802550b1b20c46c5c18aa0
```
jim-minter, line 85: I am considering being a chicken and
a) bumping golang.org/x/net to the minimum commit needed
b) not doing a wider `glide up`

pkg/entrypoint/admissioncontroller/admissioncontroller.go
```diff
@@ -70,6 +72,14 @@ func newAdmissionController(ctx context.Context, log *logrus.Entry, configFile s
 		return nil, err
 	}
 
+	if cs.Properties.PrivateAPIServer {
+		hostname := os.Getenv("masterNodeName")
+		if hostname == "" {
+			return nil, fmt.Errorf("could not read hostname from env[masterNodeName]")
+		}
+		cs.Config.MasterKubeconfig.Clusters[0].Cluster.Server = "https://" + hostname
+	}
```
jim-minter, line 81: I don't think that this contributes to the core issue, but I think it's worth fixing nonetheless.  Functionality roughly copied from the sync pod.

PR #2292, opened 2020-05-19
Disable etcd backups on upgrades

ehashman:
```release-note
Internal fix: no longer perform etcd-backup on upgrade
```

MS has been disabling this by commenting out:

https://github.com/openshift/openshift-azure/blob/c70152ea551524eb7470c19519d01a1d7035ea8d/pkg/plugin/plugin.go#L287-L293

To avoid leaving a dangling `backupEtcd` argument and an associated refactor to clean it up, I instead located everywhere that `createOrUpdateExt` is called and ensured that `backupEtcd` is always set to false. But I'm not attached to this solution, I could also just remove the code above.

Hopefully this will unblock upgrades on private clusters.

/cc @jim-minter

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -249,7 +249,8 @@ const (
 
 func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool, deployFn api.DeployFn) *api.PluginError {
 	if isUpdate {
-		return p.createOrUpdateExt(ctx, cs, updateTypeNormal, deployFn, true)
+		// NOTE(ehashman): disable backups on updates, they regularly fail
+		return p.createOrUpdateExt(ctx, cs, updateTypeNormal, deployFn, false)
 	}
 	return p.createOrUpdateExt(ctx, cs, updateTypeCreate, deployFn, false)
 }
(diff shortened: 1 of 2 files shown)


pkg/plugin/plugin.go
```diff
@@ -249,7 +249,7 @@ const (
 
 func (p *plugin) CreateOrUpdate(ctx context.Context, cs *api.OpenShiftManagedCluster, isUpdate bool, deployFn api.DeployFn) *api.PluginError {
 	if isUpdate {
-		return p.createOrUpdateExt(ctx, cs, updateTypeNormal, deployFn, true)
+		return p.createOrUpdateExt(ctx, cs, updateTypeNormal, deployFn, false)
```
jim-minter, line 252: a comment wouldn't hurt.

ehashman:
Upgrade flakes were quota-related, unit tests were a real failure (should be fixed).

ehashman:
flakes
upgrade-v16.0 timed out on provision
upgrade-v15.0 #2118 

/retest

ehashman:
@jim-minter this should be GTG if you wanna TAL

ehashman:
@mjudeikis can you TAL and make sure this goes into v18.0?

PR #2303, opened 2020-06-16
Fix ipv6 routing

mjudeikis:
```release-note
NONE
```


mjudeikis:
/override e2e-create-20190430
/override e2e-create-20191027-public 
Tests passed but no status were reported

PR #2332, opened 2020-09-30
Pre-Release update v20

mjudeikis:
```release-note
Update vm and container images
```


petrkotas (approved):
Seems good to me, if the images indeed exist in the registry.

PR #2339, opened 2021-01-06
Fix Ci

mjudeikis:
```release-note
NONE
```

* Fix Ci for secret mounting
* Remove generation validation for now


jim-minter (approved):
The following changes should be reverted in a future PR:
* make verify / validate-generated
* xtrace in ci-prepare.sh
* early return in ci-prepare.sh

